PR #2, opened 2024-10-21
Resolve "Web FTP"

### LINKED ISSUE #1: Web FTP
To natively support `FTP` features in **AIP**, we need to add the following features and components:

- Implement `FTPClient`
   - This is a component designed to handle AIP's Web FTP Command that maps to the FTP Method.
   - The following methods must be supported:

        FTP Command | Web FTP Command
        -- | --
        LIST | `list`
        MKD |  `mkd`
        CWD | `cwd`
        PWD | `pwd`
        DELE | `dele`
        RMD | `rmd`
        RNFR/RNTO | `mv`
        Not Defined | `cp`

- `openftp` Command Addition
   - An internal command that runs the `FTPClient` thread.
   - `openftp` command needs to be integrated into the `CommandRunner`.
   
- Update `upload` & `download` Command
   - The upload and download commands need to be updated to support the following features.
      - bulk upload/download
      - recursive upload/download
      - zip upload
      - unzip download

geunwoonoh:
As mentioned in https://github.com/alpacanetworks/alpamon-go/issues/1, the following changes have been made to natively replace the FTP function with AIP:

- Implement FTPClient
   - An FTPClient has been implemented to establish a Control Connection with alpacon-server.
   - FTPClient receives Web FTP Commands, which map to FTP Commands, from the user and processes them.

FTP Command | Web FTP Command
-- | --
LIST | list
MKD | mkd
CWD | cwd
PWD | pwd
DELE | dele
RMD | rmd
RNFR/RNTO | mv
Not Defined | cp

- `openftp` Command Addition
   - `openftp` Command runs `FTPClient` as go routine.
   - `openftp` command integrate into the `CommandRunner`.

- Update upload & download alpamon Command
   - Web FTP provides FTP's Data Connection-like functionality through a REST API.
   - The upload and download commands support the following features.
      - bulk upload/download
      - recursive upload/download
      - zip upload
      - unzip download

Closes #1


royroyee:
Please fix all issues flagged by lint

pkg/runner/ftp.go
```diff
@@ -0,0 +1,428 @@
...
+	groupname        string
+	homeDirectory    string
+	workingDirectory string
+	sessionID        string
+}
+
+var ftpTerminals map[string]*FtpClient
+
+func init() {
+	ftpTerminals = make(map[string]*FtpClient)
```
royroyee: A global variable `terminals` was added in pty for `session_id` reference due to resize, but if FTP doesn’t require this, please remove the global variable

pkg/runner/command.go
```diff
@@ -468,7 +489,22 @@ func (cr *CommandRunner) runFileUpload(fileName string) (exitCode int, result st
...
+		return 1, err.Error()
+	}
+
+	name, err := makeArchive(paths, bulk, recursive, sysProcAttr)
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to create archive")
+		return 1, err.Error()
+	}
+	if bulk || recursive {
+		defer os.Remove(name)
```
royroyee: Ensure errors are handled explicitly when using defer (Go convention). Review and fix other areas as needed
```go	
	defer func() { _ = os.Remove(name) }()
```

pkg/runner/command.go
```diff
@@ -586,3 +632,108 @@ func getFileData(data CommandData) ([]byte, error) {
 
 	return content, nil
 }
+
+func parsePaths(pathList []string) ([]string, bool, bool, error) {
```
royroyee: Consider adding descriptive names for return values to improve readability, especially when there are multiple return variables in Go. For example:
```go
func (cr *CommandRunner) delGroup() (exitCode int, result string) {

```

royroyee:
The current code heavily uses `map[string]interface{}`, leading to frequent type assertions. In Go, it’s better to define explicit structs for type safety and to reduce conversions.

For request/response types, using **omitempty** can simplify things.

Since invalid type assertions cause a panic in Go, it’s safer to use:

```go
size, ok := result["size"].(int64)
```

Alternatively, you can use a switch statement to determine the type beforehand:
```go
switch v := addr.(type) {
case *net.IPNet:
    ...
}
```

pkg/runner/ftp_types.go
```diff
@@ -0,0 +1,89 @@
+package runner
+
+import "strings"
+
+type returnCode struct {
+	Success int            `json:"success"`
+	Error   map[string]int `json:"error"`
+}
+
+var returnCodes = map[string]returnCode{
```
royroyee: It’s better to use const for the error messages rather than repeating them:
```go
const (
    ErrPermissionDenied = "permission denied"
    ErrInvalidArgument  = "invalid argument"
    ErrNoSuchFile       = "no such file or directory"
    ...
    )
```
Additionally, instead of returning `map[string]string`, it would be clearer to define a specific type for the return value.

geunwoonoh:
Based on your feedback, I've refactored the code and conducted additional testing in a Linux environment. Please let me know if you have any further comments or if you would like me to make any additional changes.

royroyee (changes requested):

pkg/runner/ftp_types.go
```diff
@@ -0,0 +1,136 @@
...
+type FtpContent struct {
+	Command FtpCommand `json:"command"`
+	Data    FtpData    `json:"data"`
+}
+
+type FtpResult struct {
+	Command FtpCommand  `json:"command"`
+	Success bool        `json:"success"`
+	Code    int         `json:"code,omitempty"`
+	Data    interface{} `json:"data,omitempty"`
```
royroyee: Consider defining a specific type for the `data` field instead of using an interface. Please review

royroyee (changes requested):

royroyee:
It looks like there are some []string types like `path` that need validation for empty arrays. Accessing an empty array could cause a panic

scripts/postinstall.sh
```diff
@@ -24,6 +25,32 @@ check_systemd_status() {
...
+install_zip_package() {
+  echo "Checking and installing zip package..."
+
+  if command -v zip &> /dev/null; then
+    echo "zip package is already installed."
+    return 0
+  fi
+
+  if command -v apt-get &> /dev/null; then
+    apt-get update
```
royroyee: No update needed. it’s already included in the Linux package script

geunwoonoh:
I've addressed all the code review comments. Please review and resolve if everything looks good.

geunwoonoh:
I've made additional improvements to the makeArchive() for better readability.

pkg/runner/ftp_types.go
royroyee: Permissioned -> Permission

geunwoonoh: I've fixed typo!

pkg/runner/ftp.go
```diff
@@ -0,0 +1,412 @@
...
+	cmd.SysProcAttr = fc.sysProcAttr
+	output, _ := cmd.Output()
+
+	return strings.HasPrefix(string(output), "d")
+}
+
+func (fc *FtpClient) list(rootDir string, depth int) (CommandResult, error) {
+	path := fc.parsePath(rootDir)
+	cmdResult, err := fc.listRecursive(path, depth, 0)
+	return *cmdResult, err
```
royroyee: Remove unnecessary pointer dereferencing

royroyee:
LGTM~

PR #8, opened 2024-10-31
Resolve "Refactor Web FTP Command"

### LINKED ISSUE #3: Refactor Web FTP Command
The current `Web FTP Command` has the following issues:

- Currently, it relies on the `os/exec` package to execute Linux commands based on user permissions.
- This approach not only degrades performance but also introduces a high risk of errors.
    - In particular, the `list` command is vulnerable.
    
To resolve these issues, the following refactoring is necessary:

- A new process should be created with appropriate user permissions using the low-level interface `os.StartProcess()`.
- The existing logic of the `Web FTP Command` should be refactored to utilize `Go`'s built-in packages.
- This `Web FTP Command` should run in a new process.

geunwoonoh:
As mentioned in https://github.com/alpacanetworks/alpamon-go/issues/3, we have improved the performance and reduced the error rate of `Web FTP command` by implementing the following changes:

- Refactor `Web FTP Command` to utilize Go's built-in packages.
   - To address the performance issue caused by using the `os/exec` package to execute Linux commands, the logic has been refactored to leverage Go's built-in packages.
   
- Add `ftp` alpamon command
   - Given the inefficiency of creating a new, less privileged process for each command using `os/exec`, we have decided to execute `Web FTP Command` in a separate process.
      - To achieve this, we have added the `ftp` command to alpamon, allowing `Web FTP Command` to be executed in a new, less privileged process.
      
- Add `zip` package dependency
   - As mentioned in https://github.com/alpacanetworks/alpamon-go/issues/4, `zip` package dependency has been added to `.goreleaser.yaml` file.

### DIFF
--- a/cmd/alpamon/command/ftp.go
+++ b/cmd/alpamon/command/ftp.go
@@ -1,15 +1,9 @@
 package command
 
 import (
-	"fmt"
-	"os/user"
-	"strconv"
-	"syscall"
-
 	"github.com/alpacanetworks/alpamon-go/pkg/config"
 	"github.com/alpacanetworks/alpamon-go/pkg/logger"
 	"github.com/alpacanetworks/alpamon-go/pkg/runner"
-	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
 )
 
@@ -18,10 +12,12 @@ var ftpCmd = &cobra.Command{
 	Short: "Start worker for Web FTP",
 	Args:  cobra.ExactArgs(4),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		username := args[0]
-		groupname := args[1]
-		url := args[2]
-		homeDirectory := args[3]
+		data := runner.CommandData{
+			Username:      args[0],
+			Groupname:     args[1],
+			URL:           args[2],
+			HomeDirectory: args[3],
+		}
 
 		logFile := logger.InitLogger()
 		defer func() { _ = logFile.Close() }()
@@ -29,57 +25,13 @@ var ftpCmd = &cobra.Command{
 		settings := config.LoadConfig()
 		config.InitFtpSettings(settings)
 
-		if syscall.Getuid() == 0 {
-			if username == "" || groupname == "" {
-				log.Debug().Msg("No username or groupname provided.")
-				return fmt.Errorf("No username or groupname provided.")
-			}
-
-			usr, err := user.Lookup(username)
-			if err != nil {
-				log.Debug().Msgf("There is no corresponding %s username in this server", username)
-				return fmt.Errorf("There is no corresponding %s username in this server", username)
-			}
-
-			group, err := user.LookupGroup(groupname)
-			if err != nil {
-				log.Debug().Msgf("There is no corresponding %s groupname in this server", groupname)
-				return fmt.Errorf("There is no corresponding %s groupname in this server", groupname)
-			}
-
-			uid, err := strconv.Atoi(usr.Uid)
-			if err != nil {
-				return err
-			}
-
-			gid, err := strconv.Atoi(group.Gid)
-			if err != nil {
-				return err
-			}
-
-			err = syscall.Setgroups([]int{})
-			if err != nil {
-				return err
-			}
-
-			err = syscall.Setuid(uid)
-			if err != nil {
-				return err
-			}
-
-			err = syscall.Setgid(gid)
-			if err != nil {
-				return err
-			}
-		}
-
-		RunFtpWorker(url, homeDirectory)
+		RunFtpWorker(data)
 
 		return nil
 	},
 }
 
-func RunFtpWorker(url, homeDirectory string) {
-	ftpClient := runner.NewFtpClient(url, homeDirectory)
+func RunFtpWorker(data runner.CommandData) {
+	ftpClient := runner.NewFtpClient(data)
 	ftpClient.RunFtpBackground()
 }
--- a/cmd/alpamon/command/ftp.go
+++ b/cmd/alpamon/command/ftp.go
@@ -1,7 +1,6 @@
 package command
 
 import (
-	"github.com/alpacanetworks/alpamon-go/pkg/config"
 	"github.com/alpacanetworks/alpamon-go/pkg/logger"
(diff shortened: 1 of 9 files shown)


pkg/runner/ftp.go
```diff
@@ -401,15 +369,85 @@ func (fc *FtpClient) cpDir(src, dst string) (CommandResult, error) {
...
 		Message: fmt.Sprintf("Copy %s to %s", src, dst),
 	}, nil
 }
+
+func copyFile(src, dst string) error {
+	srcFile, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer srcFile.Close()
```
royroyee: Ensure errors are handled explicitly when using defer (Go convention). Review and fix other areas as needed

pkg/runner/command.go
```diff
@@ -153,8 +153,35 @@ func (cr *CommandRunner) handleInternalCmd() (int, string) {
 			return 1, fmt.Sprintf("openftp: Not enough information. %s", err.Error())
 		}
 
-		ftpClient := NewFtpClient(cr.data)
-		go ftpClient.RunFtpBackground()
+		sysProcAttr, err := demote(data.Username, data.Groupname)
```
royroyee: Too much logic in the "openftp" case block. Recommend refactoring into separate functions for readability. Also, avoid reassigning variables like `err` multiple times in the same block.

pkg/logger/logger.go
```diff
@@ -64,6 +65,29 @@ func InitLogger() *os.File {
 	return logFile
 }
 
+// TODO : Send logs to alpamon's Logserver using a Unix domain socket
+func InitFtpLogger() {
```
royroyee: This log uses console writer, outputting only to console. Consider following existing logger setup, adjusting for dev and production environments. 
Also, using `log.Logger` as a global variable could lead to clutter.

cmd/alpamon/command/ftp.go
```diff
@@ -0,0 +1,32 @@
+package command
+
+import (
+	"github.com/alpacanetworks/alpamon-go/pkg/config"
+	"github.com/alpacanetworks/alpamon-go/pkg/logger"
+	"github.com/alpacanetworks/alpamon-go/pkg/runner"
+	"github.com/spf13/cobra"
+)
+
+var ftpCmd = &cobra.Command{
```
royroyee, line 10: `RunE` is used without returning errors. Either add error handling at each step or switch to `Run`.

pkg/runner/ftp.go
```diff
@@ -401,15 +369,85 @@ func (fc *FtpClient) cpDir(src, dst string) (CommandResult, error) {
...
+	if err != nil {
+		return err
+	}
+	defer dstFile.Close()
+
+	if _, err = io.Copy(dstFile, srcFile); err != nil {
+		return err
+	}
+
+	if err = dstFile.Close(); err != nil {
```
royroyee: Redundant `dstFile.Close()`

pkg/runner/ftp.go
```diff
@@ -401,15 +369,85 @@ func (fc *FtpClient) cpDir(src, dst string) (CommandResult, error) {
...
+
+	if err = os.Chmod(dst, srcInfo.Mode()); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func copyDir(src, dst string) error {
+	if strings.HasPrefix(dst, src) {
```
royroyee, line 420: Should the two paths (src and dst) be aligned to the same format? They could differ as relative vs absolute paths – is this accounted for?

pkg/runner/command.go
```diff
@@ -153,8 +153,35 @@ func (cr *CommandRunner) handleInternalCmd() (int, string) {
...
+			data.HomeDirectory,
+		)
+		cmd.SysProcAttr = sysProcAttr
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		if err := cmd.Start(); err != nil {
+			log.Debug().Err(err).Msg("Failed to start worker process")
+
+			return 1, fmt.Sprintf("openftp: Failed to start worker process. %s", err.Error())
```
royroyee: Use `err` directly in `fmt.Sprintf` instead of `err.Error()`.

pkg/runner/command.go
```diff
@@ -153,8 +153,35 @@ func (cr *CommandRunner) handleInternalCmd() (int, string) {
...
+			executable,
+			"ftp",
+			data.URL,
+			data.HomeDirectory,
+		)
+		cmd.SysProcAttr = sysProcAttr
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		if err := cmd.Start(); err != nil {
```
royroyee: Please review if it's necessary to manage the child process with `Wait()` here.

pkg/config/config.go
```diff
@@ -18,6 +19,7 @@ var (
 	}
 
 	GlobalSettings Settings
+	FtpSettings    Settings
```
royroyee: Since ftp runs as a separate process, consider removing global variable for init.

royroyee (changes requested):

pkg/logger/logger.go
royroyee: Will different processes started with ftp log to the same file? Not sure if this is the best approach for concurrent logging.

pkg/runner/ftp.go
geunwoonoh: To support relative paths, `parsePath()` parses relative paths into absolute paths based on the home directory. 
`copyDir()` and other `Web FTP Commands` utilize these parsed paths.

geunwoonoh:
@royroyee I've addressed all the code review comments. Please review and resolve if everything looks good.

pkg/runner/command.go
geunwoonoh: The purpose of `openftp` is to create a child process, but adding wait() would prevent the command from finishing until the child process ends. This seems inappropriate because the `command fin request` to `alpacon-server` would be sent after the child process terminates.
Instead, I added `os.Exit(1)` to the FtpClient.close() to terminate child process upon completion of the FtpClient connection.

PR #20, opened 2024-12-26
Update README with test config values

ghost:
Matters not mentioned in the issue
- The configuration file for developers has been updated in the Alpacon Server GitLab repository, but it has not been updated in the Alpamon Agent GitHub repository.
- The PR adds instructions to the README.md to include the update of id and key values in the /etc/alpamon/alpamon.conf file for the connection between Alpamon agent and Alpacon server.
- Well, I'm not sure if these values are publicly accessible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,8 @@ It is recommended to use `/etc/alpamon/alpamon.conf` for deployment, but you can
 ```ini
 [server]
 url = http://localhost:8000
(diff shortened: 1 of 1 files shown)


royroyee:
The id and key are for testing purposes. It would be good to mention that they are included in the test dataset of `alpacon-server`.

PR #21, opened 2024-12-31
Resolve "Metric Collector"

### LINKED ISSUE #9: Metric Collector
To build a real-time monitoring system using AIP, we need to add the following features and components:

- Implement MetricCollector

    - Utilize the Strategy design pattern to collect data for different types of metrics.

    - Employ the Factory design pattern to create Collector objects based on specific metrics.

- Implement Push system

   - Periodically send metric data collected by MetricCollector to the alpacon-server at specified intervals.

   - Operate asynchronously and in parallel.
   
- alpamon should collect the following data
   - CPU usage
   - RAM usage
   - disk usage
   - network traffic

geunwoonoh:
As mentioned in https://github.com/alpacanetworks/alpamon/issues/9, we have added the following packages to implement a real-time monitoring system for AIP:

- `Collector`
   - This package creates and manages the following subcomponents required for metric collection:
       - `Check`
       - `Scheduler`
       - `Transporter`

- `Check`
   - The package functions as an individual metric monitoring unit.
   - It is responsible for metric collection and data management.
   
- `Scheduler`
   - This package manages the execution timing and frequency of Checks.
   
- `Transporter`
   - This package is responsible for sending collected metrics to the `alpacon-server`.
  
Closes https://github.com/alpacanetworks/alpamon/issues/9


cmd/alpamon/command/root.go
```diff
@@ -61,6 +64,25 @@ func runAgent() {
 	// Commit
 	runner.CommitAsync(session, commissioned)
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// DB
+	client := db.InitDB(ctx)
```
royroyee: - The `collector` variable name matches the package name, which can cause confusion.
- `err :=` pattern directly for simplicity, especially in shorter code blocks
- The error handling goroutine appears unnecessary and may introduce complexity. Simplify error handling without spawning a separate goroutine.
- `context` might not be necessary here, as it’s not being effectively utilized. Remove it if not needed or refactor to use it meaningfully

pkg/collector/collector.go
```diff
@@ -0,0 +1,228 @@
...
+	err = json.Unmarshal(resp, &conf)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+	}
+
+	return conf, nil
+}
+
+func NewCollector(args collectorArgs) (*Collector, error) {
+	transporter, err := args.transportFactory.CreateTransporter(args.session)
```
royroyee: Many variables share names with their packages, causing confusion. Please use distinct names for clarity.

pkg/collector/scheduler/scheduler.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+	}
+}
+
+func (s *Scheduler) executeTask(ctx context.Context, task *ScheduledTask) {
+	defer func() {
+		task.nextRun = time.Now().Add(task.interval)
+	}()
+
+	for attempt := 0; attempt <= s.retryConf.MaxRetries; attempt++ {
```
royroyee: Please refer to the existing backoff and related implementations in Alpamon for improving readability and simplifying the retry logic.

pkg/db/client.go
```diff
@@ -0,0 +1,35 @@
...
+		client, err = ent.Open("sqlite3", url)
+		if err != nil {
+			initErr = err
+			client = nil
+		}
+	})
+	return client, initErr
+}
+
+func Close() error {
```
royroyee: This function is currently unused.

pkg/db/client.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/alpacanetworks/alpamon-go/pkg/db/ent"
+	_ "github.com/glebarez/go-sqlite"
+)
+
+var (
```
royroyee: Please avoid using global variables whenever possible.

pkg/db/migrate.go
```diff
@@ -0,0 +1,61 @@
...
+	workDir, err := atlasexec.NewWorkingDir(
+		atlasexec.WithMigrations(
+			migrationFS,
+		),
+	)
+	if err != nil {
+		log.Error().Err(err).Msgf("failed to open migration dir: %v", err)
+		return err
+	}
+	defer workDir.Close()
```
royroyee: Missing error handling in defer statement.

royroyee:
This monitoring update is extensive and adds many dependencies. It would be great to include test cases to ensure reliability and maintainability.

- refer `commit_test.go`

pkg/collector/collector.go
```diff
@@ -0,0 +1,228 @@
...
+		check, err := args.checkFactory.CreateCheck(&checkArgs)
+		if err != nil {
+			return err
+		}
+		c.scheduler.AddTask(check)
+	}
+	return nil
+}
+
+func (c *Collector) Start(ctx context.Context) error {
```
royroyee: The Start function always returns nil, but root.go is handling its error return. 

Consider revising the error handling logic.

pkg/collector/collector.go
```diff
@@ -0,0 +1,228 @@
...
+func (c *Collector) successQueueWorker(ctx context.Context) {
+	defer c.wg.Done()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-c.stopChan:
+			return
+		case metric := <-c.buffer.SuccessQueue:
```
royroyee: Nested select reduces readability; consider refactoring to simplify the logic

pkg/collector/transporter/transporter.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, statusCode, err := t.session.Post(url, data.Data, 10)
+	if err != nil {
+		return err
+	}
+
+	if statusCode > 300 {
```
royroyee: Use explicit status code checks for consistency

Our project already follows this approach in files like reporter.go and commit.go.

pkg/collector/scheduler/scheduler.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	"github.com/alpacanetworks/alpamon-go/pkg/collector/check/base"
+	"github.com/alpacanetworks/alpamon-go/pkg/utils"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	MAX_RETRIES     int           = 5
+	MAX_RETRY_TIMES time.Duration = 1 * time.Minute
```
royroyee: Type time.Duration can be omitted for simplicity.

pkg/collector/check/batch/hourly/cpu/cpu.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	buffer := c.GetBuffer()
+	buffer.SuccessQueue <- metric
+
+	return nil
+}
+
+func (c *Check) queryCPUUsage(ctx context.Context) (base.MetricData, error) {
+	queryset, err := c.getCPU(ctx)
```
royroyee: Use camel case for variable naming, and apply this change consistently throughout this pr (queryset, .. etc)

royroyee (commented):
There seem to be many redundant parts in the code. While some may be unavoidable, it would be great to minimize them or improve them over time. 

For code style, please refer to the existing implementation as much as possible for consistency. 

Since there are multiple contexts and channels in use, please test various cases and consider using pprof to check for potential goroutine leaks

cmd/alpamon/command/root.go
geunwoonoh: I have applied your feedback. Please review and resolve it:
- The context initialized in root.go has been modified to include a timeout within the InitDB() function. For detailed changes, please refer to the provided link.
   - https://github.com/alpacanetworks/alpamon/pull/21/commits/94ba7e5d42af1153056e09e15880df025f9aaf25

pkg/collector/collector.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/collector/scheduler/scheduler.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/db/client.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/db/client.go
geunwoonoh: I have added a DBClientManager to maintain the client using the singleton pattern and applied the feedback.

pkg/db/migrate.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/collector/collector.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/collector/collector.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/collector/transporter/transporter.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/collector/scheduler/scheduler.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

pkg/collector/check/batch/hourly/cpu/cpu.go
geunwoonoh: I have applied your feedback. Please review and resolve it.

cmd/alpamon/command/root.go
royroyee: The for range loop is blocking and will keep waiting until it receives an error from the channel.

In root.go, it would be better to keep the code concise and focused only on initialization logic.

Also, since other packages in root.go do not pass context, it might be better to declare and handle the context within the DB logic itself, if it is necessary, rather than declaring it in root.go.

pkg/collector/collector.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+
+			return nil
+		}
+	}
+
+	return fmt.Errorf("max retries exceeded for metric %s", metric.Type)
+}
+
+func (c *Collector) Stop() {
```
royroyee, line 223: Is this method currently in use? I couldn’t find where it is being called.

Also, when working with goroutines and channels, **panic can occur if a closed channel is accessed**. Since this PR involves multiple channels, please pay extra attention to avoid panic.

cmd/alpamon/command/root.go
geunwoonoh: The code has been modified to encapsulate context declaration within InitDB() function and to localize context declaration for the collector within its specific logic.

pkg/collector/collector.go
geunwoonoh: I've added a call to `defer metricCollector.Stop()` in `root.go`:
- https://github.com/alpacanetworks/alpamon/pull/21/commits/1cb6a8f1016246883a67124b3b55ab84813e87f7#diff-3c43b14aa7f9f7a6c37f5fb29c403060d61545082c4972bc928f29f6d9ca29a6R72
Additionally, i've refactored the code to use the "comma-ok" idiom for checking if a channel is closed when reading data:
- https://github.com/alpacanetworks/alpamon/pull/21/commits/1cb6a8f1016246883a67124b3b55ab84813e87f7#diff-82dc411a0ed86675e9efddaa43dadc4fc6a75ea0e672400429e8ea39424da44cR162
- https://github.com/alpacanetworks/alpamon/pull/21/commits/1cb6a8f1016246883a67124b3b55ab84813e87f7#diff-82dc411a0ed86675e9efddaa43dadc4fc6a75ea0e672400429e8ea39424da44cR186

PR #32, opened 2025-03-10
Refactor command package for external import and plugin support

royroyee:
Refactor command package for external import and plugin support

#### etc
- Fix first install detection by checking binary existence instead of arguments


geunwoonoh (approved):
LGTM 👍

PR #73, opened 2025-05-22
Resolve "Improve getDpkgPackage scanner robustness for large package blocks"

### LINKED ISSUE #71: Improve `getDpkgPackage` scanner robustness for large package blocks
## Description

The current `getDpkgPackage()` function uses `bufio.Scanner` to parse the dpkg status file (`/var/lib/dpkg/status`), scanning blocks of package metadata. However, `bufio.Scanner` in Go has a **default token size limit of 64KB**, and any single block exceeding that limit can cause a runtime error:

```
bufio.Scanner: token too long
```

This can happen on systems with:
- A large number of packages
- Packages with unusually large or verbose metadata
- Corrupted or malformed dpkg status files

## Problem

The function may fail silently or panic on larger systems or corner cases due to this size limit. This is especially problematic in environments with many packages or in edge cases where a package block may exceed 64KB.

## Proposed Solution

- Explicitly set a larger buffer size using `scanner.Buffer()`:

```go
scanner := bufio.NewScanner(fd)
scanner.Split(utils.ScanBlock)
buf := make([]byte, 0, 1024*1024)         // 1MB initial buffer
scanner.Buffer(buf, 1024*1024)            // set 1MB max token size
```

- Optionally validate `utils.ScanBlock` to ensure it does not allow overly large blocks to accumulate due to incorrect delimiter logic.

## Acceptance Criteria

- [ ] Avoids panic when parsing large package blocks
- [ ] Gracefully handles large or malformed entries
- [ ] Logs or surfaces a meaningful error if a block is too large
- [ ] Exclude system packages from the commit data if it failed to collect

## References

- [Go bufio.Scanner documentation](https://pkg.go.dev/bufio#Scanner)
- [Relevant issue about scanner limits](https://github.com/golang/go/issues/29725)

geunwoonoh:
As mentioned in https://github.com/alpacanetworks/alpamon/issues/71, improve getDpkgPackage() as follows:

- Avoids panic when parsing large package blocks
   - Explicitly set a larger buffer size using scanner.Buffer()

- Gracefully handles large or malformed entries
   - Add exception handling for packages that are malformed or failed to parse in the dpkg status file

- Add logs if a block is too large or file is malformed

- Exclude system packages from the commit data if it failed to collect

Closes https://github.com/alpacanetworks/alpamon/issues/71

### DIFF
--- a/pkg/runner/commit.go
+++ b/pkg/runner/commit.go
@@ -37,7 +37,8 @@ const (
 	passwdFilePath = "/etc/passwd"
 	groupFilePath  = "/etc/group"
 
-	dpkgDbPath = "/var/lib/dpkg/status"
+	dpkgDbPath     = "/var/lib/dpkg/status"
+	dpkgBufferSize = 1024 * 1024
 
 	IFF_UP          = 1 << 0 // Interface is up
 	IFF_LOOPBACK    = 1 << 3 // Loopback interface
@@ -578,6 +579,9 @@ func getDpkgPackage() ([]SystemPackageData, error) {
 	scanner := bufio.NewScanner(fd)
 	scanner.Split(utils.ScanBlock)
 
+	buf := make([]byte, 0, dpkgBufferSize)
+	scanner.Buffer(buf, dpkgBufferSize)
+
 	pkgNamePrefix := []byte("Package:")
 	for scanner.Scan() {
 		chunk := scanner.Bytes()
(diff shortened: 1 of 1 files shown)


pkg/runner/commit.go
```diff
@@ -608,10 +613,16 @@ func getDpkgPackage() ([]SystemPackageData, error) {
 			Arch:    header.Get("Architecture"),
 		}
 
+		if pkg.Name == "" || pkg.Version == "" {
+			log.Error().Msgf("Skip malformed package entry: %s", chunk)
```
royroyee: Would it be more appropriate to use debug level logging here, since this is a skipped case?

geunwoonoh: Change to debug level based on this review. Thanks!

PR #81, opened 2025-06-05
Add moduser command to handle updates in user and membership

### LINKED ISSUE #79: Add `moduser` and `modgroup` commands to handle updates in user and group details.
Description:
Add two new commands in alpamon to handle system synchronization triggered by modifications to user and group information:
- `moduser`: Handles system-level updates when a user information is modified.
- `modgroup`: Handles system-level updates when a group information is modified.

The server is synced after the server side update to reflect the latest configuration.

seowonnoh:
Description:
Add a new command in alpamon to handle system synchronization triggered by modifications to user information and membership updates:

`moduser`: Handles system-level updates when a user information is modified, or user-group membership is changed.
The server is synced after the server side update to reflect the latest configuration.

### DIFF
--- a/pkg/runner/command.go
+++ b/pkg/runner/command.go
@@ -563,19 +563,6 @@ func (cr *CommandRunner) modUser() (exitCode int, result string) {
 		if exitCode != 0 {
 			return exitCode, result
 		}
-	} else if utils.PlatformLike == "rhel" {
-		exitCode, result = runCmdWithOutput(
-			[]string{
-				"/usr/sbin/usermod",
-				"--comment", data.Comment,
-				"-G", utils.JoinUint64s(cr.data.Groups),
-				data.Username,
-			},
-			"root", "", nil, 60,
-		)
-		if exitCode != 0 {
-			return exitCode, result
-		}
 	} else {
 		return 1, "Not implemented 'moduser' command for this platform."
 	}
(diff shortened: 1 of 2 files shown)


pkg/runner/command.go
```diff
@@ -537,6 +539,51 @@ func (cr *CommandRunner) delGroup() (exitCode int, result string) {
...
 
+func (cr *CommandRunner) modUser() (exitCode int, result string) {
+	data := modUserData{
+		Username: cr.data.Username,
+		Comment:  cr.data.Comment,
+	}
+
+	err := cr.validateData(data)
+	if err != nil {
+		return 1, fmt.Sprintf("adduser: Not enough information. %s", err)
```
royroyee: It looks like there’s a small typo here — it should be moduser.

pkg/runner/command.go
```diff
@@ -537,6 +539,51 @@ func (cr *CommandRunner) delGroup() (exitCode int, result string) {
...
+	}
+
+	if utils.PlatformLike == "debian" {
+		exitCode, result = runCmdWithOutput(
+			[]string{
+				"usr/sbin/usermod",
```
royroyee: Since the usermod command works the same across both platforms, conditional branching may not be necessary. Also, please make sure to include the leading / in the path (/usr/...).

royroyee (changes requested):

PR #87, opened 2025-06-26
Resolve "Fix edge cases that occur when creating/deleting system users"

### LINKED ISSUE #86: Fix edge cases that occur when creating/deleting system users
# Description
We need to fix an edge case where a system user is deleted, and then a new system user is created with the exact same username.

geunwoonoh:
As mentioned in https://github.com/alpacanetworks/alpamon/issues/86, fix an edge case where a system user is deleted, 
and then a new system user is created with the exact same username as follows:

- Update `delUser` command
  - Add `PurgeHomeDirectory` to support system user purge.
  - If `PurgeHomeDirectory` is false, the home directory of the system user to be deleted will be moved to a backup directory.

### DIFF
--- a/pkg/runner/command.go
+++ b/pkg/runner/command.go
@@ -461,7 +461,8 @@ func (cr *CommandRunner) addGroup() (exitCode int, result string) {
 
 func (cr *CommandRunner) delUser() (exitCode int, result string) {
 	data := deleteUserData{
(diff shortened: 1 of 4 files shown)


eunyoung14 (approved):
Approved. This PR should be megerd after the corresponding update of alpacon-server get merged.

Issue #88, opened 2025-07-01
Fix compatibility between ent0.14.0 and Go 1.24.x


sangyeong01:
# Description
using ent`0.14.0` with Go `1.24.x` results in build/runtime errors unless ent is upgraded to at least `v0.14.2.`

However, our documentation currently mentions “Go `1.22` or `higher`” as the requirement, which can be misleading for developers using Go `1.24` with older ent versions.

# Proposed Solutions

- Fix by upgrading ent to `v0.14.2` or later in go.mod & Fix documentation 
go run -mod=mod entgo.io/ent/cmd/ent@`v0.14.0` generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
-> go run -mod=mod entgo.io/ent/cmd/ent@`v0.14.2` generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
- Update documentation to explicitly state that Go `1.22` is the supported version, unless ent is upgraded.

# References
- https://github.com/ent/ent/issues/4324 (official confirmation that it’s fixed in ent v0.14.2)
- https://github.com/ent/ent/issues/2155
- https://github.com/mikestefanello/pagoda/issues/98

PR #89, opened 2025-07-01
Fix ent incompatibility with Go 1.24 – update CLI usage and module versions

sangyeong01:
# Description
Related issue: #88 

As mentioned in #88 , ent fails to run properly under Go 1.24 unless upgraded to v0.14.2 or higher.

- Updated ent to v0.14.2 in go.mod
- Updated other related modules via go get, to satisfy ent dependencies
- Updated README.md 
go run -mod=mod entgo.io/ent/cmd/ent@v0.14.0 ... -> go run -mod=mod entgo.io/ent/cmd/ent@v0.14.2 ...


eunyoung14:
Could you revert commits and do them again after configuring the git `user.name` and `user.email` properly? The name should be "Sangyeong Park" and the email should be your work email.

sangyeong01:
After upgrading Ent to version v0.14.2 and updating dependencies via go get, the Go version in go.mod is automatically updated to 1.23.

Given that, I’ve considered three possible solutions:
1. The Generate Ent Schema Code with Entgo command in the README explicitly sets the Ent version via CLI, independently of the version declared in go.mod. So we could keep using ent v0.14.2 only in the command, without changing the go.mod entry. This seems like the cleanest approach.
2. Update the documentation from “Go version 1.22 or higher installed” to “Go version 1.22 or 1.23”.
3. Upgrade the project’s Go version to 1.23 entirely — though this might introduce a number of side effects, so I see this as the most risky option.

Please take a moment to review this when you have time.

README.md
```diff
@@ -40,7 +40,7 @@ git clone https://github.com/alpacanetworks/alpamon.git
 #### Generate Ent Schema Code with Entgo
 To generate Ent schema code with custom features, navigate to the root of the project and use the following command:
 ```bash
-go run -mod=mod entgo.io/ent/cmd/ent@v0.14.0 generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
+go run -mod=mod entgo.io/ent/cmd/ent@v0.14.2 generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
```
geunwoonoh, line 44: Please add the updated go.mod and go.sum files for the ent package v0.14.2 upgrade.

sangyeong01:
The program itself is compatible with Go 1.23+, but due to the toolchain directive, building requires Go 1.24.4 or higher — see [Go toolchain documentation](https://go.dev/doc/toolchain) for details.

geunwoonoh (approved):

PR #98, opened 2025-07-20
Log rotation

eunyoung14:
This pull request has two key updates.

- Log rotation: Log files are now automatically rotated based on file size and time limits.
- Reduced verbosity: In production, caller information will not be printed into logs.

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -111,19 +111,18 @@ func PrettyWriter(out io.Writer, showCaller bool) zerolog.ConsoleWriter {
 				return ""
 			}
 			callerStr := fmt.Sprint(i)
-			// Try to make path relative to "/alpamon/"
 			if idx := strings.Index(callerStr, "/alpamon/"); idx != -1 {
-				callerStr = callerStr[idx+len("/alpamon/"):] // remove up to and including /alpamon/
+				callerStr = callerStr[idx+len("/alpamon/"):]
 			}
 			return fmt.Sprintf("(%s)", callerStr)
 		}
 	} else {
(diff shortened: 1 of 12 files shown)


pkg/logger/logger.go
```diff
@@ -59,29 +60,34 @@ func InitLogger() *os.File {
...
 	if version.Version == "dev" {
-		output = zerolog.MultiLevelWriter(PrettyWriter(os.Stderr), recordWriter)
+		// In development, log to console with caller info
+		output = zerolog.MultiLevelWriter(PrettyWriter(os.Stderr, true), recordWriter)
+		log.Logger = zerolog.New(output).With().Timestamp().Caller().Logger()
 	} else {
-		output = zerolog.MultiLevelWriter(PrettyWriter(logFile), recordWriter)
+		// In production, log to file without caller info
+		output = zerolog.MultiLevelWriter(PrettyWriter(logRotate, false), recordWriter)
+		log.Logger = zerolog.New(output).With().Timestamp().Logger()
```
geunwoonoh: In the current implementation, when not in the dev environment, removing `.Caller()` causes `entry.Caller` to be empty, 
which makes `logRecordWriter.Write()` not work properly — the log is not sent to alpacon-server due to `entry.Caller == ""`.
If the goal is to hide the caller from the log file (but still include it in the structured log), then how about including `.Caller()` in `log.Logger`, and conditionally formatting it in `PrettyWriter` instead?

eunyoung14: Haven't checked the log report. Thanks for the comment!

geunwoonoh (approved):

PR #101, opened 2025-08-06
feat: implement MFA-based sudo approval system

### LINKED ISSUE #99: Sudo Privilege Verification and Centralized Management via MFA Authentication
### Description

When accessing an Alpamon server through Websh, users can log in either as `root` or as a regular user. However, when logged in as a regular user, **password input is blocked** due to Websh's security policy, making it impossible to escalate privileges using `sudo`.

To address this limitation, we propose introducing an **MFA-based `sudo` approval system** integrated with **Alpacon's centralized permission management**.

 This issue arises in the following scenarios:
 - A regular user needs to perform tasks requiring `root` privileges
 - Password-based PAM prompts are not available in the Websh environment
 - Centralized access control and enforcement is required for operational security

---

### Problem

- `sudo` privilege escalation is not possible for regular users logged in via Websh
- PAM requires password input, but Websh blocks this for security reasons
- Difficult to satisfy both **operational efficiency** and **security policy compliance**

---

### Proposed Solution

- Integrate with `pam_alpamon.so` to verify `sudo` permissions and trigger MFA requests
- Branch logic based on user role:

| User Type                          | Behavior                                       |
|-----------------------------------|------------------------------------------------|
| `superuser`, `staff`, `owner`, `manager` | Allow `sudo` after successful MFA verification |
| `member` (regular user)           | Forward `sudo` request to a higher-privilege user |

- Handle MFA requests and responses via **WebSocket communication with the Alpacon server**
- If MFA fails or times out, deny `sudo` and log the event

hyunwoo-alps:
## Summary
Implement centralized sudo privilege verification through MFA authentication system.

### 1. Auth Manager (Unix Socket Server)
- Creates Unix domain socket server at `/var/run/alpamon/auth.sock`
- Handles `check_user` requests from `pam_alpamon.so`
- Handles `sudo_approval` requests from `alpacon_approval.so`
- Manages PID-to-session mapping for Alpacon users

### 2. MFA Request Flow
- Detects sudo command execution via PAM module
- Routes MFA requests to alpacon-server for authenticated users
- Returns `{success: false, reason: "session missing"}` for non-Alpacon users
- Delivers MFA responses back to PAM module

### 3. User Management
- Automatically adds new Alpacon users to sudo group on creation
- Enables sudo access for Alpacon-managed users

### 4. Development Improvements
- Added local test user in Dockerfile for convenient testing
- Integrated PAM module build tools for easier development

## Integration Points
- **alpacon-server**: MFA verification endpoint
- **pam_alpamon.so**: User authentication module
- **alpacon_approval.so**: Sudo approval plugin

## Related Issues
Closes #99


pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	log.Info().Msg("Auth Manager started successfully")
+
+	<-am.ctx.Done()
+	log.Info().Msg("Auth Manager stopped")
+}
+
+func (am *AuthManager) startSocketListener(ctx context.Context) error {
+	const socketPath = "/var/run/alpamon.sock"
```
eunyoung14: Please use a socket file under `/var/run/alpamon` directory, and make a proper name that fits with the purpose of this socket.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,353 @@
...
+
+	if err := os.MkdirAll("/var/run", 0755); err != nil {
+		return fmt.Errorf("failed to create socket directory: %w", err)
+	}
```
eunyoung14: Please use `tmpfile` which is managed by systemd to create the socket directory.

.goreleaser.yaml
```diff
@@ -67,12 +67,16 @@ nfpms:
           - sqlite3
           - nftables | iptables
         file_name_template: '{{ .PackageName }}_{{ trimprefix .Version "v" }}_{{ .Os }}_{{ .Arch }}'
+        recommends:
```
geunwoonoh: Using `recommends` causes `alpamon-pam` to be installed automatically when `alpamon` is installed. 

It seems better to use `suggests`.

pkg/runner/control_client.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		log.Debug().Err(err).Msg("Failed to write close message to control websocket.")
+	}
+
+	_ = cc.Conn.Close()
+	cc.Conn = nil
+}
+
+// WriteJSON sends JSON data through the WebSocket connection
+func (cc *ControlClient) WriteJSON(data interface{}) error {
```
geunwoonoh, line 165: Please specify the parameter type of this function as SudoApprovalRequest instead of an interface.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+	retryBackoff := backoff.NewExponentialBackOff()
+	retryBackoff.InitialInterval = authRetryInitialInterval
+	retryBackoff.MaxInterval = authRetryMaxInterval
+	retryBackoff.MaxElapsedTime = authRetryTimeout
+	retryBackoff.RandomizationFactor = 0
+
+	ctx, cancel := context.WithTimeout(am.ctx, authRetryTimeout)
+	defer cancel()
+
+	operation := func() error {
```
geunwoonoh: How about separating `operation` into distinct functions for code readability? 
For example, like `sendSudoRequest()`.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+			return nil
+		}
+	}
+
+	return backoff.Retry(operation, backoff.WithContext(retryBackoff, ctx))
+}
+
+func (am *AuthManager) handleSudoRequest(unix_conn net.Conn) {
+	buf := make([]byte, 1024)
+	n, err := unix_conn.Read(buf)
```
geunwoonoh: Go convention recommends adhering to camel case for variable names. 
Please make the necessary corrections.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+	if err != nil {
+		log.Warn().Err(err).Msg("Failed to read sudo request")
+		am.sendIsAlpconResponse(unix_conn, "", "", 0, 0, false)
+		return
+	}
+
+	var requestData map[string]interface{}
```
geunwoonoh: It would be better to define a dedicated struct for `requestData` to enhance type safety.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+	if err := json.Unmarshal(buf[:n], &requestData); err != nil {
+		log.Warn().Err(err).Msg("Invalid JSON request")
+		unix_conn.Close()
+		return
+	}
+
+	requestType, ok := requestData["type"].(string)
```
geunwoonoh: If the above review is applied, that logic seems unnecessary.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+			unix_conn.Close()
+			return
+		}
+
+		log.Debug().Msgf("Session found for PID %d: %s", isAlpconReq.PPID, session.SessionID)
+		am.sendIsAlpconResponse(unix_conn, isAlpconReq.Username, isAlpconReq.Groupname, isAlpconReq.PID, isAlpconReq.PPID, true)
+		unix_conn.Close()
+
+	case "sudo_approval":
+		var sudo_approval_req SudoApprovalRequest
```
geunwoonoh: "Please change this to camel case as well, in accordance with Go convention.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+		if err := json.Unmarshal(buf[:n], &sudo_approval_req); err != nil {
+			log.Warn().Err(err).Msg("Invalid sudo_approval_request")
+			am.sendSudoApprovalResponse(unix_conn, sudo_approval_req, false, "Invalid sudo_approval_request")
+			unix_conn.Close()
+			return
+		}
+
+		am.mu.RLock()
+		session, exists := am.pidToSessionMap[sudo_approval_req.PPID]
+		am.mu.RUnlock()
```
geunwoonoh: There appears to be a possibility that the session could be removed between the release of the `RLock` and the acquisition of the regular `Lock`. 
It seems necessary to either protect the entire block with a single `Lock` or re-verify the session's existence within the `Lock`.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+			unix_conn.Close()
+			return
+		}
+
+		log.Debug().Msgf("sudo_approval request sent to WebSocket client, waiting for response...")
+
+		// Connection will be closed when response arrives or on timeout
+		// Do not close here as it's managed by HandleSudoApprovalResponse or cleanupTimeoutRequest
+		select {
+		case <-time.After(30 * time.Second):
```
geunwoonoh: If `HandleSudoApprovalResponse` is called before the timeout here, 
this goroutine will still face the issue of waiting for the full 30 seconds. 
Additionally, the `select` statement won't terminate upon receiving a normal response. 
To resolve this, a completion channel should be added to ensure the `select` statement exits immediately upon receiving the response.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+func (am *AuthManager) cleanupTimeoutRequest(requestID string, approved bool, reason string) {
+	am.mu.Lock()
+
+	// 1. Alpacon user
+	for _, session := range am.pidToSessionMap {
+		if req, exists := session.Requests[requestID]; exists {
+			delete(session.Requests, requestID)
+			am.mu.Unlock()
+			if req.Connection != nil {
+				am.sendSudoApprovalResponse(req.Connection, SudoApprovalRequest{RequestID: requestID}, approved, reason)
```
geunwoonoh, line 533: The Unix socket connection(`req.Connection`) remains open after calling `am.sendSudoApprovalResponse()`. 
Is this behavior intended?

pkg/runner/command.go
geunwoonoh, line 1: The changes in `pkg/runner/command.go` appear redundant with the `HandleMessage()` function in `pkg/runner/control\_client.go` and seem unnecessary.

pkg/runner/auth_manager.go
```diff
@@ -0,0 +1,501 @@
...
+	return nil
+}
+
+func (am *AuthManager) AddPIDSessionMapping(pid int, session *SessionInfo) {
+	am.mu.Lock()
+	am.pidToSessionMap[pid] = session
+	am.mu.Unlock()
+}
+
+func (am *AuthManager) RemovePIDSessionMapping(pid int) {
```
geunwoonoh, line 480: This function is currently unused. 
It should probably be called when the `PtyClient` disconnects. 
If there are no plans to use it, it would be advisable to delete it.

pkg/runner/control_client.go
junho226: Regarding changing `WriteJSON(data interface{})` to a specific struct:

Currently `ControlClient.WriteJSON` only sends `SudoApprovalRequest`, but `WebsocketClient.WriteJSON` also uses the same pattern (`interface{}`). This is also a common pattern in gorilla/websocket.

When ControlClient needs to send multiple message types in the future, adding type-specific methods would be more appropriate. For now, keeping `interface{}` for consistency.

pkg/runner/command.go
junho226: sudo_approval_response command is sent to command runner by regular websocket from alpacon's approval request model when the sudo request is approved by admin. This should be fixed to use control channel but not for this PR so I tagged it as TODO.

geunwoonoh (dismissed):
I think we can merge this PR as soon as the conflicts are resolved. Thank you!

geunwoonoh (approved):

PR #104, opened 2025-08-08
Modify URL pattern in pty client

### LINKED ISSUE #102: Modify URL pattern for pty client
The pty client needs to be updated to match the new URL pattern that the `alpacon-server` will be sending.

mingyu-00:
Since the URL provided by the `alpacon-server` will change, we need to update the `alpamon` code accordingly.

### DIFF
--- a/pkg/runner/pty.go
+++ b/pkg/runner/pty.go
@@ -59,10 +59,22 @@ func NewPtyClient(data CommandData, apiSession *scheduler.Session) *PtyClient {
 		"Authorization": {fmt.Sprintf(`id="%s", key="%s"`, config.GlobalSettings.ID, config.GlobalSettings.Key)},
 		"Origin":        {config.GlobalSettings.ServerURL},
 	}
+
+	// For local development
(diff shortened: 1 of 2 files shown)


pkg/runner/pty.go
```diff
@@ -59,10 +59,22 @@ func NewPtyClient(data CommandData, apiSession *scheduler.Session) *PtyClient {
...
 		"Origin":        {config.GlobalSettings.ServerURL},
 	}
+
+	var finalURL string
+	if strings.HasPrefix(data.URL, "ws:") || strings.HasPrefix(data.URL, "wss:") {
+		finalURL = data.URL
+	} else {
+		// For local development
+		wsURL := strings.Replace(config.GlobalSettings.ServerURL, "http", "ws", 1)
+		wsURL = strings.Replace(wsURL, ":8000", ":8080", 1)
```
eunyoung14: I think this conversion won't be needed anymore for development purposes.

eunyoung14 (approved):

PR #106, opened 2025-08-14
Resolve "Resolve the issue that ~/ path is not matching to the user's home directory"

### LINKED ISSUE #105: Resolve the issue that ~/ path is not matching to the user's home directory
The application is not correctly handling the tilde character (`~`) as a shorthand for the current user's home directory. This is causing file paths starting with `~/` to be unrecognized.

geunwoonoh:
As mentioned in https://github.com/alpacax/alpamon/issues/105,
update to support relative paths for file downloads.


mingyu-00 (approved):
Thank you!

PR #152, opened 2025-12-12
Fix wildcard (*) glob expansion in shell command execution

### LINKED ISSUE #151: Fix wildcard (*) glob expansion in shell command execution
## Problem

When executing shell commands with wildcard patterns like `rm -rf ./hello/*`, the glob expansion does not work because `exec.Command` executes commands directly without shell interpretation.

### Current Behavior
```go
cmd := exec.CommandContext(ctx, args[0], args[1:]...)
// exec.Command("rm", "-rf", "./hello/*")
// The * is passed as a literal string, not expanded
```

**Result**: `rm` tries to find a file literally named `./hello/*` → "No such file or directory"

### Expected Behavior
Commands with `*` wildcards should have glob expansion applied, so `rm -rf ./hello/*` deletes all files inside the `hello` directory.

## Solution

When arguments contain `*` wildcard, execute via shell to enable glob expansion:

```go
if needsGlobExpansion(args) {
    shellCmd := strings.Join(quotedArgs, " ")
    cmd = exec.CommandContext(ctx, "sh", "-c", shellCmd)
} else {
    cmd = exec.CommandContext(ctx, args[0], args[1:]...)
}
```

### Security Considerations

To prevent command injection when using shell execution:
1. Only detect `*` character (not other shell special characters like `;`, `|`, `&`, etc.)
2. Quote all arguments using single quotes, escaping embedded single quotes
3. Preserve `*` unquoted for glob expansion while quoting the rest

**Example transformation**:
| Input | Quoted Output |
|-------|---------------|
| `rm -rf ./hello/*` | `'rm' '-rf' './hello/'*` |
| `ls *.txt` | `'ls' *'.txt'` |
| `ls ; rm -rf /` | `'ls ; rm -rf /'` (injection prevented) |

## Changes

- `pkg/runner/shell.go`:
  - Modify `runCmdWithOutput` to detect `*` and execute via shell
  - Add `needsGlobExpansion` function to check for `*` in arguments
  - Add `quoteArgPreservingGlob` function for safe quoting while preserving glob

mingyu-00:
## Summary

Fix shell command execution to properly handle wildcard (`*`) glob patterns like `rm -rf ./hello/*`.

## Problem

`exec.Command` executes commands directly without shell interpretation, so `*` is passed as a literal string instead of being expanded.

```go
// Before: * not expanded
cmd := exec.CommandContext(ctx, "rm", "-rf", "./hello/*")
// Result: tries to find file literally named "./hello/*"
```

## Solution

When arguments contain `*`, execute via shell with proper quoting to enable glob expansion while preventing command injection:

```go
if needsGlobExpansion(args) {
    // Quote args, preserving * for glob
    cmd = exec.CommandContext(ctx, "sh", "-c", shellCmd)
}
```

### Security

- Only `*` triggers shell execution (not `;`, `|`, `&`, etc.)
- All arguments are quoted with single quotes
- `*` is preserved unquoted for glob expansion

| Input | Quoted Output |
|-------|---------------|
| `rm -rf ./hello/*` | `'rm' '-rf' './hello/'*` |
| `ls ; rm -rf /` | `'ls ; rm -rf /'` (injection prevented) |

## Changes

- `pkg/runner/shell.go`:
  - Modify `runCmdWithOutput` to detect `*` and execute via shell
  - Add `needsGlobExpansion` function
  - Add `quoteArgPreservingGlob` function

## Test Plan

- [x] `rm -rf ./test/*` removes all files in test directory
- [x] `ls *.txt` lists all txt files
- [x] `echo hello` works without shell (no `*`)

Closes #151

### DIFF
--- a/pkg/runner/shell.go
+++ b/pkg/runner/shell.go
@@ -173,9 +173,9 @@ func runCmdWithOutput(args []string, username, groupname string, env map[string]
 	defer cancel()
 
 	// Expand glob patterns (*) in arguments using filepath.Glob
-	expandedArgs := expandGlobArgs(args)
+	args = expandGlobArgs(args)
 
-	cmd := exec.CommandContext(ctx, expandedArgs[0], expandedArgs[1:]...)
+	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
 	if username != "root" {
 		sysProcAttr, err := demote(username, groupname)
 		if err != nil {
--- a/pkg/runner/shell.go
+++ b/pkg/runner/shell.go
@@ -172,10 +172,15 @@ func runCmdWithOutput(args []string, username, groupname string, env map[string]
 	}
 	defer cancel()
 
-	// Expand glob patterns (*) in arguments using filepath.Glob
-	args = expandGlobArgs(args)
+	// Check if args is empty
+	if len(args) == 0 {
+		return 1, "no command provided"
+	}
 
-	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
+	// Expand glob patterns (*) in arguments using filepath.Glob
+	expandedArgs := expandGlobArgs(args[1:])
+	cmd := exec.CommandContext(ctx, args[0], expandedArgs...)
+
 	if username != "root" {
 		sysProcAttr, err := demote(username, groupname)
 		if err != nil {
--- a/pkg/runner/shell.go
+++ b/pkg/runner/shell.go
@@ -171,7 +171,19 @@ func runCmdWithOutput(args []string, username, groupname string, env map[string]
 	}
 	defer cancel()
 
-	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
+	// Check if command contains glob wildcard (*) that requires shell expansion
+	var cmd *exec.Cmd
+	if needsGlobExpansion(args) {
+		// Quote each argument to prevent command injection, preserving * for glob expansion
+		quotedArgs := make([]string, len(args))
+		for i, arg := range args {
+			quotedArgs[i] = quoteArgPreservingGlob(arg)
+		}
+		shellCmd := strings.Join(quotedArgs, " ")
+		cmd = exec.CommandContext(ctx, "sh", "-c", shellCmd)
+	} else {
+		cmd = exec.CommandContext(ctx, args[0], args[1:]...)
+	}
 	if username != "root" {
 		sysProcAttr, err := demote(username, groupname)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


mingyu-00:
This PR will resolve https://github.com/alpacax/alpacon-cli/issues/63.

pkg/runner/shell.go
```diff
@@ -171,7 +172,10 @@
 	}
 	defer cancel()
 
-	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
+	// Expand glob patterns (*) in arguments using filepath.Glob
+	expandedArgs := expandGlobArgs(args)
+
+	cmd := exec.CommandContext(ctx, expandedArgs[0], expandedArgs[1:]...)
```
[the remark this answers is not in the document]
mingyu-00: @geunwoonoh I think this logic was exists before, i don't know why the security bot alerts now. Can you check this logic?

pkg/runner/shell.go
```diff
@@ -171,6 +172,9 @@ func runCmdWithOutput(args []string, username, groupname string, env map[string]
 	}
 	defer cancel()
 
+	// Expand glob patterns (*) in arguments using filepath.Glob
+	args = expandGlobArgs(args)
```
geunwoonoh: Since args[0] is the command to be executed, an unintended binary could be executed if it contains wildcards.
Therefore, it seems necessary to keep args[0] as is and expand from args[1] onwards, as shown below.
```go
expandedArgs := expandGlobArgs(args[1:])
cmd := exec.CommandContext(ctx, args[0], expandedArgs...)
```

pkg/runner/shell.go
```diff
@@ -204,3 +208,26 @@ func runCmdWithOutput(args []string, username, groupname string, env map[string]
...
 	return 0, string(output)
 }
+
+// expandGlobArgs expands glob patterns (*) in arguments using filepath.Glob.
+// This is safer than using shell execution as it doesn't allow command injection.
+func expandGlobArgs(args []string) []string {
+	var expandedArgs []string
+
+	for _, arg := range args {
+		if strings.Contains(arg, "*") {
```
geunwoonoh: If Glob support is being added with this PR, it would be beneficial to also support other glob patterns like `?` and  `[...]`

pkg/runner/shell.go
```diff
@@ -171,6 +172,9 @@ func runCmdWithOutput(args []string, username, groupname string, env map[string]
 	}
 	defer cancel()
 
+	// Expand glob patterns (*) in arguments using filepath.Glob
+	args = expandGlobArgs(args)
+
 	cmd := exec.CommandContext(ctx, args[0], args[1:]...)
```
geunwoonoh: A panic could occur at args[0] if args is an empty slice. Please add exception handling for this case.

pkg/runner/shell.go
geunwoonoh, line 1: Have relative paths been tested as well?
I ask because, based on this branch, line 198 sets the command execution directory to `cmd.Dir = usr.HomeDir`.
This suggests that the working directory where `filepath.Glob` is executed might differ from the actual working directory where the command is executed.
If the concern mentioned above is valid, it seems necessary to either change the working directory or convert the paths to absolute paths before performing glob expansion.

mingyu-00: Thank you for review! I found the wrong execution in relative path command, and i resolve it.

geunwoonoh (approved):

PR #156, opened 2025-12-19
feat: Support for Websh based TCP-tunneling

jisung-02:
## Summary 
- Implements TCP tunneling functionality between Alpacon console and local services. 
- Multiple TCP connections are multiplexed into a single WebSocket connection through WebSocket-based smux multiplexing, and tunnel workers run as the nobody user for enhanced security.

## Key Change
1. New Features
    - Added opentunnel / closetunnel commands - Support remote tunnel creation/termination from Alpacon console
    - Implemented WebSocket-based TCP stream multiplexing through smux library
    - Tunnel session management (activeTunnels map) and prevention of duplicate sessions
2. Security Enhancements
    - Run tunnel worker process as nobody user (privilege degradation)
    - Restrict connections to localhost only (allow 127.0.0.1 and localhost only)
         - Tunneling is only possible for services on machines where the agent is installed.
    - Prevent DoS attacks with stream metadata size limit (1KB)
    - Port range validation (1-65535)
3. Performance Optimization
    - Reduced memory allocation/GC overhead with sync.Pool-based buffer pooling
    - Optimized smux configuration (KeepAlive 10s, MaxFrameSize 32KB, MaxReceiveBuffer 4MB)
    - Improved connection stability with TCP KeepAlive and NoDelay settings


pkg/runner/tunnel_client.go
```diff
@@ -0,0 +1,257 @@
...
+	log.Info().Msgf("Connecting to tunnel server at %s...", tc.serverURL)
+
+	dialer := websocket.Dialer{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: !config.GlobalSettings.SSLVerify,
+		},
+		HandshakeTimeout: 30 * time.Second,
+	}
+
+	conn, _, err := dialer.Dial(tc.serverURL, tc.requestHeader)
```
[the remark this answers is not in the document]
jisung-02: @eunyoung14 
This warning is the same as the ptychannel configuration in websh, so it would be best to ignore it.

pkg/config/config.go
```diff
@@ -22,6 +23,17 @@ const (
 	MaxConnectInterval = 300 * time.Second
 )
 
+// GetSmuxConfig returns optimized smux configuration for tunnel connections.
+func GetSmuxConfig() *smux.Config {
```
geunwoonoh, line 32: It would be better to define the configuration values within `GetSmuxConfig()`, 
such as `cfg.KeepAliveInterval` and `cfg.KeepAliveTimeout`, as constants—similar to `MinConnectInterval` above.

pkg/runner/tunnel_client.go
```diff
@@ -0,0 +1,257 @@
...
+				return
+			}
+
+			go tc.handleStream(stream)
+		}
+	}
+}
+
+// streamMetadata contains the target port information sent by the server.
+type streamMetadata struct {
```
geunwoonoh: Please move the type and const definitions to the top of the file to improve readability and consistency.

pkg/runner/tunnel_client.go
```diff
@@ -0,0 +1,257 @@
...
+func CloseTunnel(sessionID string) error {
+	activeTunnelsMu.RLock()
+	tc, exists := activeTunnels[sessionID]
+	activeTunnelsMu.RUnlock()
+
+	if !exists {
+		return fmt.Errorf("tunnel session %s not found", sessionID)
+	}
+
+	tc.Close()
```
geunwoonoh, line 265: There seems to be a potential race condition between calling `tc.Close()` 
after `activeTunnelsMu.RUnlock()` and the `delete(activeTunnels, tc.sessionID)` logic
 in the `defer` block of `RunTunnelBackground()`. 
This could potentially lead to `tc` becoming `nil`.

pkg/runner/tunnel_client.go
```diff
@@ -0,0 +1,257 @@
...
+	return nil
+}
+
+// handleStreams accepts and processes incoming smux streams from the server.
+func (tc *TunnelClient) handleStreams() {
+	for {
+		select {
+		case <-tc.ctx.Done():
+			return
+		default:
```
geunwoonoh: If `AcceptStream()` does not block within the `default` case, it may lead to a CPU spin. 
To resolve this, I recommend handling context cancellation in a separate goroutine instead of using a default case, as shown below:

```go
func (tc *TunnelClient) handleStreams() {
    for {
        stream, err := tc.session.AcceptStream()
        if err != nil {
            select {
            case <-tc.ctx.Done():
                return
            default:
                log.Debug().Err(err).Msgf("Tunnel session %s closed.", tc.sessionID)
                return
            }
        }
        go tc.handleStream(stream)
    }
}
```

pkg/tunnel/conn.go
```diff
@@ -0,0 +1,62 @@
...
+	if n < len(msg) {
+		w.readBuffer = msg[n:]
+	}
+
+	return n, nil
+}
+
+// Write writes data to the WebSocket connection as a binary message.
+func (w *WebSocketConn) Write(b []byte) (int, error) {
+	err := w.conn.WriteMessage(websocket.BinaryMessage, b)
```
geunwoonoh, line 55: According to this [link](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency), `WriteMessage` in `gorilla/websocket` is not thread-safe for concurrent calls. 
However, when using `smux`, multiple streams may call `Write()` simultaneously. 
To address this, we should add a mutex to the `Write()`.

pkg/runner/tunnel_client.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		log.Debug().Err(err).Msgf("Failed to spawn tunnel worker for %s.", targetAddr)
+		return
+	}
+
+	defer func() {
+		stdinPipe.Close()
+		if cmd.Process != nil {
+			if err := cmd.Process.Kill(); err != nil {
+				log.Debug().Err(err).Msg("Failed to kill tunnel worker process.")
```
geunwoonoh: This log appears redundant. 
The system is reporting a 'Kill' error even for processes that have already finished, 
so we should probably suppress it.

pkg/runner/tunnel_client.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+		_ = cmd.Wait()
+	}()
+
+	log.Debug().Msgf("Tunnel worker spawned for %s.", targetAddr)
+
+	// Create combined reader: remaining buffered data + original stream
+	// This is needed because bufReader may have buffered data beyond metadata
+	remainingBuf := make([]byte, bufReader.Buffered())
+	_, _ = bufReader.Read(remainingBuf)
```
geunwoonoh: Is there a specific reason why the error from `bufReader.Read()` is being ignored? 
I'm asking because ignoring this error might lead to potential loss of buffered data.

pkg/config/config.go
jisung-02: I have applied the changes based on your review comments.

pkg/runner/tunnel_client.go
jisung-02: I have applied the changes based on your review comments.

pkg/runner/tunnel_client.go
```diff
@@ -0,0 +1,257 @@
...
+				return
+			}
+
+			go tc.handleStream(stream)
+		}
+	}
+}
+
+// maxMetadataSize is the maximum size of stream metadata to prevent DoS attacks.
+const maxMetadataSize = 1024
```
geunwoonoh: Please move this const definitions to the top of the file to improve readability and consistency too.

pkg/runner/tunnel_client.go
jisung-02: I optimized CPU usage by removing the select-default block and allowing AcceptStream() to block naturally until a new stream arrives, preventing unnecessary loop iterations.

pkg/runner/tunnel_client.go
jisung-02: I reduced log noise by filtering out 'os.ErrProcessDone' errors during the process kill operation, ensuring that redundant debug logs are not triggered for already terminated processes.

pkg/runner/tunnel_client.go
jisung-02: Great catch. I've updated the logic to use io.ReadFull to ensure all buffered data is fully captured without any loss. I also added proper error handling to maintain data integrity as suggested.

pkg/runner/tunnel_client.go
jisung-02: I've fixed this issue. I changed RLock to Lock, and now delete(activeTunnels, sessionID) is executed under Lock before calling Unlock and tc.Close(). This prevents the race condition between CloseTunnel() and the defer block in RunTunnelBackground(). Since we delete from activeTunnels first, duplicate deletion in defer is harmless, and Close() is safe to call multiple times.

pkg/tunnel/conn.go
jisung-02: I've fixed this issue. I added a writeMu sync.Mutex field to the WebSocketConn struct, and protected the WriteMessage call in Write() with writeMu.Lock() and defer writeMu.Unlock(). This ensures thread-safe operation even when multiple smux streams call Write() concurrently.

geunwoonoh (approved):

PR #189, opened 2026-01-28
feat(runner): send sudo approval requests via REST API

junho226:
## Summary
- Change sudo approval requests from WebSocket to REST API
- Update control message parsing logic for wrapped message format

## Changes
- `pkg/runner/auth_manager.go`:
  - Add HTTP session for REST API requests
  - Change `createSendOperation()` to use REST API instead of WebSocket
  - Endpoint: `POST /api/servers/servers/{id}/sudo-approval/`
- `pkg/runner/control_client.go`:
  - Add `ControlMessage` struct for wrapped message format
  - Update `HandleMessage()` to parse `{"query": "control", "data": {...}}` structure

## Message Flow
```
alpamon ──REST API──> alpacon-server
alpamon <──WebSocket── Go backhaul <──Redis control stream── alpacon-server
```

## Test plan
- [ ] Test sudo approval request via REST API
- [ ] Test control WebSocket response reception
- [ ] E2E sudo approval flow test

### DIFF
--- a/pkg/runner/auth_manager.go
+++ b/pkg/runner/auth_manager.go
@@ -10,6 +10,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/alpacax/alpamon/pkg/config"
+	"github.com/alpacax/alpamon/pkg/scheduler"
 	"github.com/cenkalti/backoff"
 	"github.com/rs/zerolog/log"
 )
@@ -93,6 +95,7 @@ type AuthManager struct {
 	listener           net.Listener
 	localSudoRequests  map[string]*SudoRequest
 	completionChannels map[string]chan struct{}
+	session            *scheduler.Session
 }
 
 const (
@@ -112,6 +115,7 @@ func GetAuthManager(controlClient *ControlClient) *AuthManager {
 			pidToSessionMap:    make(map[int]*SessionInfo),
 			localSudoRequests:  make(map[string]*SudoRequest),
 			completionChannels: make(map[string]chan struct{}),
+			session:            scheduler.InitSession(),
 		}
 	})
 
@@ -125,6 +129,10 @@ func GetAuthManager(controlClient *ControlClient) *AuthManager {
 
 	if authManager.completionChannels == nil {
 		authManager.completionChannels = make(map[string]chan struct{})
+	}
+
+	if authManager.session == nil {
+		authManager.session = scheduler.InitSession()
 	}
 
 	return authManager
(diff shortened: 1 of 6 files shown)


pkg/runner/auth_manager.go
```diff
@@ -211,17 +218,25 @@ func (am *AuthManager) createSendOperation(ctx context.Context, req SudoApproval
...
 		default:
-			if am.controlClient == nil || !am.controlClient.IsConnected() {
-				return fmt.Errorf("control WebSocket client not available")
+			if am.session == nil {
+				return fmt.Errorf("HTTP session not available")
 			}
 
-			if err := am.controlClient.WriteJSON(req); err != nil {
+			url := fmt.Sprintf("/api/websh/sessions/%s/sudo-approval/", req.SessionID)
+			_, statusCode, err := am.session.Post(url, req, 10)
```
geunwoonoh: The third parameter of `Post()` is of type `time.Duration`, so '10' is interpreted as 10 nanoseconds. 
I think it should be changed to `10 * time.Second`.

pkg/runner/auth_manager.go
```diff
@@ -112,6 +114,7 @@ func GetAuthManager(controlClient *ControlClient) *AuthManager {
 			pidToSessionMap:    make(map[int]*SessionInfo),
 			localSudoRequests:  make(map[string]*SudoRequest),
 			completionChannels: make(map[string]chan struct{}),
+			session:            scheduler.InitSession(),
```
geunwoonoh: `scheduler.InitSession()` is already being called in `cmd/alpamon/command/root.go`. 
Instead of calling it again in `GetAuthManager()`, it would be better to inject the `session` created in `root.go` into `GetAuthManager()`.

pkg/runner/auth_manager.go
```diff
@@ -211,17 +218,25 @@ func (am *AuthManager) createSendOperation(ctx context.Context, req SudoApproval
...
+			_, statusCode, err := am.session.Post(url, req, 10)
+			if err != nil {
 				nextInterval := retryBackoff.NextBackOff()
-				log.Warn().Err(err).Msgf("Failed to send sudo request, will retry in %ds", int(nextInterval.Seconds()))
+				log.Warn().Err(err).Msgf("Failed to send sudo request via REST API, will retry in %ds", int(nextInterval.Seconds()))
 				return err
 			}
 
-			log.Debug().Msg("Sudo request sent successfully")
+			if statusCode < 200 || statusCode >= 300 {
```
geunwoonoh, line 232: `createSendOperation` calls `retryBackoff.NextBackOff()` inside the operation function to log the next retry interval.
When this operation returns a non-nil error, `backoff.Retry` internally calls `retryBackoff.NextBackOff()` again to determine the actual sleep duration. This means each failed attempt advances the backoff state twice — once for logging and once for the actual sleep. Consequences:

- The logged interval does not match the actual sleep duration (actual sleep uses the next interval).
- MaxElapsedTime is reached roughly twice as fast as intended, reducing the effective number of retries.

pkg/runner/auth_manager.go
geunwoonoh, line 1: While this was fine with WebSockets, the transition to HTTP requests might cause an issue on line 318, 
where an empty string is assigned to `SessionID` for local users. 
For instance, if `SessionID` is empty, the resulting URL would be `/api/websh/sessions//sudo-approval/`,
which will likely cause the request to fail. 
Has this edge case been considered?

pkg/runner/auth_manager.go
junho226: Fixed. Changed to `10*time.Second`.

pkg/runner/auth_manager.go
junho226: Fixed. `GetAuthManager()` now accepts session as a parameter, and the session created in `root.go` is injected.

pkg/runner/auth_manager.go
junho226: Fixed. Removed `NextBackOff()` calls from logging and now just output "will retry" message. Also removed the unused `retryBackoff` parameter from `createSendOperation`.

pkg/runner/auth_manager.go
junho226: Fixed. For local users (is_alpacon_user=false), we now reject immediately with "No Authority" without sending a request to the server. Since the server always rejects local users anyway, this avoids unnecessary API calls.

geunwoonoh (approved):
LGTM 👍 Could you please backport the changes from this PR to the `release/v1.3.1` branch? 
For compatibility reasons, I believe these changes should also be included

PR #195, opened 2026-02-05
fix(tunnel,editor): optimize resource management with daemon architecture and admission control

### LINKED ISSUE #194: [Editor/Tunnel] Optimize system resource usage for editor/tunnel
## Background      
- Optimize the system resources consumed by code-server (editor) and tunneling features.
- The current architecture spawns a new process per stream, incurring significant memory overhead. Additionally, incomplete cleanup of goroutines and associated resources leads to unnecessary resource retention.

### Reduce memory footprint
  - Handle multiple streams through a single worker instead of spawning a process per stream
  - Cap Node.js (code-server) heap memory
  - Reuse buffers to reduce GC pressure

### Improve resource lifecycle management
  - Ensure complete cleanup of goroutines and related resources (pipes, streams)
  - Deliver signals at the process group level to prevent orphaned child processes
  - Make idle timeout configurable

### Strengthen stability
  - Limit concurrent streams to prevent unbounded resource growth
  - Cache repeated syscall results

### Improve network efficiency
  - Enable WebSocket compression
  - Differentiate smux settings by client type and tune KeepAlive intervals

## Constraints
  - Credential demotion for tunnel workers (running as nobody) must be preserved as a security requirement.

jisung-02:
## Summary

  - Replace per-stream worker subprocess model with persistent daemon architecture, eliminating subprocess spawn overhead
  - Add stream concurrency limits (64 per session, 256 global) to prevent resource exhaustion
  - Add CPU/memory threshold checks (90%) before tunnel session creation with fail-open policy
  - Use process group signaling for code-server to ensure child process cleanup
  - Add pre-start memory check (512MB) for code-server
  - Add configurable [editor] idle_timeout option (default: 60 minutes)

### Tunnel Daemon Architecture

  - tunnel-worker (per-stream subprocess) → tunnel-daemon (persistent subprocess per session)
  - Unix Domain Socket-based multiplexing: single process handles all streams
  - Buffer pooling via tunnel.CopyBuffered

### Resource Protection

  - CheckSystemResources(): Reject new tunnel sessions when CPU/memory exceeds 90%
  - checkEditorResources(): Reject code-server startup when available memory is below 512MB
  - Session-level and global stream semaphores for concurrency limiting
  - Warning logs when utilization reaches 75% of limits

### Process Management

  - syscall.Kill(-pid, SIGTERM/SIGKILL) for process group signaling
  - Setpgid: true on both darwin and linux

### Configuration

  - Add [editor] idle_timeout to alpamon.conf (in minutes, 0 = disabled)
Closes #194


pkg/runner/tunnel_daemon.go
```diff
@@ -0,0 +1,147 @@
...
+	}()
+
+	// TCP -> UDS
+	go func() {
+		_, err := tunnel.CopyBuffered(conn, tcpConn)
+		errChan <- err
+	}()
+
+	// Wait for one direction to complete
+	<-errChan
```
geunwoonoh, line 170: When relaying in one direction completes, the function returns and closes both `conn` and `tcpConn` via defer. However, the remaining relay goroutine may remain blocked in `CopyBuffered`. 
While the connection closure eventually triggers an error that terminates the goroutine, this same pattern is observed in `relayBidirectional` at `tunnel_client.go:461`. 
Although normal connection closures handle this, half-open states could potentially lead to goroutine leaks.

pkg/runner/tunnel_client.go
geunwoonoh, line 1: Currently, the function that sets `tc.targetPort` (`handleHealthCheck`) is different from the one that reads it (`resolveTargetPort`). 
Since there is no synchronization for the `tc.targetPort` field, a race condition may occur. 
I recommend protecting `tc.targetPort` using `atomic.Int32` or a `mutex` to prevent this.

pkg/runner/editor.go
```diff
@@ -250,15 +275,21 @@ func (m *CodeServerManager) StartAsync() {
...
 func (m *CodeServerManager) Start() error {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-
+	m.mu.RLock()
 	if m.started {
+		m.mu.RUnlock()
 		return nil
 	}
+	m.mu.RUnlock()
```
geunwoonoh: The `Start()` method checks `m.started` under an `RLock`, releases the lock, and then proceeds with the startup logic. 
This creates a **Time-of-Check-to-Time-of-Use (TOCTOU)** race window where multiple goroutines can pass the guard check simultaneously.

```
// Goroutine A                          // Goroutine B
m.mu.RLock()                            
if m.started { // false                 
    m.mu.RUnlock()                      
    return nil                          
}                                       
m.mu.RUnlock()                          m.mu.RLock()
                                        if m.started { // still false!
// ... proceeds to install/start        m.mu.RUnlock()
                                        // ... also proceeds to install/start
```
Since `RLock` allows concurrent readers, both goroutines read `m.started == false` at the same time. Neither blocks the other because `RLock` is shared. After both release the lock, they both proceed to:

1. Run `checkEditorResources()` twice
2. Attempt to install code-server concurrently (race on filesystem)
3. Call `findAvailablePort()` and get different ports
4. Start two separate code-server processes
5. Both set `m.started = true`, but only the last `m.cmd` assignment survives — the first process becomes an orphan with no reference to stop it

pkg/runner/tunnel_client.go
```diff
@@ -338,47 +439,148 @@ func (tc *TunnelClient) buildDataReader(bufReader *bufio.Reader, stream *smux.St
...
-		if err := tc.codeServerMgr.Stop(); err != nil {
-			log.Debug().Err(err).Msg("Failed to stop code-server.")
+// startTunnelDaemon starts a tunnel daemon subprocess for this session.
+// The daemon runs with demoted credentials and handles all stream relay via UDS.
+func (tc *TunnelClient) startTunnelDaemon() error {
+	if !IsValidSessionID(tc.sessionID) {
+		return fmt.Errorf("invalid session ID for tunnel daemon socket")
+	}
+
+	tc.daemonSocket = fmt.Sprintf("/tmp/alpamon-tunnel-%s.sock", tc.sessionID)
```
geunwoonoh: While the practical risk is low, there is currently no protection against symlink attacks. If a malicious user pre-creates `/tmp/alpamon-tunnel-<sessionID>.sock` as a symbolic link, `os.Remove` might delete an unintended file, or the socket could be created in an unauthorized location.

I suggest the following alternatives:

- Use `os.MkdirTemp`.

- Consider using a dedicated runtime directory for alpamon (e.g., `/run/alpamon/`).

- Apply the `O_NOFOLLOW` flag.

pkg/runner/tunnel_daemon.go
jisung-02: Fixed in both locations.                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                             
  Previously, the function relied on `defer` to close connections, which meant the                                                                                                                                                                                                                           
  second goroutine could remain blocked in `CopyBuffered` during half-open states
  until the deferred close executed on function return.

  Now both `tunnel_daemon.go` and `tunnel_client.go` explicitly close both ends
  immediately after the first direction completes, then wait for the second
  goroutine to exit:

    <-errChan          // first direction done
    conn.Close()       // unblock the other goroutine
    tcpConn.Close()
    <-errChan          // wait for second direction to exit

  This ensures no goroutine is left blocked regardless of half-open TCP states.
  Double-close is safe since the callers already defer Close().

pkg/runner/tunnel_client.go
jisung-02: Good catch. Fixed by changing targetPort from plain int to atomic.Int32.                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                             
  All access sites have been updated:                                                                                                                                                                                                                                                                        
  - Write: tc.targetPort.Store(int32(...)) in handleHealthCheck and NewTunnelClient
  - Read: tc.targetPort.Load() in resolveTargetPort and RunTunnelBackground

  Chose atomic.Int32 over mutex since this is a simple single-field read/write
  with no compound operations needed.

pkg/runner/editor.go
jisung-02: Fixed by replacing the RLock guard with sync.Once.                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                             
  The original Start() body has been moved to a private doStart() method,                                                                                                                                                                                                                                    
  and Start() now simply wraps it:

```
  func (m *CodeServerManager) Start() error {
      m.startOnce.Do(func() {
          m.startErr = m.doStart()
      })
      return m.startErr
  }
```
  sync.Once guarantees exactly one execution regardless of how many goroutines
  call Start() concurrently, eliminating the TOCTOU window entirely.
  No orphan processes, no duplicate installs, no port conflicts.

pkg/runner/tunnel_client.go
jisung-02: Addressed with a dedicated socket directory and Lstat validation.                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                             
  1. Dedicated directory: ensureTunnelSocketDir() creates /tmp/alpamon-tunnels/                                                                                                                                                                                                                              
     with 0700 permissions. On Linux, the directory is chowned to nobody                                                                                                                                                                                                                                     
     (the daemon's runtime user). This prevents other users from creating                                                                                                                                                                                                                                    
     symlinks or files inside the directory.                                                                                                                                                                                                                                                                 

  2. Lstat validation: safeRemoveSocket() verifies the file is not a symlink
     and is an actual socket before removal (defense-in-depth).

  This provides two layers of protection:
  - Directory-level: 0700 blocks unauthorized file creation
  - File-level: Lstat rejects symlinks before os.Remove

geunwoonoh (approved):

PR #200, opened 2026-02-13
fix(auth): fix timeout overflow in sudo approval request

junho226:
## Summary

Fixes the "context deadline exceeded" error that occurred immediately when sending sudo approval requests. The bug was caused by passing `10*time.Second` to `Post()` while `session.do()` multiplies the timeout by `time.Second` again, causing int64 overflow.

## Changes

- Changed `am.session.Post(url, req, 10*time.Second)` to `am.session.Post(url, req, 10)` to match other API calls in the codebase

## Root Cause

```go
// auth_manager.go:226 - was passing time.Duration
am.session.Post(url, req, 10*time.Second)  // 10,000,000,000 ns

// session.go:112 - multiplies by time.Second again
ctx, cancel := context.WithTimeout(req.Context(), timeout*time.Second)
// Result: 10,000,000,000 * 1,000,000,000 = OVERFLOW!
// → Negative value → Immediate deadline exceeded
```

## Testing

- [x] Verified sudo approval request now works correctly in local environment
- [x] Other API calls using the same pattern (integer timeout) work correctly

---
Generated with AlpacaX Claude Plugin


junho226:
Closing to recreate from main branch

PR #253, opened 2026-03-27
fix(sync): include sync hashes in commit payload

### LINKED ISSUE #251: fix: store sync hashes at commit and sync-check so hash protocol actually reduces API calls
### Summary

The hash-based sync protocol implemented in #250 does not reduce sync execution time in production. Hashes are never stored on the server because the only storage path (`X-Sync-Hash` header on write requests) is never triggered when data is already in sync — the common case for stable, commissioned servers.

### Motivation

1. **Hash protocol has no effect in production**: Sync command execution time remains at 4-5 seconds instead of the expected <1 second. Every `sync-check` returns all 9 categories as `sync_required` because the server has no stored hashes to compare against.

2. **New server installs are also affected**: `CommitSystemInfo()` sends all data via PUT without hashes. The first sync after commission finds data already matching (just committed), so no PATCH is issued and no hash is stored.

3. **Existing servers never accumulate hashes**: Commissioned servers with stable data produce no write requests during sync, so the `X-Sync-Hash` header is never sent. This creates a permanent loop where every sync is a full sync.

### Root cause

`compareData()` and `compareListData()` only attach the `X-Sync-Hash` header to PATCH/POST/DELETE requests. When local and remote data match (no diff), no write request is queued and the hash never reaches the server:

```go
// commit.go — compareData()
if updateData != nil {
    // X-Sync-Hash is here, but this only runs when data changed
    scheduler.Rqueue.PatchWithHeaders(url, data, 80, time.Time{}, h)
}
// Data matches → nothing sent → hash never stored
```

The `sync-check` endpoint only compares received hashes against stored hashes — it never stores them. `_update_sync_hash()` is only called from proc endpoint write handlers, not from `sync_check`.

### Design

Two fixes are needed to cover all cases:

#### Fix 1: Include hashes in commit (new servers)

Modify `CommitSystemInfo()` to compute and include per-category hashes in the commit payload. The server stores these hashes during commit processing, so the very first sync after commission finds matching hashes and skips all GETs.

**alpamon changes:**
- `pkg/runner/commit.go` — compute hashes via `collectSnapshot()` and include them in the commit payload

**alpacon-server changes:**
- `servers/api/views.py` or `servers/api/serializers.py` — extract and store hashes from the commit request

#### Fix 2: Store hashes at sync-check when changed (existing servers)

Modify the `sync-check` endpoint to store received hashes when they differ from stored values. This is a one-time DB write per existing server (on first sync-check after deployment), after which the hashes are in place and no further writes occur.

**alpacon-server changes:**

```python
# servers/api/views.py — sync_check action
sync_required = [
    key for key, hash_val in received_hashes.items()
    if stored_hashes.get(key) != hash_val
]

# Only write when hashes actually changed (not on every call)
if sync_required:
    server.sync_hashes = {**stored_hashes, **received_hashes}
    server.save(update_fields=['sync_hashes'])
```

DB write frequency:
- First sync-check for existing server with no stored hashes: 1 write (one-time)
- Subsequent syncs with no data changes: 0 writes (hashes match, condition skipped)
- Sync after data change: 1 write (hash updated)

#### Coverage matrix

| Scenario | Fix 1 (commit) | Fix 2 (sync-check) |
|----------|:-:|:-:|
| New server (effective from first sync) | Solved | From second sync |
| Existing commissioned server | Not solved | Solved |
| Data change followed by sync | Solved (existing X-Sync-Hash) | Solved |

### Scope

**Files changed:**

| File | Repo | Change |
|------|------|--------|
| `pkg/runner/commit.go` | alpamon | Include hashes in commit payload |
| `servers/api/views.py` | alpacon-server | Store hashes in `sync_check` when changed |
| `servers/api/serializers.py` | alpacon-server | Accept and store hashes from commit request |

**Out of scope:**
- Removing `X-Sync-Hash` from write requests (kept as secondary confirmation)
- Per-category mutex or parallel sync optimization

### Dependencies

| Dependency | Notes |
|------------|-------|
| #250 (hash-based sync protocol) | Already merged; this issue fixes the hash storage gap |
| Server `sync_hashes` field and migration | Already deployed |

### Test plan

- [ ] **New server**: install → commit → first sync returns `sync_required: []` → execution time <1s
- [ ] **Existing server (no stored hashes)**: first sync stores hashes (one-time) → second sync returns `sync_required: []` → execution time <1s
- [ ] **Data change**: modify data on agent (e.g., add user) → only changed category appears in `sync_required`
- [ ] **Agent restart**: server retains stored hashes → unchanged categories skip sync immediately
- [ ] **sync-check DB write frequency**: confirm no DB write occurs when all hashes match
- [ ] **Backward compatibility**: agent with hashes in commit against old server without hash storage → commit succeeds (extra field ignored)

geunwoonoh:
## Summary

- Include per-category sync hashes in the commit PUT payload so the server can store them at commission time
- Define `SyncHashes` type to unify `map[string]string` usage across `syncSnapshot`, `checkSyncHashes()`, and `commitData`

## Problem

The hash-based sync protocol (#250) does not reduce sync execution time in production. Hashes are only sent via `X-Sync-Hash` header on write requests (PATCH/POST/DELETE), but when data is already in sync (the common case for commissioned servers), no write occurs and the hash never reaches the server. Every `sync-check` returns all 9 categories as `sync_required`.

## Solution

Compute hashes during `collectData()` using the existing `ComputeHash()` method on each syncer, and include them in the `commitData` struct as `sync_hashes`. The server stores these hashes during commit processing, so the very first sync after commission finds matching hashes and skips all GETs.

This is one of two coordinated fixes:
- **This PR (alpamon)**: send hashes at commit (covers new servers from first sync)
- **Separate PR (alpacon-server)**: store hashes at sync-check (covers existing commissioned servers)

## Changes

| File | Change |
|------|--------|
| `pkg/runner/commit_types.go` | Define `SyncHashes` type, add field to `commitData` |
| `pkg/runner/commit.go` | Compute hashes in `collectData()` via `s.ComputeHash()` |
| `pkg/runner/syncer.go` | Use `SyncHashes` type in `syncSnapshot` and `checkSyncHashes()` |

## Backward compatibility

| Scenario | Behavior |
|----------|----------|
| New alpamon + old server | `sync_hashes` field ignored (unknown JSON field) |
| Old alpamon + new server | No `sync_hashes` in commit; server-side sync-check fix covers this |
| New alpamon + new server | Hashes stored at commit; first sync skips unchanged categories |

## Test plan

- [ ] `go build -v ./cmd/alpamon` passes
- [ ] `go test -v ./... -p 1` passes
- [ ] New server: install -> commit -> first sync returns `sync_required: []` -> execution time <1s
- [ ] Existing server: first sync stores hashes via sync-check (server-side fix) -> second sync returns `sync_required: []`
- [ ] Data change: modify data (e.g., add user) -> only changed category appears in `sync_required`
- [ ] Backward compatibility: agent with `sync_hashes` against old server -> commit succeeds

## Related

- Closes #251
- Depends on #250 (hash-based sync protocol, already merged)
- Companion server-side PR: alpacon-server sync-check hash storage

### DIFF
--- a/pkg/runner/commit.go
+++ b/pkg/runner/commit.go
@@ -247,7 +247,7 @@ func syncRequiredKeys(session *scheduler.Session, snap syncSnapshot) []string {
 // checkSyncHashes sends per-category data hashes to the server and returns the
 // list of categories that need syncing. On error (old server returning 404,
 // network failure, etc.), the caller should fall back to syncing all categories.
-func checkSyncHashes(session *scheduler.Session, hashes map[string]string) ([]string, error) {
+func checkSyncHashes(session *scheduler.Session, hashes SyncHashes) ([]string, error) {
 	payload := map[string]any{"hashes": hashes}
 
 	resp, statusCode, err := session.Post(syncCheckURL, payload, 10)
@@ -341,6 +341,12 @@ func collectData() *commitData {
 			continue
 		}
 		assignToCommitData(data, s.Key(), result)
+		if h := s.ComputeHash(result); h != "" {
+			if data.SyncHashes == nil {
+				data.SyncHashes = make(SyncHashes, len(syncers))
+			}
+			data.SyncHashes[s.Key()] = h
+		}
 	}
 
 	return data
--- a/pkg/runner/commit.go
+++ b/pkg/runner/commit.go
@@ -341,11 +341,11 @@ func collectData() *commitData {
 			continue
 		}
 		assignToCommitData(data, s.Key(), result)
-		if h := s.ComputeHash(result); h != "" {
+		if hash := s.ComputeHash(result); hash != "" {
 			if data.SyncHashes == nil {
 				data.SyncHashes = make(SyncHashes, len(syncers))
 			}
-			data.SyncHashes[s.Key()] = h
+			data.SyncHashes[s.Key()] = hash
 		}
 	}
 
--- a/pkg/runner/commit_test.go
+++ b/pkg/runner/commit_test.go
@@ -442,38 +442,35 @@ func TestTimeSyncerComputeHash(t *testing.T) {
 func TestCollectDataIncludesSyncHashes(t *testing.T) {
 	data := collectData()
 
-	// SyncHashes should contain hashes for all syncers whose Collect() succeeds.
-	assert.Greater(t, len(data.SyncHashes), 0,
-		"SyncHashes should contain at least one entry")
+	// Skip if no syncer can collect in this environment (e.g., constrained CI/container).
+	if len(data.SyncHashes) == 0 {
+		t.Skip("skipping: no syncer Collect() succeeded in this environment")
+	}
+
+	// Validate entries without re-running Collect() to avoid non-determinism.
+	validKeys := make(map[string]struct{}, len(syncers))
+	for _, s := range syncers {
+		validKeys[s.Key()] = struct{}{}
+	}
+
 	assert.LessOrEqual(t, len(data.SyncHashes), len(syncers),
 		"SyncHashes should not contain more entries than syncers")
 
-	for _, s := range syncers {
-		result, err := s.Collect()
-		if err != nil {
-			t.Logf("Collect returned error for %s: %v (skipping assertions)", s.Key(), err)
-			continue
-		}
-
-		key := s.Key()
-		hash, ok := data.SyncHashes[key]
-		assert.True(t, ok, "SyncHashes should contain entry for %s", key)
+	for key, hash := range data.SyncHashes {
+		_, ok := validKeys[key]
+		assert.True(t, ok, "SyncHashes key %s should be a known syncer key", key)
 		assert.NotEmpty(t, hash, "Hash for %s should not be empty", key)
 		assert.True(t, strings.HasPrefix(hash, "sha256:"),
 			"Hash for %s should have sha256: prefix, got %s", key, hash)
 		assert.Len(t, hash, 7+64,
 			"Hash for %s should be 71 chars (sha256: + 64 hex), got %d", key, len(hash))
-		assert.Equal(t, s.ComputeHash(result), hash,
-			"Hash for %s should match ComputeHash(result)", key)
 	}
 
-	// Verify sync_hashes is present in marshaled JSON when there are sync hashes.
-	if len(data.SyncHashes) > 0 {
-		jsonBytes, err := json.Marshal(data)
-		assert.NoError(t, err)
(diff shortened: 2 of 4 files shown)


pkg/runner/commit.go
```diff
@@ -341,6 +341,12 @@ func collectData() *commitData {
 			continue
 		}
 		assignToCommitData(data, s.Key(), result)
+		if h := s.ComputeHash(result); h != "" {
+			if data.SyncHashes == nil {
+				data.SyncHashes = make(SyncHashes, len(syncers))
```
junho226, line 346: Suggestion: `h` is a bit cryptic here. Since this is scoped to the `if` block it's not a big deal, but a more descriptive name like `hash` would improve readability at no cost.

```suggestion
		if hash := s.ComputeHash(result); hash != "" {
			if data.SyncHashes == nil {
				data.SyncHashes = make(SyncHashes, len(syncers))
			}
			data.SyncHashes[s.Key()] = hash
```

junho226 (dismissed):
## PR Review: #253 - fix(sync): include sync hashes in commit payload

### Summary

Fixes the hash-based sync protocol (#250) which wasn't reducing sync time in production. The root cause: hashes were only sent via `X-Sync-Hash` header on write requests, but when data is already in sync (common for commissioned servers), no write occurs and hashes never reach the server. This PR computes hashes during `collectData()` and includes them in the commit PUT payload as `sync_hashes`, so the server stores them at commission time and the first post-commission sync can skip unchanged categories.

### Critical Issues (0)

None.

### Warnings (0)

None.

### Suggestions (1)

See inline comment.

### Highlights

- Excellent backward compatibility design: `omitempty` ensures old servers silently ignore the new field, and the companion server-side fix covers old agents
- `SyncHashes` type alias properly unifies the `map[string]string` usage across `syncSnapshot`, `checkSyncHashes()`, and `commitData`
- Lazy initialization of `SyncHashes` map avoids allocation when no syncer succeeds
- Well-designed test: CI/container skip, hash format validation, JSON marshaling confirmation
- Minimal, surgical change: 4 files, 48 additions, 3 deletions

---

**Verdict**: APPROVED

**Checklist**:
- [x] Code follows Go conventions
- [x] Tests included and thorough
- [x] Backward compatible with old server versions

pkg/runner/commit.go
geunwoonoh: Renamed to `hash`. Agreed — the extra character costs nothing and makes the intent immediately clear.

junho226 (approved):

PR #270, opened 2026-04-06
chore: remove plan files and hardcode signing config

junho226:
## Summary
- Remove obsolete plan files (`DEVELOP.md`, `FIREWALL_COMMANDS.md`) that were committed as development references
- Remove `[signing]` section from INI config parsing — signing settings are now hardcoded constants in `pkg/config/`
- Always initialize signature verification (no conditional), ensuring fleet-wide consistency
- Mode transitions (monitor → enforce) are deployed via code release, not per-server config changes

## Test plan
- [x] `go build ./cmd/alpamon` passes
- [x] `go test -v ./pkg/config/` passes (6 tests)
- [ ] Verify alpamon starts correctly without `[signing]` in config file
- [ ] Verify existing `[signing]` section in config file is silently ignored


pkg/config/config.go
```diff
@@ -34,10 +34,12 @@ const (
...
-	// Signing configuration defaults
-	// mode: "monitor" (log warnings, always execute) or "enforce" (reject unsigned/invalid)
-	DefaultSigningMode    = "monitor"
-	DefaultKeyRefreshSecs = 3600 // 1 hour
+	// Command signature verification defaults.
+	// Hardcoded intentionally: mode transitions (monitor → enforce) should be
+	// deployed fleet-wide via code release, not per-server config changes.
+	DefaultAIServerURL    = "https://alpacon.ai"
+	DefaultSigningMode    = "monitor" // change to "enforce" when ready
+	DefaultKeyRefreshSecs = 3600     // 1 hour
```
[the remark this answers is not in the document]
junho226: These are compile-time constants, not user input. Invalid values would be caught during code review and testing, not at runtime. Adding runtime assertions for values that can never change at runtime would be dead code. Dismissing.

pkg/runner/client.go
```diff
@@ -61,20 +61,18 @@ func NewWebsocketClient(session *scheduler.Session, ctxManager *agent.ContextMan
...
-			utils.NewHTTPClient(),
-		)
-		log.Info().Str("mode", wc.signingMode).Msg("Command signature verification enabled.")
-	}
+	wc.keyManager = signing.NewKeyManager(
+		config.GlobalSettings.AIServerURL,
+		config.GlobalSettings.KeyRefreshSecs,
+		utils.NewHTTPClient(),
+	)
+	log.Info().Str("mode", wc.signingMode).Msg("Command signature verification enabled.")
```
geunwoonoh, line 75: With the addition of this logic, the nil guard at `signing.go:50` has become unnecessary dead code. 
It would be good to remove the `nil` guard, and also remove the corresponding test case `TestVerifyCommandSignature_NoKeyManager` in `signing_test.go:61-76`.

junho226: Fixed. Removed the nil guard and corresponding test since keyManager is now always initialized.

geunwoonoh (approved):

PR #282, opened 2026-04-22
feat(pam): track Command executions for deploy shell sudo authorization

junho226:
## Summary

Extends the in-memory PID tracker in `pkg/runner` so deploy shell Commands (aka alpacon-cli / MCP `command`) register their root pid alongside existing Websh PTY sessions. When the PAM module walks the ppid chain, a matching entry now sends either a `session_id` (websh) or a `command_id` (deploy shell) to alpacon-server, enabling the server to resolve the IAM user via `Command.requested_by` and apply WorkSession-bound `SudoPolicy` without a password prompt fallback.

See the server-side plan for the cross-repo contract.

## Changes

- `SessionInfo` grows `Kind` / `CommandID` / `Username` / `StartedAt` fields. Legacy entries without `Kind` are read as `websh` for backward compatibility.
- New `AuthManager` methods `AddPIDCommandMapping` / `RemovePIDCommandMapping` plus package-level `RegisterCommandPID` / `UnregisterCommandPID` helpers. The remove path guards against pid reuse by requiring a matching `command_id`.
- `SudoApprovalRequest` / `SudoApprovalResponse` gain `command_id` (`omitempty`). The request path chooses `POST /api/sudo/approval/` when `command_id` is set and `session_id` is empty, otherwise falls back to the existing `/api/websh/sessions/{id}/sudo-approval/` URL.
- `pkg/executor`: `CommandOptions.PIDHook` and `CommandExecutor.ExecWithHook` propagate the child's pid immediately after `cmd.Start()` so the tracker registration happens before the child can `exec sudo`. The shell handler registers / unregisters a tracker entry when `CommandArgs.CommandID` is set.
- `runner/command.go` plumbs `protocol.Command.ID` into `CommandArgs.CommandID` for both `internal` and `system` shells.
- `pty.go` sets `Kind: TrackerKindWebsh` explicitly on websh registrations (behavioural no-op — existing entries already resolve to websh via `effectiveKind`).

## Test plan

Automated (included):

- [x] `TestAddPIDCommandMapping_RegistersCommandKind` — Command entries populate `Kind`/`CommandID`/`Username`/`StartedAt`.
- [x] `TestAddPIDCommandMapping_IgnoresInvalidInput` — bogus pids / empty command ids are rejected.
- [x] `TestRemovePIDCommandMapping_RemovesMatchingEntry` — unregister on Command completion.
- [x] `TestRemovePIDCommandMapping_PIDReuseGuard` — mismatched command id does not drop unrelated entry.
- [x] `TestRemovePIDCommandMapping_LeavesWebshEntryAlone` — Command-style remove never touches a websh entry with the same pid.
- [x] `TestParallelCommands_TrackedIndependently` — three concurrent Commands each get their own entry.
- [x] `TestLegacyWebshEntry_ReadsAsWebsh` — entries without `Kind` default to websh on read.
- [x] `TestAddPIDSessionMapping_NormalisesWebshKind` — websh registrations clear `CommandID` defensively.
- [x] `TestRegisterCommandPID_NoopWithoutManager` / `TestRegisterCommandPID_RoundTrip` — package-level helpers.
- [x] `TestSudoApprovalRequest_JSONTagsOmitEmpty` — wire-format guardrail: deploy shell vs websh payloads.
- [x] `TestLookupPID_Missing`, `TestAddPIDCommandMapping_OverwritesStaleEntry`.
- [x] `TestShellHandler_CommandID_RegistersAndUnregistersPID` — shell handler uses the hook path and cleans up.
- [x] `TestShellHandler_NoCommandID_UsesPlainExec` — no regression for internal / non-Command shell invocations.

Manual (post-merge, requires server-side PR and alpamon-pam update):

- [ ] Deploy shell Command as an IAM user with a WorkSession-bound `allow_bypass_mfa=True` SudoPolicy succeeds without password prompt.
- [ ] Same user via Websh still hits the existing MFA-bypass path.
- [ ] Root-owned deploy shell path unchanged.

## Build / test

- `go build ./...` clean
- `go vet ./...` clean
- `go test ./... -p 1` : 630 passed / 73 packages

### DIFF
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -113,16 +113,9 @@ func (e *Executor) runCommand(cmd *exec.Cmd, pidHook func(pid int)) ([]byte, err
 		return buf.Bytes(), err
 	}
 
-	func() {
-		defer func() {
-			if r := recover(); r != nil {
-				log.Error().Interface("panic", r).Msg("PIDHook panicked; continuing")
-			}
-		}()
-		if cmd.Process != nil {
-			pidHook(cmd.Process.Pid)
-		}
-	}()
+	if cmd.Process != nil {
+		pidHook(cmd.Process.Pid)
+	}
 
 	err := cmd.Wait()
 	return buf.Bytes(), err
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -79,7 +79,7 @@ func (e *Executor) Execute(ctx context.Context, opts CommandOptions) (int, strin
 
 	// Execute command
 	start := time.Now()
-	output, err := cmd.CombinedOutput()
+	output, err := e.runCommand(cmd, opts.PIDHook)
 	exitCode := 0
 	if err != nil {
 		if ctx.Err() == context.DeadlineExceeded {
@@ -96,6 +96,38 @@ func (e *Executor) Execute(ctx context.Context, opts CommandOptions) (int, strin
 	return exitCode, string(output), err
 }
 
+// runCommand executes cmd and returns its combined stdout/stderr output.
+// When pidHook is non-nil, the command is started with cmd.Start() so the
+// child's pid can be reported before Wait blocks. When pidHook is nil,
+// the simpler cmd.CombinedOutput() path is used unchanged.
+func (e *Executor) runCommand(cmd *exec.Cmd, pidHook func(pid int)) ([]byte, error) {
+	if pidHook == nil {
+		return cmd.CombinedOutput()
+	}
+
+	var buf bytes.Buffer
+	cmd.Stdout = &buf
+	cmd.Stderr = &buf
+
+	if err := cmd.Start(); err != nil {
+		return buf.Bytes(), err
+	}
+
+	func() {
+		defer func() {
+			if r := recover(); r != nil {
+				log.Error().Interface("panic", r).Msg("PIDHook panicked; continuing")
+			}
+		}()
+		if cmd.Process != nil {
+			pidHook(cmd.Process.Pid)
+		}
+	}()
+
+	err := cmd.Wait()
+	return buf.Bytes(), err
+}
+
 // CommandOptions defines options for command execution
 type CommandOptions struct {
 	Args       []string          // Command and arguments
@@ -105,6 +137,18 @@ type CommandOptions struct {
 	WorkingDir string            // Working directory
 	Timeout    time.Duration     // Command timeout
 	Input      string            // Input to provide via stdin
+
+	// PIDHook, if non-nil, is invoked with the child's pid immediately
+	// after cmd.Start() returns successfully and before the command is
+	// waited on. It is used to register the root pid of a deploy shell
+	// Command with the PAM tracker so that sudo invoked inside the
+	// command can be attributed to the originating Command.ID.
+	//
+	// When PIDHook is set, Execute uses cmd.Start()/cmd.Wait() instead
+	// of cmd.CombinedOutput() so the pid is visible to the hook before
+	// the child can exec sudo. Any panic from the hook is recovered and
+	// logged without affecting command execution.
+	PIDHook func(pid int)
 }
 
 // substituteEnvVars replaces environment variables in arguments
@@ -217,5 +261,20 @@ func (e *Executor) Exec(ctx context.Context, args []string, username, groupname 
 		Groupname: groupname,
 		Env:       env,
 		Timeout:   timeout,
+	})
+}
+
+// ExecWithHook is like Exec but registers a PIDHook that receives the
+// child's pid immediately after cmd.Start() succeeds. This is used by
+// the shell handler to track the root pid of a deploy shell Command so
+// sudo calls made inside the command can be authorized via command_id.
+func (e *Executor) ExecWithHook(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration, pidHook func(pid int)) (int, string, error) {
+	return e.Execute(ctx, CommandOptions{
+		Args:      args,
+		Username:  username,
+		Groupname: groupname,
+		Env:       env,
+		Timeout:   timeout,
+		PIDHook:   pidHook,
 	})
 }
--- a/pkg/executor/handlers/common/interfaces.go
+++ b/pkg/executor/handlers/common/interfaces.go
@@ -42,6 +42,14 @@ type CommandExecutor interface {
 
 	// Exec executes a command with all options (user, group, env, timeout)
 	Exec(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration) (int, string, error)
+
+	// ExecWithHook is like Exec but invokes pidHook with the child's pid
+	// immediately after the process is started (before it is waited on).
+	// Callers use it to register the root pid with the PAM tracker so
+	// sudo calls issued by the child can be attributed to the originating
+	// deploy shell Command. pidHook may be nil, in which case this
+	// behaves identically to Exec.
+	ExecWithHook(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration, pidHook func(pid int)) (int, string, error)
 }
 
 // WSClient interface for WebSocket client operations
--- a/pkg/executor/handlers/common/testing.go
+++ b/pkg/executor/handlers/common/testing.go
@@ -82,6 +82,16 @@ func (m *MockCommandExecutor) Exec(ctx context.Context, args []string, username,
 	return m.lookupResult(args[0], args[1:]...)
 }
 
+// ExecWithHook mirrors Exec for mock purposes and invokes pidHook (when
+// non-nil) with a synthetic pid of 0 so handlers that rely on the hook
+// contract can be exercised without spawning real processes.
+func (m *MockCommandExecutor) ExecWithHook(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration, pidHook func(pid int)) (int, string, error) {
+	if pidHook != nil {
+		pidHook(0)
+	}
+	return m.Exec(ctx, args, username, groupname, env, timeout)
+}
+
 func (m *MockCommandExecutor) SetResult(command string, exitCode int, output string, err error) {
 	m.results[command] = CommandResult{ExitCode: exitCode, Output: output, Err: err}
 }
--- a/pkg/executor/handlers/common/testing.go
+++ b/pkg/executor/handlers/common/testing.go
@@ -40,6 +40,9 @@ type CommandResult struct {
 }
 
 func NewMockCommandExecutor(t *testing.T) *MockCommandExecutor {
+	// Reset the synthetic pid counter on cleanup so pids don't leak across
+	// tests and future tests can rely on a predictable starting point.
+	t.Cleanup(func() { mockSyntheticPID.Store(0) })
 	return &MockCommandExecutor{
 		t:        t,
 		commands: []ExecutedCommand{},
--- a/pkg/executor/handlers/common/testing.go
+++ b/pkg/executor/handlers/common/testing.go
@@ -3,9 +3,18 @@ package common
 import (
 	"context"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"time"
 )
+
+// mockSyntheticPID produces monotonically increasing positive pids for
+// MockCommandExecutor.ExecWithHook so that every invocation hands the
+// hook a distinct, positive value. The base is well above typical real
+// pids to make accidental collisions obvious in test output.
+var mockSyntheticPID atomic.Int64
+
+const mockSyntheticPIDBase = 900000
 
 // MockCommandExecutor is a mock implementation of CommandExecutor for testing.
 // It is the single source of truth for mocking in this package.
@@ -83,11 +92,14 @@ func (m *MockCommandExecutor) Exec(ctx context.Context, args []string, username,
 }
 
 // ExecWithHook mirrors Exec for mock purposes and invokes pidHook (when
-// non-nil) with a synthetic pid of 0 so handlers that rely on the hook
-// contract can be exercised without spawning real processes.
+// non-nil) with a distinct, positive synthetic pid so handlers that rely
+// on the hook contract can be exercised without spawning real processes.
+// Using a positive pid ensures the registration code paths in the PID
+// tracker (which ignores pid<=0) are actually exercised.
 func (m *MockCommandExecutor) ExecWithHook(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration, pidHook func(pid int)) (int, string, error) {
 	if pidHook != nil {
-		pidHook(0)
+		pid := int(mockSyntheticPIDBase + mockSyntheticPID.Add(1))
+		pidHook(pid)
 	}
 	return m.Exec(ctx, args, username, groupname, env, timeout)
 }
--- a/pkg/executor/handlers/shell/shell.go
+++ b/pkg/executor/handlers/shell/shell.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/alpacax/alpamon/pkg/executor/handlers/common"
+	"github.com/alpacax/alpamon/pkg/runner"
 	"github.com/alpacax/alpamon/pkg/utils"
 	"github.com/rs/zerolog/log"
 )
@@ -86,16 +87,16 @@ func (h *ShellHandler) handleShellCommand(ctx context.Context, args *common.Comm
 		} else {
 			cmdArgs = []string{"/bin/sh", "-c", command}
 		}
-		exitCode, result := h.executeCommand(ctx, cmdArgs, username, groupname, env, timeout)
+		exitCode, result := h.executeCommand(ctx, cmdArgs, username, groupname, env, timeout, args.CommandID)
 		return exitCode, result, nil
 	}
 
 	// Fallback: direct execution with manual operator parsing
-	return h.executeWithOperators(ctx, command, username, groupname, env, timeout)
+	return h.executeWithOperators(ctx, command, username, groupname, env, timeout, args.CommandID)
 }
 
 // executeWithOperators handles shell operators (&&, ||, ;)
-func (h *ShellHandler) executeWithOperators(ctx context.Context, command, username, groupname string, env map[string]string, timeout time.Duration) (int, string, error) {
+func (h *ShellHandler) executeWithOperators(ctx context.Context, command, username, groupname string, env map[string]string, timeout time.Duration, commandID string) (int, string, error) {
 	spl := strings.Fields(command)
 	var currentCmd []string
 	var results strings.Builder
@@ -107,7 +108,7 @@ func (h *ShellHandler) executeWithOperators(ctx context.Context, command, userna
 		case "&&":
 			// Execute current command
 			if len(currentCmd) > 0 {
-				exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout)
+				exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout, commandID)
 				results.WriteString(result)
 				// Stop if command fails
 				if exitCode != 0 {
@@ -118,7 +119,7 @@ func (h *ShellHandler) executeWithOperators(ctx context.Context, command, userna
 		case "||":
 			// Execute current command
 			if len(currentCmd) > 0 {
-				exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout)
+				exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout, commandID)
 				results.WriteString(result)
 				// Continue only if command fails
 				if exitCode == 0 {
@@ -129,7 +130,7 @@ func (h *ShellHandler) executeWithOperators(ctx context.Context, command, userna
 		case ";":
 			// Execute current command
 			if len(currentCmd) > 0 {
-				exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout)
+				exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout, commandID)
 				results.WriteString(result)
 				// Continue regardless of result
 				currentCmd = nil
@@ -141,21 +142,46 @@ func (h *ShellHandler) executeWithOperators(ctx context.Context, command, userna
 
 	// Execute any remaining command
 	if len(currentCmd) > 0 {
-		exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout)
+		exitCode, result = h.executeCommand(ctx, currentCmd, username, groupname, env, timeout, commandID)
 		results.WriteString(result)
 	}
 
 	return exitCode, results.String(), nil
 }
 
-// executeCommand executes a single command
-func (h *ShellHandler) executeCommand(ctx context.Context, cmdArgs []string, username, groupname string, env map[string]string, timeout time.Duration) (int, string) {
+// executeCommand executes a single command.
+// When commandID is non-empty, the child's root pid is registered with
+// the PAM tracker for the duration of the execution so sudo calls made
+// from inside the command can be authorized by command_id.
+func (h *ShellHandler) executeCommand(ctx context.Context, cmdArgs []string, username, groupname string, env map[string]string, timeout time.Duration, commandID string) (int, string) {
 	if len(cmdArgs) == 0 {
 		return 0, ""
 	}
 
-	// Execute command through the executor with full parameters (user, group, env, timeout)
-	exitCode, output, err := h.Executor.Exec(ctx, cmdArgs, username, groupname, env, timeout)
+	var (
+		exitCode int
+		output   string
+		err      error
+	)
+
+	if commandID != "" {
+		// Track the pid in the PAM tracker before the child can exec
+		// sudo; unregister after the command completes. Holds the pid
+		// in a closure so we only unregister the exact pid we tracked,
+		// guarding against pid reuse.
+		var trackedPID int
+		registered := false
+		pidHook := func(pid int) {
+			trackedPID = pid
+			registered = runner.RegisterCommandPID(pid, commandID, username)
+		}
+		exitCode, output, err = h.Executor.ExecWithHook(ctx, cmdArgs, username, groupname, env, timeout, pidHook)
+		if registered {
+			runner.UnregisterCommandPID(trackedPID, commandID)
+		}
+	} else {
+		exitCode, output, err = h.Executor.Exec(ctx, cmdArgs, username, groupname, env, timeout)
+	}
 
 	if err != nil && exitCode == -1 {
 		// Command execution error (not just non-zero exit)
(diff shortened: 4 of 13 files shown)


pkg/runner/auth_manager_testing.go
```diff
@@ -0,0 +1,27 @@
...
+// SwapAuthManagerForTest installs am as the package-level singleton for
+// the duration of t and restores the previous singleton on cleanup.
+// It returns am so callers can keep working with the installed instance.
+func SwapAuthManagerForTest(t *testing.T, am *AuthManager) *AuthManager {
+	t.Helper()
+	prev := authManager
+	authManager = am
+	t.Cleanup(func() { authManager = prev })
+	return am
+}
```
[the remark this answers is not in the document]
junho226: Fixed: moved the test-only glue out of the shipped runner package. `pkg/runner/auth_manager_testing.go` no longer imports "testing"; it only exposes two plain helpers (`NewEmptyAuthManager`, `SwapAuthManager`) that production code does not call. The `*testing.T` cleanup wrappers now live in a new `pkg/runner/runnertest` subpackage, which is only imported from `*_test.go` files so the testing dependency is no longer pulled into release binaries.

pkg/runner/auth_manager.go
```diff
@@ -527,11 +598,122 @@ func (am *AuthManager) RemovePIDSessionMapping(pid int) {
...
+// from a crash) by another kind of entry, the existing entry is
+// preserved.
+func (am *AuthManager) RemovePIDCommandMapping(pid int, commandID string) {
+	if pid <= 0 {
+		return
+	}
+	am.mu.Lock()
+	if existing, ok := am.pidToSessionMap[pid]; ok {
+		if existing.effectiveKind() == TrackerKindCommand &&
+			(commandID == "" || existing.CommandID == commandID) {
```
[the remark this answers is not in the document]
junho226: Fixed: `RemovePIDCommandMapping` now requires a non-empty commandID and rejects the call up-front (`pid <= 0 || commandID == ""`), then only deletes when `existing.CommandID == commandID`. This matches the PR description's pid-reuse guard. Updated `TestRemovePIDCommandMapping_LeavesWebshEntryAlone` to pass a concrete id so it still exercises the cross-kind defence.

pkg/runner/auth_manager.go
```diff
@@ -527,11 +598,122 @@ func (am *AuthManager) RemovePIDSessionMapping(pid int) {
...
+// Returning a boolean lets callers decide whether an Unregister is owed
+// even in setups where the AuthManager has not been wired up yet (tests,
+// early boot).
+func RegisterCommandPID(pid int, commandID, username string) bool {
+	if authManager == nil {
+		return false
+	}
+	authManager.AddPIDCommandMapping(pid, commandID, username)
+	return true
+}
```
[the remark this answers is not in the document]
junho226: Fixed: `RegisterCommandPID` now validates pid/commandID itself and returns `false` when the arguments would be rejected by `AddPIDCommandMapping`. The boolean therefore reflects whether a mapping was actually added, so callers only issue an `Unregister` when one is genuinely owed.

pkg/executor/handlers/common/testing.go
```diff
@@ -82,6 +82,16 @@ func (m *MockCommandExecutor) Exec(ctx context.Context, args []string, username,
...
 
+// ExecWithHook mirrors Exec for mock purposes and invokes pidHook (when
+// non-nil) with a synthetic pid of 0 so handlers that rely on the hook
+// contract can be exercised without spawning real processes.
+func (m *MockCommandExecutor) ExecWithHook(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration, pidHook func(pid int)) (int, string, error) {
+	if pidHook != nil {
+		pidHook(0)
+	}
+	return m.Exec(ctx, args, username, groupname, env, timeout)
+}
```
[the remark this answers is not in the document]
junho226: Fixed: `MockCommandExecutor.ExecWithHook` now hands the hook a distinct, positive synthetic pid (`900000 + atomic counter`) on every call. Because the tracker ignores `pid<=0`, the old value would have silently skipped the registration path; with this change the hook-driven Register/Unregister lifecycle is genuinely exercised by any test that uses the mock.

pkg/runner/auth_manager_tracker_test.go
```diff
@@ -0,0 +1,321 @@
...
+		Type:      "sudo_approval_request",
+		SessionID: "session-uuid",
+		PID:       1,
+		PPID:      2,
+		Username:  "alice",
+	}
+	if webshReq.CommandID != "" {
+		t.Errorf("websh request should leave CommandID empty, got %q", webshReq.CommandID)
+	}
+}
```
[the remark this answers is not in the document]
junho226: Fixed: `TestSudoApprovalRequest_JSONTagsOmitEmpty` now `json.Marshal`s the struct and asserts on the serialized keys. It checks (a) deploy shell payload contains `command_id` and omits `session_id`, (b) websh payload contains `session_id` and omits `command_id`, and (c) a request with neither id set omits both keys. A regression dropping `omitempty` on either field will now fail the test.

pkg/runner/auth_manager_testing.go
```diff
@@ -0,0 +1,25 @@
+package runner
+
+// NewEmptyAuthManager returns an AuthManager populated just enough to
+// exercise the PID tracker without starting the socket listener. It has
+// no dependency on the testing package, so it is safe to keep in the
+// shipped package; the actual test-only glue that swaps the singleton
+// lives in the pkg/runner/runnertest subpackage, which is only imported
+// from *_test.go files.
+func NewEmptyAuthManager() *AuthManager {
```
geunwoonoh, line 9: **Warning**: this file ships in production binaries — test-only API is exposed.

Because the filename is `auth_manager_testing.go` (not `_test.go`), it is compiled into the shipped binary, and `NewEmptyAuthManager` / `SwapAuthManager` become callable production APIs. The docstring says "direct use in production code is not supported," but the Go compiler can't enforce that — a misuse like `runner.SwapAuthManager(nil)` would silently replace the singleton and break every subsequent sudo request.

**Recommended fixes** (in order of increasing rigor):

1. Move into an `internal/` sub-package — the compiler blocks imports from outside `pkg/runner/…`, which is effectively a visibility firewall even though the symbols still ship:

   ```
   pkg/runner/internal/testhooks/testhooks.go
   ```

2. Gate with a build tag so the code is excluded from `go build` entirely:

   ```go
   //go:build test
   // +build test

   package runner
   ```

   (requires `-tags test` in the test invocation / CI).

Note: simply renaming to `auth_manager_test.go` would exclude these symbols from production, but then `runnertest/runnertest.go` (a regular non-`_test.go` file) would no longer see them, breaking `go build ./...`. The `internal/` sub-package is the simplest fix.

> Historical context: a previous Copilot iteration already pushed the `testing` import out of this file — this comment is about the residual production-API-surface concern.

pkg/runner/auth_manager.go
```diff
@@ -351,15 +391,28 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 
 	// Alpacon user: pidToSessionMap
 	sudoApprovalReq.IsAlpconUser = true
-	sudoApprovalReq.SessionID = session.SessionID
+	switch session.effectiveKind() {
```
geunwoonoh: **Warning**: the switch on `session.effectiveKind()` is non-exhaustive — a future new `Kind` will silently fall back to the websh path.

```go
switch session.effectiveKind() {
case TrackerKindCommand:
    sudoApprovalReq.SessionID = ""
    sudoApprovalReq.CommandID = session.CommandID
default:
    // websh (and legacy entries without Kind) keep the session-based path.
    sudoApprovalReq.SessionID = session.SessionID
    sudoApprovalReq.CommandID = ""
}
```

`effectiveKind()` already normalizes legacy (`Kind == ""`) entries to `TrackerKindWebsh`, so the `default` branch no longer needs to double as a legacy-absorber. If `TrackerKindTunnel` (or similar) is added later, it will be silently misattributed as websh.

**Recommendation**: make the switch exhaustive and reject unknown kinds explicitly.

```suggestion
	kind := session.effectiveKind()
	switch kind {
	case TrackerKindCommand:
		sudoApprovalReq.SessionID = ""
		sudoApprovalReq.CommandID = session.CommandID
	case TrackerKindWebsh:
		sudoApprovalReq.SessionID = session.SessionID
		sudoApprovalReq.CommandID = ""
	default:
		am.mu.Unlock()
		log.Warn().Str("kind", kind).Msg("Unknown tracker kind; rejecting sudo")
		am.sendSudoApprovalResponse(unixConn, sudoApprovalReq, false, "Unknown session kind")
		_ = unixConn.Close()
		return
	}
```

(This also addresses Warning 3 by caching `kind` once.)

pkg/runner/auth_manager.go
```diff
@@ -351,15 +391,28 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
...
 	session.Requests[sudoApprovalReq.RequestID] = &SudoRequest{
 		RequestID:  sudoApprovalReq.RequestID,
 		Connection: unixConn,
 	}
 	am.mu.Unlock()
 
-	log.Debug().Msgf("Alpacon user sudo request: %s for session %s", sudoApprovalReq.RequestID, session.SessionID)
+	log.Debug().
+		Str("request_id", sudoApprovalReq.RequestID).
+		Str("kind", session.effectiveKind()).
```
geunwoonoh: **Warning**: `session.effectiveKind()` is called twice in `handleSudoApprovalRequest` (here and in the switch above). The function itself is trivial (`nil` check + string compare), so performance is a non-issue — this is purely a readability note. Cache once into a `kind` local and reuse it in both the switch and the log builder. The Warning 2 suggestion block already incorporates this.

pkg/executor/handlers/shell/shell.go
```diff
@@ -141,21 +142,46 @@ func (h *ShellHandler) executeWithOperators(ctx context.Context, command, userna
...
+	)
+
+	if commandID != "" {
+		// Track the pid in the PAM tracker before the child can exec
+		// sudo; unregister after the command completes. Holds the pid
+		// in a closure so we only unregister the exact pid we tracked,
+		// guarding against pid reuse.
+		var trackedPID int
+		registered := false
+		pidHook := func(pid int) {
```
geunwoonoh, line 173: **Suggestion**: return a deferrable cleanup from `RegisterCommandPID` so the API is leak-proof by construction.

Right now the shell handler threads two closure variables (`trackedPID`, `registered`) across the hook and then reconstructs the unregister call. A caller that forgets to call `UnregisterCommandPID` leaks an entry.

```go
// in pkg/runner/auth_manager.go
func RegisterCommandPID(pid int, commandID, username string) (unregister func()) {
    if authManager == nil || pid <= 0 || commandID == "" {
        return func() {} // no-op
    }
    authManager.AddPIDCommandMapping(pid, commandID, username)
    return func() { authManager.RemovePIDCommandMapping(pid, commandID) }
}
```

Caller:

```go
var cleanup func()
pidHook := func(pid int) {
    cleanup = runner.RegisterCommandPID(pid, commandID, username)
}
exitCode, output, err = h.Executor.ExecWithHook(ctx, cmdArgs, username, groupname, env, timeout, pidHook)
if cleanup != nil {
    cleanup()
}
```

The `pid` is captured in the closure, so callers no longer need to remember the exact pid they tracked. Guarding against pid reuse is preserved by the closure, and the Register/Unregister pair becomes structurally enforced by the API rather than convention.

pkg/executor/handlers/common/interfaces.go
```diff
@@ -42,6 +42,14 @@ type CommandExecutor interface {
...
 	// Exec executes a command with all options (user, group, env, timeout)
 	Exec(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration) (int, string, error)
+
+	// ExecWithHook is like Exec but invokes pidHook with the child's pid
+	// immediately after the process is started (before it is waited on).
+	// Callers use it to register the root pid with the PAM tracker so
+	// sudo calls issued by the child can be attributed to the originating
+	// deploy shell Command. pidHook may be nil, in which case this
+	// behaves identically to Exec.
+	ExecWithHook(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration, pidHook func(pid int)) (int, string, error)
```
geunwoonoh, line 52: **Suggestion**: 7-parameter signature is hard to read and duplicates `Exec`.

```go
ExecWithHook(ctx context.Context, args []string, username, groupname string, env map[string]string, timeout time.Duration, pidHook func(pid int)) (int, string, error)
```

Go style (e.g. Effective Go, Google Go Style) recommends collapsing 4+ arguments into an options struct. `CommandOptions` already exists and already has a `PIDHook` field — consider reducing `CommandExecutor` to a single `Execute(ctx, opts)` and keeping `Exec` / `ExecWithHook` as thin package-level helpers.

Not a blocker for this PR — just flagging for a future cleanup.

pkg/executor/executor.go
```diff
@@ -96,6 +96,38 @@ func (e *Executor) Execute(ctx context.Context, opts CommandOptions) (int, strin
...
+
+	var buf bytes.Buffer
+	cmd.Stdout = &buf
+	cmd.Stderr = &buf
+
+	if err := cmd.Start(); err != nil {
+		return buf.Bytes(), err
+	}
+
+	func() {
```
geunwoonoh: **Suggestion**: the `recover()` around `pidHook` may be over-defensive.

```go
func() {
    defer func() {
        if r := recover(); r != nil {
            log.Error().Interface("panic", r).Msg("PIDHook panicked; continuing")
        }
    }()
    if cmd.Process != nil {
        pidHook(cmd.Process.Pid)
    }
}()
```

The only caller passes a closure that calls `runner.RegisterCommandPID`, which is a trivial map insert — there is no realistic panic path. If one ever appears it would be a bug, and crashing is more useful than silently swallowing it (the command would also continue with a leaked expectation that the pid is tracked).

Either remove the `recover`, or add a comment stating the recover exists only as a safety net for a potential future case where `PIDHook` is exposed to external/untrusted code.

pkg/executor/handlers/common/testing.go
```diff
@@ -3,10 +3,19 @@ package common
...
+	"sync/atomic"
 	"testing"
 	"time"
 )
 
+// mockSyntheticPID produces monotonically increasing positive pids for
+// MockCommandExecutor.ExecWithHook so that every invocation hands the
+// hook a distinct, positive value. The base is well above typical real
+// pids to make accidental collisions obvious in test output.
+var mockSyntheticPID atomic.Int64
```
geunwoonoh, line 15: **Suggestion**: `mockSyntheticPID` is global state that leaks across tests.

```go
var mockSyntheticPID atomic.Int64
const mockSyntheticPIDBase = 900000
```

The counter increases monotonically across every test in the process. No current test depends on a specific value, so it is safe today — but future tests that assert on a concrete pid will become order-dependent.

**Recommendation**: reset on `NewMockCommandExecutor` via `t.Cleanup`, or use a per-mock counter:

```go
func NewMockCommandExecutor(t *testing.T) *MockCommandExecutor {
    t.Cleanup(func() { mockSyntheticPID.Store(0) })
    // ...
}
```

pkg/executor/handlers/shell/shell_command_tracker_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/alpacax/alpamon/pkg/executor/handlers/common"
+	"github.com/alpacax/alpamon/pkg/runner"
+	"github.com/alpacax/alpamon/pkg/runner/runnertest"
+)
+
+// hookRecordingExecutor wraps a MockCommandExecutor and records whether
+// ExecWithHook was invoked and whether it passed through a non-nil hook.
+// It is deliberately local to the shell handler tests so it cannot be
+// accidentally reused by other handlers and mask real breakage.
+type hookRecordingExecutor struct {
```
geunwoonoh, line 18: **Suggestion**: `hookRecordingExecutor` overlaps with `common.MockCommandExecutor.ExecWithHook`.

This PR already adds `ExecWithHook` to `MockCommandExecutor` (with a synthetic pid), but the shell handler test wraps the mock in `hookRecordingExecutor` specifically to pin the pid to `424242` so it can later call `am.LookupPID(424242)`.

Having two implementations of the same contract (synthetic pid vs. fixed pid) can confuse future readers. Consider consolidating: for example, add a `SetHookPID(pid int)` option on `MockCommandExecutor` and drop the wrapper here.

Not a functional concern — just a tidiness note.

pkg/executor/handlers/shell/shell.go
```diff
@@ -86,16 +87,16 @@ func (h *ShellHandler) handleShellCommand(ctx context.Context, args *common.Comm
...
 
 	// Fallback: direct execution with manual operator parsing
-	return h.executeWithOperators(ctx, command, username, groupname, env, timeout)
+	return h.executeWithOperators(ctx, command, username, groupname, env, timeout, args.CommandID)
 }
 
 // executeWithOperators handles shell operators (&&, ||, ;)
-func (h *ShellHandler) executeWithOperators(ctx context.Context, command, username, groupname string, env map[string]string, timeout time.Duration) (int, string, error) {
+func (h *ShellHandler) executeWithOperators(ctx context.Context, command, username, groupname string, env map[string]string, timeout time.Duration, commandID string) (int, string, error) {
 	spl := strings.Fields(command)
```
geunwoonoh, line 100: **Suggestion** (pre-existing, not introduced by this PR — flagging for tracking):

```go
spl := strings.Fields(command)
```

`strings.Fields` splits purely on whitespace and does not recognize quoting. `echo "hello world"` becomes `["echo", "\"hello", "world\""]`. The `AllowSh` path (`/bin/sh -c command`) is unaffected, so this only matters for the fallback operator-parsing path.

No change needed here — just worth noting in the PR description or opening a follow-up issue so the fallback path's limitations are documented somewhere.

geunwoonoh (commented):
# Code Review: PR #282 — feat(pam): track Command executions for deploy shell sudo authorization

**Stack**: Go (inferred — no frontmatter in CLAUDE.md, all changed files are `.go`)
**Changes**: +849 / -18, 12 files
**Reviewed commits**: `10e5118` (initial) + `b91a543` (Copilot iteration 3) + `8a0f514` (embedded field simplification)

---

## Summary

Extends the in-memory PID→session tracker in `pkg/runner` so that deploy shell Commands (alpacon-CLI / MCP `command`) register their root pid alongside Websh PTY sessions. `SessionInfo` grows `Kind` / `CommandID` / `Username` / `StartedAt`, letting a single map (and a single PAM lookup path) hold both kinds of entries. The `executor` gains a `PIDHook` / `ExecWithHook` path that switches from `cmd.CombinedOutput()` to `cmd.Start()` → hook → `cmd.Wait()` so the child's pid is registered with the tracker **before** it can exec `sudo`, eliminating a race. The shell handler only takes the hook path when `CommandID` is non-empty, and always pairs `Register` / `Unregister`. `SudoApprovalRequest` / `Response` get an `omitempty` `command_id` field, and the REST endpoint is routed to `/api/sudo/approval/` when `command_id` is set.

Overall this is a solid implementation with thorough docs and good defensive posture: pid-reuse guard, backward-compatibility (`effectiveKind`), concurrency safety, and a JSON wire-format guardrail test. All prior Copilot feedback has been addressed in iteration 3.

---

## Findings

- **Critical**: 0
- **Warnings**: 3 (see inline comments)
- **Suggestions**: 6 (see inline comments)

## Highlights

- **PID-reuse defense**: `RemovePIDCommandMapping` requires a matching `command_id`, so a crash that leaves a stale entry can't be accidentally cleared by a later unrelated Command. The `LeavesWebshEntryAlone` test even covers cross-kind defence-in-depth.
- **Race-free hook design**: `cmd.Start()` → `pidHook` → `cmd.Wait()` is synchronous; the docstring explicitly explains the "sudo arrives at PAM before the tracker knows the pid" race, which will help future maintainers.
- **Backward compatibility**: `effectiveKind()` transparently interprets legacy `Kind == ""` entries as websh, covered by `TestLegacyWebshEntry_ReadsAsWebsh`.
- **JSON wire-format guardrail**: `TestSudoApprovalRequest_JSONTagsOmitEmpty` actually marshals the struct and inspects serialized keys, so a regression dropping `omitempty` would be caught in CI. This locks down the contract with the server-side 2-branch resolver.
- **Test coverage**: 14 test functions covering registration / removal / concurrency / legacy compat / wire format / handler lifecycle — excellent for a PR of this size.
- **Structured logging migration**: `log.Debug().Msgf(...)` → `log.Debug().Int(...).Str(...)` improves log parser ergonomics.
- **Copilot iteration**: All 5 initial Copilot findings (test-only API leak, remove-guard, register return value, mock pid=0, weak JSON guardrail) were fully addressed in iteration 3 (`b91a543`).

## Go Convention Check

| Item | Status | Note |
|------|--------|------|
| Doc comments on exported identifiers | ✅ | `TrackerKindWebsh`, `SessionInfo`, `AddPIDCommandMapping`, etc. all documented |
| Receiver naming consistency | ✅ | `am` / `e` / `h` / `m` / `r` / `pc` — consistent |
| Error wrapping (`%w`) | ✅ | Preserves existing pattern |
| Package doc | ✅ | `runnertest` has one |
| Naming (CommandID/commandID) | ✅ | Idiomatic Go |
| Context position | ✅ | First parameter |
| `sync/atomic.Int64` | ✅ | Go 1.19+ |
| Interface simplicity | ⚠️ | `CommandExecutor` now has 7 methods; consider option-struct pattern (see Suggestion 2) |
| Struct field alignment | ⚠️ | `SessionInfo` / `SudoApprovalRequest` are mixed-type; minor padding loss, not critical |

## Duplicate / Unused Code

- **No duplication**: `runCommand` reuses `cmd.CombinedOutput()` when `pidHook == nil`, so the existing path isn't reimplemented.
- **No unused code**: All new fields (`Username`, `StartedAt`) are actually read; `TrackerEntry` is returned by `LookupPID` for inspection.
- **Minor test overlap**: `TestRegisterCommandPID_RoundTrip` and `TestAddPIDCommandMapping_RegistersCommandKind` both cover register→lookup, but intentionally — one exercises the package-level helper, the other the method. Acceptable.

## Performance Impact

| Area | Impact |
|------|--------|
| `commandID == ""` path (internal shell, collectors, …) | **No change**. `executeCommand` still calls the original `Exec()`; `runCommand` falls back to `CombinedOutput()` when the hook is nil. |
| `commandID != ""` path (deploy shell) | `CombinedOutput()` → `Start()+Wait()` transition is essentially free — `CombinedOutput()` itself is `Start()+Wait()` + `bytes.Buffer` internally. Hook cost is one function-pointer call + (atomic add / map insert). |
| AuthManager lock contention | `AddPIDCommandMapping` / `RemovePIDCommandMapping` / `LookupPID` are O(1) under the RW-lock; deploy shell Command frequency is low. No meaningful contention. |
| Memory | `SessionInfo` grows by ~32 bytes/entry; active entries number in tens. Negligible. |
| `effectiveKind()` repeated calls | Trivial — the suggestion to cache is for readability, not performance (see Warning 3). |

**No performance regression.** The deploy shell path gains a single-digit-nanosecond hook overhead.

## Security Check

- **Hardcoded credentials**: none.
- **SQL injection / path traversal**: N/A.
- **Authz flow**: `command_id` is propagated from `protocol.Command.ID`, which is a server-generated UUID — not client-controlled. The tracker rejects empty `commandID`, so an empty-id bypass is impossible.
- **PID reuse attack**: `RemovePIDCommandMapping` requires matching `command_id`, so a hostile local process reusing a pid cannot drop the victim's entry.
- **Race window**: between `cmd.Start()` returning and the hook firing, the child has already exec'd (Linux stdlib synchronizes fork+exec via a pipe); the hook runs inline on the main goroutine, not a copy goroutine, so timing is deterministic.
- **Sudo fallback**: when `IsAlpconUser` is false, the `blockLocalSudo` policy still applies — unchanged by this PR.

**No significant security concerns.**

---

## Verdict

**COMMENTED** — Ready to merge; 3 Warnings recommended for follow-up (none are release-blockers).

### Checklist

- [x] Code follows Go conventions
- [x] Tests included and passing (PR description: `go test ./... -p 1` → 630/630)
- [x] Backward compatibility preserved (`effectiveKind` + legacy entry test)
- [x] JSON guardrail test for wire-format change
- [x] Documentation: all exported symbols have doc comments
- [ ] Tighten production exposure of `auth_manager_testing.go` (Warning 1)
- [ ] Make `handleSudoApprovalRequest` switch exhaustive (Warning 2)
- [ ] Cache `effectiveKind()` locally (Warning 3)

pkg/runner/auth_manager_testing.go
junho226: Fixed. Moved runnertest to `internal/runnertest/` (repo root) so the Go compiler's internal rule firewalls it from external modules. Placed under repo-root rather than `pkg/runner/internal/` because the shell handler tests in `pkg/executor/` need access, which `pkg/runner/internal/` would block. `NewEmptyAuthManager`/`SwapAuthManager` still ship in the binary (they need unexported field access), but the friendly testing-aware API is now firewalled.

pkg/runner/auth_manager.go
junho226: Fixed. Made the switch exhaustive with an explicit `TrackerKindWebsh` case; the `default` branch now rejects unknown kinds, logs a warning, sends a sudo denial, and closes the connection. Future new kinds will force a compile-time review rather than silently masquerading as websh.

pkg/executor/handlers/shell/shell.go
junho226: Fixed. `RegisterCommandPID` now returns an unregister closure that captures the exact `(pid, commandID)` pair; removed the separate `UnregisterCommandPID`. Shell handler simplified to `cleanup := runner.RegisterCommandPID(...); if cleanup != nil { cleanup() }`. Leak-proof by construction as suggested.

pkg/runner/auth_manager.go
junho226: Fixed. Cached `session.effectiveKind()` into a `kind` local and reused it in both the switch and the log builder (bundled with Warning 2).

pkg/executor/handlers/shell/shell_command_tracker_test.go
junho226: Intentional — the wrapper pins the pid to a known value (`424242`) so the test can assert `LookupPID` cleanup, whereas `MockCommandExecutor.ExecWithHook` uses a monotonically increasing synthetic pid. Keeping them separate avoids coupling unrelated handler tests to a shared pid source. Happy to consolidate if a second handler ends up needing the same pattern.

pkg/executor/handlers/shell/shell.go
junho226: Acknowledged — the `strings.Fields` limitation is pre-existing on the fallback (non-`AllowSh`) path and out of scope for this PR. Will open a follow-up issue so the limitation is documented.

pkg/executor/handlers/common/interfaces.go
junho226: Acknowledged — agreed the 7-parameter signature is awkward. Deferring the options-struct refactor to a follow-up since it cuts across `Exec`/`ExecWithHook`/`RunWithTimeout` and is better done as a dedicated cleanup PR.

pkg/executor/executor.go
junho226: Fixed. Removed the `recover()` wrapper. The hook is a trivial map insert and a panic would indicate a real bug — letting it propagate is more useful than silently swallowing.

pkg/executor/handlers/common/testing.go
junho226: Fixed. Added `t.Cleanup(func() { mockSyntheticPID.Store(0) })` in `NewMockCommandExecutor` so the counter resets between tests.

geunwoonoh (approved):

PR #283, opened 2026-04-22
feat: support uid-less adduser for Application service accounts

### LINKED ISSUE #281: feat(alpamon): support uid-less adduser for Application service accounts
## Summary

Make the `adduser` handler treat `uid`, `gid`, and `home_directory` as optional fields. When these fields are zero-valued or empty, the corresponding command-line flags (`-u`, `-g`, `--home`/`--home-dir`) are omitted so the OS auto-assigns them.

**Background:** alpacon-server's Application provisioning feature sends `adduser` commands without `uid`/`gid` for service accounts. Currently alpamon marks these fields as `validate:"required"`, so the command fails validation before execution.

**IAM User vs Application service account:**
- IAM User: uid centrally managed by alpacon-server → `useradd -u 5001` (cross-server consistency required)
- Application: uid auto-assigned by OS → `useradd gitlab-runner` (no cross-server consistency needed, matched by username)

## Motivation

- **Prerequisite for Application provisioning** — without this change, automatic service account creation is impossible
- Very small change (conditional flag omission) + fully backward compatible
- Existing IAM User uid-based adduser is unchanged

## Current behavior

Data expected by the alpamon `adduser` handler:

```json
{
  "username": "gitlab-runner",
  "uid": 5001,
  "gid": 5001,
  "groups": [1001, 1002],
  "comment": "GitLab Runner,,,,(alpacon)uuid",
  "home_directory": "/home/gitlab-runner",
  "home_directory_permission": "0755",
  "shell": "/bin/bash",
  "groupname": "alpacon"
}
```

Command built:
```bash
useradd -u 5001 -g 5001 -G 1001,1002 -d /home/gitlab-runner -s /bin/bash -c "..." gitlab-runner
```

When uid is missing, `-u` receives 0 or empty value → command fails.

## Expected behavior after change

**Application service account data (no uid/gid):**
```json
{
  "username": "gitlab-runner",
  "comment": "GitLab Runner,,,,(alpacon-app)uuid",
  "shell": "/usr/sbin/nologin"
}
```

**Command built:**
```bash
useradd -s /usr/sbin/nologin -c "..." gitlab-runner
```

→ OS auto-assigns the next available uid.

**Existing IAM User data (with uid) — no change:**
```json
{
  "username": "john",
  "uid": 5001,
  "gid": 5001,
  ...
}
```

→ `useradd -u 5001 -g 5001 ...` works as before.

### 1. Change UserData struct validate tags

**File:** `pkg/executor/handlers/user/user.go`

Current struct:
```go
type UserData struct {
    Username                string   `validate:"required"`
    UID                     uint64   `validate:"required"`     // ← needs change
    GID                     uint64   `validate:"required"`     // ← needs change
    Comment                 string   `validate:"required"`
    HomeDirectory           string   `validate:"required"`     // ← needs change
    HomeDirectoryPermission string   `validate:"omitempty"`
    Shell                   string   `validate:"required"`
    Groupname               string   `validate:"required"`
    Groups                  []uint64 `validate:"omitempty"`
}
```

After change:
```go
type UserData struct {
    Username                string   `validate:"required"`
    UID                     uint64   `validate:"omitempty"`    // required → omitempty
    GID                     uint64   `validate:"omitempty"`    // required → omitempty
    Comment                 string   `validate:"required"`
    HomeDirectory           string   `validate:"omitempty"`    // required → omitempty
    HomeDirectoryPermission string   `validate:"omitempty"`
    Shell                   string   `validate:"required"`
    Groupname               string   `validate:"required"`
    Groups                  []uint64 `validate:"omitempty"`
}
```

**Backward compatible:** existing alpacon-server always sends UID/GID/HomeDirectory, so changing to `omitempty` has no effect on current behavior. Validation is only skipped when the value is 0 or empty string.

### 2. Modify Debian path (adduser command)

In `handleAddUser()`, the Debian adduser command builder:
```go
// pkg/executor/handlers/user/user.go (~line 159-180)
// Before
args := []string{
    "--home", data.HomeDirectory,
    "--shell", data.Shell,
    "--uid", strconv.FormatUint(data.UID, 10),
    "--gid", strconv.FormatUint(data.GID, 10),
    "--gecos", data.Comment,
    "--disabled-password",
    data.Username,
}

// After
args := []string{}
if data.HomeDirectory != "" {
    args = append(args, "--home", data.HomeDirectory)
}
args = append(args, "--shell", data.Shell)
if data.UID > 0 {
    args = append(args, "--uid", strconv.FormatUint(data.UID, 10))
}
if data.GID > 0 {
    args = append(args, "--gid", strconv.FormatUint(data.GID, 10))
}
args = append(args, "--gecos", data.Comment, "--disabled-password", data.Username)
```

### 3. Modify RHEL path (useradd command)

RHEL command builder:
```go
// Before (~line 185-200)
// runs groupadd -g {GID} first, then:
args := []string{
    "--home-dir", data.HomeDirectory,
    "--shell", data.Shell,
    "--uid", strconv.FormatUint(data.UID, 10),
    "--gid", strconv.FormatUint(data.GID, 10),
    "--comment", data.Comment,
    "--create-home",
    data.Username,
}

// After
// Only run groupadd when GID > 0
if data.GID > 0 {
    // existing groupadd logic
}
args := []string{}
if data.HomeDirectory != "" {
    args = append(args, "--home-dir", data.HomeDirectory)
}
args = append(args, "--shell", data.Shell)
if data.UID > 0 {
    args = append(args, "--uid", strconv.FormatUint(data.UID, 10))
}
if data.GID > 0 {
    args = append(args, "--gid", strconv.FormatUint(data.GID, 10))
}
args = append(args, "--comment", data.Comment, "--create-home", data.Username)
```

### 4. HomeDirectory default handling

- Empty HomeDirectory: `--home`/`--home-dir` flag omitted → OS default `/home/{username}`
- HomeDirectoryPermission: existing logic unchanged (defaults to 0755 when empty)

### 5. Post-execution sync (no change)

After successful adduser, existing auto-sync behavior remains:
```go
// line 71-74
if exitCode == 0 {
    syncManager.SyncSystemInfo([]string{"groups", "users"})
}
```
This sync propagates to alpacon-server → SystemUser record created → Application.username matching → automatic binding.

### Backward compatibility

| Field | Existing (with uid) | New (without uid) | Behavior |
|-------|--------------------|--------------------|----------|
| uid | `5001` | `0` (Go zero value) | If uid>0 add `-u`, otherwise omit |
| gid | `5001` | `0` (Go zero value) | If gid>0 add `-g`, otherwise omit. Also skip groupadd |
| groups | `[1001, 1002]` | `[]` (empty array) | If empty, omit `-G` (existing omitempty behavior) |
| home_directory | `/home/john` | `""` (empty string) | If empty, omit `--home`/`--home-dir` (OS default) |
| shell | `/bin/bash` | `/usr/sbin/nologin` | Always add `--shell` (required stays) |

Existing alpacon-server IAM User adduser always sends UID/GID/HomeDirectory → **no behavior change**.

### Version bump

Bump alpamon version so alpacon-server can check `server.version >= MIN_VERSION` to determine support.

## Files to modify

| File | Change |
|------|--------|
| `pkg/executor/handlers/user/user.go` | UserData struct validate tags (UID/GID/HomeDirectory: `required` → `omitempty`) |
| `pkg/executor/handlers/user/user.go` | Debian adduser path: conditional uid/gid/home_directory flags |
| `pkg/executor/handlers/user/user.go` | RHEL useradd path: conditional uid/gid/home_directory flags + conditional groupadd |
| adduser tests | Add uid-less cases (both Debian + RHEL), regression for existing uid cases |

### Debian (adduser command)
- [ ] uid=0 → `adduser` has no `--uid` flag
- [ ] gid=0 → `adduser` has no `--gid` flag
- [ ] home_directory="" → `adduser` has no `--home` flag (OS default)
- [ ] uid=5001, gid=5001 → `adduser --uid 5001 --gid 5001 ...` as before (backward compat)

### RHEL (useradd command)
- [ ] uid=0 → `useradd` has no `--uid` flag
- [ ] gid=0 → `useradd` has no `--gid` flag + `groupadd` not executed
- [ ] home_directory="" → `useradd` has no `--home-dir` flag (OS default)
- [ ] uid=5001, gid=5001 → `useradd --uid 5001 --gid 5001 ...` as before (backward compat)

### Common
- [ ] groups=[] → no `-G` flag
- [ ] shell="/usr/sbin/nologin" → `--shell /usr/sbin/nologin`
- [ ] Account already exists → error returned (existing behavior)
- [ ] Comment field passed correctly (including `(alpacon-app)` prefix)
- [ ] On success, `SyncSystemInfo(["groups", "users"])` called automatically (existing behavior)
- [ ] After validate tag changes, existing required fields (Username, Comment, Shell, Groupname) with empty values → still error

geunwoonoh:
## Summary

This PR teaches the `adduser` handler to build commands without `-u`/`-g`/`--home`/`--home-dir` flags when the payload is flagged as an Application service account, letting the OS auto-assign uid/gid/home. Existing IAM User flows — where alpacon-server centrally assigns uid for cross-server consistency — are unchanged and, importantly, have their validation guarantees reinforced.

This unblocks alpacon-server's upcoming Application declarative provisioning feature.

## Background

- **IAM User** — alpacon-server assigns uid centrally so the same user has the same uid on every server. uid/gid/home are required.
- **Application service account** — uid is irrelevant across servers (matched by username). uid/gid/home should be optional so the OS can pick them.

The naive fix (relaxing all three fields to `omitempty`) would have silently let an IAM User payload with missing uid succeed with an OS-assigned uid, breaking cross-server consistency. This PR avoids that by using an explicit discriminator field.

## Design: `IsServiceAccount` discriminator

Add `UserData.IsServiceAccount bool` and switch UID / GID / HomeDirectory from `validate:"required"` to `validate:"required_unless=IsServiceAccount true"`.

```go
type UserData struct {
    Username                string   `validate:"required"`
    UID                     uint64   `validate:"required_unless=IsServiceAccount true"`
    GID                     uint64   `validate:"required_unless=IsServiceAccount true"`
    Comment                 string   `validate:"required"`
    HomeDirectory           string   `validate:"required_unless=IsServiceAccount true"`
    HomeDirectoryPermission string   `validate:"omitempty"`
    Shell                   string   `validate:"required"`
    Groupname               string   `validate:"required"`
    Groups                  []uint64 `validate:"omitempty"`
    IsServiceAccount        bool
}
```

### Defense in depth

1. **Validator** — IAM User payloads (`IsServiceAccount=false`, the default) still require uid/gid/home. A missing or zero uid now fails validation loudly instead of falling back to OS auto-assignment.
2. **Handler** — the flag-omission logic is `data.IsServiceAccount && data.UID == 0`, not just `data.UID == 0`. Even if somehow a validator rule were loosened in the future, an IAM User payload with UID=0 would still pass `--uid 0` to useradd and be rejected at OS level rather than silently succeeding.
3. **Tests** — both scenarios are locked in with explicit regression tests (see below).

### JSON compatibility

`CommandData.IsServiceAccount` uses `json:"is_service_account,omitempty"`, so existing IAM User callers that never send the field are fully backward-compatible. Missing key → Go zero value `false` → IAM User path, as expected.

## Changes

| Commit | Files | Summary |
|--------|-------|---------|
| `feat(protocol): add is_service_account discriminator to user commands` | `internal/protocol/command.go`, `pkg/executor/handlers/common/args.go` | Thread the new field through `CommandData` → `CommandArgs` |
| `feat(iam): gate uid/gid/home requirements on service-account flag` | `pkg/executor/handlers/user/types.go` | `required_unless` validator tags + `userDataFromArgs` constructor (with Shell / HomeDirectoryPermission defaulting) |
| `feat(iam): conditionally omit uid/gid/home flags for service accounts` | `pkg/executor/handlers/user/user.go` | Build adduser/useradd args into a `[]string` with per-flag gates; skip RHEL `groupadd` when GID is omitted; guard `os.Chmod` against empty HomeDirectory |
| `test(iam): cover uid-less service accounts and IAM User uid regressions` | `pkg/executor/handlers/user/user_test.go` | Service-account happy path (Debian/RHEL), IAM User regression, discriminator-based validation tests |

## Behavior matrix

| Scenario | alpacon-server sends | Validation | Command emitted |
|----------|----------------------|------------|-----------------|
| IAM User (current flow) | `is_service_account` absent or `false`, uid/gid/home set | ✅ required fields present | `useradd -u 5001 -g 5001 --home-dir /home/john …` |
| IAM User with missing uid (bug) | `is_service_account` absent or `false`, uid missing | ❌ validation fails | — |
| IAM User with uid=0 (bug) | `is_service_account` absent or `false`, uid=0 | ❌ validation fails | — |
| Service account, uid-less | `is_service_account: true`, uid/gid/home omitted | ✅ passes | `useradd -s /usr/sbin/nologin -c "(alpacon-app)…" gitlab-runner` (no -u/-g/--home-dir; no groupadd on RHEL) |
| Service account missing flag (contract violation) | `is_service_account` absent, uid/gid/home also missing | ❌ validation fails | Error clearly surfaces the missing contract |

## Test plan

```bash
go test -v ./pkg/executor/handlers/user/ -p 1
go test ./internal/protocol/ -p 1
gofmt -l pkg/executor/handlers/user/ pkg/executor/handlers/common/ internal/protocol/
go vet ./...
go build ./...
```

### New test cases

- `TestUserHandler_AddUser_UidLess` — on both Debian and RHEL: assert the emitted args contain `--shell`, `--gecos`/`--comment`, `--disabled-password`/`--create-home`, username; and assert `--uid`, `--gid`, `--home`/`--home-dir` are absent. On RHEL, also assert `groupadd` is not called.
- `TestUserHandler_AddUser_RhelWithUID` — regression: IAM User (uid=5001, gid=5001, home=/home/john) still produces `groupadd --gid 5001 alpacon` followed by `useradd --uid 5001 --gid 5001 --home-dir /home/john ... --create-home john`.
- `TestUserHandler_Validate` — three new cases:
  - service account with only username/comment/shell/groupname → passes
  - IAM User missing uid/gid/home → fails
  - IAM User with UID=0 → fails (the exact cross-server-consistency bug the discriminator defends against)

### End-to-end (optional)

Run alpamon in a Debian/RHEL container, send an `adduser` command with `is_service_account: true` and no uid, then verify with `id <username>` that the OS assigned a uid. Then send an IAM User payload without uid and confirm a validation error is returned.

## alpacon-server contract (for the follow-up PR)

`Server.add_service_user()` **must** include `is_service_account: True` in its data payload:

```python
data = {
    'username': application.username,
    'comment': f'{application.name},,,,(alpacon-app){application.id}',
    'shell': application.shell,
    'groupname': 'alpacon',
    'is_service_account': True,
}
if application.home_directory:
    data['home_directory'] = application.home_directory
```

`Server.add_user()` (IAM User) should continue to omit the field (or explicitly send `False`). Both are accepted; omitting is recommended for clarity.

## Risks and mitigations

| Risk | Mitigation |
|------|------------|
| alpacon-server bug drops uid from IAM User payload → servers diverge on uid | `required_unless` validator rejects the payload; regression test covers it |
| alpacon-server sends IAM User with uid=0 | Validator rejects; handler still wouldn't strip the flag because the `IsServiceAccount` gate is false; regression test covers it |
| alpacon-server forgets to set `is_service_account` on a service-account payload | Validator treats uid/gid/home as required → fails loudly, surfaces the contract violation |
| RHEL service account skips `groupadd` | Intentional — `useradd` auto-creates a per-user group; no cross-server group consistency is needed for service accounts |
| Rollout order (alpacon-server deploys before alpamon) | alpacon-server checks alpamon version via `_supports_service_user_provisioning()` and skips unsupported servers |

## Not in scope

- Deprovisioning (deletion of service accounts)
- `mod_service_user` (shell/home propagation after create)
- macOS support (pre-existing limitation of the user handler)
- Version bump — alpamon version is injected at build time (`pkg/version/version.go` is `var Version = "dev"`); the real version comes from the release tag, which is what alpacon-server's `MIN_VERSION` check compares against.

## Checklist

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `gofmt` clean on all changed files
- [x] Unit tests pass (`go test ./pkg/executor/handlers/user/ ./internal/protocol/ -p 1`)
- [x] Backward-compatible JSON wire format (existing callers unaffected)
- [x] Regression tests for the IAM User path (both happy path and uid-missing/uid=0 failure modes)
- [ ] Manual end-to-end on Debian and RHEL containers (recommended before merge)
- [ ] Coordinated alpacon-server PR uses `is_service_account: True` in `Server.add_service_user()`

### DIFF
--- a/pkg/executor/handlers/user/types.go
+++ b/pkg/executor/handlers/user/types.go
@@ -11,9 +11,11 @@ import "github.com/alpacax/alpamon/pkg/executor/handlers/common"
 //     consistency. Missing values indicate a server-side bug and must fail
 //     validation rather than silently fall back to OS auto-assignment.
 //   - IsServiceAccount=true (Application): UID/GID/HomeDirectory are optional.
-//     When omitted, the corresponding useradd/adduser flag is skipped so the
-//     OS auto-assigns values. Cross-server consistency is not required because
-//     alpacon-server matches service accounts by username.
+//     When omitted, the numeric --uid / --gid / --home flag is skipped:
+//     UID and HomeDirectory fall back to OS defaults; GID is replaced with
+//     `--ingroup <Groupname>` (Debian) or `--gid <Groupname>` (RHEL) so the
+//     primary group is set by name. Cross-server consistency is not required
+//     because alpacon-server matches service accounts by username.
 type UserData struct {
 	Username                string   `validate:"required"`
 	UID                     uint64   `validate:"required_unless=IsServiceAccount true"`
@@ -27,9 +29,10 @@ type UserData struct {
 	IsServiceAccount        bool
 }
 
-// userDataFromArgs builds a UserData from CommandArgs, applying defaults for
-// HomeDirectoryPermission and Shell. Shared by Validate and handleAddUser so
-// both paths see identical field population and defaulting.
+// userDataFromArgs builds a UserData from CommandArgs, defaulting empty
+// HomeDirectoryPermission to "0755". Shell is intentionally NOT defaulted
+// here (see comment below). Shared by Validate and handleAddUser so both
+// paths see identical field population.
 func userDataFromArgs(args *common.CommandArgs) UserData {
 	data := UserData{
 		Username:                args.Username,
@@ -47,8 +50,8 @@ func userDataFromArgs(args *common.CommandArgs) UserData {
 		data.HomeDirectoryPermission = "0755"
 	}
 	// Shell is intentionally NOT defaulted: defaulting it before
-	// ValidateStruct would mask an empty `shell` payload — the
-	// `validate:"required"` tag could never fire — and silently
+	// ValidateStruct would mask an empty `shell` payload (the
+	// `validate:"required"` tag could never fire) and silently
 	// give a service account an interactive `/bin/bash`. alpacon-server
 	// already always sends Shell explicitly (e.g. `/usr/sbin/nologin`
 	// for Application service accounts, `user.shell` for IAM Users),
--- a/pkg/executor/handlers/user/types.go
+++ b/pkg/executor/handlers/user/types.go
@@ -1,16 +1,55 @@
 package user
 
-// UserData contains data for user operations
+import "github.com/alpacax/alpamon/pkg/executor/handlers/common"
+
+// UserData contains data for user operations.
+//
+// IsServiceAccount discriminates IAM User provisioning from Application
+// service-account provisioning:
+//   - IsServiceAccount=false (default, IAM User): UID/GID/HomeDirectory are
+//     required — alpacon-server centrally assigns these for cross-server
+//     consistency. Missing values indicate a server-side bug and must fail
+//     validation rather than silently fall back to OS auto-assignment.
+//   - IsServiceAccount=true (Application): UID/GID/HomeDirectory are optional.
+//     When omitted, the corresponding useradd/adduser flag is skipped so the
+//     OS auto-assigns values. Cross-server consistency is not required because
+//     alpacon-server matches service accounts by username.
 type UserData struct {
 	Username                string   `validate:"required"`
-	UID                     uint64   `validate:"required"`
-	GID                     uint64   `validate:"required"`
+	UID                     uint64   `validate:"required_unless=IsServiceAccount true"`
+	GID                     uint64   `validate:"required_unless=IsServiceAccount true"`
 	Comment                 string   `validate:"required"`
-	HomeDirectory           string   `validate:"required"`
+	HomeDirectory           string   `validate:"required_unless=IsServiceAccount true"`
 	HomeDirectoryPermission string   `validate:"omitempty"`
 	Shell                   string   `validate:"required"`
 	Groupname               string   `validate:"required"`
 	Groups                  []uint64 `validate:"omitempty"`
+	IsServiceAccount        bool
+}
+
+// userDataFromArgs builds a UserData from CommandArgs, applying defaults for
+// HomeDirectoryPermission and Shell. Shared by Validate and handleAddUser so
+// both paths see identical field population and defaulting.
+func userDataFromArgs(args *common.CommandArgs) UserData {
+	data := UserData{
+		Username:                args.Username,
+		UID:                     args.UID,
+		GID:                     args.GID,
+		Comment:                 args.Comment,
+		HomeDirectory:           args.HomeDirectory,
+		HomeDirectoryPermission: args.HomeDirectoryPermission,
+		Shell:                   args.Shell,
+		Groupname:               args.Groupname,
+		Groups:                  args.Groups,
+		IsServiceAccount:        args.IsServiceAccount,
+	}
+	if data.HomeDirectoryPermission == "" {
+		data.HomeDirectoryPermission = "0755"
+	}
+	if data.Shell == "" {
+		data.Shell = "/bin/bash"
+	}
+	return data
 }
 
 // DeleteUserData contains data for user deletion
--- a/pkg/executor/handlers/user/types.go
+++ b/pkg/executor/handlers/user/types.go
@@ -7,7 +7,7 @@ import "github.com/alpacax/alpamon/pkg/executor/handlers/common"
 // IsServiceAccount discriminates IAM User provisioning from Application
 // service-account provisioning:
 //   - IsServiceAccount=false (default, IAM User): UID/GID/HomeDirectory are
-//     required — alpacon-server centrally assigns these for cross-server
+//     required. alpacon-server centrally assigns these for cross-server
 //     consistency. Missing values indicate a server-side bug and must fail
 //     validation rather than silently fall back to OS auto-assignment.
 //   - IsServiceAccount=true (Application): UID/GID/HomeDirectory are optional.
--- a/pkg/executor/handlers/user/types.go
+++ b/pkg/executor/handlers/user/types.go
@@ -46,9 +46,13 @@ func userDataFromArgs(args *common.CommandArgs) UserData {
 	if data.HomeDirectoryPermission == "" {
 		data.HomeDirectoryPermission = "0755"
 	}
-	if data.Shell == "" {
-		data.Shell = "/bin/bash"
-	}
+	// Shell is intentionally NOT defaulted: defaulting it before
+	// ValidateStruct would mask an empty `shell` payload — the
+	// `validate:"required"` tag could never fire — and silently
+	// give a service account an interactive `/bin/bash`. alpacon-server
+	// already always sends Shell explicitly (e.g. `/usr/sbin/nologin`
+	// for Application service accounts, `user.shell` for IAM Users),
+	// so the validator catching missing `shell` is the desired contract.
 	return data
 }
 
--- a/pkg/executor/handlers/user/user.go
+++ b/pkg/executor/handlers/user/user.go
@@ -125,7 +125,7 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 	// UID/GID/HomeDirectory flags are omitted only when IsServiceAccount=true
 	// AND the value is zero/empty, so the OS can auto-assign. For IAM User
 	// payloads (IsServiceAccount=false), these fields are validated as
-	// required upstream — if we ever reach this point with zero values, we
+	// required upstream: if we ever reach this point with zero values, we
 	// still pass whatever was provided rather than silently rewriting the
 	// command, since defense-in-depth is cheap.
 	omitUID := data.IsServiceAccount && data.UID == 0
@@ -198,24 +198,22 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 		}
 	}
 
-	// Add user to additional groups if specified
-	if len(data.Groups) > 0 && h.groupService != nil {
-		if err := h.groupService.AddUserToGroups(ctx, data.Username, data.Groups); err != nil {
-			log.Warn().Err(err).Msg("Failed to add user to additional groups")
-			return 0, fmt.Sprintf("User '%s' created but failed to add to groups: %v", data.Username, err), nil
-		}
-	}
-
-	// Service accounts skip the numeric-gid primary-group setup above, so they
-	// would not be a member of data.Groupname (e.g. "alpacon"). Add them as a
-	// supplementary member by name so privilege demotion (utils.Demote with
-	// ValidateGroup=true) succeeds for later commands. IAM User path doesn't
-	// need this because useradd --gid already sets the primary group.
-	if data.IsServiceAccount && data.Groupname != "" {
+	// Service-account Groupname setup.
+	// Runs BEFORE AddUserToGroups so the critical `Groupname` membership is
+	// established even if the optional supplementary `Groups` list fails and
+	// triggers the early return below.
+	// Gated on `omitGID`: when GID is provided, the RHEL path already ran
+	// `groupadd --gid <GID> <Groupname>` and `useradd --gid <GID>`, making the
+	// user a member of Groupname as the primary group. Re-running groupadd/
+	// usermod there would be redundant. When GID is omitted (the typical
+	// service-account payload), useradd auto-creates a per-user primary group
+	// and the user is NOT in `Groupname`, which would break later
+	// `utils.Demote(..., ValidateGroup=true)` calls.
+	if data.IsServiceAccount && omitGID && data.Groupname != "" {
 		// Ensure the group exists (groupadd -f is a no-op if it already does).
 		// Non-fatal: if this fails (missing binary, permissions, corrupt group
 		// db), usermod below will also fail and later Demote errors become
-		// hard to diagnose — so log the actual groupadd failure here.
+		// hard to diagnose, so log the actual groupadd failure here.
 		if code, out, err := h.Executor.Run(ctx, "/usr/sbin/groupadd", "-f", data.Groupname); code != 0 {
 			log.Warn().
 				Str("group", data.Groupname).
@@ -235,6 +233,14 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 				Str("output", out).
 				Err(err).
 				Msg("Failed to add service account to supplementary group")
+		}
+	}
+
+	// Add user to additional groups if specified
+	if len(data.Groups) > 0 && h.groupService != nil {
+		if err := h.groupService.AddUserToGroups(ctx, data.Username, data.Groups); err != nil {
+			log.Warn().Err(err).Msg("Failed to add user to additional groups")
+			return 0, fmt.Sprintf("User '%s' created but failed to add to groups: %v", data.Username, err), nil
 		}
 	}
 
--- a/pkg/executor/handlers/user/user.go
+++ b/pkg/executor/handlers/user/user.go
@@ -213,7 +213,17 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 	// need this because useradd --gid already sets the primary group.
 	if data.IsServiceAccount && data.Groupname != "" {
 		// Ensure the group exists (groupadd -f is a no-op if it already does).
-		_, _, _ = h.Executor.Run(ctx, "/usr/sbin/groupadd", "-f", data.Groupname)
+		// Non-fatal: if this fails (missing binary, permissions, corrupt group
+		// db), usermod below will also fail and later Demote errors become
+		// hard to diagnose — so log the actual groupadd failure here.
+		if code, out, err := h.Executor.Run(ctx, "/usr/sbin/groupadd", "-f", data.Groupname); code != 0 {
+			log.Warn().
+				Str("group", data.Groupname).
+				Int("exitCode", code).
+				Str("output", out).
+				Err(err).
+				Msg("Failed to ensure supplementary group exists (groupadd -f); usermod will likely fail")
+		}
 		// Add the user to the group as a supplementary member. Non-fatal: log
 		// a warning so the root cause surfaces instead of silently breaking
 		// later Demote calls.
@@ -221,6 +231,7 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 			log.Warn().
 				Str("user", data.Username).
 				Str("group", data.Groupname).
+				Int("exitCode", code).
 				Str("output", out).
 				Err(err).
 				Msg("Failed to add service account to supplementary group")
--- a/pkg/executor/handlers/user/user.go
+++ b/pkg/executor/handlers/user/user.go
@@ -121,28 +121,67 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 	var output string
 	var err error
 
-	// Platform-specific user addition.
-	// UID/GID/HomeDirectory flags are omitted only when IsServiceAccount=true
-	// AND the value is zero/empty, so the OS can auto-assign. For IAM User
-	// payloads (IsServiceAccount=false), these fields are validated as
-	// required upstream: if we ever reach this point with zero values, we
-	// still pass whatever was provided rather than silently rewriting the
-	// command, since defense-in-depth is cheap.
-	omitUID := data.IsServiceAccount && data.UID == 0
-	omitGID := data.IsServiceAccount && data.GID == 0
-	omitHome := data.IsServiceAccount && data.HomeDirectory == ""
+	// The omit*Flag booleans govern *flag emission* on adduser/useradd,
+	// NOT whether the underlying entity exists. When a flag is omitted:
+	//   - omitUIDFlag  → OS auto-assigns a uid (the user still gets a uid).
+	//   - omitGIDFlag  → no numeric --gid; we set the primary group by
+	//                    name via --ingroup / --gid <Groupname> below.
+	//   - omitHomeFlag → no explicit --home/--home-dir; the OS default
+	//                    path (typically /home/<username>) is used and
+	//                    the directory IS still created.
+	//
+	// We omit a flag only when IsServiceAccount=true AND the value is
+	// zero/empty. For IAM User payloads (IsServiceAccount=false), these
+	// fields are validated as required upstream; if we ever reach this
+	// point with zero values we still pass whatever was provided rather
+	// than silently rewriting the command (defense-in-depth is cheap).
+	omitUIDFlag := data.IsServiceAccount && data.UID == 0
+	omitGIDFlag := data.IsServiceAccount && data.GID == 0
+	omitHomeFlag := data.IsServiceAccount && data.HomeDirectory == ""
+
+	// Service-account primary-group bootstrap (load-bearing).
+	//
+	// When the payload omits a numeric GID, we want the user's primary
+	// group to be `Groupname` (e.g. "alpacon") so that later
+	// `utils.Demote(..., ValidateGroup=true)` succeeds. Setting it as a
+	// post-fact supplementary membership via `usermod -aG` is fragile
+	// on systems with USERGROUPS_ENAB=no / Debian USERGROUPS=no.
+	//
+	// Instead, ensure the named group exists up front, then pass it as
+	// the primary group to adduser/useradd via `--ingroup` (Debian) or
+	// `--gid <name>` (RHEL — useradd accepts a group name as well as a
+	// numeric id). If `groupadd -f` fails here, fail loudly so the
+	// caller does not see a "succeeded" provisioning that breaks at
+	// runtime.
+	// `omitGIDFlag` already implies `data.IsServiceAccount` per its definition.
+	if omitGIDFlag && data.Groupname != "" {
+		if code, out, gerr := h.Executor.Run(ctx, "/usr/sbin/groupadd", "-f", data.Groupname); code != 0 {
+			log.Error().
+				Str("group", data.Groupname).
+				Int("exitCode", code).
+				Str("output", out).
+				Err(gerr).
+				Msg("Failed to ensure service-account primary group exists")
+			return code, fmt.Sprintf("Failed to ensure group %q for service account: %s", data.Groupname, out), gerr
+		}
+	}
 
 	switch utils.PlatformLike {
 	case "debian":
 		cmdArgs := []string{}
-		if !omitHome {
+		if !omitHomeFlag {
 			cmdArgs = append(cmdArgs, "--home", data.HomeDirectory)
 		}
 		cmdArgs = append(cmdArgs, "--shell", data.Shell)
-		if !omitUID {
+		if !omitUIDFlag {
 			cmdArgs = append(cmdArgs, "--uid", strconv.FormatUint(data.UID, 10))
 		}
-		if !omitGID {
+		switch {
+		case omitGIDFlag && data.Groupname != "":
+			// Service-account path: set primary group by name so the user
+			// joins `Groupname` regardless of USERGROUPS settings.
+			cmdArgs = append(cmdArgs, "--ingroup", data.Groupname)
+		case !omitGIDFlag:
 			cmdArgs = append(cmdArgs, "--gid", strconv.FormatUint(data.GID, 10))
 		}
 		cmdArgs = append(cmdArgs, "--gecos", data.Comment, "--disabled-password", data.Username)
@@ -151,10 +190,10 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 			return exitCode, output, err
 		}
 	case "rhel":
-		// Create primary group only when an explicit GID is requested.
-		// Without GID (service-account path), useradd auto-creates a group
-		// matching the username.
-		if !omitGID {
+		// Create primary group with the explicit GID for IAM User payloads.
+		// Service-account payloads (omitGIDFlag=true) already ran `groupadd -f`
+		// above and will use `--gid <Groupname>` below.
+		if !omitGIDFlag {
 			exitCode, output, err = h.Executor.Run(
 				ctx,
 				"/usr/sbin/groupadd",
@@ -168,14 +207,18 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 		}
 
 		cmdArgs := []string{}
-		if !omitHome {
+		if !omitHomeFlag {
 			cmdArgs = append(cmdArgs, "--home-dir", data.HomeDirectory)
 		}
 		cmdArgs = append(cmdArgs, "--shell", data.Shell)
-		if !omitUID {
+		if !omitUIDFlag {
 			cmdArgs = append(cmdArgs, "--uid", strconv.FormatUint(data.UID, 10))
 		}
-		if !omitGID {
+		switch {
+		case omitGIDFlag && data.Groupname != "":
+			// Service-account path: useradd accepts a group name for --gid.
+			cmdArgs = append(cmdArgs, "--gid", data.Groupname)
+		case !omitGIDFlag:
 			cmdArgs = append(cmdArgs, "--gid", strconv.FormatUint(data.GID, 10))
 		}
 		cmdArgs = append(cmdArgs, "--comment", data.Comment, "--create-home", data.Username)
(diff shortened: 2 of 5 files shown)


pkg/executor/handlers/user/user.go
```diff
@@ -155,58 +121,77 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
...
+		}
+		cmdArgs = append(cmdArgs, "--shell", data.Shell)
+		if !omitUID {
+			cmdArgs = append(cmdArgs, "--uid", strconv.FormatUint(data.UID, 10))
+		}
+		if !omitGID {
+			cmdArgs = append(cmdArgs, "--gid", strconv.FormatUint(data.GID, 10))
+		}
+		cmdArgs = append(cmdArgs, "--comment", data.Comment, "--create-home", data.Username)
+		exitCode, output, err = h.Executor.Run(ctx, "/usr/sbin/useradd", cmdArgs...)
```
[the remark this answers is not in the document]
geunwoonoh: Good catch — fixed in 77ad74f.

You were right that skipping both `groupadd` and `--gid` left the service account out of `data.Groupname`, which would have broken `utils.Demote(..., ValidateGroup=true)` for later commands run as that user.

**Fix:** after the platform-specific `adduser`/`useradd` succeeds, the handler now runs `groupadd -f <Groupname>` (idempotent, works on both Debian and RHEL via shadow-utils) followed by `usermod -aG <Groupname> <Username>`. The service account ends up in `alpacon` as a supplementary member, so `ValidateGroup=true` passes. IAM User path is unchanged because `useradd --gid` already assigns the primary group.

Test updates:
- Dropped the stale "no groupadd" assertion on the service-account path.
- Kept an assertion that **gid-based** `groupadd` is not called on that path (that path is IAM-User-only).
- Added positive assertions for the new `groupadd -f alpacon` + `usermod -aG alpacon gitlab-runner` pair.

pkg/executor/handlers/user/user_test.go
```diff
@@ -154,6 +155,156 @@ func TestUserHandler_Execute(t *testing.T) {
...
+			forbidFlags: []string{"--uid", "--gid", "--home"},
+		},
+		{
+			name:         "rhel uid-less useradd, no groupadd",
+			platform:     "rhel",
+			wantProgram:  "/usr/sbin/useradd",
+			wantFlags:    []string{"--shell", "/usr/sbin/nologin", "--comment", "--create-home", "gitlab-runner"},
+			forbidFlags:  []string{"--uid", "--gid", "--home-dir"},
+			wantGroupadd: false,
+		},
```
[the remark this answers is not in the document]
geunwoonoh: Good catch — fixed in 77ad74f.

You were right that skipping both `groupadd` and `--gid` left the service account out of `data.Groupname`, which would have broken `utils.Demote(..., ValidateGroup=true)` for later commands run as that user.

**Fix:** after the platform-specific `adduser`/`useradd` succeeds, the handler now runs `groupadd -f <Groupname>` (idempotent, works on both Debian and RHEL via shadow-utils) followed by `usermod -aG <Groupname> <Username>`. The service account ends up in `alpacon` as a supplementary member, so `ValidateGroup=true` passes. IAM User path is unchanged because `useradd --gid` already assigns the primary group.

Test updates:
- Dropped the stale "no groupadd" assertion on the service-account path.
- Kept an assertion that **gid-based** `groupadd` is not called on that path (that path is IAM-User-only).
- Added positive assertions for the new `groupadd -f alpacon` + `usermod -aG alpacon gitlab-runner` pair.

pkg/executor/handlers/user/user.go
```diff
@@ -221,6 +206,27 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
...
 	}
 
+	// Service accounts skip the numeric-gid primary-group setup above, so they
+	// would not be a member of data.Groupname (e.g. "alpacon"). Add them as a
+	// supplementary member by name so privilege demotion (utils.Demote with
+	// ValidateGroup=true) succeeds for later commands. IAM User path doesn't
+	// need this because useradd --gid already sets the primary group.
+	if data.IsServiceAccount && data.Groupname != "" {
+		// Ensure the group exists (groupadd -f is a no-op if it already does).
+		_, _, _ = h.Executor.Run(ctx, "/usr/sbin/groupadd", "-f", data.Groupname)
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 9f08af8. `groupadd -f` now has its exit code, output, and error captured; on non-zero exit we log a warning with all three plus a hint that the subsequent `usermod` will likely fail. Still non-fatal (matches the pattern of the usermod branch right below), but the real cause is now visible in logs instead of guessed from a downstream `Demote` error.

pkg/executor/handlers/user/user.go
```diff
@@ -221,6 +206,38 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 		}
 	}
 
+	// Service accounts skip the numeric-gid primary-group setup above, so they
+	// would not be a member of data.Groupname (e.g. "alpacon"). Add them as a
+	// supplementary member by name so privilege demotion (utils.Demote with
+	// ValidateGroup=true) succeeds for later commands. IAM User path doesn't
+	// need this because useradd --gid already sets the primary group.
+	if data.IsServiceAccount && data.Groupname != "" {
+		// Ensure the group exists (groupadd -f is a no-op if it already does).
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Moved the service-account `Groupname` setup to run **before** the `AddUserToGroups` block. The critical membership (`groupadd -f` + `usermod -aG`) is now established first, so even if a service-account payload somehow carries a `Groups` list that fails to apply, the supplementary `Groupname` membership is already in place and later `Demote(..., ValidateGroup=true)` will still succeed.

pkg/executor/handlers/user/user.go
```diff
@@ -221,6 +206,38 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 		}
 	}
 
+	// Service accounts skip the numeric-gid primary-group setup above, so they
+	// would not be a member of data.Groupname (e.g. "alpacon"). Add them as a
+	// supplementary member by name so privilege demotion (utils.Demote with
+	// ValidateGroup=true) succeeds for later commands. IAM User path doesn't
+	// need this because useradd --gid already sets the primary group.
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Gated the block on `omitGID` so it only runs when the primary-group step was skipped (i.e. service-account payloads without a numeric GID). When GID is provided, `groupadd --gid <GID> <Groupname>` + `useradd --gid <GID>` already made the user a primary member of `Groupname`, so the extra `groupadd -f` / `usermod -aG` would have been redundant. Comment updated to match the actual behavior.

pkg/executor/handlers/user/types.go
```diff
@@ -1,16 +1,55 @@
...
+
+// UserData contains data for user operations.
+//
+// IsServiceAccount discriminates IAM User provisioning from Application
+// service-account provisioning:
+//   - IsServiceAccount=false (default, IAM User): UID/GID/HomeDirectory are
+//     required — alpacon-server centrally assigns these for cross-server
+//     consistency. Missing values indicate a server-side bug and must fail
+//     validation rather than silently fall back to OS auto-assignment.
+//   - IsServiceAccount=true (Application): UID/GID/HomeDirectory are optional.
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Replaced the spaced em-dash with a colon (or rephrased) to match the repo writing convention documented in CLAUDE.md.

pkg/executor/handlers/user/user_test.go
```diff
@@ -154,6 +155,177 @@ func TestUserHandler_Execute(t *testing.T) {
...
+// provisioning (no uid/gid/home_directory) results in adduser/useradd commands
+// without the corresponding flags, letting the OS auto-assign.
+func TestUserHandler_AddUser_UidLess(t *testing.T) {
+	baseArgs := &common.CommandArgs{
+		Username:         "gitlab-runner",
+		Comment:          "GitLab Runner,,,,(alpacon-app)abc",
+		Shell:            "/usr/sbin/nologin",
+		Groupname:        "alpacon",
+		IsServiceAccount: true,
+		// UID, GID, HomeDirectory intentionally omitted — OS auto-assigns
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Replaced the spaced em-dash with a colon (or rephrased) to match the repo writing convention documented in CLAUDE.md.

pkg/executor/handlers/user/user.go
```diff
@@ -221,6 +206,38 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
...
+	if data.IsServiceAccount && data.Groupname != "" {
+		// Ensure the group exists (groupadd -f is a no-op if it already does).
+		// Non-fatal: if this fails (missing binary, permissions, corrupt group
+		// db), usermod below will also fail and later Demote errors become
+		// hard to diagnose — so log the actual groupadd failure here.
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Replaced the spaced em-dash with a colon (or rephrased) to match the repo writing convention documented in CLAUDE.md.

pkg/executor/handlers/user/user_test.go
```diff
@@ -182,7 +354,50 @@ func TestUserHandler_Validate(t *testing.T) {
...
+		{
+			name: "adduser IAM User path must still require uid/gid/home",
+			cmd:  "adduser",
+			args: &common.CommandArgs{
+				Username:  "john",
+				Comment:   "John,,,,(alpacon)uuid",
+				Shell:     "/bin/bash",
+				Groupname: "alpacon",
+				// IsServiceAccount=false (default)
+				// UID/GID/HomeDirectory missing — must fail
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Replaced the spaced em-dash with a colon (or rephrased) to match the repo writing convention documented in CLAUDE.md.

pkg/executor/handlers/user/user_test.go
```diff
@@ -182,7 +354,50 @@ func TestUserHandler_Validate(t *testing.T) {
...
+			cmd:  "adduser",
+			args: &common.CommandArgs{
+				Username:      "john",
+				UID:           0, // bug: alpacon-server sent zero
+				GID:           5001,
+				Comment:       "John,,,,(alpacon)uuid",
+				HomeDirectory: "/home/john",
+				Shell:         "/bin/bash",
+				Groupname:     "alpacon",
+				// IsServiceAccount=false — uid=0 must be rejected
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Replaced the spaced em-dash with a colon (or rephrased) to match the repo writing convention documented in CLAUDE.md.

pkg/executor/handlers/user/user.go
```diff
@@ -155,58 +121,77 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
 	var exitCode int
 	var output string
 
-	// Platform-specific user addition
+	// Platform-specific user addition.
+	// UID/GID/HomeDirectory flags are omitted only when IsServiceAccount=true
+	// AND the value is zero/empty, so the OS can auto-assign. For IAM User
+	// payloads (IsServiceAccount=false), these fields are validated as
+	// required upstream — if we ever reach this point with zero values, we
```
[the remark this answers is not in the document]
geunwoonoh: Fixed in 6910cd4. Replaced the spaced em-dash with a colon (or rephrased) to match the repo writing convention documented in CLAUDE.md.

pkg/executor/handlers/user/user.go
```diff
@@ -155,64 +121,121 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
...
+	// established even if the optional supplementary `Groups` list fails and
+	// triggers the early return below.
+	// Gated on `omitGID`: when GID is provided, the RHEL path already ran
+	// `groupadd --gid <GID> <Groupname>` and `useradd --gid <GID>`, making the
+	// user a member of Groupname as the primary group. Re-running groupadd/
+	// usermod there would be redundant. When GID is omitted (the typical
+	// service-account payload), useradd auto-creates a per-user primary group
+	// and the user is NOT in `Groupname`, which would break later
+	// `utils.Demote(..., ValidateGroup=true)` calls.
+	if data.IsServiceAccount && omitGID && data.Groupname != "" {
```
jisung-02: Redundant check

omitGID := data.IsServiceAccount && data.GID == 0 at L132 guarantees that whenever omitGID is true, data.IsServiceAccount is also true.  The data.IsServiceAccount check can be removed from this branch.

pkg/executor/handlers/user/user.go
```diff
@@ -121,24 +104,7 @@ func (h *UserHandler) Validate(cmd string, args *common.CommandArgs) error {
...
-	}
-	if data.HomeDirectoryPermission == "" {
-		data.HomeDirectoryPermission = "0755"
-	}
-	if data.Shell == "" {
-		data.Shell = "/bin/bash"
-	}
+	data := userDataFromArgs(args)
 
 	err := h.Validate(common.AddUser.String(), args)
```
jisung-02: Redundant 

userDataFromArgs is called twice in handleAddUser — once directly at L107, and again inside Validate at L109. Replacing h.Validate(...) with h.ValidateStruct(data) reuses the already-built data and removes the duplication.

geunwoonoh: Fixed in fbd240b. Replaced the `h.Validate(...)` call with `h.ValidateStruct(data)` so the already-built `UserData` is reused instead of rebuilt inside the `Validate` AddUser branch. Added a `var err error` below to preserve the variable that the platform-specific `Executor.Run` branches reassign, since the previous `err := h.Validate(...)` used to declare it.

pkg/executor/handlers/user/user.go
geunwoonoh: Fixed in fbd240b. Dropped the redundant `data.IsServiceAccount` check — `omitGID` (defined as `data.IsServiceAccount && data.GID == 0` at the top of the function) already implies it. Added a one-line comment documenting the invariant so future readers do not reintroduce the check.

pkg/executor/handlers/user/types.go
```diff
@@ -1,16 +1,55 @@
...
+		Groupname:               args.Groupname,
+		Groups:                  args.Groups,
+		IsServiceAccount:        args.IsServiceAccount,
+	}
+	if data.HomeDirectoryPermission == "" {
+		data.HomeDirectoryPermission = "0755"
+	}
+	if data.Shell == "" {
+		data.Shell = "/bin/bash"
+	}
```
eunyoung14: **Blocker.** The `Shell` default (`/bin/bash`) is applied inside `userDataFromArgs` *before* `ValidateStruct` runs, so the `validate:"required"` tag on `Shell` can never fire. A payload that omits `shell` silently gets an interactive `/bin/bash` — which is exactly what we don't want for a service account, and arguably wrong for any path.

Two options:

1. Move the `Shell` default into `handleAddUser` after validation succeeds (and drop the `required` tag if we want it to be truly defaultable), or
2. Keep the default in one place but let the validator see the raw input. For example:

```go
func userDataFromArgs(args *common.CommandArgs) UserData {
    data := UserData{ /* fields as-is, no default */ }
    return data
}

// In handleAddUser, after ValidateStruct:
if data.Shell == "" {
    data.Shell = "/bin/bash"
}
if data.HomeDirectoryPermission == "" {
    data.HomeDirectoryPermission = "0755"
}
```

Either way, either `required` should hold, or it should be removed; today it's silently bypassed.

pkg/executor/handlers/user/user.go
```diff
@@ -154,65 +119,124 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
...
+			cmdArgs = append(cmdArgs, "--home-dir", data.HomeDirectory)
+		}
+		cmdArgs = append(cmdArgs, "--shell", data.Shell)
+		if !omitUID {
+			cmdArgs = append(cmdArgs, "--uid", strconv.FormatUint(data.UID, 10))
+		}
+		if !omitGID {
+			cmdArgs = append(cmdArgs, "--gid", strconv.FormatUint(data.GID, 10))
+		}
+		cmdArgs = append(cmdArgs, "--comment", data.Comment, "--create-home", data.Username)
```
eunyoung14, line 224: When `omitHome` is true, the `--home-dir`/`--home` flag is skipped, but `adduser --disabled-password <user>` on Debian and `useradd --create-home <user>` on RHEL both still default to creating `/home/<username>`. So the command emits no home flag, yet a home directory *is* created on disk. That can surprise the alpacon-server side if it assumes "no home" means "no home dir".

Options:
- Pass `--no-create-home` (RHEL) / `--no-create-home-dir` (Debian's `adduser` doesn't have a flag but `useradd --no-create-home` does via low-level) for service accounts, or
- Document this as intentional in the PR description / type comment so the contract is explicit.

Low priority but worth calling out in the design notes.

pkg/executor/handlers/user/user.go
```diff
@@ -154,65 +119,124 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
...
+	// triggers the early return below.
+	// Gated on `omitGID`: when GID is provided, the RHEL path already ran
+	// `groupadd --gid <GID> <Groupname>` and `useradd --gid <GID>`, making the
+	// user a member of Groupname as the primary group. Re-running groupadd/
+	// usermod there would be redundant. When GID is omitted (the typical
+	// service-account payload), useradd auto-creates a per-user primary group
+	// and the user is NOT in `Groupname`, which would break later
+	// `utils.Demote(..., ValidateGroup=true)` calls.
+	// omitGID already implies data.IsServiceAccount (see definition above).
+	if omitGID && data.Groupname != "" {
```
eunyoung14: The current logic relies on `useradd`/`adduser` auto-creating a per-user primary group and then adds `Groupname` as a supplementary membership after the fact. This works on default Debian/RHEL but diverges on systems with `USERGROUPS_ENAB no` in `/etc/login.defs` (or Debian `USERGROUPS=no` in `/etc/adduser.conf`), which do exist in hardened enterprise environments.

A cleaner shape that avoids the supplementary fixup entirely:

- Debian: `adduser --ingroup <Groupname> ...`  (omits both `--uid` and `--gid`, uses `Groupname` as primary)
- RHEL: `groupadd -f <Groupname>` → `useradd --gid <Groupname> ...`  (named gid rather than numeric)

Effects:
- Primary group = `Groupname` → `Demote(..., ValidateGroup=true)` satisfied without any supplementary step.
- No per-user private group, which is the expected service-account shape anyway.
- OS `USERGROUPS_ENAB` setting becomes irrelevant.
- Lines 213-238 collapse to roughly zero.

Downside: slightly different from the Linux "every user has a private group" convention. For Application service accounts I think this is actually *more* correct (they're shared accounts, not personal ones), but it's a small design call worth confirming with the server team.

Follow-up OK, but worth having on the radar before the design ossifies.

pkg/executor/handlers/user/user.go
```diff
@@ -154,65 +119,124 @@ func (h *UserHandler) handleAddUser(ctx context.Context, args *common.CommandArg
...
+		// later Demote calls.
+		if code, out, err := h.Executor.Run(ctx, "/usr/sbin/usermod", "-aG", data.Groupname, data.Username); code != 0 {
+			log.Warn().
+				Str("user", data.Username).
+				Str("group", data.Groupname).
+				Int("exitCode", code).
+				Str("output", out).
+				Err(err).
+				Msg("Failed to add service account to supplementary group")
+		}
```
eunyoung14: These failures are currently `log.Warn` and the handler still returns `User 'X' added successfully`. That matches the existing `Groups` supplementary pattern (line 241-246), but for service accounts the `Groupname` membership is *load-bearing* — subsequent `utils.Demote(..., ValidateGroup=true)` calls will fail, and alpacon-server will see a "succeeded" provisioning with a broken runtime.

Suggestions:
1. For service accounts specifically, make these failures surface a non-zero exit code or a warning in the success message so the caller can act.
2. Or, mirror the Groups pattern exactly: return `0, fmt.Sprintf("User '%s' created but failed to add to groupname: %v", ...)`.

Pre-existing pattern, so not blocking — but worth a note in the follow-up issue.

pkg/executor/handlers/user/user_test.go
```diff
@@ -154,6 +155,177 @@ func TestUserHandler_Execute(t *testing.T) {
...
+				t.Fatalf("Execute() unexpected error: %v", err)
+			}
+			if exitCode != 0 {
+				t.Fatalf("Execute() exitCode = %d, want 0", exitCode)
+			}
+
+			executed := mock.GetExecutedCommands()
+			var target *common.ExecutedCommand
+			var sawGidGroupadd, sawGroupaddDashF, sawUsermodAppendGroup bool
+			for i := range executed {
```
eunyoung14, line 222: Two gaps worth closing:

1. `IsServiceAccount=true` **with** UID/GID/home provided — current `omitUID` logic is `IsServiceAccount && UID == 0`, so a payload that sets both the flag and a numeric uid should honor the numeric uid. A test lock-in here prevents a future refactor from accidentally omitting the flag when both are present.
2. `groupadd -f` / `usermod -aG` failure on the service-account path — assert that a non-zero exit code from the supplementary group step produces the `log.Warn` and that `handleAddUser` still returns success (matching the intentional non-fatal behavior). Keeps the "silent success on failure" pattern visible and intentional.

Optional but small.

jisung-02 (dismissed):

pkg/executor/handlers/user/types.go
geunwoonoh: Fixed in 82ddedb. Removed the Shell default from `userDataFromArgs` so the `validate:"required"` tag on `Shell` actually fires when a payload omits it. The default was both dead code (alpacon-server always sends `shell` explicitly for IAM User and for Application service accounts) and a dangerous silent fallback to `/bin/bash` that could land an interactive shell on a service account. Added a comment so future maintainers don't reintroduce the default.

pkg/executor/handlers/user/user.go
geunwoonoh: Fixed in 82ddedb. The behavior is intentional — Application service accounts (gitlab-runner, etc.) typically need a home, and the contract on the alpacon-server side is "OS default home", not "no home". Two changes to make this explicit:

1. **Renamed** `omitUID`/`omitGID`/`omitHome` → `omitUIDFlag`/`omitGIDFlag`/`omitHomeFlag` so the variable names themselves encode the actual semantic: we are omitting the *flag* on adduser/useradd, not the underlying uid/group/home directory. The same ambiguity applied to all three (e.g. `omitUIDFlag=true` still results in a uid being assigned, just by the OS), so renaming all three keeps the symmetry.
2. Added a header comment listing what each `omit*Flag` actually means, including that `omitHomeFlag` uses the OS default path (typically `/home/<username>`) and the directory is still created.

If we ever need a true `--no-create-home` mode, that would be a separate explicit flag rather than overloading the missing-`home_directory` signal.

pkg/executor/handlers/user/user.go
geunwoonoh: Refactored as suggested in 82ddedb. Service-account creation now ensures the named group up front (`groupadd -f <Groupname>`) and passes the group as the primary group directly to the create command (`adduser --ingroup <Groupname>` on Debian, `useradd --gid <Groupname>` on RHEL — `useradd` accepts a group name as well as a numeric id). The post-fact `usermod -aG` step is gone, which removes the `USERGROUPS_ENAB no` / Debian `USERGROUPS=no` failure mode entirely.

pkg/executor/handlers/user/user.go
geunwoonoh: Addressed by the same refactor in 82ddedb. With `groupadd -f` now running before `adduser`/`useradd`, a failure on this step returns a non-zero exit code with `log.Error` (not `log.Warn`), so alpacon-server sees the failure rather than a "succeeded" provisioning that would break at `utils.Demote(..., ValidateGroup=true)` time. The new `TestUserHandler_AddUser_ServiceAccountGroupaddFails` test (in 86ed689) asserts both the non-zero return AND that `useradd` is not invoked when `groupadd` fails.

pkg/executor/handlers/user/user_test.go
geunwoonoh: Done in 86ed689. Added two tests covering the gaps:

- `TestUserHandler_AddUser_ServiceAccountWithExplicitUID`: a payload with `IsServiceAccount=true` AND explicit `UID/GID/HomeDirectory` honors all three numeric values and does NOT take the by-name `--gid alpacon` path, locking in the contract that `IsServiceAccount` alone does not strip flags — the value must also be zero/empty.
- `TestUserHandler_AddUser_ServiceAccountGroupaddFails`: a failing `groupadd -f` on the service-account path now produces a non-zero exit from `handleAddUser` and `useradd` is never invoked, matching the load-bearing semantics from the refactor.

Also updated the existing `TestUserHandler_AddUser_UidLess` to assert the new shape: `groupadd -f` runs BEFORE `adduser`/`useradd`, the create command sets the primary group by name (`--ingroup alpacon` on Debian, `--gid alpacon` on RHEL), and post-fact `usermod` is no longer invoked.

eunyoung14 (approved):

PR #297, opened 2026-05-06
ci: bump action versions to Node 24 runtimes and reorder checkout/setup-go

### LINKED ISSUE #266: CI: Resolve Node.js 20 deprecation warnings and cache restore failures
## Overview

The CI build's Annotations tab shows 34 warnings. The build itself passes, but the Node.js 20 deprecation must be addressed before June 2026.

## 1. Node.js 20 deprecation warnings

GitHub Actions runners will force-upgrade from Node.js 20 to Node.js 24 on June 2, 2026, and Node.js 20 will be fully removed on September 16, 2026. The following actions need to be upgraded to their latest major versions.

| Action | Current | Latest | Files |
|---|---|---|---|
| `actions/checkout` | v4 | **v6** | All |
| `actions/setup-go` | v5 | **v6** | All |
| `actions/upload-artifact` | v4 | **v7** | release.yml |
| `actions/download-artifact` | v4 | **v8** | release.yml |
| `golangci/golangci-lint-action` | v7 | **v9** | lint.yml |
| `goreleaser/goreleaser-action` | v6 | **v7** | release.yml |

## 2. Cache restore failure warnings

In `build-and-test.yml`, the `setup-go` step runs before `checkout`, so `go.sum` does not exist yet when the cache restore is attempted.

```
Dependencies file is not found in /_w/alpamon/alpamon. Supported file pattern: go.sum
```

**Fix**: Move the `checkout` step before `setup-go`. This is safe because git is already installed in the preceding dependency installation step for container jobs.

## Tasks

- [ ] Upgrade action versions across all workflow files
- [ ] Reorder checkout and setup-go steps in `build-and-test.yml`
- [ ] Verify warnings are resolved after CI run

Image

jisung-02:
## Summary

Resolves #266. Bumps every GitHub Action pinned to a Node.js 20 runtime to its latest Node 24 major, and fixes the `setup-go` cache-restore warning by reordering `checkout` before `setup-go`.

GitHub Actions runners force-upgrade Node.js 20 → 24 on **2026-06-02** and remove Node.js 20 entirely on **2026-09-16**. Without this change the workflows will break in June.

### 1. Action version bumps

| Action | From | To | Files |
|---|---|---|---|
| `actions/checkout` | v4 | **v6** | all workflows |
| `actions/setup-go` | v5 | **v6** | all workflows |
| `actions/upload-artifact` | v4 | **v7** | `release.yml` (4 places) |
| `actions/download-artifact` | v4 | **v8** | `release.yml` (2 places) |
| `golangci/golangci-lint-action` | v7 | **v9** | `lint.yml` |
| `goreleaser/goreleaser-action` | v6 | **v7** | `release.yml` |

### 2. Cache-restore failure fix (`build-and-test.yml`)

Previously `setup-go` ran before `checkout`, so `go.sum` was missing when the cache restore was attempted, producing:

```
Dependencies file is not found in /_w/alpamon/alpamon. Supported file pattern: go.sum
```

Reordered `Check out code` to run before `Set up Go` in all three jobs (Linux container matrix, Windows, macOS). The container jobs already install `git` in the preceding dependency-install step, so `checkout` has `git` available when it runs first.

## Breaking-change review

Release notes for each action were checked. Nothing in this codebase's usage pattern is impacted:

| Change | Impact here |
|---|---|
| `checkout@v6`: credentials stored in `$RUNNER_TEMP` instead of git config | `release.yml`'s `git fetch --tags` is read-only public — no impact |
| `setup-go@v6`: improved toolchain handling | `go.mod` has no `toolchain` directive and `GO_VERSION` matches `go 1.25.7` — no impact |
| `upload-artifact@v7`: ESM migration | No user-facing change |
| `download-artifact@v8`: hash mismatch is now an error; Content-Type-based auto-decompression | Same-workflow v7 upload → v8 download chain — no mismatch in normal operation |
| `golangci-lint-action@v9`: requires `golangci-lint >= v2.1.0` | Currently pinned to `v2.9.0` — satisfied. `working-directory` is not used |
| `goreleaser-action@v7`: ESM migration | Inputs unchanged — no impact |

## Test plan

Workflow changes can only be validated by an actual CI run.

- [ ] All 17 `Build and Test` Linux container matrix entries pass
  - [ ] Pay special attention to **rocky-8 / rhel-8** — glibc 2.28 sits at the lower bound of Node 24's prebuilt-binary requirements
- [ ] `Build and Test (windows-amd64)` passes
- [ ] `Build and Test (macos-arm64)` passes
- [ ] `golangci-lint` passes
- [ ] CI Annotations tab shows Node.js 20 deprecation warnings and the `go.sum not found` cache warnings are gone
- [ ] `release.yml` will be exercised on the next tag — monitor that release run

Closes #266

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -113,16 +113,16 @@ jobs:
         if: matrix.pkg_manager == 'dnf'
         run: dnf install -y --allowerasing git curl gcc glibc-devel tar
 
-      - name: Set up Go
-        uses: actions/setup-go@v5
-        with:
-          go-version: ${{ env.GO_VERSION }}
-
       - name: Check out code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Mark workspace as safe for Git
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: ${{ env.GO_VERSION }}
 
       - name: Generate go code
         run: go run -mod=mod entgo.io/ent/cmd/ent@${{ env.ENT_VERSION }} generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
@@ -138,13 +138,13 @@ jobs:
     name: Build and Test (windows-amd64)
     runs-on: windows-latest
     steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      - name: Check out code
-        uses: actions/checkout@v4
 
       - name: Generate go code
         run: go run -mod=mod entgo.io/ent/cmd/ent@${{ env.ENT_VERSION }} generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
@@ -185,13 +185,13 @@ jobs:
     name: Build and Test (macos-arm64)
     runs-on: macos-14
     steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      - name: Check out code
-        uses: actions/checkout@v4
 
       - name: Generate go code
         run: go run -mod=mod entgo.io/ent/cmd/ent@${{ env.ENT_VERSION }} generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -113,11 +113,11 @@ jobs:
         if: matrix.pkg_manager == 'dnf'
         run: dnf install -y --allowerasing git curl gcc glibc-devel tar
 
-      - name: Check out code
-        uses: actions/checkout@v6
-
       - name: Mark workspace as safe for Git
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Check out code
+        uses: actions/checkout@v6
 
       - name: Set up Go
         uses: actions/setup-go@v6
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,8 +25,8 @@ jobs:
       CGO_ENABLED: 0
 
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
         with:
           go-version: ${{ env.GO_VERSION }}
 
@@ -34,7 +34,7 @@ jobs:
         run: go run -mod=mod entgo.io/ent/cmd/ent@${{ env.ENT_VERSION }} generate --feature sql/modifier --target ./pkg/db/ent ./pkg/db/schema
 
       - name: GolangCI-Lint
-        uses: golangci/golangci-lint-action@v7
+        uses: golangci/golangci-lint-action@v9
         with:
           version: v2.9.0
           args: --timeout=5m
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,8 +25,11 @@ jobs:
       CGO_ENABLED: 0
 
     steps:
-      - uses: actions/checkout@v6
-      - uses: actions/setup-go@v6
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
         with:
           go-version: ${{ env.GO_VERSION }}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,7 +48,7 @@ jobs:
           echo "tag=${TAG_WITHOUT_V}" >> $GITHUB_OUTPUT
 
       - name: Check out code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           fetch-depth: 0
 
@@ -56,12 +56,12 @@ jobs:
         run: git fetch --force --tags
 
       - name: Set up Go
-        uses: actions/setup-go@v5
+        uses: actions/setup-go@v6
         with:
           go-version: ${{ env.GO_VERSION }}
 
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@v6
+        uses: goreleaser/goreleaser-action@v7
         with:
           distribution: goreleaser
           version: latest
@@ -129,25 +129,25 @@ jobs:
         run: echo "TAG_NAME=${{ steps.channel.outputs.tag }}" >> $GITHUB_ENV
 
       - name: Upload AMD64 DEB artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: alpamon_${{ env.TAG_NAME }}_linux_amd64.deb
           path: dist/alpamon_${{ env.TAG_NAME }}_linux_amd64.deb
 
       - name: Upload ARM64 DEB artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: alpamon_${{ env.TAG_NAME }}_linux_arm64.deb
           path: dist/alpamon_${{ env.TAG_NAME }}_linux_arm64.deb
 
       - name: Upload AMD64 RPM artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: alpamon-${{ env.TAG_NAME }}-1.linux.amd64.rpm
           path: dist/alpamon-${{ env.TAG_NAME }}-1.linux.amd64.rpm
 
       - name: Upload ARM64 RPM artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: alpamon-${{ env.TAG_NAME }}-1.linux.arm64.rpm
           path: dist/alpamon-${{ env.TAG_NAME }}-1.linux.arm64.rpm
@@ -179,7 +179,7 @@ jobs:
           fi
 
       - name: Download artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v8
         with:
           name: ${{ env.PKG_NAME }}
 
@@ -341,7 +341,7 @@ jobs:
           fi
 
       - name: Download artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v8
         with:
           name: ${{ env.PKG_NAME }}
 


.github/workflows/release.yml
```diff
@@ -129,25 +129,25 @@ jobs:
         run: echo "TAG_NAME=${{ steps.channel.outputs.tag }}" >> $GITHUB_ENV
 
       - name: Upload AMD64 DEB artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
```
geunwoonoh, line 132: Critical: `upload-artifact@v7` here is paired with `download-artifact@v8` later in the same workflow. GitHub requires upload and download artifact actions to share the same major version within a run; mixing v7 upload with v8 download invites the exact `hash mismatch is now an error` regression v8 introduced. Pin both to the same major.

```suggestion
        uses: actions/upload-artifact@v8
```
(or downgrade the `download-artifact` calls below to v7 — either direction works as long as they match.)

.github/workflows/release.yml
```diff
@@ -179,7 +179,7 @@ jobs:
           fi
 
       - name: Download artifact
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v8
```
geunwoonoh, line 182: Critical (paired with the upload-artifact comment above): `download-artifact@v8` does not match `upload-artifact@v7` used to produce these artifacts. Match the majors.

```suggestion
        uses: actions/download-artifact@v7
```

.github/workflows/release.yml
```diff
@@ -341,7 +341,7 @@ jobs:
...
```
geunwoonoh, line 344: Critical (same as above): keep this `download-artifact` major aligned with the `upload-artifact` major used in the producing job.

.github/workflows/build-and-test.yml
```diff
@@ -113,17 +113,17 @@ jobs:
...
         run: dnf install -y --allowerasing git curl gcc glibc-devel tar
 
-      - name: Set up Go
-        uses: actions/setup-go@v5
-        with:
-          go-version: ${{ env.GO_VERSION }}
-
       - name: Check out code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
```
geunwoonoh: Warning: please verify `actions/checkout@v6` is a real published tag (same for `setup-go@v6`, `upload-artifact@v7`, `download-artifact@v8`, `golangci-lint-action@v9`, `goreleaser-action@v7` elsewhere in this PR). Several of these jump 2+ majors in a single PR; a typo'd major silently resolves to "tag not found" only at runtime. Per AlpacaX security baseline, prefer pinning to a full commit SHA or `vX.Y.Z`.

.github/workflows/build-and-test.yml
```diff
@@ -113,17 +113,17 @@ jobs:
...
 
       - name: Mark workspace as safe for Git
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
geunwoonoh, line 117: Warning: with the reorder, this `safe.directory` step now runs *after* `checkout@v6`. The action it was meant to protect is now ahead of it. It works today because `checkout` configures its own per-repo safe.directory, but if `checkout@v6` ever reports a "dubious ownership" failure on rocky/rhel containers, the only fix is to move this step before `checkout` (or fold it into the `Install dependencies` step). Defensive option:

```suggestion
      - name: Mark workspace as safe for Git
        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"

      - name: Check out code
        uses: actions/checkout@v6
```
(moved above the checkout step)

.github/workflows/lint.yml
```diff
@@ -25,16 +25,16 @@ jobs:
       CGO_ENABLED: 0
 
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
```
geunwoonoh: Suggestion: anonymous step entries (no `name:`) are inconsistent with the rest of the repo's workflows, which name every step.

```suggestion
      - name: Check out code
        uses: actions/checkout@v6
      - name: Set up Go
        uses: actions/setup-go@v6
        with:
          go-version: ${{ env.GO_VERSION }}
```

geunwoonoh (changes requested):
## Review summary

**Stack**: github-actions (YAML workflows; no Go code touched)
**Changes**: +26 -26 across 3 files

Bumps GitHub Actions to majors that ship Node.js 24 runtimes (ahead of the 2026-06-02 forced upgrade) and reorders `checkout` to run before `setup-go` in `build-and-test.yml` so the Go module cache restore can find `go.sum`. No application code changed, so Go conventions / dead-code / runtime-perf review is N/A — review is scoped to workflow correctness.

---

### Critical issues (1)

1. **`upload-artifact@v7` paired with `download-artifact@v8` — major-version mismatch.** GitHub's official guidance (and the historical hard-fail behavior since v4) is that upload and download artifact actions must share the same major version within a workflow run. Even if v8's "Content-Type-based auto-decompression" handles v7 archives in the happy path, the PR description's own table notes "hash mismatch is now an error" in v8 — exactly the regression risk a major mismatch invites. Pin both to the same major.

### Warnings (3)

1. **Action version pins should be verified against actually-published releases.** `actions/checkout@v6`, `actions/setup-go@v6`, `actions/upload-artifact@v7`, `actions/download-artifact@v8`, `golangci/golangci-lint-action@v9`, `goreleaser/goreleaser-action@v7` — several jump 2+ majors at once (download-artifact v4→v8, golangci-lint-action v7→v9). Before merging, confirm each tag exists on the corresponding repo's releases page (a typo'd major silently resolves to "tag not found" only at runtime). Ideally pin to a full SHA or `vX.Y.Z` per AlpacaX security baseline rather than the floating major.

2. **`Mark workspace as safe for Git` now runs *after* `checkout@v6`.** Previously irrelevant because checkout ran before setup-go without a safe.directory issue, but with the reorder the `checkout` action itself executes git commands inside the container before this step. If `checkout@v6` ever reports a "dubious ownership" failure on rocky/rhel containers, the only fix is to move this step *before* `checkout` (or merge it into the `Install dependencies` step). It works today because `checkout` configures its own per-repo safe.directory, but the safety net the step was added for now sits behind the action it was meant to protect.

3. **`release.yml`'s `Determine release channel` step still runs before checkout** — unchanged behavior, but worth noting: if that step ever needs repo state it will silently see an empty workspace. Not introduced by this PR, just adjacent.

### Suggestions (2)

1. **lint.yml uses anonymous step entries** (no `name:`), inconsistent with the rest of the repo's workflows which name every step. Minor consistency nit.

2. **Single concern per PR**: this PR mixes (a) Node 24 prep and (b) the `setup-go` cache-restore reorder fix. They're both small and the description separates them clearly, so this is acceptable, but if action-version verification (warning #1) reveals issues on any single bump, splitting would let the cache-restore fix land independently.

### Highlights

- Good motivation in PR body — dates, force-upgrade timeline, per-action breaking-change table, and explicit call-out for rocky-8 / rhel-8 glibc bound on Node 24.
- The cache-restore reorder is a genuine perf win: `setup-go`'s module cache key is derived from `go.sum`, which previously didn't exist when `setup-go` ran first → cache always missed.
- Container jobs install `git` in the package-manager step, so `checkout` running first is safe.
- Description's "Breaking-change review" table per action is exactly the right level of due diligence for a CI bump.

---

The artifact upload/download major mismatch is the only blocker. Everything else is verification (tag existence) and a latent ordering concern around `safe.directory`.

**Checklist**:
- [ ] Match `upload-artifact` and `download-artifact` majors
- [ ] Verify each action tag is a real published release
- [ ] Move `Mark workspace as safe for Git` before `checkout` (defensive, optional)
- [ ] Watch the first CI run on rocky-8 / rhel-8 for Node 24 glibc issues per PR description

.github/workflows/release.yml
jisung-02: v7 upload + v8 download is a combination GitHub released as a deliberate pair. The [`download-artifact@v8` release notes](https://github.com/actions/download-artifact/releases/tag/v8.0.0) state the change was made "To support direct uploads in `actions/upload-artifact`", and the [official download-artifact README](https://github.com/actions/download-artifact/blob/main/README.md) uses `upload@v7` + `download@v8` as its standard example. The README also explicitly states "only supports files uploaded with `actions/upload-artifact@v7`".

Keeping the v7/v8 pair as-is. Full evidence in the review-summary reply.

.github/workflows/release.yml
jisung-02: v7 upload + v8 download is a combination GitHub released as a deliberate pair. The [`download-artifact@v8` release notes](https://github.com/actions/download-artifact/releases/tag/v8.0.0) state the change was made "To support direct uploads in `actions/upload-artifact`", and the [official download-artifact README](https://github.com/actions/download-artifact/blob/main/README.md) uses `upload@v7` + `download@v8` as its standard example. The README also explicitly states "only supports files uploaded with `actions/upload-artifact@v7`".

Keeping the v7/v8 pair as-is. Full evidence in the review-summary reply.

.github/workflows/release.yml
jisung-02: v7 upload + v8 download is a combination GitHub released as a deliberate pair. The [`download-artifact@v8` release notes](https://github.com/actions/download-artifact/releases/tag/v8.0.0) state the change was made "To support direct uploads in `actions/upload-artifact`", and the [official download-artifact README](https://github.com/actions/download-artifact/blob/main/README.md) uses `upload@v7` + `download@v8` as its standard example. The README also explicitly states "only supports files uploaded with `actions/upload-artifact@v7`".

Keeping the v7/v8 pair as-is. Full evidence in the review-summary reply.

jisung-02:
@geunwoonoh Thanks for the review. After verifying, the "upload/download artifact major version must match" concern doesn't apply to the v7/v8 pair. Here's the evidence:

### 1. v8 was explicitly designed to consume v7 uploads

`download-artifact@v8.0.0` release notes:
https://github.com/actions/download-artifact/releases/tag/v8.0.0

> "**To support direct uploads in `actions/upload-artifact`**, the action will no longer attempt to unzip all downloaded files. Instead, the action checks the `Content-Type` header ahead of unzipping and skips non-zipped files."

v8 download is the deliberately-paired counterpart to v7 upload's new `archive: false` behavior. The majors aren't out of sync — download bumped its major one extra step to add the compatibility handling for v7's new direct-upload feature.

### 2. The official `download-artifact` README uses v7 + v8 as the canonical example

https://github.com/actions/download-artifact/blob/main/README.md

Verbatim from the standard example:

```yaml
- name: Upload Artifact
  uses: actions/upload-artifact@v7
  ...
- name: Download All Artifacts
  uses: actions/download-artifact@v8
```

The same README also contains:

> "Directly download a non-zipped file (**only supports files uploaded with `actions/upload-artifact@v7`** and `archive: false` set)"

The maintainers explicitly document that v8 download supports files produced by v7 upload.

### Conclusion

- Current stable latest: `upload-artifact@v7.0.1` / `download-artifact@v8.0.1`
- These two majors were intentionally released as a pair, and the official README uses the combination as its recommended example.

I'll keep the v7 upload + v8 download combination in this PR as-is. I'll address Warning #2 (`safe.directory` ordering) and Suggestion #1 (`lint.yml` step names) separately.

.github/workflows/build-and-test.yml
jisung-02: Verified each action tag against the upstream releases API — all real:

- `actions/checkout@v6` (latest v6.0.2)
- `actions/setup-go@v6` (latest v6.4.0)
- `actions/upload-artifact@v7` (latest v7.0.1)
- `actions/download-artifact@v8` (latest v8.0.1)
- `golangci/golangci-lint-action@v9` (latest v9.2.0)
- `goreleaser/goreleaser-action@v7` (latest v7.2.1)

On the SHA / `vX.Y.Z` pinning point — the repo currently isn't pinning any action this way, so this PR kept the existing convention. Should we adopt pinning across all workflows? If yes, I'll open a follow-up PR for it.

.github/workflows/build-and-test.yml
jisung-02: Fixed in 2b3e3bc — moved `Mark workspace as safe for Git` to run before `actions/checkout@v6` in the Linux container matrix job, so the workspace is marked safe before checkout executes any git commands inside the container.

.github/workflows/lint.yml
jisung-02: Fixed in 2b3e3bc — named the previously-anonymous `checkout` and `setup-go` steps (`Check out code` / `Set up Go`) to match the rest of the repo's workflows.

geunwoonoh (approved):

PR #319, opened 2026-05-25
feat: pull-based plugin-config SDK (configreceiver + pluginclient)

eunyoung14:
## Summary

Phase 0 SDK additions used by ``alpamon-dhcp-plugin``, ``alpamon-dns-plugin``, and ``alpamon-proxy-plugin``. Two new packages:

**``pkg/configreceiver``** — agent-side pull flow. Plugins embed a ``Receiver`` and supply an ``Applier``. On a ``config_updated`` WS event the receiver fetches the body via ``GET /api/plugins/configs/{id}/``, verifies the sha256, and applies. Reports back via ``POST /applied/``. This replaces the legacy pattern where the server pushed full config text inside the WS payload.

**``pkg/pluginclient``** — shared WebSocket command loop. dhcp / dns / proxy each copy-pasted ~100 lines of read-limit + query whitelist + dispatch + reconnect + ping handling. The new ``Client`` exposes a small ``RunForever(ctx)`` + ``HandleMessage(ctx, msg)`` surface; plugins keep only their ``Applier`` and a small legacy-reconfigure callback that decodes their plugin-private ``Command`` payload during the dual-emit rollout window.

Side effects of the move: dhcp inherits the 10 MiB read limit + query whitelist that dns/proxy already had (they were missing in the hand-rolled dhcp loop).

## PR ordering

**#3 of 9.** Independent of #1/#2. Required by:
- #4 alpamon-dhcp-plugin
- #5 alpamon-dns-plugin
- #6 alpamon-proxy-plugin

Merge before any of those three. They have ``replace github.com/alpacax/alpamon => ../alpamon`` directives for local dev that need to be removed and the ``require`` line bumped once this lands and is tagged.

## Test plan

- [x] ``go test ./pkg/configreceiver/...`` (6 tests: success, hash mismatch, decode failure, applier failure, fetch failure, sha256 sanity)
- [x] ``go test ./pkg/pluginclient/...`` (handler size validation, query whitelist, missing-id rejection, ``New`` shape, etc.)
- [x] e2e against dns and cert agents — both connect, receive frames, apply via ``configreceiver``


pkg/configreceiver/receiver.go
```diff
@@ -0,0 +1,201 @@
...
+	fetchURLFmt   = "/api/plugins/configs/%s/"
+	appliedURLFmt = "/api/plugins/configs/%s/applied/"
+	// requestTimeout caps each REST call in seconds.
+	//
+	// alpamon's scheduler.Session multiplies the caller's timeout by
+	// time.Second internally (see session.do), so this value is a
+	// plain int count of seconds and is intentionally NOT a
+	// time.Duration — passing a Duration here would overflow int64
+	// and trip the context deadline immediately.
+	requestTimeout time.Duration = 30
```
junho226, line 50: Warning: the type and the comment contradict each other. The const is declared as `time.Duration`, but the comment two lines up says it's "intentionally NOT a time.Duration." That's comment-vs-type drift — a future reader looking at the type sees one contract, the next person reading the comment sees a different one.

Suggest changing the type to `int` so the documented constraint is enforced by the compiler:

```suggestion
	requestTimeout int = 30
```

(The comment block above stays accurate once the type matches.)

pkg/configreceiver/receiver.go
```diff
@@ -0,0 +1,201 @@
...
+
+// maxReportedErrorLen caps the error string posted to the server.
+// Applier and fetch errors can include full upstream response bodies
+// or large stack traces; truncating prevents DB bloat and accidental
+// secret leakage via log/error fields.
+const maxReportedErrorLen = 512
+
+func (r *Receiver) reportError(id, errMsg string) {
+	if len(errMsg) > maxReportedErrorLen {
+		errMsg = errMsg[:maxReportedErrorLen] + "...(truncated)"
```
junho226: Warning: `errMsg[:maxReportedErrorLen]` slices by bytes, not runes. If the 512th byte lands inside a multi-byte UTF-8 sequence — likely whenever upstream surfaces a Korean stack trace, a non-ASCII config path, or a binary-ish blob via `err.Error()` — the truncated string is invalid UTF-8. Go's JSON encoder silently replaces with U+FFFD, but the alpacon UI ends up showing a garbled last character right before `...(truncated)`.

```suggestion
		errMsg = strings.ToValidUTF8(errMsg[:maxReportedErrorLen], "") + "...(truncated)"
```

(also add `"strings"` to the import block). Same fix is worth applying to `preview[:200]` at `pkg/pluginclient/client.go:122`.

junho226 (commented):
## PR Review: #319 — feat: pull-based plugin-config SDK (configreceiver + pluginclient)

**Author**: @eunyoung14
**Branch**: `feat/pluginclient` → `main`
**Stack**: golang
**Changes**: +813 −0 across 4 files (2 new packages, 2 test files)

---

### Summary

Introduces two new packages that consolidate copy-pasted logic across the dhcp/dns/proxy plugins:

- **`pkg/configreceiver`** — pull-based config flow. Receives a `config_updated` WS event, fetches the body via REST, verifies the sha256, decodes the `{files, metadata}` envelope, calls a plugin-supplied `Applier`, and reports success/failure.
- **`pkg/pluginclient`** — shared WebSocket command loop. Owns size limit, query whitelist, dispatch, ping/quit/reconnect/restart, and bridges to `configreceiver`. Plugins now bring only an `Applier` and an optional legacy `reconfigure` callback.

The design is well-documented (package comments explain the *why*, not just the *what*), the dual-emit rollout strategy is thoughtful, and `configreceiver` test coverage is solid.

### Critical Issues (0)

None — nothing here blocks merge.

### Warnings (2)

1. **`pkg/configreceiver/receiver.go:48`** — `requestTimeout` type contradicts its own comment. Declared as `time.Duration` but the comment explicitly says "intentionally NOT a time.Duration." Future readers see two different contracts on the same identifier. Either change the type to `int` (lets the compiler enforce the documented constraint) or fix the comment.
2. **`pkg/configreceiver/receiver.go:174`** — byte-slice truncation can split UTF-8 mid-rune. `errMsg[:maxReportedErrorLen]` slices by *bytes*; if the 512th byte lands inside a multi-byte rune (very plausible with a Korean stack trace or a binary-ish blob in `err.Error()`), the result is invalid UTF-8. JSON encoding will silently swap in U+FFFD, but the alpacon UI ends up rendering a mojibake last char. Same concern at `pkg/pluginclient/client.go:122` (`preview[:200]`). `strings.ToValidUTF8` or `utf8.RuneStart`-based backoff fixes both.

### Suggestions (3)

1. **`pkg/pluginclient/client.go:152`** — `go c.Receiver.Handle(ctx, ...)` spawns an unbounded goroutine per event. `applyMu` serialises them, but if `Apply` blocks (systemctl reload stuck, etc.) every subsequent `config_updated` queues another live goroutine. A 1-deep buffered channel + single worker would give natural coalescing (only the latest config ID survives, which is the right behaviour since newer configs supersede older ones). Worth deciding now since dhcp/dns/proxy will inherit this.
2. **`pkg/configreceiver/receiver.go` `Handle`** — `ctx` is accepted but never forwarded to `fetch`/`post`. If the caller cancels mid-apply, in-flight REST calls keep running. If `scheduler.Session` doesn't accept a context today, a `TODO` would at least mark this for when it does.
3. **`pkg/pluginclient/client_test.go`** — rejection-path coverage is good, but there's no positive test confirming a well-formed `config_updated` actually reaches `Receiver.Handle`. Making `Receiver` an interface (or accepting a function) for the dispatch test would lock the dispatch wiring in.

### Highlights

- Package-level comments are unusually thorough — they document the *why* (release-coordination rationale, dual-emit window) not just the API surface.
- `applyMu` serialisation with an explicit rationale comment about why interleaving would be incorrect.
- The PR description's "side effect" call-out (dhcp inheriting the 10 MiB read limit and whitelist) is exactly the kind of behaviour-change disclosure reviewers need.
- Defensive nil-handling in `setReadLimit`, `Handle`, `HandleMessage`, and `RunForever`.
- `TestSha256HexMatchesStdlib` as a paranoia sanity check is a nice touch.
- `fakeServer` + `recordingApplier` keep the receiver tests easy to read.

---

**Verdict**: COMMENTED (non-blocking)

The two warnings are clarity/edge-case concerns, not breakage. Worth fixing the `requestTimeout` type drift before merge since this PR is the foundation for three downstream plugin PRs (#4–#6) and the footgun will propagate.

**Checklist**:
- [x] Code follows golang conventions
- [x] Tests included for new packages
- [x] No security vulnerabilities found
- [x] Documentation / package comments are excellent

---
🤖 Auto-review by junho-bot during junho226's leave. If this review looks off, please disregard — junho will follow up upon return.

pkg/pluginclient/client.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+		if payload.PluginConfigID == "" {
+			log.Error().Msg("config_updated event missing plugin_config_id")
+			return
+		}
+		if c.Receiver == nil {
+			log.Warn().Msg("Cannot process config_updated: Receiver is nil")
+			return
+		}
+		go c.Receiver.Handle(ctx, payload.PluginConfigID)
```
junho226: Suggestion: this spawns an unbounded goroutine per `config_updated` event. They all serialise on `applyMu`, but if `Apply` blocks (e.g., a systemctl reload waiting on DHCP leases), every subsequent `config_updated` queues another live goroutine that's still pinning `ctx` and the payload.

A 1-deep buffered channel fed from here and drained by a single worker gives you natural coalescing — if N events arrive during a slow apply, only the latest plugin_config_id survives, which is the correct behaviour since the newest config supersedes older ones anyway.

Not blocking, but worth deciding now since dhcp/dns/proxy will all inherit this behaviour from the shared dispatcher.

eunyoung14:
@junho226 thanks for the deep review — all 8 items now addressed across [f17894a](https://github.com/alpacax/alpamon/commit/f17894a), [1f3c5ee](https://github.com/alpacax/alpamon/commit/1f3c5ee), and [fa12e37](https://github.com/alpacax/alpamon/commit/fa12e37):

**Critical / Warnings**

- **Receiver.Handle nil-deref panic** (Copilot, receiver.go:97) — guard added at the top of Handle: if r/Session/Applier is nil, log and return without panicking.
- **requestTimeout type vs comment** (junho, receiver.go:49) — comment rewritten to describe the actual scheduler.Session.do quirk (multiplies by time.Second, so bare 30 resolves to 30s) and explicitly state the type stays `time.Duration` because that's the Session method signature.
- **UTF-8 byte slicing in reportError** (junho, receiver.go:174) — switched to rune-aware truncation (`[]rune(errMsg)[:512]`). Korean stack traces or non-ASCII paths can't split mid-codepoint anymore.
- **JSON unmarshal 10 MiB allocation on bad frame** (Copilot, client.go:128) — slice the byte buffer before converting to string for the log preview (`string(message[:200])`), so a 10 MiB invalid frame doesn't allocate 10 MiB just to render 200 chars.
- **reportError sanitization** (Copilot, receiver.go:174) — added `maxReportedErrorLen = 512` rune cap (see above). fetch errors with full upstream bodies and applier errors no longer flood the server's DB / leak details.
- **RunForever initial Connect ignores ctx** (Copilot, client.go:204) — doc comment now spells out that the *initial* `WsClient.Connect()` call honours its internal retry timeout, not ctx; cancellation kicks in once the read loop starts. Matches existing `runner.WebsocketClient.Connect` semantics.

**Unbounded goroutine pile** (junho + Copilot, client.go:158) — `go c.Receiver.Handle(...)` is gone. Replaced with a single `configWorker` goroutine plus a buffered-of-1 channel; new events use last-write-wins (drop the pending value, enqueue newest) because PluginConfig versions are absolute and monotonic — applying the latest supersedes any pending intermediates. One live goroutine per Client regardless of event volume, and slow Applies can no longer back up live goroutines.

**Module path migration** — separately, the v2 import bump merged on main (#317); this branch rebased on main and updated `github.com/alpacax/alpamon/pkg/...` → `.../v2/pkg/...`. Downstream plugin agents got the same bump on their feature branches.

Re-review please when you have a moment.

junho226 (approved):
## Re-Review: #319 — feat: pull-based plugin-config SDK (configreceiver + pluginclient)

**Author**: @eunyoung14
**Branch**: `feat/pluginclient` → `main`
**Stack**: golang
**Re-review of commits**: f17894a, 1f3c5ee, fa12e37 (head: e608938)

---

### Summary

Re-review of the eight items raised in the [first pass](https://github.com/alpacax/alpamon/pull/319#pullrequestreview-PRR_kwDOM3YzSM8AAAABA6ikcg) (mine + Copilot). All eight are now addressed; the fixes are thorough and a couple go beyond the minimum:

| # | Item | Status |
|---|------|--------|
| 1 | `requestTimeout` type vs comment drift (receiver.go:48) | ✅ Comment rewritten to describe the `scheduler.Session.do` ×`time.Second` quirk and explicitly justify why the type stays `time.Duration` |
| 2 | UTF-8 byte-slicing in `reportError` (receiver.go:174) | ✅ New `truncateUTF8` helper walks back to the nearest rune start; covered by `TestTruncateUTF8` with Korean rune cases and `TestReportErrorTruncates` with `utf8.ValidString` assertion |
| 3 | Unbounded goroutine pile on `config_updated` (client.go:158) | ✅ Replaced with single `configWorker` + buffered-of-1 channel, last-write-wins coalescing. One live goroutine per `Client` regardless of event rate; correct given config versions are absolute and monotonic |
| 4 | JSON unmarshal 10 MiB allocation on bad frame (client.go:128) | ✅ Slices the byte buffer first, then converts the preview to string |
| 5 | `Receiver.Handle` nil-deref panic (receiver.go:97) | ✅ Guard added at the top of `Handle` — logs and returns instead of panicking |
| 6 | `RunForever` initial `Connect` ignores ctx (client.go:204) | ✅ Doc comment spells out the asymmetry (initial connect uses internal retry; ctx kicks in once the read loop starts), matching existing `runner.WebsocketClient.Connect` semantics |
| 7 | `reportError` sanitization / DB bloat (receiver.go:174) | ✅ `maxReportedErrorBytes = 2048` cap with rune-safe truncation. The chosen byte-based cap (vs rune-based) is justified inline — avoids allocating a full rune slice for a 10 MiB error string |
| 8 | Fetch body preview cap (receiver.go:160-164) | ✅ Bonus: 4xx/5xx upstream bodies are now also truncated via `truncateUTF8(..., maxBodyPreviewBytes)`. Wasn't strictly asked for, but the same flood concern applied here |

### Notes on the goroutine refactor (item 3)

The `enqueueConfigUpdate` / `configWorker` pattern is the right shape:

- **Bounded memory** — one worker goroutine, one slot in the channel, regardless of event burst rate.
- **Last-write-wins is correct here** — `PluginConfig` versions are absolute, so applying only the newest pending ID is equivalent to applying all of them. The inline comment makes the invariant clear for future readers.
- **Lazy worker start via `sync.Once`** — avoids a goroutine for plugins that never receive a `config_updated` (e.g. during tests). Subtle but reasonable.
- **Triple-`select` saturation handling** — the third-select default branch ("Worker raced us") only fires under concurrent producers, which doesn't happen in single-reader `RunForever`; harmless safety net.

### Minor observations (non-blocking)

- `pkg/configreceiver/receiver.go` `fetch`/`post` still don't take `ctx`. As before, this is a `scheduler.Session` limitation — not addressable in this PR. No action needed.
- `TestEnqueueConfigUpdate_LastWriteWins` cleverly tests the saturation path by killing the worker with a pre-cancelled ctx, but there's no positive test of the worker actually draining and forwarding to `Receiver.Handle`. Not blocking, but worth adding when convenient.

### Highlights of this revision

- Comments explain *why* (the `do` multiplier quirk, byte-vs-rune cap tradeoff, last-write-wins rationale) rather than just *what*.
- New tests added for the new code paths (`TestTruncateUTF8`, `TestReportErrorTruncates`, `TestEnqueueConfigUpdate_LastWriteWins`) — fixes don't ship without coverage.
- The fix for item 7 (`reportError` cap) was extended to item 8 (`fetch` body preview) without being asked. Good instinct.

---

**Verdict**: APPROVE ✅

Ready to merge. Downstream PRs (#4 dhcp, #5 dns, #6 proxy) can now have their `replace github.com/alpacax/alpamon => ../alpamon` directives removed and the `require` line bumped once this is tagged.

**Checklist**:
- [x] All 8 prior-review items addressed
- [x] Tests added for new code paths
- [x] No new issues introduced
- [x] golang conventions followed
- [x] No security concerns

---
🤖 Auto-review by junho-bot during junho226's leave. If this review looks off, please disregard — junho will follow up upon return.

PR #347, opened 2026-07-03
fix(collector): correct realtime metric query window unit

jisung-02:
## Background

The realtime `net` and `disk_io` send checks computed their query window as `now.Add(-1 * interval * time.Second)`. But `interval` is already a `time.Duration`, built in `collector.go` as `time.Duration(entry.Interval) * time.Second`. The extra `* time.Second` applied the unit a second time.

This line has existed since the realtime NET/DISK_IO refactor (788e527, 2025-01-07) and contradicts how `scheduler.go` uses the same value directly via `now.Add(interval)` — which is the correct pattern. It is a unit-confusion bug, not intended behavior.

## Changes

Removed the redundant `* time.Second` in `net_send.go` and `io_send.go` and subtract the interval directly:

```go
from := now.Add(-interval)
```

The window now covers exactly the previous scheduler cycle.

## Impact

`interval * time.Second` is a `time.Duration` (int64 nanoseconds) multiplication, so for interval values of 10s or more it overflows int64 and wraps via two's-complement, landing on either side of "now" depending on the exact value — not simply "future for 10s+, past otherwise" as originally described here. Verified by execution:

| Configured interval | Resulting `from` (before fix) | Effect |
| --- | --- | --- |
| 1s – 9s | ~1741 – 1994 (past) | Whole-table aggregation every tick |
| 10s, 30s | ~2244 – 2294 (future) | `from > now`, nothing sent |
| 20s, 60s, 300s | ~1872 – 1977 (past) | Whole-table aggregation (stale/incorrect values still sent) |

In every case the query window was wrong; whether it silently sent incorrect aggregates or silently sent nothing depended on the specific interval's overflow arithmetic. After the fix, `from` is always `now - interval`, so only the previous cycle is aggregated regardless of the configured interval.

## Tests

`go test ./pkg/collector/check/realtime/disk/io/... ./pkg/collector/check/realtime/net/...` passes for both changed packages. The `pkg/db/ent/migrate` Atlas-dialect go.sum error surfaced by the local hooks is a pre-existing issue unrelated to this change.

Added regression tests (`TestGetTraffic`, `TestGetDiskIO`) that seed a row at `now - 2s` and assert it's excluded from the queryset — the previous tests only inserted a row at `now`, so they passed against both the buggy and the fixed window and couldn't catch a reintroduction of the bug.

### DIFF
--- a/pkg/collector/check/realtime/disk/io/io_send.go
+++ b/pkg/collector/check/realtime/disk/io/io_send.go
@@ -62,7 +62,7 @@ func (c *SendCheck) getDiskIO(ctx context.Context) ([]base.DiskIOQuerySet, error
 	client := c.GetClient()
 	interval := c.GetInterval()
 	now := time.Now()
-	from := now.Add(-1 * interval * time.Second)
+	from := now.Add(-interval)
 
 	var querySet []base.DiskIOQuerySet
 	err := client.DiskIO.Query().
--- a/pkg/collector/check/realtime/disk/io/io_send.go
+++ b/pkg/collector/check/realtime/disk/io/io_send.go
@@ -74,9 +74,6 @@ func (c *SendCheck) getDiskIO(ctx context.Context) ([]base.DiskIOQuerySet, error
 			ent.As(ent.Mean(diskio.FieldReadBps), "avg_read_bps"),
 			ent.As(ent.Mean(diskio.FieldWriteBps), "avg_write_bps"),
 		).Scan(ctx, &querySet)
-	if err != nil {
-		return querySet, err
-	}
 
-	return querySet, nil
+	return querySet, err
 }
--- a/pkg/collector/check/realtime/net/net_send.go
+++ b/pkg/collector/check/realtime/net/net_send.go
@@ -82,9 +82,6 @@ func (c *SendCheck) getTraffic(ctx context.Context) ([]base.TrafficQuerySet, err
 			ent.As(ent.Mean(traffic.FieldOutputPps), "avg_output_pps"),
 			ent.As(ent.Mean(traffic.FieldOutputBps), "avg_output_bps"),
 		).Scan(ctx, &querySet)
-	if err != nil {
-		return querySet, err
-	}
 
-	return querySet, nil
+	return querySet, err
 }
--- a/pkg/collector/check/realtime/net/net_send.go
+++ b/pkg/collector/check/realtime/net/net_send.go
@@ -66,7 +66,7 @@ func (c *SendCheck) getTraffic(ctx context.Context) ([]base.TrafficQuerySet, err
 	client := c.GetClient()
 	interval := c.GetInterval()
 	now := time.Now()
-	from := now.Add(-1 * interval * time.Second)
+	from := now.Add(-interval)
 
 	var querySet []base.TrafficQuerySet
 	err := client.Traffic.Query().
--- a/pkg/collector/check/realtime/disk/io/io_test.go
+++ b/pkg/collector/check/realtime/disk/io/io_test.go
@@ -89,9 +89,21 @@ func (suite *DiskIOCheckSuite) TestGetDiskIO() {
 		SetWriteBps(rand.Float64()).Exec(suite.ctx)
 	assert.NoError(suite.T(), err, "Failed to create disk io.")
 
+	staleDevice := "stale-" + uuid.NewString()
+	err = suite.collectCheck.GetClient().DiskIO.Create().
+		SetTimestamp(time.Now().Add(-2 * time.Second)).
+		SetDevice(staleDevice).
+		SetReadBps(rand.Float64()).
+		SetWriteBps(rand.Float64()).Exec(suite.ctx)
+	assert.NoError(suite.T(), err, "Failed to create stale disk io.")
+
 	querySet, err := suite.sendCheck.getDiskIO(suite.ctx)
 	assert.NoError(suite.T(), err, "Failed to get disk io queryset.")
 	assert.NotEmpty(suite.T(), querySet, "Disk IO queryset should not be empty")
+
+	for _, row := range querySet {
+		assert.NotEqual(suite.T(), staleDevice, row.Device, "Disk IO queryset should exclude rows outside the interval window")
+	}
 }
 
 func TestDiskIOCheckSuite(t *testing.T) {
--- a/pkg/collector/check/realtime/net/net_test.go
+++ b/pkg/collector/check/realtime/net/net_test.go
@@ -93,9 +93,23 @@ func (suite *NetCheckSuite) TestGetTraffic() {
 		SetOutputBps(rand.Float64()).Exec(suite.ctx)
 	assert.NoError(suite.T(), err, "Failed to create traffic.")
 
+	staleName := "stale-" + uuid.NewString()
+	err = suite.sendCheck.GetClient().Traffic.Create().
+		SetTimestamp(time.Now().Add(-2 * time.Second)).
+		SetName(staleName).
+		SetInputPps(rand.Float64()).
+		SetInputBps(rand.Float64()).
+		SetOutputPps(rand.Float64()).
+		SetOutputBps(rand.Float64()).Exec(suite.ctx)
+	assert.NoError(suite.T(), err, "Failed to create stale traffic.")
+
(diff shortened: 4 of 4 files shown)


pkg/collector/check/realtime/net/net_send.go
```diff
@@ -66,7 +66,7 @@ func (c *SendCheck) getTraffic(ctx context.Context) ([]base.TrafficQuerySet, err
 	client := c.GetClient()
 	interval := c.GetInterval()
 	now := time.Now()
-	from := now.Add(-1 * interval * time.Second)
+	from := now.Add(-interval)
```
geunwoonoh, line 69: Warning: the fix on this line is correct, but `TestGetTraffic` (net_test.go:86) does not pin it—verified empirically that the test also passes on the old buggy code. The test interval is 1s, so the row inserted at `time.Now()` falls inside both the old window (~31.7 years) and the new one (1s).

A regression test that seeds a row outside the corrected window and asserts it is excluded would catch any reintroduction:

```go
// Excluded by the new code, included by the old code—catches a regression
err := suite.sendCheck.GetClient().Traffic.Create().
    SetTimestamp(time.Now().Add(-2 * time.Second)).
    SetName("stale-row").
    SetInputPps(rand.Float64()).
    SetInputBps(rand.Float64()).
    SetOutputPps(rand.Float64()).
    SetOutputBps(rand.Float64()).Exec(suite.ctx)
// ... assert that "stale-row" is absent from the queryset
```

SQLite timestamp precision (TEXT with nanoseconds preserved) and the GTE boundary inclusivity were both checked—this test discriminates correctly in both directions.

pkg/collector/check/realtime/disk/io/io_send.go
```diff
@@ -62,7 +62,7 @@ func (c *SendCheck) getDiskIO(ctx context.Context) ([]base.DiskIOQuerySet, error
 	client := c.GetClient()
 	interval := c.GetInterval()
 	now := time.Now()
-	from := now.Add(-1 * interval * time.Second)
+	from := now.Add(-interval)
```
geunwoonoh, line 65: Warning: same as the net check—`TestGetDiskIO` (io_test.go:84) does not pin this fix; it passes on the old buggy code as well (1s test interval, row seeded at `time.Now()` falls inside both the old and new windows). Recommend adding a regression test that seeds a `DiskIO` row at `time.Now().Add(-2 * time.Second)` with a distinct device and asserts it is excluded from the queryset.

pkg/collector/check/realtime/net/net_send.go
geunwoonoh, line 69: Suggestion: pre-existing design behavior, now observable after this fix (applies to io_send.go:65 as well). The scheduler runs on a 1s ticker with a strict `now.After(nextRun)` comparison (scheduler.go:106), so actual execution spacing always exceeds the interval, while this query window is exactly one interval wide anchored at execution time. Consecutive windows can therefore gap (rows never aggregated) or overlap (double-counted, since GTE/LTE are both inclusive). Previously this was masked by the decades-wide buggy window.

Not a defect of this diff, and no data is permanently lost (the hourly batch reads the raw tables independently)—but worth recording as a follow-up issue.

geunwoonoh (commented):
## Summary

This PR removes a unit-confusion bug in the realtime `net` and `disk_io` send checks: `interval` is already a `time.Duration`, but the query window multiplied it by `time.Second` a second time. The fix subtracts the interval directly via `now.Add(-interval)`.

**The fix itself is correct and complete.** Verified directly:

- `GetInterval()` returns `time.Duration` (`base.go:46`), built at `collector.go:125` as `time.Duration(entry.Interval) * time.Second`—so the old expression did apply the unit twice.
- Consistent with the codebase's existing patterns: `now.Add(task.interval)` in `scheduler.go:107` and `-1 * c.GetInterval()` in `alert.go:27`.
- **Repo-wide sweep for the bug pattern**: no other site has the double-unit bug (batch checks correctly use literal `-1 * time.Hour` etc.). Nothing was missed.
- The origin commit cited in the PR body (788e527, 2025-01-07 realtime NET/DISK_IO refactor) checks out.
- Build, `go vet`, and `gofmt` are clean; the full test suite (`go test ./... -p 1`) passes (exit 0).
- An adversarial review loop (Security + Correctness lenses, 2 rounds) found no issues beyond the notes below. One reviewer built origin/main in a separate worktree and empirically confirmed the old code's behavior.

### Critical issues (0)

None.

### Warnings (2)

1. **[pkg/collector/check/realtime/net/net_test.go:86]** `TestGetTraffic` does not pin this fix—verified empirically that it also passes on the old code (origin/main). The test interval is 1s, so the row inserted at `time.Now()` falls inside both the old window (~31.7 years) and the new one (1s). A regression test asserting that rows outside the window are excluded is recommended (see inline comment).
2. **[pkg/collector/check/realtime/disk/io/io_test.go:84]** `TestGetDiskIO` fails to pin the fix for the same reason. The same kind of exclusion test is recommended.

### Suggestions (3)

1. **[PR description impact table]** "10s or more → `from` pushed into the future → nothing sent" only holds for some values. The overflow wrap direction depends on the interval, verified by execution:

   | Interval | Old `from` | Actual result |
   |---|---|---|
   | 1s–9s | 1994–1741 (past) | Whole-table aggregation |
   | 10s, 30s | 2294, 2244 (future) | Nothing sent |
   | 20s, 60s, 300s | 1977, 1878, 1872 (past) | Whole-table aggregation (incorrect values still sent) |

   Correcting the PR body to something like "10s or more overflows int64 and wraps, landing either in the future (transmission stops) or in the distant past (whole-table aggregation) depending on the value" would keep the change record accurate. This does not affect the direction of the fix.

2. **[net_send.go:85-89, io_send.go:77-81]** (pre-existing, out of scope) The `return querySet, nil` after `if err != nil { return querySet, err }` can be collapsed into a single `return querySet, err`. The two files being nearly structurally identical is also a future consolidation candidate.

3. **[net_send.go:69, io_send.go:65]** (pre-existing design, now observable after this fix) The scheduler runs on a 1s ticker with a strict `now.After(nextRun)` comparison (`scheduler.go:106`), so actual execution spacing always exceeds the interval, while the query window is exactly one interval wide anchored at execution time. Consecutive windows can therefore gap (rows never aggregated) or overlap (double-counted, since GTE/LTE are both inclusive). Not a defect of this diff, and no data is lost since the hourly batch reads the raw tables independently—but worth recording as a follow-up issue.

### Highlights

- A minimal 2-line diff that removes exactly the root cause. `now.Add(-interval)` is idiomatic Go and matches the codebase's existing patterns (scheduler, alert).
- **Performance improvement**: for configurations where the interval wrapped into the past, every send cycle used to aggregate the entire retained table (up to ~25 hours of rows depending on cleanup timing); it now scans only one cycle's worth. No dead code or new duplication introduced.
- The PR body is well-grounded: origin commit, contrast with the correct pattern, and test results are all documented.

---

**Verdict**: COMMENTED (no critical issues; the fix itself is approvable, with a recommendation to strengthen regression tests)

**Checklist**:
- [x] Code follows stack conventions (gofmt/vet clean, idiomatic)
- [ ] Tests included and passing—tests pass, but no regression test pins this fix
- [x] No security vulnerabilities (Security lens, 2 rounds, no findings)
- [x] Documentation updated if needed—recommend correcting part of the PR body impact table

jisung-02:
Thanks for the thorough review. Addressed the actionable points:

- **Warnings 1 & 2** (`net_test.go`, `io_test.go` didn't pin the fix): extended `TestGetTraffic` and `TestGetDiskIO` to seed a row at `now - 2s` and assert it's excluded from the queryset. Verified these now fail against the old `-1 * interval * time.Second` formula and pass against the fix (462867d).
- **Suggestion 1** (impact table accuracy): rewrote the Impact section with the verified per-interval overflow table instead of the oversimplified "10s+ → future" framing.
- **Suggestion 2** (redundant error check): collapsed `if err != nil { return querySet, err }; return querySet, nil` into `return querySet, err` in both `net_send.go` and `io_send.go` (f16b29f).
- **Suggestion 3** (scheduler window gap/overlap): agreed this is a pre-existing design issue orthogonal to this diff, not something to fold in here. Filed as #349.

geunwoonoh (approved):

PR #354, opened 2026-07-07
ci: add govulncheck and go.mod tidy checks, bump Go to 1.25.11

jisung-02:
## Background

alpacon-cli's CI runs two dependency checks that alpamon was missing. alpamon runs as a root server agent, so scanning its dependencies for known CVEs with govulncheck is especially worthwhile. The existing lint only runs golangci-lint, which covers static analysis of our own code but not known vulnerabilities in the libraries we pull in.

## Changes

lint.yml gets a govulncheck job. It installs `golang.org/x/vuln/cmd/govulncheck` and runs `govulncheck ./...`.

build-and-test.yml gets a verify-tidy job that runs `go mod tidy` and fails on any `go.mod`/`go.sum` diff, guarding against dependency drift. It runs as a single ubuntu job rather than inside the distro matrix, since tidiness is platform-independent and the matrix would just repeat the same work.

Both new jobs run `go generate ./pkg/db/ent` before their check. alpamon's ent output is gitignored (`pkg/db/ent/*`, with only `entc.go` and `generate.go` tracked), and the tree imports the generated ent subpackages (`pkg/db/ent/cpu`, etc.). Without generating them first, `go mod tidy` fails to resolve those internal packages and `govulncheck ./...` fails to load them. alpacon-cli has no ent, which is why its equivalent jobs don't need this step.

GO_VERSION is bumped from 1.25.10 to 1.25.11 across build-and-test.yml, lint.yml, release.yml, and the `go` directive in go.mod. govulncheck flagged two called stdlib vulnerabilities on 1.25.10, both fixed in 1.25.11; without the bump the new job would fail on the first run.

## Safety

The two vulnerabilities govulncheck reported on 1.25.10 are both reachable from our code (called, not just imported):

- GO-2026-5039 (net/textproto): reached from the WebSocket dial path in `pkg/runner/ftp.go` via `ReadMIMEHeader`
- GO-2026-5037 (crypto/x509): reached from certificate-verification paths in file, ftp, and auth code

After the bump to 1.25.11, re-running govulncheck reports zero vulnerabilities that affect the code.

`actions/checkout` stays at `@v6` to match every other job in the repo. The latest major is v7, but v6 already runs on the node24 runtime, so there is no runtime benefit and mixing majors within one repo would be inconsistent. govulncheck is installed with `@latest` so it uses the newest scanner and vulnerability database; this favors security currency over build reproducibility and can be pinned to a specific version if reproducibility is preferred.

## Testing

Verified locally with the Go 1.25.11 toolchain, from a clean regeneration (deleted ent output, then `go generate`):

- `go build ./cmd/alpamon`: passes
- `go mod tidy`: only the `go` directive line changes, no dependency or go.sum drift
- `govulncheck ./...`: zero vulnerabilities affecting the code (was two on 1.25.10)

## Checklist

- [ ] verify-tidy and govulncheck jobs green in CI
- [ ] full distro matrix build/test green

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -14,9 +14,31 @@ permissions:
   contents: read
 
 env:
-  GO_VERSION: '1.25.10'
+  GO_VERSION: '1.25.11'
 
 jobs:
+  verify-tidy:
+    name: Verify go.mod and go.sum are tidy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      # Generate the ent package first: the tree imports gitignored ent
+      # subpackages, so go mod tidy fails to resolve them otherwise.
+      - name: Generate go code
+        run: go generate ./pkg/db/ent
+
+      - name: Verify go.mod and go.sum are tidy
+        run: |
+          go mod tidy
+          git diff --exit-code go.mod go.sum
+
   build-and-test:
     name: Build and Test (${{ matrix.distro }})
     runs-on: ${{ matrix.runner }}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -6,7 +6,7 @@ on:
       - "docs/**"
 
   pull_request:
-    types: [opened, ready_for_review, synchronize]
+    types: [opened, reopened, ready_for_review, synchronize]
 
   workflow_call:
 
@@ -24,15 +24,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - name: Set up Go and generate ent code
+        uses: ./.github/actions/setup-go-generate
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      # Generate the ent package first: the tree imports gitignored ent
-      # subpackages, so go mod tidy fails to resolve them otherwise.
-      - name: Generate go code
-        run: go generate ./pkg/db/ent
 
       - name: Verify go.mod and go.sum are tidy
         run: |
@@ -140,13 +135,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - name: Set up Go and generate ent code
+        uses: ./.github/actions/setup-go-generate
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      - name: Generate go code
-        run: go generate ./pkg/db/ent
 
       - name: Build
         run: go build -v .
@@ -162,13 +154,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - name: Set up Go and generate ent code
+        uses: ./.github/actions/setup-go-generate
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      - name: Generate go code
-        run: go generate ./pkg/db/ent
 
       - name: Build
         run: go build -v .
@@ -253,13 +242,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - name: Set up Go and generate ent code
+        uses: ./.github/actions/setup-go-generate
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      - name: Generate go code
-        run: go generate ./pkg/db/ent
 
       - name: Build
         run: go build -v .
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,13 +27,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - name: Set up Go and generate ent code
+        uses: ./.github/actions/setup-go-generate
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      - name: Generate go code
-        run: go generate ./pkg/db/ent
 
       - name: GolangCI-Lint
         uses: golangci/golangci-lint-action@v9
@@ -49,15 +46,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v6
 
-      - name: Set up Go
-        uses: actions/setup-go@v6
+      - name: Set up Go and generate ent code
+        uses: ./.github/actions/setup-go-generate
         with:
           go-version: ${{ env.GO_VERSION }}
-
-      # Generate the ent package first: the tree imports gitignored ent
-      # subpackages, so govulncheck ./... fails to load them otherwise.
-      - name: Generate go code
-        run: go generate ./pkg/db/ent
 
       - name: Install govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,7 +14,7 @@ permissions:
   contents: read
 
 env:
-  GO_VERSION: '1.25.10'
+  GO_VERSION: '1.25.11'
 
 jobs:
   linter:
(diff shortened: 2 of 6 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -14,9 +14,31 @@ permissions:
...
+
+      # Generate the ent package first: the tree imports gitignored ent
+      # subpackages, so go mod tidy fails to resolve them otherwise.
+      - name: Generate go code
+        run: go generate ./pkg/db/ent
+
+      - name: Verify go.mod and go.sum are tidy
+        run: |
+          go mod tidy
+          git diff --exit-code go.mod go.sum
```
geunwoonoh, line 35: Warning: This check is correct on a clean checkout, but now that tidiness is enforced it introduces a contributor footgun. Anyone with a stale `pkg/db/ent/migrate/` in their local tree (left over from before Migrate was disabled — it's gitignored, so `go generate` never removes it) will see `go mod tidy` *add* ~13 modules (`ariga.io/atlas`, `hcl`, `cty`, `inflect`, …) via `pkg/db/ent/migrate → entgo.io/ent/dialect/sql/schema → ariga.io/atlas`. If they commit that, CI (clean checkout) removes them again and this `git diff --exit-code` fails — the classic "tidy on my machine, red in CI."

Suggestion: add a note in CLAUDE.md / contributing docs to clean stale output before tidying:

```
git clean -fdx pkg/db/ent && go generate ./pkg/db/ent
go mod tidy
```

.github/workflows/lint.yml
```diff
@@ -40,3 +40,27 @@ jobs:
...
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      # Generate the ent package first: the tree imports gitignored ent
+      # subpackages, so govulncheck ./... fails to load them otherwise.
+      - name: Generate go code
+        run: go generate ./pkg/db/ent
+
+      - name: Install govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
```
geunwoonoh, line 55: Warning: `@latest` + the live vulnerability DB means a newly-published stdlib/dependency CVE can turn a PR (or a `main` push) red even when it's unrelated to dependencies. That's inherent to continuous scanning, but it can act as a merge blocker for a small team.

Suggestion (team-policy call): pin govulncheck to a specific version for reproducibility, and/or add a `schedule:` run so unrelated PRs aren't blocked by newly-disclosed CVEs:

```suggestion
        run: go install golang.org/x/vuln/cmd/govulncheck@v1.1.4
```

geunwoonoh (dismissed):
## Summary

This PR adds two dependency checks to alpamon CI:
- A `govulncheck` job in `lint.yml` to scan for known CVEs
- A `verify-tidy` job in `build-and-test.yml` to block `go mod tidy` drift
- Bumps Go 1.25.10 → 1.25.11 across the three workflows and `go.mod`

Both new jobs run `go generate ./pkg/db/ent` before their check. Since the ent output is gitignored, without generating it first `go mod tidy` / `govulncheck` can't load the internal packages — that reasoning is correct.

**What I verified locally (Go 1.25.11):**
- ✅ On a clean checkout, `go generate ./pkg/db/ent` → `go mod tidy` → `git diff --exit-code go.mod go.sum` = **clean**. `verify-tidy` passes on a fresh CI checkout.
- ✅ `go build ./cmd/alpamon` passes; `govulncheck ./...` loads packages fine.
- ✅ No lingering `1.25.10` references (full-repo scan); all `GO_VERSION` references consistent.

---

### Critical Issues (0)

None. (I initially suspected `go mod tidy` drift would fail CI on every run, but tracing it down showed it was a **stale `pkg/db/ent/migrate/` directory in my local tree**. entc.go disables `gen.Migrate`, so a clean checkout has no `migrate/` and CI passes cleanly.)

---

### Warnings (2)

**1. `verify-tidy` — stale local ent output can confuse contributors** (`build-and-test.yml`)

Now that tidiness is *enforced*, this scenario becomes real:
- A contributor who built alpamon before Migrate was disabled has a leftover `pkg/db/ent/migrate/` locally (it's gitignored, so `go generate` never deletes it).
- In that state, `go mod tidy` **adds** ~13 modules to go.mod via `migrate/` → `entgo.io/ent/dialect/sql/schema` → `ariga.io/atlas` (+ hcl, cty, inflect, …). If they commit that, CI (clean checkout) **removes** them again → `git diff` fires → `verify-tidy` fails. The classic "tidy on my machine, red in CI."
- Evidence: `go mod why -m ariga.io/atlas` → `pkg/db/ent/migrate → entgo.io/ent/dialect/sql/schema → ariga.io/atlas/...`

CI itself is correct. But since this job is being introduced, a line in `CLAUDE.md` / contributing docs — "before `go mod tidy`, run `git clean -fdx pkg/db/ent && go generate ./pkg/db/ent` to drop stale output" — would prevent recurring confusion.

**2. `govulncheck @latest` — reproducibility + unrelated PRs going red** (`lint.yml`)

`go install ...govulncheck@latest` combined with the live vuln DB means that when a new stdlib/dependency CVE is published tomorrow, a PR (or a `main` push) **unrelated to dependencies can suddenly fail**. That's inherent to (and a benefit of) continuous scanning, but for a small team it can act as a merge blocker. The author already calls out this trade-off in the PR description and leaves the door open to pinning. This is a team-policy call: pin a version (reproducibility) and/or add a `schedule:` run (so unrelated PRs aren't blocked).

---

### Suggestions (3)

**1. Boilerplate duplication** — the checkout + setup-go + `go generate ./pkg/db/ent` trio now repeats across ~6 places (linter, govulncheck, verify-tidy, build-and-test, windows, macos, release). A composite action (`.github/actions/setup-go-generate`) would let a Go-version or generate-logic change be made in one spot. (Optional cleanup.)

**2. Register as actual merge gates** — if the new jobs run but aren't in branch-protection required checks, a failure won't block merge. Add `govulncheck` and `Verify go.mod and go.sum are tidy` as required checks to get the intended gating. (Repo setting, outside the diff.)

**3. Trigger mismatch (pre-existing)** — `build-and-test.yml` is missing `reopened` in its `pull_request` types while `lint.yml` has it. So on PR reopen, govulncheck runs but verify-tidy doesn't. Not introduced by this PR, but aligning the two gate jobs' trigger conditions would be good.

_(Note: the Dockerfiles pin only the minor `golang:1.25`. Thanks to the `go 1.25.11` directive + `GOTOOLCHAIN=auto`, they'll automatically use a ≥1.25.11 toolchain, so no problem. Outside this PR's scope.)_

---

### Highlights

- Placing `go generate` before the checks correctly solves the "ent output is gitignored" problem, and the reasoning is documented in both comments and the PR description.
- Keeping `verify-tidy` as a single ubuntu job rather than in the distro matrix is the right call (platform-independent check).
- Inherits top-level `permissions: contents: read`, so the new jobs carry no excess permissions.
- The version bump is consistently reflected in all 4 locations (0 lingering references).
- Given alpamon runs as a root server agent, adding govulncheck is a particularly worthwhile security improvement.

---

**Verdict**: **APPROVED** — no merge-blocking issues. I reproduced and verified the CI behavior locally (verify-tidy passes clean, build passes). The 2 warnings are documentation / team-policy items and don't require code changes to merge.

<sub>Reviewed with local verification on Go 1.25.11.</sub>

jisung-02:
Thanks for the thorough review with local verification. Addressed W1, W2, S1, and S3.

**W1 (stale local ent output)** — 3d79c91. Documented in `CLAUDE.md`: run `git clean -fdx pkg/db/ent && go generate ./pkg/db/ent` before `go mod tidy` to drop leftover gitignored output (e.g. a `migrate/` directory from before entc.go disabled `gen.Migrate`), with the "tidy on my machine, red in CI" reasoning inline.

**W2 (`govulncheck@latest` — unrelated PRs going red)** — 485beb2. Added `continue-on-error: true` to the `Run govulncheck` step so a newly published CVE in a transitive dep or the Go stdlib no longer turns unrelated PRs red. Dependabot already handles proactive dependency bumps; govulncheck is kept as the stdlib + call-graph reachability signal (which Dependabot doesn't cover), surfaced in the step log rather than acting as a merge gate. The setup/install steps above stay hard-failing so real CI breakage is still caught. Left `@latest` as-is rather than pinning, since it's no longer a gate.

**S1 (boilerplate duplication)** — 3d79c91. Extracted the setup-go + `go generate ./pkg/db/ent` pair into a local composite action `.github/actions/setup-go-generate`, now reused across all six jobs (linter, govulncheck, verify-tidy, build-and-test matrix, windows, macos). A Go-version or codegen change now lives in one place. `checkout` stays inline in each job since a local action requires the repo to be checked out first.

**S3 (trigger mismatch)** — 3d79c91. Added `reopened` to `build-and-test.yml`'s `pull_request` types so verify-tidy runs on PR reopen, matching `lint.yml`.

**S2 (register as required checks)** — leaving as-is. main's ruleset has no `required_status_checks` rule today, so these jobs were never merge blockers before this PR either; nothing regresses by leaving it unchanged. Turning any job into a hard gate is a separate branch-protection change that should go through team approval. (And with W2 now non-blocking, govulncheck would in any case stay off the required list.)

geunwoonoh (approved):

PR #355, opened 2026-07-08
feat(pty): accept optional shell parameter in openpty command

### LINKED ISSUE #257: feat(pty): accept optional shell parameter in openpty command
## Background

alpacon-server now sends a `shell` field in the openpty command data when the target SystemUser has a known shell (e.g., `/bin/zsh`, `/bin/bash`). Currently alpamon ignores this field and always uses `utils.DefaultShell()`.

## Proposal

Allow the openpty command to accept an optional `shell` parameter, so users connect with their configured shell instead of the system default.

### Requirements

1. **Backwards compatibility**: If `shell` is empty or absent in the command data, fall back to `utils.DefaultShell()` (current behavior).

2. **Validation before use**: Before spawning the PTY with the requested shell, verify it is safe:
   - **Linux**: Check that the shell path exists in `/etc/shells`
   - **macOS**: Check via `/etc/shells` (macOS also uses this file)
   - If validation fails, fall back to `DefaultShell()` and log a warning

3. **No shell metacharacter risk**: Use `exec.Command(shell)` directly — never pass through `sh -c`. The shell path should be an absolute path with no arguments.

### Example flow

```
Server sends: {"shell": "/bin/zsh", "home_directory": "/Users/deploy", ...}

alpamon:
  1. Parse shell from CommandData
  2. If shell != "":
     a. Read /etc/shells into a set
     b. If shell is in the set → use it
     c. Else → log warning, use DefaultShell()
  3. Else → use DefaultShell()
  4. exec.Command(selectedShell)
```

### Files to modify

- `pkg/runner/pty.go` — Accept shell from PTYData, validate, use in exec
- `pkg/executor/handlers/terminal/terminal.go` — Pass shell field from CommandData to PTYData

## Related

- alpacax/alpacon-server branch `254-fix-macos-uid-threshold` — sends `shell` field in openpty data
- Part of macOS support improvements

jisung-02:
## Background

Until now the `openpty` command always launched the shell via `utils.DefaultShell()`. There was no way for the server to select a per-user login shell. Following alpacax/alpamon#257, the `openpty` command now accepts an optional `shell` field sent by the server and, when it passes validation, starts the PTY session with that shell.

## What changed

The `shell` field is wired from the protocol through to PTY launch. `pkg/executor/handlers/terminal/terminal.go` passes the command argument's `Shell` into `CommandData`, and `PtyClient` resolves it via `resolveShell` when the session starts.

Behavior of `resolveShell` (`pkg/runner/pty_shell.go`):

- If `shell` is empty or absent, it uses `utils.DefaultShell()` + `utils.DefaultShellArgs()` exactly as before (backwards compatible).
- If `shell` matches the host's valid login shells exactly, it launches that shell with no extra args. The default args (`-il` / `-NoLogo`) are bash/zsh/powershell specific and would break dash, tcsh, and cmd.exe.
- If it is not in the list, it logs a warning and falls back to the default shell.

The valid-shell list is per platform. Unix reads `/etc/shells` (extracted into `utils.LoadValidShells`, consolidating the reader that was duplicated across `commit_linux.go` and `commit_darwin.go`); Windows uses the fixed list in `commit_windows.go` (powershell.exe, cmd.exe, pwsh.exe). Windows executable names are case-insensitive, so matching uses `strings.EqualFold` on that platform only.

The session's `SHELL` environment variable is now set to the resolved shell in `pkg/runner/pty.go`, and the static `SHELL` assignment in `getDefaultEnv` was removed.

On macOS, when `/etc/shells` is empty or unreadable, `loadValidShells` now returns the common default shells (/bin/bash, /bin/zsh, /bin/sh). Previously it returned nil, which rejected every requested shell.

## Safety

A server-provided shell value never passes through `sh -c`. Unix runs it via `exec.Command(shell, args...)` and Windows places it as a discrete token in the ConPTY command line, so shell-metacharacter injection is not possible. Validation is an exact-match whitelist against the per-platform valid-shell list, so an arbitrary path is always rejected and falls back to the default (not fail-open). The empty-list behavior differs by platform: on Linux the list comes straight from `/etc/shells`, so if it is missing or unreadable the list is empty and every requested shell is rejected; on macOS `/etc/shells` is less reliable, so `loadValidShells` falls back to a fixed set of trusted default shells (/bin/bash, /bin/zsh, /bin/sh); Windows uses a fixed list (powershell.exe, cmd.exe, pwsh.exe). In every case the accepted set is a trusted whitelist, never the caller's raw input.

## Testing

Unit tests:

- `pkg/runner/pty_shell_test.go`: covers the empty/valid/invalid branches of `resolveShell` plus the Windows-only case-insensitive match.
- `pkg/utils/shell_unix_test.go`: `/etc/shells` parsing (skips comments and blank lines, trims whitespace, no partial matches, nil on missing file).
- Cross-compiles on linux/darwin/windows, `go vet`, and the full test suite pass.

End-to-end verification:

Built this branch for linux/arm64, deployed it over the packaged binary on a live Ubuntu host, and opened real Websh sessions from the console. The shell the server sends is the target system user's login shell (`proc.models.SystemUser.shell` in alpacon-server, synced from `/etc/passwd`); it is not selectable per session. Each case changed the target user's login shell, restarted alpamon to force a re-sync, then opened a new Websh session.

| Case | Target user login shell | Expected | Observed |
| --- | --- | --- | --- |
| Valid non-default shell | `/usr/bin/dash` (in `/etc/shells`) | that shell launched, no `-il` | `echo $SHELL` -> `/usr/bin/dash`, `readlink /proc/$$/exe` -> `/usr/bin/dash`, `ps -p $$ -o comm=` -> `dash` |
| Default shell | `/bin/bash` (in `/etc/shells`) | bash launched without login args | `shopt -q login_shell` -> 1 (not a login shell), `$0` -> `/bin/bash` (not `-bash`), `echo $SHELL` -> `/bin/bash` |
| Invalid shell (fallback) | `/bin/dash` (not in `/etc/shells`) | fallback to default + warning | not reachable end-to-end, see note below |

Note on the invalid-shell fallback: it cannot be exercised through the real server. alpamon syncs each user's shell together with the `/etc/shells` list, and alpacon-server's `compute_login_enabled` (`proc/utils.py`) marks any user whose shell is not in `/etc/shells` as login-disabled. Such a session is rejected at creation with `system_user_access_denied` (`websh/mixins.py`) before any `openpty` command is sent, so the server only ever sends a shell that is already in `/etc/shells`. The `resolveShell` fallback branch remains as defense-in-depth (sync races, older alpamon, `/etc/shells` changed after a sync) and is covered by the unit tests above.

## Checklist

- [x] Both Unix and Windows paths implemented
- [x] Backwards compatible (existing behavior when shell is unset)
- [x] Confirmed no shell-metacharacter injection
- [x] Builds and tests pass on all three platforms
- [x] Verified end-to-end on a live host via Websh (valid and default shells)

Closes #257

### DIFF
--- a/pkg/runner/commit_darwin.go
+++ b/pkg/runner/commit_darwin.go
@@ -1,8 +1,6 @@
 package runner
 
 import (
-	"bufio"
-	"os"
 	"os/exec"
 	"strconv"
 	"strings"
@@ -11,32 +9,13 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// loadValidShells reads /etc/shells on macOS (which exists and follows the same format as Linux).
+// loadValidShells returns the host's valid login shells from /etc/shells,
+// falling back to the common macOS defaults when the file is unavailable.
 func loadValidShells() []string {
-	const shellsFilePath = "/etc/shells"
-	var shells []string
-
-	file, err := os.Open(shellsFilePath)
-	if err != nil {
-		log.Debug().Err(err).Msg("Failed to open /etc/shells, using defaults")
+	shells := utils.LoadValidShells()
+	if len(shells) == 0 {
 		return []string{"/bin/bash", "/bin/zsh", "/bin/sh"}
 	}
-	defer func() { _ = file.Close() }()
-
-	scanner := bufio.NewScanner(file)
-	for scanner.Scan() {
-		line := strings.TrimSpace(scanner.Text())
-		if line == "" || strings.HasPrefix(line, "#") {
-			continue
-		}
-		shells = append(shells, line)
-	}
-
-	if err := scanner.Err(); err != nil {
-		log.Debug().Err(err).Msg("Error reading /etc/shells")
-		return []string{"/bin/bash", "/bin/zsh", "/bin/sh"}
-	}
-
 	return shells
 }
 
--- a/pkg/runner/commit_darwin.go
+++ b/pkg/runner/commit_darwin.go
@@ -10,7 +10,7 @@ import (
 )
 
 // loadValidShells returns the host's valid login shells from /etc/shells,
-// falling back to the common macOS defaults when the file is unavailable.
+// falling back to the common macOS defaults when the list is empty or unavailable.
 func loadValidShells() []string {
 	shells := utils.LoadValidShells()
 	if len(shells) == 0 {
--- a/pkg/runner/pty.go
+++ b/pkg/runner/pty.go
@@ -39,6 +39,7 @@ type PtyClient struct {
 	username      string
 	groupname     string
 	homeDirectory string
+	shell         string
 	sessionID     string
 	wsToPty       chan []byte
 	ptyToWs       chan []byte
@@ -69,6 +70,7 @@ func NewPtyClient(data protocol.CommandData, apiSession *scheduler.Session, mana
 		username:      data.Username,
 		groupname:     data.Groupname,
 		homeDirectory: data.HomeDirectory,
+		shell:         data.Shell,
 		sessionID:     data.SessionID,
 		wsToPty:       make(chan []byte, bufferSize),
 		ptyToWs:       make(chan []byte, bufferSize),
@@ -92,7 +94,8 @@ func (pc *PtyClient) initializePtySession() error {
 		return fmt.Errorf("failed to connect Websh server: %w", err)
 	}
 
-	pc.cmd = exec.Command(utils.DefaultShell(), utils.DefaultShellArgs()...)
+	shell, args := resolveShell(pc.shell, loadValidShells())
+	pc.cmd = exec.Command(shell, args...)
 	uid, gid, groupIds, env, err := pc.getPtyUserAndEnv()
 	if err != nil {
 		return fmt.Errorf("failed to get user/env: %w", err)
--- a/pkg/runner/pty.go
+++ b/pkg/runner/pty.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync/atomic"
@@ -96,6 +97,9 @@ func (pc *PtyClient) initializePtySession() error {
 
 	shell, args := resolveShell(pc.shell, loadValidShells())
 	pc.cmd = exec.Command(shell, args...)
+	// login(1) convention: a leading "-" in argv[0] makes any shell a login
+	// shell, so profile files are sourced without shell-specific -l flags.
+	pc.cmd.Args[0] = "-" + filepath.Base(shell)
 	uid, gid, groupIds, env, err := pc.getPtyUserAndEnv()
 	if err != nil {
 		return fmt.Errorf("failed to get user/env: %w", err)
--- a/pkg/runner/pty.go
+++ b/pkg/runner/pty.go
@@ -100,6 +100,7 @@ func (pc *PtyClient) initializePtySession() error {
 	if err != nil {
 		return fmt.Errorf("failed to get user/env: %w", err)
 	}
+	env["SHELL"] = shell
 	if err = pc.setPtyCmdSysProcAttrAndEnv(uid, gid, groupIds, env); err != nil {
 		return fmt.Errorf("failed to set process credentials: %w", err)
 	}
--- a/pkg/runner/pty_shell.go
+++ b/pkg/runner/pty_shell.go
@@ -14,7 +14,7 @@ import (
 // bash/zsh/powershell-specific and break other shells (dash, tcsh, cmd.exe).
 func resolveShell(requested string, validShells []string) (shell string, args []string) {
 	if requested != "" {
-		if shellAllowed(requested, validShells) {
+		if shellAllowed(requested, validShells, runtime.GOOS == "windows") {
 			return requested, nil
 		}
 		log.Warn().Str("requested", requested).
@@ -23,10 +23,9 @@ func resolveShell(requested string, validShells []string) (shell string, args []
 	return utils.DefaultShell(), utils.DefaultShellArgs()
 }
 
-// shellAllowed matches case-insensitively on Windows, where executable
-// names are case-insensitive; exact paths elsewhere.
-func shellAllowed(requested string, validShells []string) bool {
-	if runtime.GOOS == "windows" {
+// caseInsensitive is set on Windows, where executable names are case-insensitive.
+func shellAllowed(requested string, validShells []string, caseInsensitive bool) bool {
+	if caseInsensitive {
 		return slices.ContainsFunc(validShells, func(s string) bool {
 			return strings.EqualFold(s, requested)
 		})
--- a/pkg/runner/pty_shell.go
+++ b/pkg/runner/pty_shell.go
@@ -1,25 +1,35 @@
 package runner
 
 import (
+	"runtime"
 	"slices"
+	"strings"
 
 	"github.com/alpacax/alpamon/v2/pkg/utils"
 	"github.com/rs/zerolog/log"
 )
 
-// resolveShell picks the shell to launch for a PTY session. A non-empty
-// requested shell is used with no extra args when it appears in the host's
-// valid login shells; otherwise the platform default shell and its default
-// args are used. Requested shells run without DefaultShellArgs because those
-// flags (-il / -NoLogo) are bash/zsh/powershell-specific and break other
-// shells (dash, tcsh, cmd.exe).
-func resolveShell(requested string, valid []string) (shell string, args []string) {
+// resolveShell picks the shell to launch for a PTY session. A requested shell
+// runs without DefaultShellArgs because those flags (-il / -NoLogo) are
+// bash/zsh/powershell-specific and break other shells (dash, tcsh, cmd.exe).
+func resolveShell(requested string, validShells []string) (shell string, args []string) {
 	if requested != "" {
-		if slices.Contains(valid, requested) {
+		if shellAllowed(requested, validShells) {
 			return requested, nil
 		}
 		log.Warn().Str("requested", requested).
 			Msg("Requested shell not in valid shells; falling back to default.")
 	}
 	return utils.DefaultShell(), utils.DefaultShellArgs()
+}
+
+// shellAllowed matches case-insensitively on Windows, where executable
+// names are case-insensitive; exact paths elsewhere.
+func shellAllowed(requested string, validShells []string) bool {
+	if runtime.GOOS == "windows" {
+		return slices.ContainsFunc(validShells, func(s string) bool {
+			return strings.EqualFold(s, requested)
+		})
+	}
+	return slices.Contains(validShells, requested)
 }
--- a/pkg/runner/pty_shell.go
+++ b/pkg/runner/pty_shell.go
@@ -1,0 +1,25 @@
+package runner
+
+import (
+	"slices"
+
+	"github.com/alpacax/alpamon/v2/pkg/utils"
+	"github.com/rs/zerolog/log"
+)
+
+// resolveShell picks the shell to launch for a PTY session. A non-empty
+// requested shell is used with no extra args when it appears in the host's
+// valid login shells; otherwise the platform default shell and its default
+// args are used. Requested shells run without DefaultShellArgs because those
+// flags (-il / -NoLogo) are bash/zsh/powershell-specific and break other
+// shells (dash, tcsh, cmd.exe).
+func resolveShell(requested string, valid []string) (shell string, args []string) {
+	if requested != "" {
+		if slices.Contains(valid, requested) {
+			return requested, nil
+		}
+		log.Warn().Str("requested", requested).
+			Msg("Requested shell not in valid shells; falling back to default.")
+	}
+	return utils.DefaultShell(), utils.DefaultShellArgs()
+}
--- a/pkg/utils/shell_unix.go
+++ b/pkg/utils/shell_unix.go
@@ -19,7 +19,7 @@ func LoadValidShells() []string { return loadValidShellsFrom(validShellsFilePath
 func loadValidShellsFrom(path string) []string {
 	file, err := os.Open(path)
 	if err != nil {
-		log.Debug().Err(err).Msg("Failed to open /etc/shells, skipping shell validation")
+		log.Debug().Err(err).Str("path", path).Msg("Failed to open valid shells file, skipping shell validation")
 		return nil
 	}
 	defer func() { _ = file.Close() }()
@@ -34,7 +34,7 @@ func loadValidShellsFrom(path string) []string {
 		shells = append(shells, line)
 	}
 	if err := scanner.Err(); err != nil {
-		log.Debug().Err(err).Msg("Error reading /etc/shells")
+		log.Debug().Err(err).Str("path", path).Msg("Error reading valid shells file")
 		return nil
 	}
 	return shells
--- a/pkg/utils/shell_unix.go
+++ b/pkg/utils/shell_unix.go
@@ -1,0 +1,41 @@
+//go:build !windows
+
+package utils
+
+import (
+	"bufio"
+	"os"
+	"strings"
+
+	"github.com/rs/zerolog/log"
+)
+
+const validShellsFilePath = "/etc/shells"
+
+// LoadValidShells reads /etc/shells and returns the listed login shell paths.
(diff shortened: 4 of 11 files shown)


pkg/utils/shell_unix.go
```diff
@@ -0,0 +1,41 @@
...
+// LoadValidShells reads /etc/shells and returns the listed login shell paths.
+// Returns nil if the file cannot be read or parsed.
+func LoadValidShells() []string { return loadValidShellsFrom(validShellsFilePath) }
+
+func loadValidShellsFrom(path string) []string {
+	file, err := os.Open(path)
+	if err != nil {
+		log.Debug().Err(err).Msg("Failed to open /etc/shells, skipping shell validation")
+		return nil
+	}
```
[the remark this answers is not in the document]
jisung-02: Fixed in 7482ea8: both failure logs now attach the actual `path` via `.Str("path", path)` and the message is generalized (no longer hard-codes /etc/shells), so logs match the file actually read by tests and future callers.

pkg/utils/shell_unix.go
```diff
@@ -0,0 +1,41 @@
...
+		line := strings.TrimSpace(scanner.Text())
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		shells = append(shells, line)
+	}
+	if err := scanner.Err(); err != nil {
+		log.Debug().Err(err).Msg("Error reading /etc/shells")
+		return nil
+	}
```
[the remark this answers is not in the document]
jisung-02: Fixed in 7482ea8: this read-error log now includes the actual `path` via `.Str("path", path)` and drops the hard-coded /etc/shells, so the failing file is identifiable and behavior is consistent with the test paths.

pkg/runner/pty.go
```diff
@@ -92,11 +94,13 @@
 		return fmt.Errorf("failed to connect Websh server: %w", err)
 	}
 
-	pc.cmd = exec.Command(utils.DefaultShell(), utils.DefaultShellArgs()...)
+	shell, args := resolveShell(pc.shell, loadValidShells())
+	pc.cmd = exec.Command(shell, args...)
```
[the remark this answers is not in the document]
jisung-02: False positive: the requested shell is validated against an allowlist before
reaching exec.Command, and no shell interpreter is involved.

- resolveShell (pkg/runner/pty_shell.go) returns the server-provided value only
  when shellAllowed() confirms an exact match against the host's login shells
  from /etc/shells (utils.LoadValidShells). Matching is exact (slices.Contains;
  strings.EqualFold on Windows only), not substring, so an arbitrary path cannot
  pass. Any non-listed value falls back to the default shell.
- exec.Command invokes the binary directly and does not spawn /bin/sh -c, so
  shell metacharacters (;, |, $(...)) cannot be interpreted. The requested shell
  is also launched with no extra args.

Worst case is that one of the login shells already registered in /etc/shells is
started — not arbitrary command execution or injection. The taint reaches the
sink, but the allowlist neutralizes it; CodeQL does not model shellAllowed() as
a sanitizer.

pkg/runner/pty_shell.go
```diff
@@ -0,0 +1,35 @@
...
+	"strings"
+
+	"github.com/alpacax/alpamon/v2/pkg/utils"
+	"github.com/rs/zerolog/log"
+)
+
+// resolveShell picks the shell to launch for a PTY session. A requested shell
+// runs without DefaultShellArgs because those flags (-il / -NoLogo) are
+// bash/zsh/powershell-specific and break other shells (dash, tcsh, cmd.exe).
+func resolveShell(requested string, validShells []string) (shell string, args []string) {
```
geunwoonoh, line 15: Warning: Every production session becomes a non-login shell.

`resolveShell` returns `(requested, nil)` for an allowed shell — launched *without* `-il`. Per the PR description the server *always* sends the user's login shell (present in `/etc/shells`), so in production `requested` is effectively always non-empty and allowed. That makes the `-il` path (empty request) dead code and means **all Websh sessions now run as non-login shells**, whereas before they were always `bash -il`.

Consequence: `/etc/profile`, `~/.bash_profile` (and zsh's `/etc/zprofile`, `~/.zprofile`, `~/.zlogin`) are no longer sourced — only `~/.bashrc` / `~/.zshrc`. PATH/env from `/etc/profile.d/*` may be lost for existing users.

The author verified `shopt -q login_shell -> 1` as expected, so this is intentional — but it affects every existing default-shell user. Since `-il` was dropped because it breaks dash/tcsh/cmd.exe, consider mapping login args per known shell (bash/zsh/sh -> `-l`, others -> none) to preserve login semantics where supported, and at minimum document this change. Please confirm with the team before merge.

pkg/runner/pty_shell.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+		log.Warn().Str("requested", requested).
+			Msg("Requested shell not in valid shells; falling back to default.")
+	}
+	return utils.DefaultShell(), utils.DefaultShellArgs()
+}
+
+// shellAllowed matches case-insensitively on Windows, where executable
+// names are case-insensitive; exact paths elsewhere.
+func shellAllowed(requested string, validShells []string) bool {
```
geunwoonoh: Warning: Windows matches on base name only.

The Windows valid list (`commit_windows.go`) is `["powershell.exe", "cmd.exe", "pwsh.exe"]` — bare names. If the server sends a full path or a name without `.exe`, `EqualFold` won't match and it silently falls back to powershell. It fails closed (safe), but the requested-shell feature is effectively inert on Windows unless the server sends exactly one of those three bare names (the PR notes the Windows path isn't e2e-verified). Consider aligning the server contract or normalizing to basename before comparing.

pkg/runner/pty_shell.go
```diff
@@ -0,0 +1,35 @@
...
+	if runtime.GOOS == "windows" {
+		return slices.ContainsFunc(validShells, func(s string) bool {
```
geunwoonoh, line 29: Suggestion: The Windows case-insensitive branch is effectively untested. It only runs when `runtime.GOOS == "windows"`; on linux/mac CI (and cross-compile, which doesn't run tests) it's never exercised — I confirmed the `case-differing` test takes the fallback path on darwin.

```suggestion
func shellAllowed(requested string, validShells []string, caseInsensitive bool) bool {
	if caseInsensitive {
		return slices.ContainsFunc(validShells, func(s string) bool {
			return strings.EqualFold(s, requested)
		})
	}
	return slices.Contains(validShells, requested)
}
```

Then pass `runtime.GOOS == "windows"` from `resolveShell`, so both matching modes become unit-testable on any OS.

pkg/runner/pty.go
```diff
@@ -92,11 +94,13 @@ func (pc *PtyClient) initializePtySession() error {
 		return fmt.Errorf("failed to connect Websh server: %w", err)
 	}
 
-	pc.cmd = exec.Command(utils.DefaultShell(), utils.DefaultShellArgs()...)
+	shell, args := resolveShell(pc.shell, loadValidShells())
```
geunwoonoh, line 98: Suggestion (low priority): `loadValidShells()` opens `/etc/shells` from disk on every openpty. Session open isn't a hot path and reading fresh is safer (shells can change), so keeping it as-is is fine — noted only in case session-open latency ever matters (then cache with invalidation).

pkg/utils/shell_unix_test.go
```diff
@@ -0,0 +1,48 @@
+//go:build !windows
+
+package utils
+
+import (
+	"os"
+	"path/filepath"
```
geunwoonoh, line 7: Suggestion (low priority): test-style inconsistency. This file uses stdlib `reflect.DeepEqual` + `t.Errorf`, while `pty_shell_test.go` uses testify `assert`. The runner package uses testify elsewhere; unifying on one style would be nice.

pkg/runner/commit_darwin.go
```diff
@@ -11,32 +9,13 @@ import (
...
-		if line == "" || strings.HasPrefix(line, "#") {
-			continue
-		}
-		shells = append(shells, line)
-	}
-
-	if err := scanner.Err(); err != nil {
-		log.Debug().Err(err).Msg("Error reading /etc/shells")
+	shells := utils.LoadValidShells()
+	if len(shells) == 0 {
```
geunwoonoh, line 16: Suggestion (low priority): the empty-list -> defaults fallback here isn't covered by a test. Worth a small case asserting that an empty/unreadable list yields `[/bin/bash, /bin/zsh, /bin/sh]` (and, in `shell_unix_test.go`, a "comments/blank lines only -> nil" case).

geunwoonoh (commented):
## PR Review: #355 - feat(pty): accept optional shell parameter in openpty command

**Stack**: golang · **Changes**: +182 -56 across 10 files

### Summary

The `openpty` command now accepts an optional server-provided `shell` field and, when it exactly matches the host's valid login shells (`/etc/shells` on Unix, a fixed list on Windows), starts the PTY session with that shell; otherwise it falls back to `DefaultShell()` + `DefaultShellArgs()`. The `/etc/shells` parsing was consolidated into `utils.LoadValidShells`, removing the duplication across `commit_linux.go`/`commit_darwin.go`, and the static `SHELL` env var was replaced with the resolved shell.

Overall this is a clean, safe PR. There are **no merge-blocking issues**, but there is **one behavioral change the team should confirm before merge**.

### Critical Issues (0)

None. The CodeQL "command built from user-controlled sources" warning on `pty.go:98` is a **confirmed false positive** — exact-match allowlist + no `sh -c` + fail-closed fallback. The author's response comment is accurate.

### Warnings (2)

**1. Every production session becomes a non-login shell (confirm before merge).** `resolveShell` returns `(requested, nil)` for an allowed shell — i.e. launched *without* `-il`. Per the PR description the server *always* sends the user's login shell (synced from `/etc/passwd`, present in `/etc/shells`), so in production `requested` is effectively always non-empty and allowed. As a result the `-il` path (empty request) becomes dead code and **all Websh sessions now run as non-login shells**. Previously they were always `bash -il` (login + interactive). Consequence: `/etc/profile`, `~/.bash_profile` (and zsh's `/etc/zprofile`, `~/.zprofile`, `~/.zlogin`) are no longer sourced — only `~/.bashrc` / `~/.zshrc`. PATH/env set in `/etc/profile.d/*` may be lost. The author verified `shopt -q login_shell -> 1` as expected, so this is intentional, but it affects every existing default-shell user. Since `-il` was dropped because it breaks dash/tcsh/cmd.exe, consider mapping login args per known shell (bash/zsh/sh -> `-l`, others -> none) to preserve login semantics where supported, and at minimum document this change.

**2. Windows matches on base name only.** The Windows valid list is `["powershell.exe", "cmd.exe", "pwsh.exe"]` (bare names). If the server sends a full path or a name without `.exe`, `EqualFold` won't match and it silently falls back to powershell. It fails closed (safe), but the requested-shell feature is effectively inert on Windows unless the server sends exactly one of those three bare names (the PR notes the Windows path is not e2e-verified). Consider aligning the server contract or normalizing to basename before comparing.

### Suggestions (4)

**1. The Windows case-insensitive branch is effectively untested.** The `EqualFold` branch only runs when `runtime.GOOS == "windows"`; on linux/mac CI (and cross-compile, which doesn't run tests) it's never exercised — I confirmed the `case-differing` test takes the fallback path on darwin. Refactor `shellAllowed(requested, validShells, caseInsensitive bool)` and pass `runtime.GOOS == "windows"` from `resolveShell` so both matching modes are unit-testable on any OS.

**2. `/etc/shells` is read from disk on every openpty.** `loadValidShells()` opens the file per session. Session open isn't a hot path and reading fresh is safer (shells can change), so **keeping it as-is is fine** — noted only in case session-open latency ever matters.

**3. Test-style inconsistency.** `pty_shell_test.go` uses testify `assert` while `shell_unix_test.go` uses stdlib `reflect.DeepEqual` + `t.Errorf`. The runner package uses testify elsewhere; unifying would be nice. (Low priority.)

**4. Missing edge-case tests.** The darwin `loadValidShells` empty-list -> defaults fallback, and the "comments/blank lines only -> nil" case, are uncovered. (Low priority.)

### Highlights

- Great de-duplication: `/etc/shells` parsing consolidated into `utils.LoadValidShells` (~50 lines of duplication removed).
- Solid security posture: exact-match allowlist + direct `exec.Command` (no `sh -c`) + fail-closed fallback. Metacharacter injection is not possible.
- Correct env handling: `SHELL` moved from a static value to the resolved shell (`pty.go:103`), and the executor path keeps its own `SHELL` — no regression (verified).
- Clean build-tag hygiene: `utils.LoadValidShells` is `!windows`, Windows uses a fixed list, and the now-unused `utils` import was correctly dropped from `pty_windows.go`.
- Honest, detailed PR description with a real e2e verification table and an explicit note on why the invalid-shell fallback can't be reproduced against the live server.

**Verification (run locally):** `go vet` clean, the relevant unit tests all PASS, and linux/darwin/windows cross-compiles all succeed.

---

**Verdict**: COMMENTED — no blocking issues. Warning 1 (non-login shell switch) is an intentional but broad UX change; please confirm with the team before merge.

**Checklist**:
- [x] Code follows stack conventions (Go 1.25, appropriate `slices`/`strings` usage)
- [x] Tests included and passing (but the Windows branch is unverified — Suggestion 1)
- [x] No security vulnerabilities (CodeQL warning is a confirmed FP)
- [ ] Documentation updated — recommend documenting the non-login shell change (Warning 1)

pkg/runner/pty_shell.go
jisung-02: Fixed in 274d59c. Requested shells now start as login shells via the login(1)/sshd convention: argv[0] is prefixed with `-` (`cmd.Args[0] = "-" + filepath.Base(shell)`), which any Unix shell (bash, zsh, dash, tcsh, fish) recognizes as a login marker. This avoids the per-shell flag table, and the original problem of `-il` breaking dash/tcsh does not reappear since no flags are added. The prefix applies uniformly, so the default `-il` fallback keeps its behavior (the redundant login marking is harmless).

Verified through creack/pty (the same startup path the PTY session uses): bash `shopt -q login_shell` and zsh `[[ -o login ]]` both report login with the prefix and non-login without it. The interactive half of `-il` is not needed for requested shells — a shell turns interactive automatically when stdin is a TTY.

Per-shell flag mapping was considered and rejected: it needs a maintained table and leaves unlisted shells (fish, dash) non-login.

pkg/runner/pty_shell.go
jisung-02: Kept exact matching intentionally, documented in 82d0246 (comment on the Windows `loadValidShells`).

- The contract is aligned by construction: the Windows list is the same bare-name list Alpamon itself reports to the server as `ValidShells`, and the server round-trips one of those exact values back in openpty. Windows has no /etc/shells equivalent, so this fixed list is the single source on both sides — a full path cannot legitimately arrive.
- Basename normalization would weaken the allowlist rather than fix anything: any path whose basename matches (e.g. `C:\anywhere\cmd.exe`) would pass and be exec'd directly. Fail-closed exact match is the safer behavior.

pkg/runner/pty_shell.go
jisung-02: Adopted in 82d0246. `shellAllowed` now takes a `caseInsensitive bool` and `resolveShell` passes `runtime.GOOS == "windows"`, so both matching modes are unit-tested on any OS (`TestShellAllowed`). The GOOS-conditional test case was replaced with direct coverage of both modes.

pkg/runner/pty.go
jisung-02: Keeping as-is per your note — session open is not a hot path, and a fresh read stays correct when /etc/shells changes.

pkg/utils/shell_unix_test.go
jisung-02: Kept stdlib style intentionally: pkg/utils has no testify usage at all — every test file in the package uses stdlib `t.Errorf`/`t.Fatalf`, so converting this file would break package-local consistency. The testify convention holds within pkg/runner, where `pty_shell_test.go` already follows it.

pkg/runner/commit_darwin.go
jisung-02: Partially adopted in 82d0246: added the comments/blank-lines-only -> nil case to `shell_unix_test.go`. Skipped the darwin empty-list -> defaults test — the fallback is a 3-line guard, and testing it would require an injection seam around `utils.LoadValidShells()`, which is not worth the structure for that logic.

geunwoonoh (approved):

PR #364, opened 2026-07-15
chore(runner): adopt Go modernize idioms (any, range over int)

### LINKED ISSUE #362: chore(runner): adopt Go modernize idioms (any, range over int)
## Background

The PR #360 review suggested changing `map[string]interface{}` to `map[string]any` at `pkg/runner/pty.go:432`. That line is not one #360 touched (it was last modified in 5814e7d), and converting only that single occurrence would split the notation within the same package, so it was deferred to a follow-up.

## Current state

Counts across the repo, excluding generated code under `pkg/db/ent/`.

| Item | Count | Notes |
| --- | --- | --- |
| `map[string]interface{}` | 31 across 13 files | `internal/protocol`, `pkg/config`, `pkg/runner`, `pkg/scheduler`, `pkg/executor/handlers/*` |
| `interface{}` overall | 84 | includes the 31 above |
| `for i := 0; i < N; i++` | 47 | candidates for range over int |

go.mod declares `go 1.25.11`, so both `any` and range over int are available. The gopls modernize analyzer already flags both patterns.

## Proposal

Converting the whole surface in one mechanical pass is the only way to avoid mixed notation. It is a pure substitution with no behavior change, so review is light relative to the diff size.

- Apply `modernize -fix ./...`, then confirm generated code is excluded.
- Verify no behavior change with `go build ./... && go vet ./... && go test ./... -p 1`.

## Decisions needed

- One PR for everything, or split per package.
- Whether to target all 84 `interface{}` occurrences or only the 31 `map[string]interface{}` ones.

Refs #360

jisung-02:
## Background

PR #360 review suggested `map[string]interface{}` → `map[string]any` at a single line in `pkg/runner/pty.go`. Fixing one occurrence would split the notation within the package, so it was deferred here (#362) to modernize the whole surface at once. Two patterns flagged by the gopls modernize analyzer are in scope: `interface{}` → `any` and `for i := 0; i < N; i++` → `for i := range N`. Generated code under `pkg/db/ent/` is excluded.

## Changes

Three commits, applied per category so each stays easy to review.

1. `interface{}` → `any` (25 files). All 84 non-generated occurrences, applied with `modernize -any -fix`. Pure alias substitution, no behavior change. This unifies notation within each package, which was the original concern from #360.
2. range over int (18 files). The 46 counting loops flagged by `modernize -rangeint -fix`. One production loop in `pkg/collector/scheduler`; the rest are tests. Loop variable and bound semantics are identical.
3. Typed structs for fixed-shape payloads (4 files). Two `map[string]any` sites had a fixed key set but were built as maps, so a key typo would only surface on the wire: the editor `settings.json` body and the config-apply report bodies. They are now `codeServerSettings` in `editor.go` and `applySuccess`/`applyError` in `configreceiver`. Struct field order follows the alphabetical key order `json.Marshal` emits for a map, so output is byte-identical; golden tests lock that.

The remaining `any` usages are left as-is on purpose: dynamic JSON parsing (`nft -j` output in `firewall/nftables.go`), runtime command args (`common/base.go` `GetStringArg` etc.), and arbitrary SSL options (`config.SSLOpt`) are genuine dynamic-typing boundaries where `any` is the correct idiom.

## Scope note

This intentionally does not run `modernize -fix ./...` wholesale. That would also apply ~36 other rewrites (`t.Context`, `atomic.Int32`, `omitempty` removal, `slices.Contains`, `maps.Copy`, etc.), several of which are not pure substitutions and need individual review. Those are left for a follow-up.

## Tests

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- `go test ./... -p 1` passes after each commit.
- New golden tests: `TestToSettingsJSONGolden` (editor) and `TestApplyReportWireFormat` (configreceiver) assert the JSON output is byte-identical to the prior map literals.
- golangci-lint reports only two pre-existing findings (`logPath`, `expireDate`) that exist unchanged on `main` and are outside this diff; the changed files add no new findings.

## Checklist

- [x] Notation unified per package (no mixed `interface{}`/`any`)
- [x] Generated code untouched
- [x] Payload wire format verified byte-identical via golden tests

Refs #360
Closes #362


pkg/runner/editor.go
```diff
@@ -119,17 +119,31 @@ func (c *CodeServerConfig) ToConfigYAML() string {
 		c.Auth, c.DisableTelemetry, c.DisableUpdateCheck)
 }
 
+// codeServerSettings is the settings.json body for the code-server editor.
+// Fields are declared in the alphabetical key order that json.MarshalIndent
+// emits for a map, so the rendered file is byte-identical to the prior literal.
+type codeServerSettings struct {
```
geunwoonoh, line 125: Suggestion (optional): the "declare fields in alphabetical JSON-key order so output stays byte-identical to the map" invariant is self-imposed and slightly fragile. Key order is semantically irrelevant to a JSON consumer, so the ordering isn't functionally required — the struct's real win is compile-time key safety. It's safe as-is (well documented + locked by `TestToSettingsJSONGolden`) and matches the PR's zero-wire-diff goal, so keeping it is defensible; just flagging the coupling for future edits.

pkg/runner/editor.go
```diff
@@ -119,17 +119,31 @@ func (c *CodeServerConfig) ToConfigYAML() string {
...
+	WorkspaceTrust  bool   `json:"security.workspace.trust.enabled"`
```
geunwoonoh: Naming nit (optional): this field holds the *enabled* state (`!c.DisableWorkspaceTrust`), so `WorkspaceTrustEnabled` would read a touch clearer, mirroring the neighboring `OpenWalkthrough`. Not required.

geunwoonoh (dismissed):
## Summary

This refactor applies the two idioms flagged by the gopls `modernize` analyzer across non-generated code (`interface{}` → `any`, and `for i := 0; i < N; i++` → `for range N`), and replaces two fixed-shape `map[string]any` JSON payloads with typed structs. It is split into three per-category commits, which makes it easy to review.

All three changes are behavior-preserving, and the higher-risk struct changes ship with golden tests that lock the wire format. I reviewed each category adversarially and found no functional defects.

**Local verification:** `gofmt` clean · `go vet` clean · golden tests pass · firewall/scheduler tests pass · `go build ./cmd/alpamon` succeeds.

## Critical Issues (0)

None.

## Warnings (0)

None.

## Suggestions (2, optional)

1. **`codeServerSettings` field ordering (editor.go)** — "declare fields in alphabetical JSON-key order so the output stays byte-identical to the map" is a self-imposed, somewhat fragile invariant. Key order is semantically irrelevant to a JSON consumer, so this ordering constraint is not functionally required; the struct's real value is compile-time key safety. That said, the rationale is clearly documented and `TestToSettingsJSONGolden` catches any violation, so it is safe — and it aligns with this PR's "zero wire diff" goal, so keeping it is a defensible defensive choice. (The 2-field configreceiver structs are negligible.)
2. **Naming nit (editor.go)** — `WorkspaceTrust bool` actually holds the *enabled* state (`!c.DisableWorkspaceTrust`), so `WorkspaceTrustEnabled` would read slightly clearer, mirroring the neighboring `OpenWalkthrough`. Not required.

## Review by requested dimension

- **Go conventions**: Conformant. `any` is a Go 1.18+ type alias (identical semantics), and `for range N` is the Go 1.22+ idiom. Notably, the genuine dynamic-typing boundaries are correctly left as `any` (`nftables.go` `nft -j` parsing, `common/base.go` runtime args, `config.SSLOpt`) — converting those to structs would have been over-engineering.
- **Duplicate / unused code**: None. No dead code or unused symbols introduced; the replaced map literals are removed. The `Success` field appearing in both `applySuccess` and `applyError` is justified since they are distinct payload shapes.
- **Performance**: No regressions, with a slight net improvement. `any` vs `interface{}` is an alias (zero runtime difference); `for range N` compiles to the same code; map→struct marshaling drops the map allocation, key sort, and value boxing, so struct marshaling is marginally faster (negligible in practice given low call frequency, but strictly positive).
- **Correctness / edge cases**: `for range N` runs zero iterations when `N <= 0`, matching the old `for i := 0; i < N; i++` (including `scheduler.Start`'s `workerCount`). Loops that use the index inside the body correctly keep `i := range N`; only index-free loops were shortened to `range N`. I also confirmed `Session.newRequest` routes both maps and structs through the same `default → json.Marshal` branch (only `string`/`[]byte` are special-cased), so the struct payloads marshal identically to the old maps.

## Highlights

- **Strong PR discipline**: scope is limited to pure substitutions rather than a wholesale `modernize -fix ./...` (the ~36 rewrites needing individual review — `t.Context`, `atomic.Int32`, `slices.Contains`, etc. — are deferred to a follow-up), and the work is split into three per-category commits.
- **Only the risky changes are guarded by golden tests**: `TestToSettingsJSONGolden` and `TestApplyReportWireFormat` lock the wire format byte-for-byte.
- **Exemplary PR description**: documents the background (#360), the excluded scope, and the rationale for each place `any` was retained.

---

**Verdict**: APPROVED

- [x] Follows stack conventions (gopls modernize idioms)
- [x] Tests included and passing (new golden tests, re-verified locally)
- [x] No security vulnerabilities (pure refactor, no trust-boundary change)
- [x] Documentation/description thorough

pkg/runner/editor.go
jisung-02: Addressed in 7e12d8b. Reordered the fields to follow CodeServerConfig's editor-settings order and dropped the "declare in alphabetical JSON-key order" invariant. TestToSettingsJSONGolden now locks the emitted output directly, so field order is no longer a compiler-unchecked coupling — any real change to the wire format is caught by the golden test instead. Key order is irrelevant to code-server, so the reordered settings.json is functionally identical.

pkg/runner/editor.go
jisung-02: Renamed to WorkspaceTrustEnabled in d319b5d to match the enabled state it holds, mirroring OpenWalkthrough. The JSON tag is unchanged, so the wire format is unaffected.

geunwoonoh (approved):

Issue #10, opened 2024-11-10
Fix bugs related to Web FTP


geunwoonoh:
We need to fix the following bugs in the Web FTP functionality:

- The system does not support Korean characters in file paths.
- Downloading large ZIP files is failing.
- Downloading specific files causes a panic.
- The system is unable to load settings for the FtpClient.
- An error occurs during Alpamon installation due to the requirement to select Yes/No.

PR #365, opened 2026-07-20
feat(runner): detect non-Alpacon access sessions (Phase 1)

junho226:
## Summary

Adds the alpamon (agent) side of non-Alpacon access detection (Phase 1). alpamon
receives `session_event` messages from the `alpamon-pam` PAM session hook over
the existing `/var/run/alpamon/auth.sock`, decides whether the session originated
from Alpacon (Websh/Command) using the existing `pidToSessionMap`, acks the socket
immediately, and asynchronously POSTs a `NonAlpaconAccessEvent` to alpacon-server.
The whole feature is gated behind the server-synced `detect_local_access` policy
flag (default off) and is fail-open: detection never blocks or delays a login.

A `sshd_use_pam` self-check is added to the sync payload so the console can flag
hosts where sshd bypasses PAM (`UsePAM no`), which would defeat PAM-based detection.

Part of a three-repo change: `alpamon-pam` (PAM session hook), this repo (agent),
and `alpacon-server` (ingestion endpoint + policy flag).

## Pipeline results

- Tests: PASS — `go test ./... -p 1` all packages ok, 0 failures; the two touched
  packages (`pkg/runner`, `pkg/utils`) re-run with `-count=1` and `-race`, all pass.
- Review: APPROVED (adversarial multi-lens review, 0 critical). Applied fixes B and D
  below from the review.

## Changes

- `pkg/runner/auth_session_event.go` (new): `SessionEventRequest`/`SessionEventResponse`/
  `NonAlpaconAccessEvent` types, `resolveSessionEvent` suppression logic, socket handler
  with immediate ack, and the async emitter with bounded best-effort retry (404 = server
  endpoint not deployed → dropped, not retried).
- `pkg/runner/auth_manager.go`: dispatch `session_event`; `detect_local_access` flag
  setter mirroring `block_local_sudo`; bounded emit-goroutine semaphore.
- `pkg/runner/commit.go`, `commit_types.go`: sync `detect_local_access` policy and
  `sshd_use_pam` self-check.
- `pkg/utils/pam.go`: `GetSSHDUsePAM()` (`yes`/`no`/`""`).

### Review fixes applied
- **Fix B**: the expected-until-Phase-2 404 now logs at Debug (was Warn), so enabling
  the flag before the server endpoint ships does not spam a warning per login.
- **Fix D**: emit goroutines are bounded by a non-blocking semaphore; on a login burst
  against an unreachable server, excess events are dropped (Debug-logged) rather than
  accumulating goroutines. The immediate-ack ordering is preserved.

## Testing

- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./... -p 1` all pass; session-event tests pass under `-race`
- [ ] Manual Linux e2e per the coverage matrix (`ssh`, `ssh host cmd`, sftp, `ssh -N`,
      local console, `su`, `su`-inside-Websh) — deferred, needs a deployed Linux host

## Follow-ups (not in this PR)

- Manual Linux e2e coverage verification (records the `ssh -N` verdict in the spec).
- Phase 2 (alpacon-server): `/api/events/access/` ingestion is a separate PR; until it
  ships, alpamon treats 404 as "not deployed" and drops the event.

- Review cleanup (socket-writer dedup, deterministic session-event negative assertions): tracked in #384.


pkg/utils/pam.go
```diff
@@ -66,3 +67,69 @@ func queryPamVersion() string {
...
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), pamQueryTimeout)
+	defer cancel()
+
+	// sshd -T prints the effective config; requires root, which alpamon
+	// runs as in production. Errors simply yield "unknown".
+	out, err := exec.CommandContext(ctx, sshdPath, "-T").Output()
+	if err != nil {
+		return ""
```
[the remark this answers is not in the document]
junho226: Fixed. The comment now says errors yield `""` (undeterminable), matching both the return value and the `GetSSHDUsePAM` docstring.

docs/superpowers/specs/2026-07-02-non-alpacon-access-detection-design.md
```diff
@@ -0,0 +1,345 @@
+# Non-Alpacon access detection design
+
+Date: 2026-07-02 (updated 2026-07-20)
+Status: Reconciled with handbook team decisions (2026-07-03/04) — awaiting review
```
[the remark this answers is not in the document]
junho226: Fixed. Changed to an unspaced em-dash per the CLAUDE.md writing convention.

docs/superpowers/specs/2026-07-02-non-alpacon-access-detection-design.md
```diff
@@ -0,0 +1,345 @@
...
+Approach: A (PAM `session` hook), reusing the existing `alpamon-pam` auth-socket infrastructure
+Team decisions: alpacon-handbook `reference/off-platform-access-detection-2026-07.md` — detect committed Q3 2026 (firm, customer requirement), enforce committed Q4 2026, backstop layer out of scope (EDR/CSPM lanes)
```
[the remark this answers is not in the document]
junho226: Fixed. Changed to an unspaced em-dash per the CLAUDE.md writing convention.

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,217 @@
...
+
+	var req SessionEventRequest
+	if err := json.Unmarshal(data, &req); err != nil {
+		log.Warn().Err(err).Msg("Invalid session_event request")
+		am.sendSessionEventResponse(unixConn, false)
+		return
+	}
+
+	event, emit := am.resolveSessionEvent(req)
+
```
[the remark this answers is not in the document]
junho226: Fixed. `handleSessionEvent` now rejects a well-formed envelope carrying an unusable event — empty `username`/`service` or non-positive `pid` — before it reaches `resolveSessionEvent`, logging it and acking `received=false`. The ack still goes out on that path, so PAM never waits.

docs/superpowers/specs/2026-07-02-non-alpacon-access-detection-design.md
```diff
@@ -0,0 +1,345 @@
...
+
+### Key insight: these hooks fire only for non-Alpacon sessions by construction
+
+Websh PTYs and deploy Commands are spawned **directly by alpamon**; they do not
+pass through the `sshd` / `login` / `su` PAM stacks. Therefore any session that
+triggers this hook is inherently outside the Alpacon paths.
+
+We still re-check `pidToSessionMap` (via the existing `lookupSessionLocked`) as a
+safety net for edge cases — most importantly `su` executed *inside* a Websh
+session, whose ppid chain traces back to a tracked PTY and must be suppressed.
```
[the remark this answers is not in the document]
junho226: Fixed. Removed the spaces around every em-dash in both documents (31 occurrences across the spec and the plan), and switched the itemized section headings (`### 1.`, `### 2.`) to colons per the CLAUDE.md convention. The remaining em-dashes introduce prose asides rather than itemized descriptions, so they stay as unspaced em-dashes.

docs/superpowers/specs/2026-07-02-non-alpacon-access-detection-design.md
```diff
@@ -0,0 +1,345 @@
...
+
+Note: even for uncaught shells, a subsequent `su` or `sudo` from that shell
+does traverse PAM and is caught by the existing/new hooks — privilege
+escalation remains visible.
+
+## Architecture
+
+Three repositories are involved.
+
+### 1. `alpamon-pam` — `pam_alpamon.c`, `install.sh`, `uninstall.sh`
```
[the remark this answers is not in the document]
junho226: Fixed. Removed the spaces around every em-dash in both documents (31 occurrences across the spec and the plan), and switched the itemized section headings (`### 1.`, `### 2.`) to colons per the CLAUDE.md convention. The remaining em-dashes introduce prose asides rather than itemized descriptions, so they stay as unspaced em-dashes.

docs/superpowers/specs/2026-07-02-non-alpacon-access-detection-design.md
```diff
@@ -0,0 +1,345 @@
...
+- `session` detect events stay active in block mode for audit continuity.
+
+Future possibility unlocked by this layout: reuse the sudo-approval
+round-trip so an SSH login can be held pending real-time console approval
+(access approval) in the account phase.
+
+## Phasing
+
+- **Phase 1** (`alpamon` + `alpamon-pam`, committed Q3 2026 — firm, customer
+  requirement): `pam_sm_open_session`, `session_event` socket handler, async
```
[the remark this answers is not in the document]
junho226: Fixed. Removed the spaces around every em-dash in both documents (31 occurrences across the spec and the plan), and switched the itemized section headings (`### 1.`, `### 2.`) to colons per the CLAUDE.md convention. The remaining em-dashes introduce prose asides rather than itemized descriptions, so they stay as unspaced em-dashes.

docs/superpowers/plans/2026-07-02-non-alpacon-access-detection-phase1.md
```diff
@@ -0,0 +1,1267 @@
...
+}
+```
+
+- [ ] **Step 3: Run tests to verify they fail**
+
+```bash
+go test -v ./pkg/runner -run 'TestSessionEventRequest|TestResolveSessionEvent' -p 1
+```
+
+Expected: FAIL — `undefined: SessionEventRequest`, `undefined: resolveSessionEvent`.
```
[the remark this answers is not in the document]
junho226: Fixed. Removed the spaces around every em-dash in both documents (31 occurrences across the spec and the plan), and switched the itemized section headings (`### 1.`, `### 2.`) to colons per the CLAUDE.md convention. The remaining em-dashes introduce prose asides rather than itemized descriptions, so they stay as unspaced em-dashes.

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,230 @@
...
+		_, statusCode, err := am.session.Post(nonAlpaconAccessEventURL, event, 10)
+		if err != nil {
+			return err
+		}
+		if statusCode == http.StatusNotFound {
+			return retry.Permanent(errAccessEndpointNotDeployed)
+		}
+		if statusCode < 200 || statusCode >= 300 {
+			return fmt.Errorf("access event failed with status code: %d", statusCode)
+		}
```
geunwoonoh, line 434: **Warning:** Every 4xx except 404 is retried for ~25s.

Only 404 is permanent here; 400/401/403 fall through to this generic branch and go through the full backoff (1→2→4→8→10s, ~5 attempts), holding one of the 16 semaphore slots the whole time.

Token expiry / permission problems (401, 403) and schema mismatches (400) will never succeed on retry. Under a login burst with a 401, all 16 slots stay occupied for 25s and every event arriving in that window is dropped — in an audit feature, that is audit loss.

```suggestion
		if statusCode >= 400 && statusCode < 500 {
			return retry.Permanent(fmt.Errorf("access event rejected with status code: %d", statusCode))
		}
		if statusCode < 200 || statusCode >= 300 {
			return fmt.Errorf("access event failed with status code: %d", statusCode)
		}
```

Note that fixing this changes the error path, so the Warn branch below should distinguish "rejected, dropped" from "unreachable after retries".

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,230 @@
...
+	defer cancel()
+
+	b := &retry.ExponentialBackoff{
+		InitialInterval: authRetryInitialInterval,
+		MaxInterval:     authRetryMaxInterval,
+		MaxElapsedTime:  authRetryTimeout,
+	}
+
+	err := retry.Retry(ctx, b, func() error {
+		_, statusCode, err := am.session.Post(nonAlpaconAccessEventURL, event, 10)
```
geunwoonoh, line 406: **Warning:** Retries can duplicate events, and there is no idempotency key.

`session.do()` returns a non-nil error even when the server processed the request successfully — if the response body read fails or the client timeout trips (`session.go:132-135`). The retry then fires and the same access event is recorded more than once.

`NonAlpaconAccessEvent` carries no unique identifier, so the only dedupe key available server-side is a guess at `(host, pid, timestamp)`. With the Phase 2 server contract not yet frozen, this is the cheapest moment to add the field — `github.com/google/uuid` is already in go.mod:

```go
type NonAlpaconAccessEvent struct {
    EventID   string    `json:"event_id"`   // uuid.NewString() in resolveSessionEvent
    Username  string    `json:"username"`
    ...
}
```

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,230 @@
...
+	// otherwise spawn goroutines without limit. Acquire a slot without
+	// blocking (the ack was already sent above); drop the event if the
+	// budget is exhausted.
+	select {
+	case am.emitSem <- struct{}{}:
+	default:
+		log.Debug().
+			Str("username", event.Username).
+			Str("service", event.Service).
+			Msg("access event dropped: emit concurrency limit reached")
```
geunwoonoh, line 333: **Warning:** Audit event loss is only recorded at Debug.

A semaphore-exhaustion drop means "the server has been unreachable for a while and logins keep arriving" — an abnormal state — yet at Debug level it vanishes under default log settings.

The 404 drop at Debug (line 219) is fine: that is an expected pre-Phase-2 steady state. This one is a different kind of event and deserves to be visible. Raising it to Warn, or keeping a drop counter with a periodic aggregate log if Warn spam is a concern, would surface the loss.

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+}
+
+// emitAccessEvent POSTs a non-Alpacon access event to alpacon-server
+// with bounded best-effort retry (same backoff envelope as the sudo
+// approval path). It runs on its own goroutine; failures are logged and
+// dropped so detection never blocks logins. A 404 means the server does
+// not implement the endpoint yet (Phase 2 not deployed) and is not
+// retried.
+func (am *AuthManager) emitAccessEvent(event NonAlpaconAccessEvent) {
```
geunwoonoh, line 382: **Warning:** `emitAccessEvent` has zero test coverage.

`auth_session_event_test.go` covers parsing, suppression, ack, gating, and the semaphore well — but this function, the only one that actually talks to the server, is entirely unverified. Uncovered branches:

- 404 → `errAccessEndpointNotDeployed` permanent handling (the headline behavior of this PR)
- non-2xx retry followed by the Warn drop
- the `am.session == nil` guard just below

An `httptest.NewServer`-backed `scheduler.Session` makes all three deterministic. Fixing the 4xx handling below makes a permanent-error test necessary anyway.

geunwoonoh (commented):
## Code review: Phase 1 non-Alpacon access detection

**Stack**: golang · **Changes**: +636 -1 in 11 files

**Local verification** (run in a worktree on this branch):

```
gofmt -l pkg/runner pkg/utils           -> (no output, clean)
go vet ./pkg/runner/... ./pkg/utils/... -> exit 0
go test ./pkg/runner/... ./pkg/utils/... -count=1 -race
  ok  github.com/alpacax/alpamon/v2/pkg/runner  29.630s
  ok  github.com/alpacax/alpamon/v2/pkg/utils    1.458s
```

---

### Summary

alpamon receives `session_event` from alpamon-pam's `pam_sm_open_session` hook over `auth.sock`, suppresses Alpacon-originated (Websh/Command) sessions via the existing `pidToSessionMap` tracker, and asynchronously POSTs the rest to alpacon-server as non-Alpacon access events. The whole path is gated behind the `detect_local_access` policy flag (default off) and is fail-open. It additionally reports `sshd -T`'s `usepam` value in the sync payload so the console can flag hosts where PAM is bypassed.

**The design itself is solid.** Ack-before-emit ordering, semaphore-bounded goroutines, default-off policy, and a socket answer on every path — the "never block a login" requirement is held consistently across both the code and the tests. The three earlier Copilot findings are already addressed.

---

### Critical issues (0)

None. I found no path that blocks a login, panics, or deadlocks. (Even an `AuthManager` with a nil `emitSem` only drops events, thanks to the `select` + `default` — no panic.)

---

### Warnings (4)

**1. `pkg/runner/auth_session_event.go:210-211` — every 4xx except 404 is retried for ~25s**

Only 404 is permanent; 400/401/403 fall through to the generic error branch and go through the full backoff (1→2→4→8→10s, ~5 attempts, up to ~25s), holding one of the 16 semaphore slots the whole time.

- Token expiry / permission problems (401, 403) and schema mismatches (400) will **never** succeed on retry.
- Under a login burst with a 401, all 16 slots stay occupied for 25s and every event arriving in that window is dropped. In an audit feature, that is audit loss.

**2. `pkg/runner/auth_session_event.go:202-214` — retries can duplicate events, and there is no idempotency key**

`session.do()` returns a non-nil error even when the server processed the request successfully, if the response body read or the client timeout trips. The retry then fires and the same access event is recorded more than once.

`NonAlpaconAccessEvent` carries no unique identifier, so the only dedupe key available server-side is a guess at `(host, pid, timestamp)`. With the Phase 2 server contract not yet frozen, this is the cheapest moment to add the field (`github.com/google/uuid` is already in go.mod).

**3. `pkg/runner/auth_session_event.go:148-152` — audit event loss is only recorded at Debug**

A semaphore-exhaustion drop means "the server has been unreachable for a while and logins keep arriving" — an **abnormal** state, yet at Debug level it vanishes under default log settings. The 404 drop at Debug is fine (an expected pre-Phase-2 state); this one is a different kind of event.

Raising it to Warn, or keeping a drop counter with a periodic aggregate log if Warn spam is a concern, would make the loss visible.

**4. Test coverage — `emitAccessEvent` has zero tests**

`auth_session_event_test.go` covers parsing, suppression, ack, gating, and the semaphore well, but `emitAccessEvent` — the only function that actually talks to the server — is entirely unverified. Uncovered branches:

- 404 → `errAccessEndpointNotDeployed` permanent handling (the headline behavior of this PR, untested)
- non-2xx retry followed by the Warn drop
- the `am.session == nil` guard

An `httptest.NewServer`-backed `scheduler.Session` makes all three deterministic. Fixing #1 makes a 4xx-permanent test necessary anyway.

---

### Suggestions (7)

**5. `pkg/runner/auth_manager_tracker_test.go:13-20` — `newTestAuthManager()` is now identical to `NewEmptyAuthManager()`**

Adding `emitSem` to both in this PR made their field sets match exactly. It is now pure duplication, and every future field has to be added in two places.

**6. `pkg/utils/pam.go:83-84` — `GetSSHDUsePAM` shares `pamCacheMutex` with `GetPamVersion`**

Both **exec an external process while holding the mutex**, each for up to `pamQueryTimeout` (3s). `collectData()` / `syncServerData()` call them back to back, so a cold cache serializes into a worst case of ~6s. A separate mutex (`sshdCacheMutex`) removes the coupling.

Relatedly, `InvalidatePamCache()` does not touch `sshdUsePAMCache`. That is correct — an alpamon-pam upgrade has nothing to do with sshd config — but sharing a mutex suggests the two caches are managed together. Splitting the mutex clears that up too.

**7. `pkg/runner/auth_session_event.go:124-135` — the policy gate could move ahead of `resolveSessionEvent`**

`detect_local_access` defaults to off, yet every PAM session still costs a `unix.Getsid` syscall, an RLock, and two map lookups, plus a Debug log for suppressed events that will never be emitted. It also takes the RLock twice (once inside `resolveSessionEvent`, once here).

The raw performance win is one syscall, so it is negligible — the real value is that "flag off means do nothing" becomes readable straight from the code. If the suppression Debug log is deliberately kept for troubleshooting while the flag is off, that is a fair reason to keep the current order; please note it in a comment if so.

**8. `pkg/runner/auth_session_event.go:193` — the `authRetryTimeout` reuse rationale does not apply on this path**

The constant is documented at `auth_manager.go:155` as `// Less than PAM 30s timeout`. But emit runs on its own goroutine after the ack has already been sent, so the PAM timeout is irrelevant here. The value is reasonable; the mismatch between the documented reason and the actual use is the problem — someone later tuning it against the PAM timeout would silently change emit behavior. A separate `accessEventEmitTimeout` (same value is fine), or at minimum a comment on why it is reused.

**9. `pkg/runner/auth_session_event.go:160-175` — third copy of the socket response writer**

`sendIsAlpconResponse`, `sendSudoApprovalResponse`, and `sendSessionEventResponse` all share the same "marshal → log and return on failure → write → log on failure" shape. A single `writeSocketResponse(conn net.Conn, v any, kind string)` would cover all three.

This is not a problem this PR introduced — it follows the existing pattern — so **please do not refactor it in this PR**; a separate cleanup is the right place.

**10. `pkg/runner/auth_manager.go:172` vs `192-194` — duplicated `emitSem` initialization**

The init inside `authManagerOnce.Do` and the nil guard after it are redundant with each other. `localSudoRequests`, `completionChannels`, and `session` follow the same duplicated pattern, so this is **consistent with existing style** and does not need to be broken here. Worth being aware, though, that the nil-guard assignment happens without holding `am.mu`, so concurrent `GetAuthManager` calls carry a theoretical race (as they already do for the other fields).

**11. `pkg/runner/auth_session_event_test.go:171,234` — sleep-based negative assertions**

Two places use `time.Sleep(100 * time.Millisecond)` to "give a wrong emit a chance to fire". That can false-negative on a loaded CI runner and adds 200ms to the suite. Negative assertions are genuinely hard to make deterministic, so this is not a blocker — but since `handleSessionEvent` ends with `defer unixConn.Close()`, waiting for the client-side `Read` to hit EOF would confirm "handler finished" without a sleep.

---

### Nits (2)

**12. `pkg/runner/commit_types.go:61,166` — acronym casing diverges within one feature**

`utils.GetSSHDUsePAM()` (all caps, matching Go convention) vs `SshdUsePam` (mixed). The struct spelling matches the existing `PamVersion` style, so keeping it as-is is fine — just noting that both spellings now coexist in the same change. If picking one, staying with the existing struct style (`SshdUsePam`) is the more surgical choice.

**13. `CLAUDE.md:112` — bullet starts lowercase**

The neighboring entries (`PTY session lifecycle`, `Tunnel operations`, `PAM authentication and sudo approval`) all start with a capital; this line starts with `non-Alpacon access detection...`.

---

### Performance review

Things that **add** load:

- `sshd -T` process spawn added to the sync path → mitigated by the 3-hour TTL cache. **But it shares a mutex with `GetPamVersion`, so a cache miss serializes** (#6).
- One `Getsid` syscall and two RLocks per PAM session even with the flag off (#7). Measured impact is negligible.
- 4xx retries occupying a semaphore slot for 25s (#1). **This is the only material performance risk.**

Things that are **well designed**:

- Writing the ack before emit → zero added login latency.
- The 16-slot semaphore bounds goroutine growth, and the non-blocking `select` guarantees it never stalls the ack path.
- `sshdUsePAMCache` with a 3-hour TTL avoids an exec on every sync cycle.

`parseSSHDUsePAM`'s `strings.Split` plus a per-line `strings.ToLower` runs once every 3 hours over ~100 lines — **no optimization needed**.

### Duplicate / unused code review

- **Real duplication**: `newTestAuthManager` ↔ `NewEmptyAuthManager` (#5) — introduced by this PR, worth cleaning up here.
- **Pre-existing pattern duplication**: the three socket response writers (#9), the double `emitSem` init (#10) — better handled in a separate PR.
- **`SessionEventRequest.Type`**: parsed but unused. `IsAlpconRequest.Type` is the same, and it documents the wire protocol, so keeping it is correct.
- **Unused functions/variables**: `go vet` clean, no dead code found.

---

### Highlights

- **The ack-first ordering is pinned down three ways** — code, comments, and tests. `TestHandleSessionEvent_SuppressedStillAcks`, `_FlagOffDoesNotEmit`, and `_DropsWhenEmitConcurrencyExhausted` each explicitly assert "ack is still true even when suppressed or dropped". These target exactly the invariant that matters most in this feature.
- **The `errAccessEndpointNotDeployed` comment explains *why* `errors.Is` is used** ("retry.Retry unwraps the PermanentError"). I checked it against `internal/retry/retry.go:66` (`return permanent.Err`) — it is accurate.
- **`TestHandleSessionEvent_DropsWhenEmitConcurrencyExhausted` narrows `emitSem` to 1 for a deterministic check.** Writing a concurrency-bound test without flake is nicely done.
- Reusing the existing `lookupSessionLocked` keeps the sudo-approval path and the suppression decision on a single source of truth, removing any room for the two to drift.
- The `sshd_use_pam` self-check is a good addition: it lets the console identify hosts where PAM-based detection is defeated, shoring up the trustworthiness of the detection feature itself.

---

**Verdict**: COMMENTED

Nothing here blocks a merge. Of the four warnings, **#1 (4xx permanent handling) and #2 (`event_id`)** are far cheaper to handle before the Phase 2 server contract solidifies, so I would suggest folding them into this PR. #3 (drop log level) is a one-line change, and #4 (emit tests) follows naturally once #1 is fixed. All suggestions are optional — in particular #9 and #10 are better left out of this PR's scope.

**Checklist**:
- [x] Code follows stack conventions (`gofmt` / `go vet` clean; only a minor acronym-casing inconsistency)
- [x] Tests included and passing (green under `-race`, though `emitAccessEvent` is uncovered)
- [x] No security vulnerabilities (no credential logging; `auth.sock` is 0600 root-owned)
- [x] Documentation updated (`CLAUDE.md` reflects the change)
- [ ] Audit event delivery reliability (retry duplication / unbounded 4xx retry / loss visibility)

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,230 @@
...
+			Msg("Incomplete session_event request; dropping")
+		am.sendSessionEventResponse(unixConn, false)
+		return
+	}
+
+	event, emit := am.resolveSessionEvent(req)
+
+	am.sendSessionEventResponse(unixConn, true)
+
+	am.mu.RLock()
```
geunwoonoh, line 292: **Suggestion:** The policy gate could move ahead of `resolveSessionEvent` (line 124).

`detect_local_access` defaults to off, yet every PAM session still costs a `unix.Getsid` syscall, an RLock, and two map lookups — plus a Debug log for suppressed events that will never be emitted. It also takes the RLock twice: once inside `resolveSessionEvent`, once here.

To be honest about the size of this: the raw performance win is one syscall, so it is negligible. The real value is that "flag off means do nothing" becomes readable straight from the code, and the double RLock collapses into one.

If the suppression Debug log is deliberately kept for troubleshooting while the flag is off, that is a fair reason to keep the current order — please note it in a comment if so.

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,230 @@
...
+	if am.session == nil {
+		log.Warn().Msg("HTTP session not available; dropping access event")
+		return
+	}
+
+	baseCtx := am.ctx
+	if baseCtx == nil {
+		baseCtx = context.Background()
+	}
+	ctx, cancel := context.WithTimeout(baseCtx, authRetryTimeout)
```
geunwoonoh, line 396: **Suggestion:** The `authRetryTimeout` reuse rationale does not apply on this path.

The constant is documented at `auth_manager.go:155` as `// Less than PAM 30s timeout`. But emit runs on its own goroutine after the ack has already been sent, so the PAM timeout is irrelevant here.

The value itself is reasonable — the mismatch between the documented reason and the actual use is the problem. Someone later tuning this against the PAM timeout would silently change emit behavior. A separate `accessEventEmitTimeout` (same value is fine), or at minimum a comment explaining why it is reused.

Minor related note: `session.Post` builds its own context from `req.Context()`, so `am.ctx` cancellation stops the retry loop between attempts but not an in-flight POST. On shutdown that means up to 16 goroutines lingering for up to the 10s HTTP timeout — bounded and harmless, just worth knowing.

pkg/runner/auth_session_event.go
```diff
@@ -0,0 +1,230 @@
...
+			Msg("access event dropped: emit concurrency limit reached")
+		return
+	}
+	go func() {
+		defer func() { <-am.emitSem }()
+		emitFn(event)
+	}()
+}
+
+func (am *AuthManager) sendSessionEventResponse(conn net.Conn, received bool) {
```
geunwoonoh, line 352: **Suggestion:** Third copy of the socket response writer.

`sendIsAlpconResponse`, `sendSudoApprovalResponse`, and this one all share the same "marshal → log and return on failure → write → log on failure" shape. A single `writeSocketResponse(conn net.Conn, v any, kind string)` would cover all three.

This is not a problem this PR introduced — it follows the existing pattern — so **please do not refactor it here**. A separate cleanup PR is the right place.

pkg/runner/auth_manager_tracker_test.go
```diff
@@ -15,6 +15,7 @@ func newTestAuthManager() *AuthManager {
 		pidToSessionMap:    make(map[int]*SessionInfo),
 		localSudoRequests:  make(map[string]*SudoRequest),
 		completionChannels: make(map[string]chan struct{}),
+		emitSem:            make(chan struct{}, emitConcurrencyLimit),
```
geunwoonoh: **Suggestion:** `newTestAuthManager()` is now identical to `NewEmptyAuthManager()`.

Adding `emitSem` to both in this PR made their field sets match exactly. It is now pure duplication, and every future field has to be added in two places.

```go
func newTestAuthManager() *AuthManager { return NewEmptyAuthManager() }
```

pkg/utils/pam.go
```diff
@@ -66,3 +67,69 @@ func queryPamVersion() string {
...
+)
+
+// GetSSHDUsePAM reports sshd's effective UsePAM setting: "yes", "no",
+// or "" when it cannot be determined (no sshd, non-linux, query error).
+// A host reporting anything but "yes" bypasses PAM-based access
+// detection for pubkey SSH logins, so the value is surfaced to the
+// server via sync. Cached with the same TTL as the pam version to avoid
+// spawning sshd -T on every sync cycle.
+func GetSSHDUsePAM() string {
+	pamCacheMutex.Lock()
```
geunwoonoh: **Suggestion:** Sharing `pamCacheMutex` with `GetPamVersion` serializes two external process execs.

Both functions exec an external process **while holding this mutex**, each for up to `pamQueryTimeout` (3s). `collectData()`, `collectEssentialData()`, and `syncServerData()` all call `GetPamVersion()` and `GetSSHDUsePAM()` back to back, so a cold cache serializes into a worst case of ~6s. A separate `sshdCacheMutex` removes the coupling.

Relatedly, `InvalidatePamCache()` does not touch `sshdUsePAMCache`. That is correct — an alpamon-pam upgrade has nothing to do with sshd config — but sharing a mutex suggests the two caches are managed together. Splitting the mutex clears that up too.

pkg/runner/auth_manager.go
```diff
@@ -157,6 +169,7 @@ func GetAuthManager(controlClient *ControlClient, session *scheduler.Session) *A
 			localSudoRequests:  make(map[string]*SudoRequest),
 			completionChannels: make(map[string]chan struct{}),
 			session:            session,
+			emitSem:            make(chan struct{}, emitConcurrencyLimit),
```
geunwoonoh, line 185: **Suggestion:** This init is redundant with the `if authManager.emitSem == nil` guard added at lines 192-194 — the guard already covers the `Once` path.

`localSudoRequests`, `completionChannels`, and `session` follow the same duplicated pattern, so this is **consistent with existing style** and does not need to be broken here.

Worth being aware, though, that the nil-guard assignment happens without holding `am.mu`, so concurrent `GetAuthManager` calls carry a theoretical race — as they already do for the other fields.

pkg/runner/auth_session_event_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	server, client := net.Pipe()
+	raw := []byte(`{"type":"session_event","username":"alice","service":"su","pid":424242,"ppid":5555}`)
+	go am.handleSessionEvent(raw, server)
+
+	resp := readSessionEventAck(t, client)
+	if !resp.Received {
+		t.Errorf("suppressed events must still ack true, got %+v", resp)
+	}
+	time.Sleep(100 * time.Millisecond) // give a wrong emit a chance to fire
```
geunwoonoh, line 388: **Suggestion:** Sleep-based negative assertion (also at line 234).

This can false-negative on a loaded CI runner and adds 200ms to the suite across the two occurrences. Negative assertions are genuinely hard to make deterministic, so this is not a blocker.

Since `handleSessionEvent` ends with `defer unixConn.Close()`, waiting for the client-side `Read` to hit EOF would confirm "handler finished" without a sleep — turning the timing assumption into an actual synchronization point.

pkg/runner/commit_types.go
```diff
@@ -58,6 +58,7 @@ var commitDefs = map[string]commitDef{
 type ServerData struct {
 	Version    string  `json:"version"`
 	PamVersion string  `json:"pam_version,omitempty"`
+	SshdUsePam string  `json:"sshd_use_pam,omitempty"`
```
geunwoonoh: **Nit:** Acronym casing diverges within one feature — `utils.GetSSHDUsePAM()` (all caps, matching Go convention) vs `SshdUsePam` (mixed), also at line 166.

The struct spelling matches the existing `PamVersion` style, so keeping it as-is is fine. Just noting that both spellings now coexist in the same change. If picking one, staying with the existing struct style is the more surgical choice.

CLAUDE.md
```diff
@@ -109,6 +109,7 @@ Commands from the Alpacon console flow through:
 - `terminal_manager.go`: PTY session lifecycle
 - `tunnel_client.go`, `tunnel_daemon.go`: Tunnel operations
 - `auth_manager.go`: PAM authentication and sudo approval
+- `auth_session_event.go`: non-Alpacon access detection (PAM session events via auth.sock, gated by `detect_local_access` policy)
```
geunwoonoh: **Nit:** This bullet starts lowercase while its neighbors (`PTY session lifecycle`, `Tunnel operations`, `PAM authentication and sudo approval`) all start with a capital.

```suggestion
- `auth_session_event.go`: Non-Alpacon access detection (PAM session events via auth.sock, gated by `detect_local_access` policy)
```

pkg/runner/auth_session_event.go
junho226: Fixed in f274512. Every 4xx is now `retry.Permanent`, so a rejected event is dropped immediately instead of pinning an emit slot for the backoff window.

Also split the log as you suggested: a rejection logs "Access event rejected by server; dropping event" via a new `errAccessEventRejected` sentinel, while retry exhaustion logs "server unreachable". 404 keeps its own quiet Debug path.

pkg/runner/auth_session_event.go
junho226: Fixed in 5f055b7 (client) and alpacon-server efc6e45 (server).

Confirmed the mechanism you described in `session.go`: both `Client.Do` and `io.ReadAll(resp.Body)` return an error after the server may already have committed, and `retry.Retry` re-invokes the closure on any error — so the duplicate window is real. The 4xx fix above removes 400/401/403 from that set, leaving transport errors and 5xx, which are exactly the ambiguous cases.

`NonAlpaconAccessEvent` now carries `event_id`, generated once in `resolveSessionEvent` so every retry of the same session reuses it. Server-side, `(server, event_id)` is unique and ingestion is a `get_or_create`, so a replay returns the stored row instead of recording a second login. Tests cover both the replay and the distinct-id case.

geunwoonoh (dismissed):

eunyoung14 (commented):
Reviewed together with alpamon-pam#37 (producer side of the `session_event` contract) and the server-side ingestion PR. No merge blockers from my side — the two fixes that mattered most are real, and the cross-repo contract checks out.

**Verified against the code at head:**

- **4xx retry storm fix (f274512): confirmed.** Every 4xx now returns `retry.Permanent(...)`; checked `internal/retry` that a `*PermanentError` short-circuits immediately, so a rejected event costs one POST and releases its emit slot instead of pinning it for the ~25s backoff window.
- **Duplicate-event fix (5f055b7): confirmed, and the two halves interlock.** `event_id` is generated once in `resolveSessionEvent` and the same value is reused across every retry attempt (closure capture in `emitAccessEvent`). Cross-checked the server side: a replayed POST with the same `event_id` is deduplicated and answered with a 2xx — which matters because this client treats any non-404 4xx as a permanent rejection, so a 409-style dedup response would have logged spurious warnings. The current pairing is correct.
- Contract with the PAM hook: field names/types match exactly, and missing `rhost`/`tty` keys unmarshal to `""` as intended. The required-field validation (`username`/`service`/`pid`) is a good guard; note the hook can theoretically emit `service: ""` (NULL PAM item fallback), which this side would drop — acceptable, since `PAM_SERVICE` is always populated in practice.
- Version skew is safe in both directions: 404 from an older server is dropped quietly at Debug, and an older agent ignores the unknown policy key.

**One new finding (non-blocking):**

- `InvalidatePamCache()` clears `pamCache`/`pamCacheTime` but not `sshdUsePAMCache`/`sshdUsePAMCacheTime`, so after an invalidation the `sshd_use_pam` value can stay stale for up to the full cache TTL even though the function's doc comment promises the next call re-queries the system. Suggest clearing both caches there (and this dovetails with the open `pamCacheMutex` thread — a dedicated mutex/cache pair for the sshd check would resolve both).

**+1 on two open threads before the flag is enabled for real traffic** (fine to land as follow-ups given `detect_local_access` defaults to off):

- The semaphore-exhaustion drop is logged at Debug. That state means "server unreachable for a while AND logins keep arriving" — audit events are being lost invisibly under default log settings. A rate-limited Warn would make the loss observable.
- `emitAccessEvent` has no direct test coverage — every test stubs `emitAccessEventFn`, so the 404-permanent, 4xx-permanent, and retry/backoff paths of the only function that talks to the server are unverified. An `httptest.Server`-based test would cover all three cheaply.

junho226:
@eunyoung14 Thank you for the review. I've addressed the non-blocking finding.

**InvalidatePamCache / sshd cache (90a83c0):** Split the sshd UsePAM check into its own mutex/cache, as you suggested. It's now fully independent of the pam-version cache, so (a) `sshd -T` no longer blocks `GetPamVersion` callers, and (b) `InvalidatePamCache` clears only the pam version — a pam-package upgrade has nothing to do with sshd config, so that's the correct behavior (documented). This also resolves the open `pamCacheMutex` thread.

**Two follow-ups** (both flagged as fine to defer since `detect_local_access` defaults to off, and both have since been done): ① the semaphore-exhaustion drop now logs at Warn (rate-limited to once/minute) instead of Debug, so audit loss is visible — 468a997; ② `emitAccessEvent` now has direct httptest coverage for the 404 / 4xx-permanent / retry paths — b303849. The remaining test/refactor cleanup is tracked in #384.

Thanks for verifying the four-way contract and the 4xx/idempotency interaction.

pkg/runner/auth_session_event.go
junho226: Fixed in f274512 (all 4xx permanent, Warn split into rejected-vs-unreachable). Resolving.

pkg/runner/auth_session_event.go
junho226: Fixed: event_id in 5f055b7 (client) + server efc6e45 ((server, event_id) unique + get_or_create). Resolving.

pkg/runner/auth_session_event.go
junho226: Fixed in 468a997. The emit-drop now logs at Warn, rate-limited to once/minute (a sustained outage keeps intervening drops at Debug), so the audit loss is visible without flooding the log.

pkg/runner/auth_session_event.go
junho226: Done in b303849: added TestEmitAccessEvent_DropsOn404, _DropsOnRejection, and _RetriesOn429, all backed by an httptest.Server-wired scheduler.Session — covering the 404-permanent, 4xx-permanent, and 429-retry branches.

pkg/runner/auth_session_event.go
junho226: Kept the current order and documented why in 468a997: resolveSessionEvent runs first so its suppression Debug trace stays available even while the flag is off. Comment added per your note.

pkg/runner/auth_session_event.go
junho226: Added a comment in 468a997 explaining that authRetryTimeout is reused only for its value here — the "less than PAM 30s" rationale does not apply because emit runs on its own goroutine after the ack. Also noted the bounded in-flight-POST-on-shutdown behavior you flagged.

pkg/runner/auth_session_event.go
junho226: Agreed it should not be refactored in this PR. Tracked as a separate cleanup in #384. Resolving.

pkg/runner/auth_manager_tracker_test.go
junho226: Fixed in 468a997: newTestAuthManager now returns NewEmptyAuthManager().

pkg/utils/pam.go
junho226: Done in 90a83c0: split out a dedicated sshdUsePAMMutex, so sshd -T no longer serializes with GetPamVersion and the two caches are visibly independent. InvalidatePamCache stays pam-only (documented).

pkg/runner/auth_manager.go
junho226: Leaving as-is — as you note it matches the existing localSudoRequests/completionChannels/session pattern, and the nil-guard covers the Once path. Resolving.

pkg/runner/auth_session_event_test.go
junho226: Valid, and there is a subtlety (a wrongful emit is on a child goroutine, so an EOF wait alone can still false-negative). Tracked with a note in #384 rather than patched hastily. Resolving.

pkg/runner/commit_types.go
junho226: Keeping the struct spelling (SshdUsePam) to match the existing PamVersion style, as you suggested is the more surgical choice. Resolving.

CLAUDE.md
junho226: Fixed in 468a997 (capitalized to match the neighbouring bullets).

eunyoung14 (commented):
Re-reviewed at `468a997`, cross-checked against alpamon-pam#37 and alpacon-server#2575 at their heads. I verified all 21 resolved threads against code rather than trusting the flags — **every substantive claim holds**, including the ones that are easy to fake:

- **4xx → permanent (f274512)**: `auth_session_event.go:260-263` returns `retry.Permanent` for 400-499, and `internal/retry/retry.go:65-68` does the `errors.As` unwrap **before** the `MaxElapsedTime` check and before `NextBackOff()` — so it really is zero backoff, not a shorter one. `TestEmitAccessEvent_DropsOnRejection` pins call count at 1.
- **`event_id` (5f055b7)**: generated once at `:110` inside `resolveSessionEvent`, captured by the emit closure at `:191`, re-marshaled identically on every attempt. Not per-attempt. Server side enforces `UniqueConstraint(server, event_id)`.
- **Drop-log rate limit**: a single package-level `atomic.Int64` with `Load`→`CompareAndSwap` (`:179-180`). Correct under concurrent drops (exactly one CAS wins), no map, no lock, no I/O under a lock.
- **`emitAccessEvent` coverage**: three httptest-backed tests now drive a real `scheduler.Session` through the real retry loop (`:268-327`). This was the function that talks to the server and had zero coverage.
- **pam mutex split** (`pkg/utils/pam.go:80`) and the `InvalidatePamCache` doc (`:26-28`) — both fixed, and the comment now matches the behavior instead of promising something else.

I also confirmed the never-block-a-login invariant end to end rather than by inspection alone: the ack precedes every blocking operation on all six branches of `handleSessionEvent`, and I audited every `am.mu.Lock()` in `auth_manager.go` (lines 200, 210, 423, 575, 655, 694, 713, 777, 833) for I/O or channel waits under the write lock — there are none; `handleSudoApprovalRequest` unlocks at `:428`/`:460` before writing to the socket. On the producer side, `pam_alpamon.c:775` bounds the round trip at 2s and `:878` returns `PAM_SUCCESS` through `cleanup:` unconditionally, so both skew directions are safe.

Local run: `go build`, `go vet`, `gofmt -l` all clean; `go test ./pkg/runner/... ./pkg/utils/... -count=1 -race` green (36.3s / 1.5s).

Not blocking — the flag is default-off and the server endpoint isn't deployed yet. But three of these should be settled before `detect_local_access` is turned on for real traffic, and one is an e2e row I'd move to the top of the list.

## Before enabling the flag

### 1. `auth_session_event.go:253-255` — the 429 retry provably cannot succeed against the throttle the server PR actually ships, and it pins an emit slot for ~33s while failing

alpacon-server#2575 lands `LocalAccessEventThrottle` at **`600/minute`** — a DRF `SimpleRateThrottle`, so a 60-second window. This client's budget is `MaxElapsedTime: authRetryTimeout` = 25s with backoff 1→2→4→8→10s.

**A 60s throttle window cannot clear inside a 25s budget.** Every retry is guaranteed to re-throttle, and the `Retry-After` header DRF returns is ignored.

Failure scenario: a host crosses 600 events/min (Ansible/CI fan-out, or a reboot re-opening sessions). Each throttled event occupies one of the 16 semaphore slots for its full budget, all 16 saturate, and *newly arriving* events are then dropped at the semaphore — which is the audit-loss pattern @geunwoonoh's W1 was about, re-entering through 429 instead of 4xx.

Either honour `Retry-After`, or — given the server's window structurally exceeds the client's budget — treat 429 as a permanent drop with a Warn. Same audit outcome, no slot pinning.

(Related: the comment at `auth_manager.go:144-148` says a slot is held "up to `authRetryTimeout`" = 25s. Actual worst case is ~33s — `retry.go:70` checks `MaxElapsedTime` only *after* an attempt returns, and each attempt carries its own 10s HTTP timeout.)

### 2. `commit_types.go:61,166` — `sshd_use_pam,omitempty` makes "undeterminable" indistinguishable from "unchanged", and it fails in the reassuring direction

`GetSSHDUsePAM` returns `""` on no-sshd, non-linux, or exec error (`pkg/utils/pam.go:103-129`). With `omitempty`, `""` omits the key. On the server, `servers/api/serializers.py:1019` gates on `if 'sshd_use_pam' in self.validated_data:` — an absent key **preserves the previous value**. The model's own help text (`servers/models.py:110`) promises "Null when alpamon could not determine it", which only holds on the very first sync.

Failure scenario: a host reports `usepam yes`; later sshd is removed, or `sshd -T` starts failing (unreadable host keys, a config parse error after an edit). alpamon sends nothing, the console keeps showing `sshd_use_pam = True`, and the operator keeps believing PAM-based detection covers that host — while it no longer does.

For a self-check whose entire job is telling you when your detection is structurally bypassed, that's the worst direction to go stale in. `*string` (or dropping `omitempty`) transmits `""`/`null` explicitly.

### 3. `pkg/utils/pam.go:107-119` — a non-nil `LookPath` error is ignored while its return value is still used

```go
sshdPath, err := exec.LookPath("sshd")
if err != nil {
    for _, p := range []string{"/usr/sbin/sshd", "/usr/local/sbin/sshd"} { ... }
}
if sshdPath == "" { return "" }
```

Since Go 1.19, `LookPath` returns **a non-empty path together with `exec.ErrDot`** when the match resolved relative to the current directory (a `.` entry or an empty entry — e.g. a trailing `:` — in `PATH`). If neither absolute fallback exists, `sshdPath` keeps that relative value and `:126` execs it as root.

Exploitation needs write access to alpamon's cwd, which is already root-equivalent, so this is narrow — but the error handling is unambiguously wrong. The better fix inverts the order: try `/usr/sbin/sshd` and `/usr/local/sbin/sshd` **first**, fall back to `LookPath`. sbin is the correct location and it removes `PATH` from the decision entirely.

## The e2e row I'd run first

### 4. `auth_session_event.go:81-115` — the headline suppression claim (`su` inside Websh) likely breaks under sudo's `use_pty`

Suppression matches `unix.Getsid(req.PID)` against a registered PTY-leader pid (`session_unix.go:17`, `auth_manager.go:340-348`), and alpamon-pam sends `getpid()` of the process that opened the session (`pam_alpamon.c:854`) — for `su`, the `su` process itself.

Plain `su -` inside Websh doesn't `setsid`, so its sid is the Websh leader and it's suppressed, as documented. But **`sudo su -` under sudo ≥ 1.9.14 with `use_pty`** (the default on current Debian/Ubuntu) forks a monitor that calls `setsid()` and allocates a new pty. `su` then runs in a new session whose sid isn't in `pidToSessionMap`, and whose `getppid()` is the sudo monitor, also not in the map. Both lookup keys miss → emitted as non-Alpacon.

Result: a false alarm on the single most common Websh admin action. This is the *safe* direction (noise, not a missed detection) — but it's the behavior the doc comment promises, and `su`-inside-Websh is on this PR's own deferred coverage matrix. The unit tests can't catch it because they hand-set `PPID: 5555` to a pre-registered pid (`auth_session_event_test.go:94,108`), which encodes the assumption rather than testing the process topology.

**UNVERIFIED** — reasoned from sudo semantics, I can't run Linux here. Please make `sudo su -` inside Websh (with `use_pty` on) the first row of the manual e2e.

Same class, worth naming in the doc comment: any `setsid`/`nohup`/daemonized invocation inside a Websh session escapes both keys and reports as non-Alpacon.

## Non-blocking

5. **`auth_manager.go:281` and `auth_session_event.go:189` — no `recover()` on either goroutine.** The repo has the opposite convention at `internal/pool/pool.go:93`, `pkg/executor/executor.go:167,375`, `pkg/executor/handlers/file/multipart_stream.go:171,241`. I traced the new path for a reachable panic and found none by construction (`effectiveKind` is nil-safe at `auth_manager.go:52`, `am.session` is nil-checked at `:219`), so this isn't urgent. What changed is blast radius: auth.sock used to see traffic on sudo invocations, now it sees a connection on every PAM session open. A `defer recover()` at the top of `handleSudoRequest` covers all three request types for a couple of lines.

6. **`auth_session_event.go:246-247, 270-277` — 404 now has two meanings and both are dropped at Debug.** alpacon-server#2575 makes the endpoint return **404 when the authenticated APIClient has no live `Server` row** — and `Server.objects` filters `deleted_at__isnull=True`, so a *console-deleted server whose agent is still running* gets 404 on every POST. That is arguably the highest-value alert this feature can produce, and it's the one case dropped invisibly at Debug, indistinguishable in the logs from "Phase 2 not deployed". Separately, the "not deployed" heuristic has no sunset: once the server rollout completes, 404 stops meaning that and the quiet drop becomes permanently wrong. Gating the quiet path on "have I ever seen a 2xx from this endpoint?" would cover both.

7. **`auth_manager.go:351` — 1024-byte read buffer, single `Read`, no framing.** A worst-case alpamon-pam frame with all four items at `PAM_ITEM_MAX_LEN 256` is ≈1133 bytes of plain ASCII before JSON escaping (jansson expands control bytes 6×). Over 1024 → `json.Unmarshal` fails → `Warn` + close → producer reads EOF → `PAM_SUCCESS`, event lost. Fail-open holds so this isn't urgent, but combined with the server's `max_length` caps (128/64/255/64) there are now three independent size ceilings that can silently eat an audit event, and none of the three truncates. Capping in `resolveSessionEvent` to the server's limits turns silent loss into a truncated-but-recorded event.

8. **`pkg/utils/pam.go:132-140` — `parseSSHDUsePAM` forwards `fields[1]` verbatim into the sync payload.** `sshd -T` only ever prints `yes`/`no` for `usepam`, but if it ever printed anything else, DRF would 400 the **whole** sync PATCH — taking `version`, `pam_version` and `load` down with it. Far wider blast radius than one dropped event. Clamp to `{"yes","no",""}` at the source.

9. **`pkg/utils/pam.go:83-88`** — `sshd -T` reads the on-disk config, not the running daemon's. An operator who edits `UsePAM no` without reloading gets `"no"` from a daemon still running with PAM. Not cheaply fixable; the doc just shouldn't say "sshd's effective UsePAM setting". "as configured on disk" is accurate.

10. **`auth_session_event.go:44-52,65-79`** — mirror the pid caveat from `pam_alpamon.c:848-853` into the Go doc comments. `NonAlpaconAccessEvent.PID/PPID` is what an operator sees in the console, and someone will read it as "the intruder's shell" when for sshd it's the daemon child.

11. **`auth_session_event.go:134`** — `PPID` is the one contract field never validated (`Username`/`Service`/`PID` are), and the server model has it non-nullable. A `PPID` of 0 posts fine. Harmless, just asymmetric.

12. **No emit drain on shutdown.** `Stop()` (`auth_manager.go:823-830`) cancels the context and closes the listener but doesn't wait on up to 16 in-flight emits; separately `scheduler.Session.newRequest` uses `http.NewRequest`, not `NewRequestWithContext` (`pkg/scheduler/session.go:109`), so `am.ctx` never reaches the HTTP layer. Both pre-existing and repo-level, bounded at ≤16 × ≤10s. Fine for best-effort audit — just don't claim otherwise.

13. **`auth_session_event.go:177-182`** — when Debug is enabled and the Warn branch wins, the `*zerolog.Event` from `log.Debug()` is discarded without `.Msg()` and never returns to the pool. Once a minute at most. `if warn { dropLog = log.Warn() } else { dropLog = log.Debug() }`.

14. **`.gitignore:11-13`** — `docs/superpowers/` is a local agent tool's scratch output being ignored in a **public** repo, and it means the design spec earlier review rounds referred to is no longer in the tree. `.git/info/exclude` or a global gitignore is the better home.

## Cross-repo / release notes

- **Deploy order: alpacon-server#2575 → this PR → alpamon-pam#37.** If this ships before the server, emit gets 404 → `retry.Permanent(errAccessEndpointNotDeployed)` → Debug drop, which is exactly what that path is for. Safe.
- **Do not publish the alpamon-pam package before the alpamon release carrying this PR.** `handleUpgrade` (`pkg/executor/handlers/system/system.go:156-196`) upgrades `alpamon` and `alpamon-pam` independently, so a session-hook alpamon-pam genuinely can land on an older alpamon. Fail-open holds, but the `default:` arm (`auth_manager.go:404`) then logs one Warn per sshd/login/su session, fleet-wide.
- **`sshd_use_pam` sync-path gap is closed on the server side** (`servers/api/serializers.py:1333` + `test_patch_sync_sshd_use_pam`), so existing servers will now actually record it rather than only newly-bootstrapped ones.
- **No conflict with #388** — zero file overlap (`pkg/utils/pam.go` vs `pkg/utils/platform.go`), and no semantic coupling: `GetSSHDUsePAM` doesn't read `PlatformLike`/`PackageManager`, and `queryPamVersion`'s `rpm` fallback works on SUSE. I'd merge #388 first since it's the smaller independent fix and it's the one that restructures `pkg/utils/`.
- **SUSE coverage caveat for the release note**: alpacon-server's `install_pam_on_connected_servers` composes `yum install -y alpamon-pam` for `platform == 'rhel'`, and #388 makes openSUSE report as `rhel` — so on SUSE hosts the PAM module is never auto-installed and detection stays dark until someone runs `zypper install alpamon-pam` by hand. I'm filing a server-side follow-up for the broader yum/zypper gap.
- **This branch is 60 commits behind main.** The only file main and this PR both touched is `pkg/runner/commit.go`, and the changed functions are disjoint (main: `commitAndNotify`, `getNetworkAddresses`, `calculateBroadcastAddress`; this PR: `collectEssentialData`, `syncServerData`, `syncAccessPolicy`, `collectData`) — no semantic conflict. Still worth a rebase so CI runs against current main.

## Highlights

- **The fail-open story is verified on both sides, not asserted.** The ack precedes every blocking operation on all six branches here, and the producer funnels every error path — missing socket, IO failure, non-root peer, JSON failure — through `cleanup:` to an unconditional `PAM_SUCCESS`. Including the awkward skew (new pam + old alpamon hitting `default:` with no reply).
- **The idempotency handshake was designed across the repo boundary in one round, and it interlocks.** Client-side `event_id` generated once, server-side `UniqueConstraint(server, event_id)` + `get_or_create` returning **201** on replay. That last detail matters: this client treats any non-404 4xx as permanent, so a 409-style dedup response would have turned normal deduplication into dropped audit events. Getting that right without a round trip is good cross-repo work.
- **`sshd_use_pam` is the right instinct.** Shipping a self-check that tells you when your detection mechanism is structurally bypassed (`UsePAM no` → pubkey SSH never reaches PAM) is what makes the honesty claim in ADR 0028 real rather than rhetorical. Finding #2 is about keeping that signal fresh, not about the idea.
- **`TestHandleSessionEvent_DropsWhenEmitConcurrencyExhausted`** narrows `emitSem` to 1 for a deterministic concurrency-bound assertion. Writing that without flake is not easy.
- **Comments explain *why***: `errAccessEndpointNotDeployed`'s note on `errors.Is` unwrapping `PermanentError` is accurate against `internal/retry/retry.go:66`, and the resolve-before-gate rationale answers the question the previous review asked rather than just complying with it.

junho226:
## e2e measurements — `use_pty` topology confirmed, entry-path coverage recorded

Ran the parts of the manual e2e checklist that do not need a live alpacon-server or a Websh session. Environment: `ubuntu:22.04` (aarch64) in Docker, `--cap-add=AUDIT_CONTROL` so `pam_loginuid` (which is `required` in the stock sshd stack) does not fail the session.

**Instrument.** Phase 1 uses `session optional pam_exec.so` appended to the same PAM services the installer hooks. `pam_exec` fires at the same point in the stack as `pam_sm_open_session`, so it observes exactly the PAM items `pam_alpamon.so` reads, plus the live `/proc` topology at that instant. This keeps the measurement independent of both agents — no alpamon, no server, no patched binary. Phase 2 (below) uses the real module.

### 1. `sudo su -` under `use_pty` — the analysis in the last review round is correct

`Defaults use_pty` is on by default in stock `ubuntu:22.04` (`/etc/sudoers:12`), so this reproduces with no configuration.

Observed at `pam_sm_open_session` time for the `su` service:

```
  pid    ppid   sid    comm
  3855   3854   3854   su      <- C
  3854   3831   3854   sudo    <- M : pid == sid, so it called setsid(): new session leader
  3831   1      1      sudo    <- P1
     1   -      1      bash    <- S  (the Websh leader's position)
```

`PAM_TTY` also moves from `/dev/console` on the front-end to `/dev/pts/1` on `su`, which is the pty `use_pty` allocated.

Both suppression keys miss, exactly as described:

- `getsid(3855)` = `3854` = **M**, not in the session map
- `ppid(3855)` = `3854` = **M**, not in the session map

So the pre-fix behaviour really was broken for this path, and the ancestor walk reaches `S` in **3 hops** — matching the write-up, and comfortably inside `maxSessionAncestorDepth = 8`.

One caveat worth recording: an earlier run of mine showed no monitor process and `su` still in `S`'s session. That run had been detached with `setsid`, so sudo had no controlling terminal and `use_pty` never engaged. **Without a controlling tty this is not a `use_pty` test at all** — worth knowing for whoever repeats it.

### 2. Entry-path coverage — what the module would report per path

| case | service | user | rhost | tty | pid/ppid/sid |
|---|---|---|---|---|---|
| ssh interactive (PTY) | `sshd` | tester | 127.0.0.1 | `ssh` | 3539/3533/3539 |
| ssh non-interactive (`ssh host id`) | `sshd` | tester | 127.0.0.1 | `ssh` | 3587/3533/3587 |
| scp | `sshd` | tester | 127.0.0.1 | `ssh` | 3636/3533/3636 |
| sftp | `sshd` | tester | 127.0.0.1 | `ssh` | 3685/3533/3685 |
| `ssh -N` (no command, no shell) | `sshd` | tester | 127.0.0.1 | `ssh` | 3733/3533/3733 |
| `su - tester` (login form) | `su-l` | tester | *absent* | /dev/pts/0 | 3780/1/1 |
| `su tester` (non-login form) | `su` | tester | *absent* | /dev/pts/0 | 3806/1/1 |
| `sudo su - tester` | `sudo` then `su-l` | root / tester | *absent* | /dev/console → /dev/pts/1 | 3855/3854/3854 |

Three things fall out of this that are not in the spec:

- **`ssh -N` does open a PAM session.** The plan lists this as an open question ("confirm whether"). It does, so forward-only sessions are covered.
- **`service` is `su-l`, not `su`, for login-mode `su`** — the common admin form. The server accepts it (free-string column), and coverage is fine because `/etc/pam.d/su-l` does `session include su`, so the line the installer puts in `su` is what fires. But `alpacon-server events/models.py` enumerates only `'sshd'`, `'login'`, `'su'` in its comment, and `sudo` and `su-l` both reach it in practice.
- **`tty` is the literal string `ssh` for every sshd path**, so scp, sftp, `ssh -N` and an interactive shell are indistinguishable in the stored record. Not a defect, but the audit trail cannot answer "did they get a shell or copy a file".

`rhost` is absent for every `su`/`sudo` form, matching the documented omission semantics.

### 3. Fail-open on a real sshd (tier 3)

Built `pam_alpamon.so` from alpamon-pam#37 at `fab5170` (`make pam-only`, exit 0, **0 warnings**), installed it with the real `install.sh` — which registered `session optional pam_alpamon.so` in `sshd`, `su` and `login` — and drove logins with alpamon in three states:

| state | ssh | su |
|---|---|---|
| no socket at all | rc=0, 138 ms | rc=0, 4 ms |
| accepts, never acks | rc=0, **2229 ms** | rc=0, **2064 ms** |
| stale socket, nothing listening | rc=0, 165 ms | rc=0, 4 ms |

`sudo -u tester true` on the bootstrap path (socket absent): rc=0, 4 ms.

Every state succeeds. The ~2.2 s rows are the documented `SESSION_TIMEOUT_SEC` residual, reproduced here on a real sshd rather than a synthetic stack. Those rows are also the proof the module is loaded and reaching the socket.

### Not covered here

Websh suppression — a Websh session has to register `S` in `pidToSessionMap`, and `pty.go:135` is the only production writer, so it needs a real server. That will be run on dev. Given the algorithm is unit-tested against an injected topology and the topology itself is now measured against reality, this is the remaining gap rather than the risky one.

eunyoung14 (approved):

geunwoonoh (approved):

PR #368, opened 2026-07-22
feat(system): route Windows upgrade through self-update

### LINKED ISSUE #367: feat(system): support agent self-update on Windows
## Background

On Windows servers, the `upgrade alpamon` command does not take the self-update path and instead returns `Platform 'windows' not supported.`. As a result, the Windows agent is effectively excluded from automatic upgrades, and the `upgrade alpamon` command dispatched by alpacon-server's weekly agent-upgrade scheduler is recorded as a failure without ever being handled. This repeats every week in environments that have Windows servers.

## Current state

`handleUpgrade` in `pkg/executor/handlers/system/system.go` branches the upgrade method per platform:

- debian → apt-get
- rhel → yum
- darwin → `selfUpdate` (self-update from a GitHub Releases binary)
- everything else (including Windows) → "not supported"

Meanwhile, the Windows self-update infrastructure is already in place:

- `pkg/updater/replace_windows.go` — Windows-specific replace that moves the running `.exe` aside to `.old` via `MoveFileEx` and swaps in the new binary
- `pkg/updater/format_windows.go` — PE/COFF binary validation
- `pkg/updater/cleanup_windows.go` — cleanup of leftover `.old` files
- SCM Recovery Actions-based restart path (`restartAgent` in `platform_windows.go`)

In other words, the `selfUpdate` path used by darwin is already built to work on Windows, but Windows is not yet wired into `handleUpgrade`'s platform branch. The Windows runbook in #284 also assumes that "alpamon self-updater handles this automatically", so finishing Windows self-update support looks like the natural direction.

## Expected behavior

On Windows, `upgrade alpamon` runs through the `selfUpdate` path just like darwin, so Windows servers are properly included in the automatic upgrade schedule.

## Proposal

Add a `case "windows"` to the switch in `handleUpgrade` and route it to `selfUpdate`. Confirm the following two points along the way:

1. Whether `wsClient.Restart()` (`pkg/runner/client.go`, closing `RestartChan`) after self-update actually leads to a restart under an SCM-managed environment — i.e. whether it interlocks with the Recovery Actions (status 1 exit → SCM restart).
2. `needPam` (alpamon-pam) handling. On darwin, when `!needAlpamon`, it falls through to a "pam not supported" message. Decide whether the same guard is needed on Windows depending on whether alpamon-pam self-upgrade is supported there.

## Verification

- Add unit tests for the `upgrade` handler in the `windows-latest` CI (added in #284).
- Register a version lower than the latest release on a Windows server, run `upgrade alpamon`, and confirm the self-update and service restart.

## References

- `pkg/executor/handlers/system/system.go:175-187`
- Self-update infrastructure: commit `ffe7029`
- Related issues: #255, #258, #284

jisung-02:
## Background

alpamon installed on Windows could not run the upgrade command. The platform switch in `handleUpgrade` had no windows case, so it fell through to default and returned `Platform 'windows' not supported`. Meanwhile the self-update infrastructure in `pkg/updater` (MoveFileEx-based binary replacement, PE/COFF format validation, stale `.old` cleanup) already supported Windows, so only the wiring was missing.

## Changes

Merged windows into the existing darwin case in the `handleUpgrade` switch (`case "darwin", "windows":`) so both route to `selfUpdate`.

The "already up-to-date" early return that used to live in the darwin case was dead code. `handleUpgrade` already returns earlier at the `!needAlpamon && !needPam` guard (system.go:160), and `needPam` is always false on non-linux, so by the time the switch is reached `needAlpamon` is always true. Removed that dead code and documented the invariant with a comment instead.

Introduced a `selfUpdateFn` injection seam on `SystemHandler` (defaulting to `updater.SelfUpdate`) so the darwin/windows routing can be unit-tested without performing a real GitHub Releases download. The test swaps this field for a fake and verifies routing only, with no network access.

## Safety

`needPam` depends on `queryPamVersion()`, which always returns an empty string when `runtime.GOOS != "linux"`, so it is always false on non-linux. The switch dispatches on `utils.PlatformLike`, which in production is derived from `runtime.GOOS`, so `PlatformLike ∈ {darwin, windows}` guarantees `GOOS ≠ linux`. The darwin/windows case is therefore only reached when `needAlpamon == true`, and the up-to-date case is already handled by the guard before the switch.

## Deployment note

Windows agents registered before this change ships need a one-time manual `alpamon register` re-run on the server to pick up the self-update path. Newly registered agents need no action.

## Testing

- `TestSystemHandler_Upgrade_SelfUpdate` (darwin and windows subtests): verifies both platforms route through `selfUpdateFn` and do not return "not supported". Passing.
- `go build ./cmd/alpamon`, `go vet`, and `go test ./pkg/executor/handlers/system/ ./pkg/updater/ -p 1` all pass.
- CI: the `windows-latest` job already runs the full suite, so no workflow change is needed.

## Checklist

- [x] Unit tests added and passing
- [x] build / vet passing
- [ ] Communicate the manual re-registration step for existing Windows agents

Closes #367

### DIFF
--- a/pkg/executor/handlers/system/system.go
+++ b/pkg/executor/handlers/system/system.go
@@ -180,7 +180,8 @@ func (h *SystemHandler) handleUpgrade(ctx context.Context) (int, string, error) 
 	case "rhel":
 		cmd = fmt.Sprintf("yum update -y %s", pkgList)
 	case "darwin", "windows":
-		// needAlpamon is always true here: needPam is always false on non-linux (pam unsupported; see pkg/utils/pam.go) and the switch is reached only when needAlpamon||needPam.
+		// needAlpamon is always true here: needPam is always false on non-linux
+		// (pam unsupported; see pkg/utils/pam.go) and the switch is reached only when needAlpamon||needPam.
 		return h.selfUpdate(ctx, latestVersion)
 	default:
 		return 1, fmt.Sprintf("Platform '%s' not supported.", utils.PlatformLike), nil
--- a/pkg/executor/handlers/system/system.go
+++ b/pkg/executor/handlers/system/system.go
@@ -37,6 +37,7 @@ type SystemHandler struct {
 	pool            *pool.Pool
 	versionResolver common.VersionResolver
 	apiSession      common.APISession
+	selfUpdateFn    updater.SelfUpdateFunc // defaults to updater.SelfUpdate; tests inject a fake
 }
 
 // NewSystemHandler creates a new system handler.
@@ -65,6 +66,7 @@ func NewSystemHandler(cmdExecutor common.CommandExecutor, wsClient common.WSClie
 		pool:            pool,
 		versionResolver: versionResolver,
 		apiSession:      apiSession,
+		selfUpdateFn:    updater.SelfUpdate,
 	}
 	return h
 }
@@ -177,10 +179,8 @@ func (h *SystemHandler) handleUpgrade(ctx context.Context) (int, string, error) 
 		cmd = fmt.Sprintf("apt-get update -y -o Acquire::Retries=3 && apt-get install --only-upgrade %s -y -o Acquire::Retries=3", pkgList)
 	case "rhel":
 		cmd = fmt.Sprintf("yum update -y %s", pkgList)
-	case "darwin":
-		if !needAlpamon {
-			return 0, fmt.Sprintf("Already up-to-date (alpamon: %s). alpamon-pam upgrade is not supported on macOS.", version.Version), nil
-		}
+	case "darwin", "windows":
+		// needAlpamon is always true here: needPam is always false on non-linux (pam unsupported; see pkg/utils/pam.go) and the switch is reached only when needAlpamon||needPam.
 		return h.selfUpdate(ctx, latestVersion)
 	default:
 		return 1, fmt.Sprintf("Platform '%s' not supported.", utils.PlatformLike), nil
@@ -196,7 +196,7 @@ func (h *SystemHandler) handleUpgrade(ctx context.Context) (int, string, error) 
 
 // selfUpdate downloads and replaces the binary from GitHub Releases, then triggers restart.
 func (h *SystemHandler) selfUpdate(ctx context.Context, latestVersion string) (int, string, error) {
-	if err := updater.SelfUpdate(ctx, latestVersion, updater.Options{}); err != nil {
+	if err := h.selfUpdateFn(ctx, latestVersion, updater.Options{}); err != nil {
 		return 1, fmt.Sprintf("Self-update failed: %v", err), err
 	}
 
--- a/pkg/executor/handlers/system/system_test.go
+++ b/pkg/executor/handlers/system/system_test.go
@@ -12,6 +12,8 @@ import (
 	"github.com/alpacax/alpamon/v2/internal/pool"
 	"github.com/alpacax/alpamon/v2/pkg/agent"
 	"github.com/alpacax/alpamon/v2/pkg/executor/handlers/common"
+	"github.com/alpacax/alpamon/v2/pkg/updater"
+	"github.com/alpacax/alpamon/v2/pkg/utils"
 	"github.com/alpacax/alpamon/v2/pkg/version"
 )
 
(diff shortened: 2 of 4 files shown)


pkg/executor/handlers/system/system.go
```diff
@@ -177,10 +179,8 @@ func (h *SystemHandler) handleUpgrade(ctx context.Context) (int, string, error)
 		cmd = fmt.Sprintf("apt-get update -y -o Acquire::Retries=3 && apt-get install --only-upgrade %s -y -o Acquire::Retries=3", pkgList)
 	case "rhel":
 		cmd = fmt.Sprintf("yum update -y %s", pkgList)
-	case "darwin":
-		if !needAlpamon {
-			return 0, fmt.Sprintf("Already up-to-date (alpamon: %s). alpamon-pam upgrade is not supported on macOS.", version.Version), nil
-		}
+	case "darwin", "windows":
+		// needAlpamon is always true here: needPam is always false on non-linux (pam unsupported; see pkg/utils/pam.go) and the switch is reached only when needAlpamon||needPam.
```
geunwoonoh: Suggestion: this invariant comment is a single ~190-char line. Wrapping it improves readability and keeps diffs narrow (optional, non-blocking):

```suggestion
		// needAlpamon is always true here: on non-linux needPam is always false
		// (pam unsupported; see pkg/utils/pam.go), and the switch is reached only when needAlpamon || needPam.
```

geunwoonoh (dismissed):
## Summary

Fixes the Windows `upgrade` failure: `handleUpgrade`'s switch had no `windows` case, so it fell through to `default` and returned `Platform 'windows' not supported`. Since `pkg/updater` already ships full Windows self-update infrastructure (MoveFileEx-based replacement, PE/COFF validation, stale `.old` cleanup), only the wiring was missing. This PR:

1. Merges windows into the darwin case (`case "darwin", "windows":`) so both route to `selfUpdate`.
2. Removes the now-unreachable "already up-to-date" early return from the darwin case and documents the invariant instead.
3. Adds a `selfUpdateFn` injection seam plus a unit test covering darwin/windows routing without a real download.
4. (Incidental) Refactors `isMachO` to `slices.Contains` and the oversize test to `min()`.

---

### Verification (checked directly)

- **Dead-code removal is sound**: the guard at `system.go:159` returns when `!needAlpamon && !needPam`, so the switch is only reached when `needAlpamon || needPam`. `queryPamVersion()` returns `""` when `runtime.GOOS != "linux"` (`pkg/utils/pam.go`), so `needPam` is always false on non-linux. In production `PlatformLike` is derived from `runtime.GOOS`, so `PlatformLike ∈ {darwin, windows}` ⇒ `GOOS ≠ linux` ⇒ `needAlpamon == true`. The removed early return was genuinely unreachable.
- **Wiring reaches working code**: `replace_windows.go` / `format_windows.go` / `cleanup_windows.go` exist, and `.goreleaser.yaml` explicitly publishes a `.tar.gz` archive for Windows via `format_overrides`, so the self-updater's hardcoded `.tar.gz` URL will not 404.
- **Build/checks**: `go vet` clean; `TestSystemHandler_Upgrade_SelfUpdate` (darwin/windows) and `TestDownloadFile_Oversize` pass. `go.mod` is `go 1.25.12`, so `slices` / `min` are available.

### Critical Issues (0)

None.

### Warnings (0)

None.

### Suggestions (1, optional)

1. **[system.go:183]** The invariant comment is a single ~190-char line. Go has no line-length limit, but wrapping it to two lines would help readability and diffs (see inline comment).

### Highlights

- Instead of deleting the dead branch by intuition, the author **proved the invariant** and left the reasoning as a comment, so a reviewer can reproduce the safety argument.
- The **`selfUpdateFn` injection seam** is a minimal, idiomatic way to unit-test routing without network access; introducing the `updater.SelfUpdateFunc` type is appropriate.
- The test is **table-driven** across darwin/windows and asserts `called` / `gotVersion` / `exitCode` / absence of `"not supported"`, restoring the `PlatformLike` global via `t.Cleanup` (safe under `-p 1` sequential runs).
- `slices.Contains` / `min()` refactors are behavior-preserving and idiomatic for Go 1.25.
- The PR description's **deployment note** (existing Windows agents need a one-time `alpamon register` re-run) is clearly called out and honestly left unchecked in the checklist.

**Duplicate/unused code**: new imports (`updater`, `utils`) and the `SelfUpdateFunc` type are all used; no unused or duplicated code — this is a net reduction that removes dead code.

**Performance**: control-flow change only, no runtime impact. `slices.Contains` is the same O(n) over 6 elements as the previous loop — no regression or improvement.

**Security**: no hardcoded secrets or injection. The self-update path already validates checksum and binary format (PE/COFF + architecture).

---

**Verdict**: APPROVED

- [x] Follows stack conventions (Go 1.25 idioms, injection seam)
- [x] Tests included and passing (darwin/windows routing)
- [x] Docs/deployment note captured (re-registration step in the PR body)

pkg/executor/handlers/system/system.go
jisung-02: Thanks — done in afda40f. Wrapped the invariant to two lines so it stays readable in side-by-side diffs. Comment-only change, no behavior impact.

geunwoonoh (approved):

PR #369, opened 2026-07-22
test: modernize test code to Go 1.25 stdlib idioms

jisung-02:
## Background

The `go.mod` baseline is Go 1.25, but a number of test files still used pre-1.21/1.25 patterns. This PR refreshes the test code to the matching stdlib idioms. It is split out from a broader modernization branch so the test-only changes, which touch no production code, can be reviewed and merged on their own.

## Changes

Applied across 8 test files:

- `atomic.Int32` typed values instead of `atomic.AddInt32`/`atomic.LoadInt32` on a bare `int32`
- `sync.WaitGroup.Go` instead of manual `wg.Add(1)` / `go func(){ defer wg.Done() ... }()`
- `maps.Copy` instead of a manual copy loop
- `slices.Contains` instead of a nested membership loop
- `t.Context()` instead of manual `context.WithCancel` + `defer cancel()`
- Dropped the `tt := tt` range-variable capture lines (unneeded since Go 1.22)

Files: `internal/pool/pool_leak_test.go`, `internal/pool/pool_test.go`, `pkg/executor/handlers/user/user_test.go`, `pkg/executor/integration_test.go`, `pkg/executor/resource_test.go`, `pkg/logsink/writer_test.go`, `pkg/migrate/migrate_test.go`, `pkg/runner/commit_test.go`.

Not every `wg.Add(1)` was converted: the calls in `resource_test.go`, `integration_test.go`, `pool_test.go`, and `pool_leak_test.go` follow the `Add(1)` -> `pool.Submit(...)` -> compensating `Done()` on submit error pattern. `wg.Go` cannot express that, because the goroutine is started by the pool rather than by the test, so those are intentionally left as-is.

## Tests

Behavior is unchanged; these are idiom swaps only. Verified against the current `main` source tree (no production code in this PR):

- `go build ./...` clean
- `go vet` clean on the affected packages
- `go test -p 1` passes for `internal/pool`, `pkg/executor/...`, `pkg/logsink`, `pkg/migrate`, `pkg/runner`
- `go test -race -count=5 -cpu=1,2 ./internal/pool/... ./pkg/logsink/...` clean

## Checklist

- [x] Test-only change, no production code
- [x] Build, vet, and tests pass on the affected packages

## Follow-up

Test assertion style is inconsistent across the suite: 27 test files already use testify (`assert`/`require`) while 72 still hand-roll `if got != want { t.Errorf(...) }` and `if err != nil { t.Fatalf(...) }`. A follow-up PR will unify the value-assertion-heavy files on testify, starting with the highest-repetition ones (`pkg/executor/handlers/user`, `handlers/shell`, `handlers/system`, `pkg/updater`, `pkg/signing`, `pkg/utils/archive`, etc.). Files whose core is concurrency, timing, or network setup (e.g. `pkg/logsink/writer_test.go`, `internal/pool/pool_leak_test.go`) are intentionally left as-is, since testify buys little there. This is out of scope here to keep the PR a pure Go 1.25 idiom refresh.

### DIFF
--- a/internal/pool/pool_leak_test.go
+++ b/internal/pool/pool_leak_test.go
@@ -183,7 +183,7 @@ func TestPool_MaxGoroutineEnforcement(t *testing.T) {
 	pool := NewPool(maxWorkers, 100)
 	defer func() { _ = pool.Shutdown(5 * time.Second) }()
 
-	var maxConcurrent int32
+	var maxConcurrent atomic.Int32
 	var concurrent atomic.Int32
 
 	ctx := context.Background()
@@ -195,8 +195,8 @@ func TestPool_MaxGoroutineEnforcement(t *testing.T) {
 
 			// Track max concurrent
 			for {
-				max := atomic.LoadInt32(&maxConcurrent)
-				if current <= max || atomic.CompareAndSwapInt32(&maxConcurrent, max, current) {
+				observed := maxConcurrent.Load()
+				if current <= observed || maxConcurrent.CompareAndSwap(observed, current) {
 					break
 				}
 			}
@@ -209,7 +209,7 @@ func TestPool_MaxGoroutineEnforcement(t *testing.T) {
 	// Wait for all jobs to complete
 	time.Sleep(600 * time.Millisecond)
 
-	if int(maxConcurrent) > maxWorkers {
-		t.Errorf("max concurrent workers %d exceeded limit %d", maxConcurrent, maxWorkers)
+	if int(maxConcurrent.Load()) > maxWorkers {
+		t.Errorf("max concurrent workers %d exceeded limit %d", maxConcurrent.Load(), maxWorkers)
 	}
 }
--- a/internal/pool/pool_leak_test.go
+++ b/internal/pool/pool_leak_test.go
@@ -47,7 +47,7 @@ func TestPool_NoLeakAfterPanic(t *testing.T) {
 	pool := NewPool(3, 20)
 	ctx := context.Background()
 
-	var completed int32
+	var completed atomic.Int32
 
 	// Submit jobs that panic
 	for range 10 {
@@ -59,7 +59,7 @@ func TestPool_NoLeakAfterPanic(t *testing.T) {
 	// Submit normal jobs after panics
 	for range 10 {
 		_ = pool.Submit(ctx, func() error {
-			atomic.AddInt32(&completed, 1)
+			completed.Add(1)
 			return nil
 		})
 	}
@@ -72,7 +72,7 @@ func TestPool_NoLeakAfterPanic(t *testing.T) {
 		t.Errorf("shutdown failed: %v", err)
 	}
 
-	if atomic.LoadInt32(&completed) == 0 {
+	if completed.Load() == 0 {
 		t.Error("no normal jobs completed after panics - workers may have died")
 	}
 }
@@ -184,14 +184,14 @@ func TestPool_MaxGoroutineEnforcement(t *testing.T) {
 	defer func() { _ = pool.Shutdown(5 * time.Second) }()
 
 	var maxConcurrent int32
-	var concurrent int32
+	var concurrent atomic.Int32
 
 	ctx := context.Background()
 
 	for range 100 {
 		_ = pool.Submit(ctx, func() error {
-			current := atomic.AddInt32(&concurrent, 1)
-			defer atomic.AddInt32(&concurrent, -1)
+			current := concurrent.Add(1)
+			defer concurrent.Add(-1)
 
 			// Track max concurrent
 			for {
--- a/internal/pool/pool_test.go
+++ b/internal/pool/pool_test.go
@@ -97,14 +97,14 @@ func TestPoolConcurrency(t *testing.T) {
 		}
 	}()
 
-	var concurrent int32
+	var concurrent atomic.Int32
 	var maxConcurrent int32
 
 	// Submit many jobs
 	for range 50 {
 		if err := pool.Submit(context.Background(), func() error {
-			current := atomic.AddInt32(&concurrent, 1)
-			defer atomic.AddInt32(&concurrent, -1)
+			current := concurrent.Add(1)
+			defer concurrent.Add(-1)
 
 			// Track max concurrent
 			for {
@@ -138,7 +138,7 @@ func TestPoolPanicRecovery(t *testing.T) {
 		}
 	}()
 
-	var completed int32
+	var completed atomic.Int32
 
 	// Submit job that panics
 	if err := pool.Submit(context.Background(), func() error {
@@ -149,7 +149,7 @@ func TestPoolPanicRecovery(t *testing.T) {
 
 	// Submit normal job after panic
 	err := pool.Submit(context.Background(), func() error {
-		atomic.AddInt32(&completed, 1)
+		completed.Add(1)
 		return nil
 	})
 
@@ -160,7 +160,7 @@ func TestPoolPanicRecovery(t *testing.T) {
 	// Wait for job completion
 	time.Sleep(100 * time.Millisecond)
 
-	if atomic.LoadInt32(&completed) != 1 {
+	if completed.Load() != 1 {
 		t.Error("normal job did not complete after panic")
 	}
 }
--- a/internal/pool/pool_test.go
+++ b/internal/pool/pool_test.go
@@ -16,14 +16,14 @@ func TestPoolBasic(t *testing.T) {
 		}
 	}()
 
-	var counter int32
+	var counter atomic.Int32
 	var wg sync.WaitGroup
 
 	// Submit 10 jobs
 	for i := range 10 {
 		wg.Add(1)
 		err := pool.Submit(context.Background(), func() error {
-			atomic.AddInt32(&counter, 1)
+			counter.Add(1)
 			wg.Done()
 			time.Sleep(10 * time.Millisecond)
 			return nil
@@ -37,8 +37,8 @@ func TestPoolBasic(t *testing.T) {
 	// Wait for all jobs to complete
 	wg.Wait()
 
-	if atomic.LoadInt32(&counter) != 10 {
-		t.Errorf("expected 10 jobs to complete, got %d", counter)
+	if counter.Load() != 10 {
+		t.Errorf("expected 10 jobs to complete, got %d", counter.Load())
 	}
 }
 
@@ -98,7 +98,7 @@ func TestPoolConcurrency(t *testing.T) {
 	}()
 
 	var concurrent atomic.Int32
-	var maxConcurrent int32
+	var maxConcurrent atomic.Int32
 
 	// Submit many jobs
 	for range 50 {
@@ -108,8 +108,8 @@ func TestPoolConcurrency(t *testing.T) {
 
 			// Track max concurrent
 			for {
-				max := atomic.LoadInt32(&maxConcurrent)
-				if current <= max || atomic.CompareAndSwapInt32(&maxConcurrent, max, current) {
+				observed := maxConcurrent.Load()
+				if current <= observed || maxConcurrent.CompareAndSwap(observed, current) {
 					break
 				}
 			}
@@ -125,8 +125,8 @@ func TestPoolConcurrency(t *testing.T) {
 	time.Sleep(200 * time.Millisecond)
 
 	// Max concurrent should not exceed worker count
-	if int(maxConcurrent) > maxWorkers {
-		t.Errorf("exceeded max concurrency: got %d, want <= %d", maxConcurrent, maxWorkers)
+	if int(maxConcurrent.Load()) > maxWorkers {
+		t.Errorf("exceeded max concurrency: got %d, want <= %d", maxConcurrent.Load(), maxWorkers)
 	}
 }
 
@@ -168,13 +168,13 @@ func TestPoolPanicRecovery(t *testing.T) {
 func TestPoolShutdown(t *testing.T) {
 	pool := NewPool(2, 10)
 
-	var completed int32
+	var completed atomic.Int32
 
 	// Submit several jobs
 	for range 5 {
 		if err := pool.Submit(context.Background(), func() error {
 			time.Sleep(50 * time.Millisecond)
-			atomic.AddInt32(&completed, 1)
+			completed.Add(1)
 			return nil
 		}); err != nil {
 			t.Errorf("failed to submit job: %v", err)
@@ -188,8 +188,8 @@ func TestPoolShutdown(t *testing.T) {
 	}
 
 	// All jobs should have completed
-	if atomic.LoadInt32(&completed) != 5 {
-		t.Errorf("not all jobs completed: got %d, want 5", completed)
+	if completed.Load() != 5 {
+		t.Errorf("not all jobs completed: got %d, want 5", completed.Load())
 	}
 
 	// New submissions should fail
--- a/pkg/logsink/writer_test.go
+++ b/pkg/logsink/writer_test.go
@@ -58,7 +58,9 @@ func startTestServer(t *testing.T) (path string, frames <-chan []byte, stop func
 		wg    sync.WaitGroup
 	)
 
+	acceptDone := make(chan struct{})
 	wg.Go(func() {
+		defer close(acceptDone)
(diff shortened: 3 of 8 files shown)


pkg/logsink/writer_test.go
```diff
@@ -57,9 +58,7 @@ func startTestServer(t *testing.T) (path string, frames <-chan []byte, stop func
 		wg    sync.WaitGroup
 	)
 
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
+	wg.Go(func() {
 		for {
 			conn, err := ln.Accept()
 			if err != nil {
```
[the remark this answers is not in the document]
jisung-02: Addressed in afb9d7c. The accept loop now signals completion via an `acceptDone` channel (`defer close(acceptDone)` when it returns), and `stop()` blocks on `<-acceptDone` immediately after `ln.Close()`, before locking/closing conns and calling `wg.Wait()`. Since the accept loop is guaranteed to have exited before `wg.Wait()`, no `wg.Add` can run concurrently with `wg.Wait()`.

pkg/logsink/writer_test.go
```diff
@@ -85,7 +84,7 @@ func startTestServer(t *testing.T) (path string, frames <-chan []byte, stop func
 				}
 			}(conn)
 		}
-	}()
+	})
 
 	stop = func() {
 		_ = ln.Close()
```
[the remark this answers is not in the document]
jisung-02: Addressed in afb9d7c. `stop()` now waits on `<-acceptDone` after `ln.Close()` and before locking/closing conns and calling `wg.Wait()`, so the accept loop has provably exited (no further conn appends or `wg.Add`) before shutdown proceeds. Verified with `go test -race -count=5 ./pkg/logsink/...`.

internal/pool/pool_test.go
```diff
@@ -97,14 +97,14 @@ func TestPoolConcurrency(t *testing.T) {
 		}
 	}()
 
-	var concurrent int32
+	var concurrent atomic.Int32
 	var maxConcurrent int32
```
geunwoonoh: **Warning:** `concurrent` on the line above became `atomic.Int32`, but `maxConcurrent` stayed a bare `int32`. Two problems.

First, the inconsistency across three lines.

Second, and more concretely: the assertion at L128 reads `int(maxConcurrent)` non-atomically while worker goroutines may still be CAS-ing it. Nothing joins them first — the test only does `time.Sleep(200ms)`, and `pool.Shutdown` is deferred, so it runs *after* the check. That is a data race by the memory model even though the detector doesn't currently catch it (I ran `-race -count=15 -cpu=1,2` and it stayed quiet; the sleep margin is generous enough in practice).

Converting it closes the race and removes the last `atomic.LoadInt32`/`CompareAndSwapInt32` calls in this function:

```go
var maxConcurrent atomic.Int32
...
for {
    observed := maxConcurrent.Load()
    if current <= observed || maxConcurrent.CompareAndSwap(observed, current) {
        break
    }
}
...
if int(maxConcurrent.Load()) > maxWorkers {
    t.Errorf("exceeded max concurrency: got %d, want <= %d", maxConcurrent.Load(), maxWorkers)
}
```

Not offering this as an applyable suggestion because the single-line change alone would break the build — L111-112 still call `atomic.LoadInt32(&maxConcurrent)`, and those lines are outside this diff hunk.

Side note: the existing `max := ...` shadows the Go 1.21 `max` builtin. Renaming it to `observed` (as above) is a natural part of a modernization pass.

internal/pool/pool_test.go
```diff
@@ -97,14 +97,14 @@ func TestPoolConcurrency(t *testing.T) {
 		}
 	}()
 
-	var concurrent int32
+	var concurrent atomic.Int32
```
geunwoonoh, line 100: **Warning:** this file is now half converted. `TestPoolConcurrency` (here) and `TestPoolPanicRecovery` (L141) moved to `atomic.Int32`, but `counter` in `TestPoolSubmit` (L19, L26, L40) and `completed` in `TestPoolShutdown` (L170, L177, L191) still use `var x int32` + `atomic.AddInt32`/`atomic.LoadInt32`.

Two styles in one file is a worse state than none converted — the next reader can't tell whether the remaining bare-`int32` ones were a deliberate exception or just missed. Please convert all four, or leave all four for a follow-up.

internal/pool/pool_leak_test.go
```diff
@@ -184,14 +184,14 @@ func TestPool_MaxGoroutineEnforcement(t *testing.T) {
 	defer func() { _ = pool.Shutdown(5 * time.Second) }()
 
 	var maxConcurrent int32
```
geunwoonoh: **Warning:** same as `pool_test.go:101`. `concurrent` on the next line was converted while `maxConcurrent` stayed a bare `int32`, and the assertion at L212 reads it non-atomically while workers may still be running — the test only sleeps 600ms (100 jobs / 10 workers x 50ms = 500ms, a 100ms margin) and `pool.Shutdown` is deferred, so it runs after the check.

```go
var maxConcurrent atomic.Int32
...
for {
    observed := maxConcurrent.Load()
    if current <= observed || maxConcurrent.CompareAndSwap(observed, current) {
        break
    }
}
...
if int(maxConcurrent.Load()) > maxWorkers {
    t.Errorf("max concurrent workers %d exceeded limit %d", maxConcurrent.Load(), maxWorkers)
}
```

pkg/logsink/writer_test.go
```diff
@@ -85,10 +86,11 @@ func startTestServer(t *testing.T) (path string, frames <-chan []byte, stop func
 				}
 			}(conn)
```
geunwoonoh: **Warning:** the accept loop was converted to `wg.Go`, but this per-connection reader immediately inside it still uses `wg.Add(1)` + `go func(c net.Conn){ defer wg.Done() ... }(conn)`.

`conn` is declared inside the loop body, so it's a fresh variable per iteration — no Go 1.22 capture concern — and this converts cleanly, dropping the explicit parameter along with it:

```go
wg.Go(func() {
    var hdr [4]byte
    for {
        if _, err := io.ReadFull(conn, hdr[:]); err != nil {
            return
        }
        n := binary.BigEndian.Uint32(hdr[:])
        body := make([]byte, n)
        if _, err := io.ReadFull(conn, body); err != nil {
            return
        }
        ch <- body
    }
})
```

This is safe: the accept goroutine itself holds the WaitGroup counter at >= 1 for its whole lifetime, so an `Add` issued from inside it can never race with `Wait` observing zero.

pkg/logsink/writer_test.go
```diff
@@ -85,10 +86,11 @@ func startTestServer(t *testing.T) (path string, frames <-chan []byte, stop func
...
 		}
-	}()
+	})
 
 	stop = func() {
 		_ = ln.Close()
+		<-acceptDone
```
geunwoonoh, line 93: **Suggestion:** this is a real fix, not defensive padding — without it, an in-flight `Accept()` can append to `conns` after `stop()` has already closed them all, leaving that connection's reader parked forever in `ReadFull` and hanging `wg.Wait()`.

But with no comment, the next reader will see `<-acceptDone` as redundant with the `wg.Wait()` five lines down and delete it. Worth one line:

```suggestion
		// Drain the accept loop first: an in-flight Accept can append to
		// conns after stop() has already closed them, stranding its reader.
		<-acceptDone
```

pkg/runner/commit_test.go
```diff
@@ -83,11 +84,8 @@ func TestLoadValidShells(t *testing.T) {
...
 		for _, commonShell := range commonShells {
-			for _, shell := range shells {
-				if shell == commonShell {
-					foundAny = true
-					break
-				}
+			if slices.Contains(shells, commonShell) {
+				foundAny = true
 			}
 			if foundAny {
```
geunwoonoh: **Suggestion:** the `slices.Contains` swap left the old loop scaffolding behind. Setting `foundAny = true` and then testing it again in a separate `if foundAny { break }` is a leftover of the inner-loop break that no longer exists.

The whole block collapses to one expression:

```go
foundAny := slices.ContainsFunc(commonShells, func(s string) bool {
    return slices.Contains(shells, s)
})
```

If you'd rather keep the loop, at minimum move the `break` inside the `if`. The current shape is correct but reads like a half-finished refactor, which is especially visible in a PR that exists to clean up idioms.

(Not an applyable suggestion — the replacement spans L85-93 and the closing braces at L92-93 fall outside this diff hunk.)

geunwoonoh (commented):
## Review

### Summary

The `go.mod` baseline is Go 1.25 but the test code still used pre-1.21/1.22 patterns; this PR refreshes them: `atomic.Int32`, `sync.WaitGroup.Go`, `maps.Copy`, `slices.Contains`, `t.Context()`, and dropping `tt := tt`. No production code touched.

**Verified locally:**

- `go vet ./internal/pool/... ./pkg/executor/... ./pkg/logsink/... ./pkg/migrate/... ./pkg/runner/...` — clean
- `go test -count=1` on `internal/pool`, `pkg/logsink`, `pkg/migrate`, `pkg/runner` — all pass
- `go test -race -count=15 -cpu=1,2 ./internal/pool` — no race reported

**Performance:** zero runtime delta. `atomic.Int32.Add` compiles to the same instructions as `atomic.AddInt32`, `maps.Copy` is the same range loop internally, and `slices.Contains` is the same linear scan. This is purely a readability change — neither a regression nor an improvement in speed.

---

### Critical issues (0)

None.

### Warnings (3)

**1. `internal/pool/pool_test.go:101`, `internal/pool/pool_leak_test.go:186` — `maxConcurrent` left as a bare `int32`, leaving an unsynchronized read**

Its sibling `concurrent`, three lines away, was converted to `atomic.Int32` but `maxConcurrent` was not. Beyond the inconsistency, a real data race remains: neither test joins the workers before asserting (only `time.Sleep`, and `pool.Shutdown` is deferred so it runs *after* the check), yet both read `int(maxConcurrent)` non-atomically while worker goroutines may still be CAS-ing it.

I ran `-race -count=15 -cpu=1,2` and the detector did not fire — the sleeps are generous enough in practice — so this is latent rather than currently breaking. But it is exactly the class of problem this PR sets out to clean up.

Also worth noting: the existing `max := atomic.LoadInt32(...)` shadows the Go 1.21 `max` builtin. A modernization PR is a good moment to rename it.

**2. `internal/pool/pool_test.go:19,26,40` and `:170,177,191` — only half the file was converted**

The counters in `TestPoolConcurrency` (L100) and `TestPoolPanicRecovery` (L141) were converted, but `counter` in `TestPoolSubmit` and `completed` in `TestPoolShutdown` still use `var x int32` + `atomic.AddInt32`/`LoadInt32`. Two styles in one file is a worse state than none converted: the next reader can't tell whether the remaining bare-`int32` ones were deliberate. Convert all four, or leave all four.

**3. `pkg/logsink/writer_test.go:72-87` — the goroutine nested inside the converted block was not converted**

The accept loop moved to `wg.Go`, but the per-connection reader immediately inside it still uses `wg.Add(1)` + `go func(c net.Conn){ defer wg.Done() ... }(conn)`. `conn` is declared inside the loop body (fresh per iteration, no Go 1.22 capture issue), so it converts cleanly and the explicit `conn` parameter goes away with it. Safety confirmed: the accept goroutine holds the WaitGroup counter at >= 1 for its entire lifetime, so an `Add` from inside can never race with `Wait` reaching zero.

### Suggestions (4)

**4. `pkg/runner/commit_test.go:85-93` — the old loop scaffolding survived the `slices.Contains` swap**

`foundAny = true` followed by a separate `if foundAny { break }` is a leftover of the inner-loop break. Correct, but it reads like a half-finished refactor — which stands out in a PR whose entire purpose is idiom cleanup. See the inline comment.

**5. `pkg/logsink/writer_test.go:61,93` — `acceptDone` needs a comment explaining why it exists**

This is a genuine bug fix: without it, an in-flight `Accept()` can append to `conns` after `stop()` has already closed them, leaving that connection's reader parked in `ReadFull` and hanging `wg.Wait()`. With no comment, the next reader will see `<-acceptDone` as redundant with `wg.Wait()` and delete it. One line is enough — inline suggestion below.

**6. The remaining `wg.Add(1)` calls are not misses — worth stating in the PR body**

`resource_test.go:161`, `integration_test.go:242`, `pool_test.go:24`, and `pool_leak_test.go:23,89` correctly keep `wg.Add(1)`. They follow the `Add(1)` -> `pool.Submit(...)` -> compensating `Done()` on submit error pattern, which `wg.Go` cannot express because the goroutine is started by the pool, not by the test. A line in the PR description would stop a reviewer from reading them as oversights.

**7. Nit (pre-existing):** `pkg/logsink/writer_test.go:288` — `_ = path` is a no-op; `path` is already used at L264. Free deletion while you're in the file, though it's outside this PR's scope.

### Highlights

- **The `t.Context()` conversion was applied case by case, not blanket.** `TestStartWatchdog_StopsOnContextCancel` (migrate_test.go:381) keeps `context.WithCancel` because it has to cancel mid-test, which `t.Context()` can't express. That discrimination is the hard part and it was done right.
- **The `t.Context()` timing shift is safe here.** Cancellation moves from "end of the test body" to "after return, before cleanups" — still ahead of `setupTempDataDir`'s `t.Cleanup` restore and the `t.TempDir()` removal. No behavior change.
- **The PR description is honest about scope**, deferring the testify unification with a stated rationale for which files to leave alone.
- No production code touched; vet clean, tests pass.

---

**Verdict: COMMENTED**

Nothing blocking. That said, warnings 1-3 leave *new* inconsistencies in a PR whose stated purpose is consistency, so one more commit before merge would be worthwhile. Warning 2 (half-converted file) and warning 1 (inconsistency plus a latent race) are cheapest to fix now.

**Checklist**

- [x] Follows Go conventions (vet clean, Go 1.25 idioms used correctly)
- [x] Tests verified passing (ran locally)
- [x] No security concerns (test-only; no secrets or input handling)
- [x] No performance impact (identical compiled output)
- [ ] Consistency — unconverted leftovers remain in the same files (#1, #2, #3)
- [x] No documentation changes needed

internal/pool/pool_test.go
jisung-02: Fixed in 8b5dfb7. `maxConcurrent` is now `atomic.Int32`; the tracking loop uses `Load`/`CompareAndSwap` and the assertion reads `maxConcurrent.Load()`, closing the non-atomic read you flagged. Renamed the `max` local to `observed` so it no longer shadows the builtin. Re-ran `go test -race -count=5 -cpu=1,2 ./internal/pool/...` — clean.

internal/pool/pool_test.go
jisung-02: Fixed in 8b5dfb7. Converted the remaining two: `counter` in `TestPoolSubmit` and `completed` in `TestPoolShutdown` now use `atomic.Int32` with `Add`/`Load`. All four counters in the file are consistent now.

internal/pool/pool_leak_test.go
jisung-02: Fixed in 8b5dfb7. Same conversion as `pool_test.go`: `maxConcurrent` is `atomic.Int32` with `Load`/`CompareAndSwap`, the assertion reads `maxConcurrent.Load()`, and `max` was renamed to `observed`.

pkg/logsink/writer_test.go
jisung-02: Fixed in 8b5dfb7. The per-connection reader is now `wg.Go` and the explicit `conn` parameter is gone (the loop-body `conn` is captured directly). Verified with `go test -race` on `pkg/logsink`.

pkg/logsink/writer_test.go
jisung-02: Done in 8b5dfb7. Added your suggested comment above `<-acceptDone` explaining that it drains the accept loop so an in-flight Accept can't append to conns after they're closed.

pkg/runner/commit_test.go
jisung-02: Done in 8b5dfb7. Collapsed the block to `slices.ContainsFunc(commonShells, func(s string) bool { return slices.Contains(shells, s) })`, dropping the leftover `foundAny = true` / `if foundAny { break }` scaffolding.

geunwoonoh (approved):

PR #371, opened 2026-07-23
refactor: modernize production code to Go 1.25 stdlib idioms

jisung-02:
## Background

The `go.mod` baseline is Go 1.25, but a number of production source files still used pre-1.21/1.25 patterns. This PR refreshes the production code to the matching stdlib idioms. It is the production-code counterpart to #369 (test-only), split out so each set can be reviewed and merged on its own.

## Changes

All changes are behavior-preserving idiom swaps, applied across 11 files:

- `slices.Backward` for reverse iteration over the rollback stack (`register.go`)
- `maps.Copy` instead of manual map-copy loops (`register.go`, `logsink/writer.go`)
- range-over-int (`for range N`) instead of a counted index loop (`service_windows.go`)
- `strings.SplitSeq` instead of `strings.Split` where the result is only range-iterated (`detection.go`, `iptables.go`, `commit_darwin.go`, `commit_windows.go`, `utils/firewall.go`)
- `strings.CutPrefix` instead of `HasPrefix`+`TrimPrefix` pairs (`iptables.go`, `utils/firewall.go`)
- `slices.Sort` + `slices.Reverse` instead of a hand-rolled bubble sort in `ReorderChains` (`iptables.go`)
- `fmt.Appendf(nil, ...)` instead of `[]byte(fmt.Sprintf(...))` (`commit.go`)
- dropped dead `omitempty` on struct-typed JSON fields: `PingResponse.Timestamp` (`internal/protocol/message.go`) and `FtpResult.Data` (`pkg/runner/ftp_types.go`)

Files: `cmd/alpamon/command/register/register.go`, `cmd/alpamon/command/register/service_windows.go`, `internal/protocol/message.go`, `pkg/executor/handlers/firewall/detection.go`, `pkg/executor/handlers/firewall/iptables.go`, `pkg/logsink/writer.go`, `pkg/runner/commit.go`, `pkg/runner/commit_darwin.go`, `pkg/runner/commit_windows.go`, `pkg/runner/ftp_types.go`, `pkg/utils/firewall.go`.

## Safety

The two `omitempty` removals touch structs that are JSON-serialized over the WebSocket connection to Alpacon, but the change is byte-identical on the wire. `encoding/json` never honors `omitempty` on struct-kind fields (`time.Time` and `CommandResult` are both structs), so these fields were always serialized; dropping the tag only removes dead metadata. A search of alpacon-server turned up no raw-JSON deserializer for these payloads, so no consumer change is required.

## Tests

Behavior is unchanged; these are idiom swaps only.

- `go build ./...` clean
- `go vet ./...` clean
- `gofmt -l` clean on all changed files
- `go test -p 1` passes for the affected packages (`internal/protocol`, `pkg/runner`, `pkg/executor/handlers/firewall`, `pkg/utils`, `pkg/logsink`)

## Checklist

- [x] Behavior-preserving idiom swaps, no functional change
- [x] Build, vet, gofmt, and tests pass on the affected packages
- [x] Wire format verified unchanged for the `omitempty` removals

### DIFF
--- a/internal/protocol/message.go
+++ b/internal/protocol/message.go
@@ -45,7 +45,7 @@ type CommandChunk struct {
 // PingResponse represents a ping response
 type PingResponse struct {
 	Query     string    `json:"query"`
-	Timestamp time.Time `json:"timestamp,omitempty"`
+	Timestamp time.Time `json:"timestamp"`
 }
 
 // NewPingResponse creates a new ping response
--- a/pkg/executor/handlers/firewall/iptables.go
+++ b/pkg/executor/handlers/firewall/iptables.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -130,10 +131,8 @@ func (s *IptablesBackend) DeleteRule(ctx context.Context, ruleID, chainName stri
 	}
 
 	// Find rule with matching rule ID in comment
-	lines := strings.Split(output, "\n")
 	lineNum := 0
-
-	for _, line := range lines {
+	for line := range strings.SplitSeq(output, "\n") {
 		if strings.Contains(line, ruleID) {
 			// Extract line number from the beginning of the line
 			parts := strings.Fields(line)
@@ -213,9 +212,7 @@ func (s *IptablesBackend) ListRules(ctx context.Context, chainName string) ([]co
 // parseIptablesSaveOutput parses iptables-save output to extract rules
 func (s *IptablesBackend) parseIptablesSaveOutput(output, filterChain string) []common.FirewallRule {
 	var rules []common.FirewallRule
-	lines := strings.Split(output, "\n")
-
-	for _, line := range lines {
+	for line := range strings.SplitSeq(output, "\n") {
 		line = strings.TrimSpace(line)
 
 		// Skip non-rule lines
@@ -242,10 +239,10 @@ func (s *IptablesBackend) parseIptablesSaveOutput(output, filterChain string) []
 // parseIptablesSaveRuleLine parses a single iptables-save rule line
 func (s *IptablesBackend) parseIptablesSaveRuleLine(line string) *common.FirewallRule {
 	// Remove -A or -I prefix
-	if strings.HasPrefix(line, "-A ") {
-		line = strings.TrimPrefix(line, "-A ")
-	} else if strings.HasPrefix(line, "-I ") {
-		line = strings.TrimPrefix(line, "-I ")
+	if after, ok := strings.CutPrefix(line, "-A "); ok {
+		line = after
+	} else if after, ok := strings.CutPrefix(line, "-I "); ok {
+		line = after
 	}
 
 	parts := strings.Fields(line)
@@ -308,8 +305,7 @@ func (s *IptablesBackend) parseIptablesSaveRuleLine(line string) *common.Firewal
 			}
 		case "--dports":
 			if i+1 < len(parts) {
-				portStrs := strings.Split(parts[i+1], ",")
-				for _, ps := range portStrs {
+				for ps := range strings.SplitSeq(parts[i+1], ",") {
 					if port, err := strconv.Atoi(ps); err == nil {
 						rule.DPorts = append(rule.DPorts, port)
 					}
@@ -373,9 +369,7 @@ func (s *IptablesBackend) ReorderChains(ctx context.Context, chainNames []string
 
 	// Find alpacon jump rule line numbers
 	var jumpLines []int
-	lines := strings.Split(output, "\n")
-
-	for _, line := range lines {
+	for line := range strings.SplitSeq(output, "\n") {
 		parts := strings.Fields(line)
 		if len(parts) < 3 {
 			continue
@@ -401,13 +395,8 @@ func (s *IptablesBackend) ReorderChains(ctx context.Context, chainNames []string
 	}
 
 	// Sort in reverse order (delete from bottom to preserve line numbers)
-	for i := 0; i < len(jumpLines); i++ {
-		for j := i + 1; j < len(jumpLines); j++ {
-			if jumpLines[i] < jumpLines[j] {
-				jumpLines[i], jumpLines[j] = jumpLines[j], jumpLines[i]
-			}
-		}
-	}
+	slices.Sort(jumpLines)
+	slices.Reverse(jumpLines)
 
 	// Delete old jump rules
 	for _, lineNum := range jumpLines {
--- a/pkg/logsink/writer.go
+++ b/pkg/logsink/writer.go
@@ -7,6 +7,7 @@ package logsink
 import (
 	"encoding/binary"
 	"encoding/json"
+	"maps"
 	"net"
 	"os"
 	"path/filepath"
@@ -59,9 +60,7 @@ type Writer struct {
 // Connection failure at construction time is non-fatal; the writer will retry.
 func New(program string, handlers map[string]int) *Writer {
 	h := make(map[string]int, len(handlers))
-	for k, v := range handlers {
-		h[k] = v
-	}
+	maps.Copy(h, handlers)
 	w := &Writer{
 		program:  program,
 		pid:      os.Getpid(),
--- a/pkg/runner/commit.go
+++ b/pkg/runner/commit.go
@@ -179,10 +179,10 @@ func CommitSystemInfo() {
 // commitAndNotify sends commit data to the server and posts a commit event.
 func commitAndNotify(data *commitData) {
 	scheduler.Rqueue.Put(commitURL, data, 80, time.Time{})
-	scheduler.Rqueue.Post(eventURL, []byte(fmt.Sprintf(`{
+	scheduler.Rqueue.Post(eventURL, fmt.Appendf(nil, `{
 		"reporter": "alpamon",
 		"record": "committed",
-		"description": "Committed system information. version: %s"}`, version.Version)), 80, time.Time{})
+		"description": "Committed system information. version: %s"}`, version.Version), 80, time.Time{})
 }
 
 // collectEssentialData collects only the essential categories (info, os)
--- a/pkg/runner/ftp_types.go
+++ b/pkg/runner/ftp_types.go
@@ -63,7 +63,7 @@ type FtpResult struct {
 	Command FtpCommand    `json:"command"`
 	Success bool          `json:"success"`
 	Code    int           `json:"code,omitempty"`
-	Data    CommandResult `json:"data,omitempty"`
+	Data    CommandResult `json:"data"`
 }
 
 type CommandResult struct {
--- a/cmd/alpamon/command/register/register.go
+++ b/cmd/alpamon/command/register/register.go
@@ -9,11 +9,13 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"slices"
 	"strings"
 	"time"
 
@@ -214,8 +216,8 @@ func runRegister(cmd *cobra.Command, args []string) error {
 			return
 		}
 		var errs []error
-		for i := len(rollbacks) - 1; i >= 0; i-- {
-			if e := rollbacks[i](); e != nil {
+		for _, rollback := range slices.Backward(rollbacks) {
+			if e := rollback(); e != nil {
 				errs = append(errs, e)
 			}
 		}
@@ -438,12 +440,8 @@ func mergeCloudAndUserTags(auto, user map[string]string) map[string]string {
 		return nil
 	}
 	out := make(map[string]string, len(auto)+len(user))
-	for k, v := range auto {
-		out[k] = v
-	}
-	for k, v := range user {
-		out[k] = v
-	}
+	maps.Copy(out, auto)
+	maps.Copy(out, user)
 	return out
 }
 
--- a/cmd/alpamon/command/register/service_windows.go
+++ b/cmd/alpamon/command/register/service_windows.go
@@ -216,7 +216,7 @@ func stopRunningService(s *mgr.Service) {
 			fmt.Printf("Warning: failed to signal service stop: %v\n", err)
 		}
 	}
-	for i := 0; i < serviceStopPollAttempts; i++ {
+	for range serviceStopPollAttempts {
 		if status, qerr := s.Query(); qerr == nil && status.State == svc.Stopped {
 			return
 		}
--- a/internal/pool/pool.go
+++ b/internal/pool/pool.go
@@ -50,7 +50,7 @@ func NewPool(maxWorkers int, queueSize int) *Pool {
 
 // startWorkers launches the worker goroutines
 func (p *Pool) startWorkers() {
-	for i := 0; i < p.maxWorkers; i++ {
+	for range p.maxWorkers {
 		p.wg.Add(1)
 		go p.worker()
 	}
--- a/pkg/collector/collector.go
+++ b/pkg/collector/collector.go
@@ -144,7 +144,7 @@ func (c *Collector) Start() {
 
 	go c.scheduler.Start(c.ctx, c.buffer.Capacity)
 
-	for i := 0; i < c.buffer.Capacity; i++ {
+	for range c.buffer.Capacity {
 		c.wg.Add(1)
 		go c.successQueueWorker(c.ctx)
 	}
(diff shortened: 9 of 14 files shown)


pkg/executor/handlers/firewall/iptables.go
```diff
@@ -401,13 +395,8 @@ func (s *IptablesBackend) ReorderChains(ctx context.Context, chainNames []string
...
 
 	// Sort in reverse order (delete from bottom to preserve line numbers)
-	for i := 0; i < len(jumpLines); i++ {
-		for j := i + 1; j < len(jumpLines); j++ {
-			if jumpLines[i] < jumpLines[j] {
-				jumpLines[i], jumpLines[j] = jumpLines[j], jumpLines[i]
-			}
-		}
-	}
+	slices.Sort(jumpLines)
```
geunwoonoh, line 398: Warning: this is the only change in the PR that is not a mechanical 1:1 swap, and it is the one with 0% test coverage.

I verified the equivalence by hand and it is correct: the original was a descending selection sort, and `Sort` + `Reverse` produces the same descending order. Duplicates are structurally impossible here — the inner `for _, chainName` loop `break`s after the first append, so each output line contributes at most one entry, and `iptables -L --line-numbers` line numbers are unique. So the multiset and the ordering both match.

But `go tool cover -func` reports `ReorderChains` at 0.0%, while the untouched `nftables.go` equivalent sits at 45.7%. Nothing here is wrong; it just means the behavior-preserving claim rests on review rather than on tests. A small table test over the ordering would close that gap.

Also noted, and I agree with the choice you made: `slices.SortFunc(jumpLines, func(a, b int) int { return b - a })` would collapse this into one call, but `Sort` + `Reverse` reads better next to the "delete from bottom to preserve line numbers" comment. Current form is the better one.

pkg/logsink/writer.go
```diff
@@ -59,9 +60,7 @@ type Writer struct {
 // Connection failure at construction time is non-fatal; the writer will retry.
 func New(program string, handlers map[string]int) *Writer {
 	h := make(map[string]int, len(handlers))
-	for k, v := range handlers {
-		h[k] = v
-	}
+	maps.Copy(h, handlers)
```
geunwoonoh, line 63: Suggestion (optional): these two lines are exactly `maps.Clone`, with one difference — `maps.Clone(nil)` returns nil, whereas `make` + `Copy` always returns a non-nil map.

That difference is immaterial here: `w.handlers` is only ever read (`writer.go:86`), and reads from a nil map are safe, so the behavior is identical either way.

```suggestion
	h := maps.Clone(handlers)
```

That said, the current form does explicitly guarantee non-nil, which has its own value. **Fine to leave as-is** — flagging only for consideration.

For contrast, `mergeCloudAndUserTags` in `register.go` correctly *cannot* use `Clone` since it merges two sources; `make` + two `Copy` calls is the only right answer there.

pkg/runner/commit.go
```diff
@@ -179,10 +179,10 @@ func CommitSystemInfo() {
 // commitAndNotify sends commit data to the server and posts a commit event.
 func commitAndNotify(data *commitData) {
 	scheduler.Rqueue.Put(commitURL, data, 80, time.Time{})
-	scheduler.Rqueue.Post(eventURL, []byte(fmt.Sprintf(`{
+	scheduler.Rqueue.Post(eventURL, fmt.Appendf(nil, `{
 		"reporter": "alpamon",
 		"record": "committed",
-		"description": "Committed system information. version: %s"}`, version.Version)), 80, time.Time{})
+		"description": "Committed system information. version: %s"}`, version.Version), 80, time.Time{})
```
geunwoonoh, line 185: Suggestion (low priority, pre-existing): this JSON payload is assembled by string formatting, so a `"` or `\` in `version.Version` would emit invalid JSON.

In practice `version.Version` is injected via build-time ldflags, so it is a controlled value and the real risk is low. This also is not a problem the PR introduced — the `fmt.Appendf` swap itself is a strict improvement (it drops 2 allocations and copies down to 1). Raising it only because the line is being touched anyway.

internal/protocol/message.go
```diff
@@ -45,7 +45,7 @@ type CommandChunk struct {
 // PingResponse represents a ping response
 type PingResponse struct {
 	Query     string    `json:"query"`
-	Timestamp time.Time `json:"timestamp,omitempty"`
+	Timestamp time.Time `json:"timestamp"`
```
geunwoonoh, line 48: Verified: `time.Time` is a struct kind, and `encoding/json` never honors `omitempty` on struct fields, so this field was always serialized. Dropping the tag is byte-identical on the wire. The safety argument in the PR description is correct.

Worth calling out that resisting Go 1.24's `omitzero` here was the right call rather than an oversight: `NewPingResponse()` always sets `Timestamp: time.Now()`, so the field can never be zero and `omitzero` would have been equally dead metadata. Deletion is the honest fix.

pkg/runner/ftp_types.go
```diff
@@ -63,7 +63,7 @@ type FtpResult struct {
 	Command FtpCommand    `json:"command"`
 	Success bool          `json:"success"`
 	Code    int           `json:"code,omitempty"`
-	Data    CommandResult `json:"data,omitempty"`
+	Data    CommandResult `json:"data"`
```
geunwoonoh, line 66: Verified: `CommandResult` is a struct kind, so `omitempty` was inert and this is byte-identical on the wire.

This one is more interesting than the `PingResponse` case, and the choice to delete rather than "fix" is load-bearing. At `ftp.go:142` a `FtpResult` is built with only `Command` and `Success` set, and on the error path `result.Data` comes from `GetFtpErrorCode` — so `Data` genuinely can be the zero struct. That means switching to `omitzero` would have **changed the wire format**, dropping `"data":{}` from error responses that the server currently receives.

Deletion was the only safe option here. Good call.

geunwoonoh (dismissed):
## Summary

A pure refactor bringing production code up to the Go 1.25 stdlib baseline; the production counterpart to #369. **I verified all 11 file changes individually — every one is genuinely behavior-preserving.**

### Verification I ran myself

Rather than taking the description at face value, I reproduced the checks. Note that this worktree had no generated ent code, so `go generate ./pkg/db/ent` was needed first.

```
go build ./...   -> exit 0 on darwin / linux / windows
go vet ./...     -> exit 0 on darwin / linux / windows
gofmt -l         -> clean on all 11 changed files
go test -p 1     -> ok on all 6 affected packages
   internal/protocol, pkg/runner, pkg/executor/handlers/firewall,
   pkg/utils, pkg/logsink, cmd/alpamon/command/register
```

`service_windows.go` and `commit_windows.go` are Windows-only, so I added a `GOOS=windows` cross-build — it passes.

**Per-substitution verification:**

- `slices.Backward` — LIFO rollback order preserved
- `maps.Copy(out, auto)` then `maps.Copy(out, user)` — user-wins-on-conflict ordering intact, matching the contract stated in the `register.go:436-438` doc comment
- `for range serviceStopPollAttempts` — confirmed `i` is unused in the loop body
- `strings.SplitSeq` x8 — element sequence identical to `strings.Split`, including the empty-input case. `break` (`DeleteRule`), `continue` (`ReorderChains`), and `return` (`detection.go:204`) all behave correctly under range-over-func
- `strings.CutPrefix` x4 — identical
- `slices.Sort` + `slices.Reverse` — the original was a descending selection sort; the replacement produces the same descending order
- `omitempty` removals — I confirmed in-source that both `time.Time` and `CommandResult` are struct kinds. `encoding/json` does not apply `omitempty` to struct fields, so the wire format is byte-identical. The safety argument in the description is correct.

---

## Critical issues (0)

None.

## Warnings (2)

### 1. `nftables.go` was skipped, leaving the sweep incomplete

`pkg/executor/handlers/firewall/nftables.go:137` and `:390` still contain the exact pattern that was fixed in `iptables.go`:

```go
lines := strings.Split(output, "\n")   // :137
handleRegex := regexp.MustCompile(`# handle (\d+)`)
for _, line := range lines {           // :140 -- the only use of `lines`
```

`grep -n "lines" nftables.go` returns only 137/140 and 390/393, confirming both are range-only. Both loops `continue` on most lines, and `DeleteRule` exits early once it finds a handle, so the lazy-iteration benefit of `SplitSeq` is actually *larger* here than in `iptables.go`.

Modernizing one of two sibling backends in the same directory leaves the next reader unable to tell whether nftables was excluded deliberately. This is a two-line change.

### 2. The changed iptables/utils functions have 0% test coverage

Measured with `go tool cover -func`:

```
iptables.go:119  DeleteRule                 0.0%
iptables.go:213  parseIptablesSaveOutput    0.0%
iptables.go:240  parseIptablesSaveRuleLine  0.0%   <- CutPrefix substitution
iptables.go:358  ReorderChains              0.0%   <- sort algorithm replacement
utils/firewall.go:63  ParseFirewallComment  0.0%   <- CutPrefix substitution
```

Ironically, the untouched `nftables.go` equivalents sit at 45–48%. The one change in this PR that is *not* a mechanical 1:1 swap — the sort replacement in `ReorderChains` — is exactly the one at 0%.

I verified the equivalence by hand (duplicates are structurally impossible: at most one append per line, and line numbers are unique), so it is safe. But it is worth recording that "behavior-preserving" currently rests on review rather than on tests. Adding tests probably belongs to the #369 scope; at minimum, a table test for the `ReorderChains` ordering would be worthwhile.

## Suggestions (4)

### 1. Two remaining `for range N` candidates

Same case as the one applied in `service_windows.go` (I confirmed `i` is unused in both):

```
internal/pool/pool.go:53        for i := 0; i < p.maxWorkers; i++ {
pkg/collector/collector.go:147  for i := 0; i < c.buffer.Capacity; i++ {
```

For reference, these three were **correctly** left alone — no need to re-check them:

- `pkg/scheduler/reporter.go:43` — uses `i` in `NewReporter(i, session)`
- `pkg/signing/verifier.go:73` — lookahead indexing
- `pkg/utils/metrics.go:134` — reverse index used for string slicing (`slices.Backward` does not apply to strings)

### 2. `logsink/writer.go:62-63` could be a single `maps.Clone` (optional)

See the inline comment. Behaviorally identical here, but the current form is defensible — fine either way.

### 3. JSON assembled by string formatting in `commitAndNotify` (pre-existing)

See the inline comment on `pkg/runner/commit.go`. Low priority.

### 4. The `slices.SortFunc` alternative is better left unused

`slices.SortFunc(jumpLines, func(a, b int) int { return b - a })` would collapse the two calls into one, but `Sort` + `Reverse` reads more clearly alongside the "delete from bottom to preserve line numbers" comment. **The current choice is the better one** — noted only to record that the alternative was considered.

---

## Out of scope: a real bug in `calculateBroadcastAddress`

This is in a file the PR touches (`pkg/runner/commit.go:670`) but is **a pre-existing bug unrelated to this PR**. I found it during the performance pass and am reporting it here so it does not get lost.

```go
broadcast := make(net.IP, len(ip.To4()))   // 4
for i := 0; i < len(ip.To4()); i++ {
    broadcast[i] = ip[i] | ^mask[i]        // <- ip[i], not ip.To4()[i]
}
```

If `net.IP` is in the 16-byte IPv4-in-IPv6 form, `ip[0..3]` is not the address — it is `0,0,0,0`. I extracted the function verbatim and measured it on a macOS host:

```
4-byte  IP len=4  -> "192.168.1.255"   (correct)
16-byte IP len=16 -> "0.0.0.255"       (wrong)
HOST en0  ip=10.1.0.28 len(ip)=16 mask=ffffff00 -> broadcast="0.0.0.255"   <- should be 10.1.0.255
HOST lo0  ip=127.0.0.1 len(ip)=16 mask=ff000000 -> broadcast="0.255.255.255"
```

`iface.Addrs()` returns 16-byte IPs on darwin, so **on macOS the broadcast address reported to Alpacon during system-info commit is always wrong.** Linux is unaffected (netlink yields 4-byte IPs). There is also a minor inefficiency: `ip.To4()` is called on every iteration. The fix is to hoist `v4 := ip.To4()` once and index `v4[i]`.

**I recommend NOT folding this into this PR.** The value of this PR is the guarantee that it is behavior-preserving; mixing in a behavior change would destroy that property. A separate issue/PR is the right home.

---

## Performance review

**Improvements:**

- `strings.SplitSeq` x8 — eliminates one `[]string` allocation per call. The substrings themselves are zero-copy in both versions, so the gain is precisely the slice header array (n x 16 bytes). This is material in `parseIptablesSaveOutput` and `ReorderChains`, where `iptables-save` / `iptables -L` output can run to thousands of lines. `DeleteRule` benefits most: it `break`s early, so where it previously split the entire output and used only the first part, it now splits only as far as needed.
- `fmt.Appendf(nil, ...)` — drops 2 allocations and copies (`Sprintf`'s `[]byte -> string`, then `[]byte(...)`'s `string -> []byte`) down to 1.
- `slices.Sort` + `Reverse` — O(n^2) to O(n log n). n here is the number of jump rules per chain (typically under 10), so there is no measurable difference in practice; what matters is that it is not a regression.

**Neutral:** `maps.Copy` (a generic loop, identical to the hand-written one) and `for range N` (identical codegen).

**Micro-regression (immaterial):** `slices.Backward` carries theoretical iterator overhead versus an index loop, but the compiler inlines it and this path runs once per failed registration over 2–3 elements. Not a concern.

**No performance regressions found.**

## Duplicate / unused code review

- All 4 new imports (`maps` x2, `slices` x2) are actually used — no unused imports
- No dangling references to the removed `lines :=` bindings (the `lines` symbol is fully gone from `iptables.go` and `utils/firewall.go`)
- No functions became unreachable
- `[]byte(fmt.Sprintf(`, `HasPrefix` + `TrimPrefix` pairs, and range-only `strings.Split` are now **entirely eliminated from production code except in `nftables.go`** (Warning 1)
- `pkg/scheduler/session.go:149`'s `for k, v := range headers` calls `req.Header.Set` (which normalizes keys), so it is not a `maps.Copy` candidate — correctly excluded

---

## Highlights

- The `omitempty` safety argument in the description is accurate. It gets right that `omitempty` is inert on struct kinds, that the wire format is therefore byte-identical, and it went as far as checking the consumer side (alpacon-server) — exactly the evidence this class of change needs.
- Resisting the temptation to "fix" these with Go 1.24's `omitzero` was the right call. `PingResponse.Timestamp` is always set to `time.Now()` in `NewPingResponse()` and can never be zero, and `FtpResult.Data` **can** be a zero struct on error paths — so `omitzero` would have actually changed the wire format. Deletion was the only safe option.
- The commit granularity and the split between #369 (tests) and #371 (production) make this genuinely easy to review.
- Correctly identifying that the original `ReorderChains` loop was a *descending* sort and substituting `Sort` + `Reverse`. Using `slices.Sort` alone is the common mistake here, and it was avoided.

---

**Verdict: APPROVED** — nothing blocking. The two warnings are worth follow-up.

**Checklist**

- [x] Go idioms — all 1.25 stdlib substitutions correct
- [x] build / vet / gofmt — verified directly on darwin, linux, and windows
- [x] Tests pass — ran the 6 affected packages directly
- [ ] Test coverage of the changed logic — 0% on the changed iptables/utils functions (Warning 2)
- [x] No security issues — no hardcoded secrets or injection paths; pure substitution
- [x] Wire format unchanged — verified that `omitempty` is inert on struct kinds
- [ ] Sweep completeness — `nftables.go` missed (Warning 1)

**Suggested follow-ups (separate PRs)**

1. `SplitSeq` substitution in `nftables.go:137` and `:390` — fine to include in this PR too
2. `for range N` in `pool.go:53` and `collector.go:147`
3. The `calculateBroadcastAddress` darwin bug — **must be a separate PR** (behavior change)

jisung-02:
Thanks for the thorough review. Pushed 496b4fe addressing Warning 1 and Suggestion 1 — the two follow-ups you flagged as fine to include here.

- nftables.go:137 and :390 — both range-only strings.Split loops now use strings.SplitSeq, matching the iptables.go change and completing the sweep across both firewall backends.
- pool.go:53 and collector.go:147 — both index-only loops now use for range N, matching service_windows.go; the index is unused in both bodies.

Verified on the affected packages: go build ./..., go vet, and gofmt -l are all clean, and tests pass (internal/pool, pkg/executor/handlers/firewall; pkg/collector has no test files).

Deliberately left out, per your guidance:
- Warning 2 (0% coverage on the changed iptables/utils functions) — belongs to the #369 test scope.
- Suggestion 2 (maps.Clone in writer.go) — keeping the explicit non-nil guarantee.
- Suggestion 3 (JSON string formatting in commit.go) — pre-existing, ldflags-controlled value.
- Suggestion 4 (slices.SortFunc) — agreed, Sort + Reverse reads better next to the comment.
- The calculateBroadcastAddress darwin bug — will file a separate PR so this one stays strictly behavior-preserving.

geunwoonoh (approved):

PR #374, opened 2026-07-23
feat(updater): gate Windows self-update on a verified SCM restart path

jisung-02:
## Background

Windows self-update replaces the running `alpamon.exe` via a rename-aside sequence and then exits the process. In service mode, SCM Recovery Actions must relaunch the process right after it exits for the new binary to start. If no prompt restart is configured (an operator disabled it, or the service predates `register`), the replaced binary stays dead and a server reachable only over its outbound WebSocket becomes unreachable. Starting a self-update without a guaranteed restart is itself the hazard.

The worker pool can also dispatch two upgrade commands concurrently. Self-update uses fixed staging paths (`.new`/`.old`), so concurrent runs can delete each other's files mid-replace and kill the agent.

## Changes

Add a preflight guard `ensureSelfRestartable()` that runs before self-update touches anything (`pkg/updater/guard_windows.go`).

- Interactive (non-service) mode returns immediately: `restartAgent` re-execs itself via `exec.Command`.
- Service mode queries SCM Recovery Actions and checks whether the first action is a prompt restart. SCM applies element `[N-1]` on the Nth failure, so relaunching after the self-update exit (failure 1 in steady state) requires `[0]` to be a restart. A restart scheduled beyond `maxAcceptableRestartDelay` (60s) is not a guarantee and is treated like a missing restart, so the server cannot be left unreachable for minutes or hours after the exit.
- If the first action is not a prompt restart, it restores the defaults `register` uses (`Restart 5s` / `Restart 5s` / `NoAction`, reset 60s), logs the replaced actions so an overwritten operator config can be reconstructed, re-queries to confirm the change took effect, and only then proceeds. Any failure to restore or re-query aborts the self-update (fail-closed).

The guard opens the SCM with least privilege: `SC_MANAGER_CONNECT` + `SERVICE_QUERY_CONFIG` for the read path, escalating to a dedicated `SERVICE_CHANGE_CONFIG | SERVICE_START` handle only when a heal is actually needed. The `svc/mgr` wrapper hardcodes `*_ALL_ACCESS`, so the guard uses the raw `windows.OpenSCManager`/`OpenService` and wraps the handle in `mgr.Service` to reuse `RecoveryActions()`/`SetRecoveryActions()`. This keeps a hardened deployment from being permanently blocked on the common read-only path.

Unix is a no-op (`pkg/updater/guard_unix.go`): `restartAgent` execs in place via `syscall.Exec`.

Concurrent runs are serialized with a process-local `atomic.Bool` CAS (`pkg/updater/updater.go`), exposed as `ErrSelfUpdateInProgress` so a caller can treat a rejected duplicate as a benign no-op rather than a failure. The latch is held through a successful run (the pending restart resets it) and released only on failure, closing the window where a second upgrade could re-run mid-restart; the system handler releases it explicitly if it cannot schedule the restart.

The shared Windows service identity and recovery defaults (service name, reset period, restart delay, default recovery actions) now live in a single `pkg/svcdef` package that `register`, the SCM handler, and the guard all import, replacing the three duplicated copies and their sync comments.

## Safety

- Fail-closed: if a prompt restart is not guaranteed, the guard aborts before touching the binary.
- Least privilege: the read path never requires write access, so tightening the service DACL cannot block a self-update whose recovery config is already correct.
- The self-heal restores the exact values `register` writes at install time, not a new policy. It does overwrite a deliberately-disabled setting (a tradeoff over letting the server die unrelaunched) and logs the replaced actions so they can be reconstructed.
- Preflight limits: it cannot cover a config change after the query, or a failure streak that has advanced SCM to a later action — permanent only when the reset period is INFINITE, which `mgr.RecoveryActions()` cannot report. Documented in the `ensureSelfRestartable` comment.
- Concurrent self-updates are serialized via the process-local CAS.

## Testing

- `guard_windows_test.go`: `firstActionRestarts` including delay-bound cases (register's 5s, exactly 60s, past the bound, 6h), `describeActions`, and `TestEnsureRecoveryRestart` through a `recoveryConfigurer` seam, now asserting the healed values against `svcdef.DefaultRecoveryActions()` and the reset period in every heal case. These run in the Windows CI job.
- `updater_test.go`: a rejected concurrent run returns `ErrSelfUpdateInProgress`, a failed run clears the latch, and `ReleaseSelfUpdateLatch` clears it. `system_test.go`: an in-progress run is a benign no-op (exit 0), and a failed restart schedule releases the latch.
- `go test ./... -p 1` passes; gofmt/lint clean; `GOOS=windows` build/vet/test-compile clean.
- Real-server E2E (Windows Server 2022, LocalSystem): installed a dev build, set the first recovery action to a 6h-delay restart (`sc.exe failure alpamon reset= 60 actions= restart/21600000`), then triggered a console upgrade. The guard rejected the 6h delay as non-prompt and self-healed the `register` defaults via the `SERVICE_CHANGE_CONFIG | SERVICE_START` write handle; the binary was replaced (dev to 2.3.2) and SCM restarted the process, which reconnected. `sc.exe qfailure alpamon` then showed `RESET_PERIOD 60` / `RESTART 5000ms` x2 restored. The least-privilege read path and the write heal are both confirmed on real hardware; a hardened non-LocalSystem logon account is not yet covered.

## Checklist

- [x] Fail-closed guard before the binary is touched
- [x] Least-privilege SCM access, escalating only to heal
- [x] Restart-delay upper bound enforced
- [x] Concurrency serialized (process-local)
- [x] Windows CI unit tests + real-server E2E

### DIFF
--- a/pkg/updater/guard_windows.go
+++ b/pkg/updater/guard_windows.go
@@ -2,21 +2,11 @@ package updater
 
 import (
 	"fmt"
-	"time"
 
+	"github.com/alpacax/alpamon/v2/pkg/svcdef"
 	"github.com/rs/zerolog/log"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
-)
-
-// scmServiceName duplicates register's serviceName and svc_windows.go's svcName;
-// pkg can't import cmd without inverting layering.
-const scmServiceName = "alpamon"
-
-// Recovery defaults from 'alpamon register', copied for the same layering reason.
-const (
-	recoveryResetSeconds = 60
-	recoveryRestartDelay = 5 * time.Second
 )
 
 // recoveryConfigurer is the subset of *mgr.Service ensureRecoveryRestart needs,
@@ -45,9 +35,9 @@ func ensureSelfRestartable() error {
 	}
 	defer func() { _ = m.Disconnect() }()
 
-	s, err := m.OpenService(scmServiceName)
+	s, err := m.OpenService(svcdef.ServiceName)
 	if err != nil {
-		return abortf("failed to open service %q: %w", scmServiceName, err)
+		return abortf("failed to open service %q: %w", svcdef.ServiceName, err)
 	}
 	defer func() { _ = s.Close() }()
 
@@ -68,12 +58,8 @@ func ensureRecoveryRestart(rc recoveryConfigurer) error {
 
 	// Not a new policy: register already applies exactly this configuration.
 	log.Warn().Msg("SCM recovery actions missing; restoring the defaults set by 'alpamon register'.")
-	defaults := []mgr.RecoveryAction{
-		{Type: mgr.ServiceRestart, Delay: recoveryRestartDelay},
-		{Type: mgr.ServiceRestart, Delay: recoveryRestartDelay},
-		{Type: mgr.NoAction},
-	}
-	if err := rc.SetRecoveryActions(defaults, recoveryResetSeconds); err != nil {
+	defaults := svcdef.DefaultRecoveryActions()
+	if err := rc.SetRecoveryActions(defaults, svcdef.RecoveryResetSeconds); err != nil {
 		return abortf("automatic restart is not configured and restoring it failed: %w; run 'alpamon register' to configure it", err)
 	}
 
--- a/pkg/updater/guard_windows.go
+++ b/pkg/updater/guard_windows.go
@@ -61,8 +61,10 @@ func ensureRecoveryRestart(rc recoveryConfigurer) error {
 		return nil
 	}
 
-	// Not a new policy: register already applies exactly this configuration.
-	log.Warn().Msg("SCM recovery actions missing; restoring the defaults set by 'alpamon register'.")
+	// register already applies exactly this config, but SetRecoveryActions replaces
+	// the whole array—record what's lost, since this log is the only way back.
+	log.Warn().Strs("replaced", describeActions(actions)).
+		Msg("First SCM recovery action is not a restart; restoring the defaults set by 'alpamon register'.")
 	defaults := svcdef.DefaultRecoveryActions()
 	if err := rc.SetRecoveryActions(defaults, svcdef.RecoveryResetSeconds); err != nil {
 		return abortf("automatic restart is not configured and restoring it failed: %w; run 'alpamon register' to configure it", err)
@@ -82,6 +84,29 @@ func ensureRecoveryRestart(rc recoveryConfigurer) error {
 // abortf formats a fail-closed reason under the shared "self-update aborted" prefix.
 func abortf(format string, args ...any) error {
 	return fmt.Errorf("self-update aborted: "+format, args...)
+}
+
+// describeActions renders recovery actions for the overwrite log; mgr encodes
+// RecoveryAction.Type as a bare int that means nothing to a reader.
+func describeActions(actions []mgr.RecoveryAction) []string {
+	out := make([]string, len(actions))
+	for i, a := range actions {
+		var name string
+		switch a.Type {
+		case mgr.NoAction:
+			name = "none"
+		case mgr.ComputerReboot:
+			name = "reboot"
+		case mgr.ServiceRestart:
+			name = "restart"
+		case mgr.RunCommand:
+			name = "run-command"
+		default:
+			name = fmt.Sprintf("unknown(%d)", a.Type)
+		}
+		out[i] = fmt.Sprintf("%s after %s", name, a.Delay)
+	}
+	return out
 }
 
 // firstActionRestarts reports whether the first recovery action is a
--- a/pkg/updater/guard_windows.go
+++ b/pkg/updater/guard_windows.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/alpacax/alpamon/v2/pkg/svcdef"
 	"github.com/rs/zerolog/log"
+	"golang.org/x/sys/windows"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
 )
@@ -14,8 +15,9 @@ import (
 // the self-update exit; a slower restart is healed to register's defaults, not trusted.
 const maxAcceptableRestartDelay = 60 * time.Second
 
-// recoveryConfigurer is the subset of *mgr.Service ensureRecoveryRestart needs,
-// an interface so the self-heal path is testable without a live SCM.
+// recoveryConfigurer is the RecoveryActions/SetRecoveryActions subset that
+// ensureRecoveryRestart needs—implemented by scmRecovery in production and a
+// fake in tests, so the self-heal path runs without a live SCM.
 type recoveryConfigurer interface {
 	RecoveryActions() ([]mgr.RecoveryAction, error)
 	SetRecoveryActions(actions []mgr.RecoveryAction, resetPeriod uint32) error
@@ -39,19 +41,69 @@ func ensureSelfRestartable() error {
 		return nil
 	}
 
-	m, err := mgr.Connect()
+	// Open with least access; mgr.Connect/OpenService would demand *_ALL_ACCESS.
+	// scmRecovery escalates to write access only when a heal is required.
+	scm, err := windows.OpenSCManager(nil, nil, windows.SC_MANAGER_CONNECT)
 	if err != nil {
 		return abortf("failed to connect to service manager: %w", err)
 	}
-	defer func() { _ = m.Disconnect() }()
+	defer func() { _ = windows.CloseServiceHandle(scm) }()
 
-	s, err := m.OpenService(svcdef.ServiceName)
-	if err != nil {
+	rc := &scmRecovery{scm: scm, name: svcdef.ServiceName}
+	if err := rc.openQuery(); err != nil {
 		return abortf("failed to open service %q: %w", svcdef.ServiceName, err)
 	}
-	defer func() { _ = s.Close() }()
+	defer rc.close()
 
-	return ensureRecoveryRestart(s)
+	return ensureRecoveryRestart(rc)
+}
+
+// scmRecovery opens each SCM handle with the least access the call needs, so the
+// read path never demands SERVICE_CHANGE_CONFIG. It satisfies recoveryConfigurer.
+type scmRecovery struct {
+	scm   windows.Handle
+	name  string
+	query *mgr.Service // opened with SERVICE_QUERY_CONFIG
+}
+
+func (r *scmRecovery) openQuery() error {
+	h, err := r.openService(windows.SERVICE_QUERY_CONFIG)
+	if err != nil {
+		return err
+	}
+	r.query = &mgr.Service{Name: r.name, Handle: h}
+	return nil
+}
+
+func (r *scmRecovery) openService(access uint32) (windows.Handle, error) {
+	namePtr, err := windows.UTF16PtrFromString(r.name)
+	if err != nil {
+		return 0, err
+	}
+	return windows.OpenService(r.scm, namePtr, access)
+}
+
+func (r *scmRecovery) close() {
+	if r.query != nil {
+		_ = r.query.Close()
+	}
+}
+
+func (r *scmRecovery) RecoveryActions() ([]mgr.RecoveryAction, error) {
+	return r.query.RecoveryActions()
+}
+
+// SetRecoveryActions opens a dedicated write handle so the read path stays
+// query-only. SERVICE_START is required alongside SERVICE_CHANGE_CONFIG because
+// the restored policy sets an SC_ACTION_RESTART action.
+func (r *scmRecovery) SetRecoveryActions(actions []mgr.RecoveryAction, resetPeriod uint32) error {
+	h, err := r.openService(windows.SERVICE_CHANGE_CONFIG | windows.SERVICE_START)
+	if err != nil {
+		return err
+	}
+	s := &mgr.Service{Name: r.name, Handle: h}
+	defer func() { _ = s.Close() }()
+	return s.SetRecoveryActions(actions, resetPeriod)
 }
 
 // ensureRecoveryRestart passes when rc's first recovery action is a prompt
--- a/pkg/updater/guard_windows.go
+++ b/pkg/updater/guard_windows.go
@@ -9,31 +9,33 @@ import (
 	"golang.org/x/sys/windows/svc/mgr"
 )
 
-// scmServiceName must match serviceName in cmd/alpamon/command/register;
-// copied because importing a cmd package from pkg would invert layering.
+// scmServiceName mirrors serviceName in cmd/alpamon/command/register; copied
+// because importing cmd from pkg would invert layering.
 const scmServiceName = "alpamon"
 
-// Recovery defaults applied by 'alpamon register'; copies for the same
-// layering reason as scmServiceName.
+// Recovery defaults from 'alpamon register', copied for the same layering reason.
 const (
 	recoveryResetSeconds = 60
 	recoveryRestartDelay = 5 * time.Second
 )
 
-// ensureSelfRestartable verifies something will relaunch alpamon after the
-// self-update exits: under SCM the first recovery action must be a restart,
-// or the replaced binary dies unrelaunched and a remote-only server goes
-// unreachable. A missing restart action is self-healed with register's
-// defaults, trusting only a confirming re-query; any failure aborts
-// (fail-closed). Preflight only: a later config change or a recent failure
-// streak (SCM steps to later actions) can still leave the service stopped.
+// recoveryConfigurer is the subset of *mgr.Service ensureRecoveryRestart needs,
+// an interface so the self-heal path is testable without a live SCM.
+type recoveryConfigurer interface {
+	RecoveryActions() ([]mgr.RecoveryAction, error)
+	SetRecoveryActions(actions []mgr.RecoveryAction, resetPeriod uint32) error
+}
+
+// ensureSelfRestartable aborts the self-update unless a relaunch is guaranteed
+// afterward—otherwise the replaced binary stays dead and a remote-only server
+// goes unreachable. Preflight only: config drift or a failure streak past the
+// first action can still leave the service stopped.
 func ensureSelfRestartable() error {
 	isSvc, err := svc.IsWindowsService()
 	if err != nil {
 		return abortf("failed to determine service mode: %w", err)
 	}
 	if !isSvc {
-		// Interactive mode restarts itself via exec.Command (restartAgent).
 		return nil
 	}
 
@@ -50,13 +52,6 @@ func ensureSelfRestartable() error {
 	defer func() { _ = s.Close() }()
 
 	return ensureRecoveryRestart(s)
-}
-
-// recoveryConfigurer is the subset of *mgr.Service that ensureRecoveryRestart
-// needs; an interface so the self-heal path is unit-testable without a live SCM.
-type recoveryConfigurer interface {
-	RecoveryActions() ([]mgr.RecoveryAction, error)
-	SetRecoveryActions(actions []mgr.RecoveryAction, resetPeriod uint32) error
 }
 
 // ensureRecoveryRestart passes when rc's first recovery action is a restart,
--- a/pkg/updater/guard_windows.go
+++ b/pkg/updater/guard_windows.go
@@ -23,6 +23,9 @@ type recoveryConfigurer interface {
 func ensureSelfRestartable() error {
 	isSvc, err := svc.IsWindowsService()
 	if err != nil {
+		// Deliberately stricter than runningAsWindowsService(), which assumes
+		// interactive on this error: guessing interactive on a real service would
+		// skip the restart check and leave the replaced binary with nothing to relaunch it.
 		return abortf("failed to determine service mode: %w", err)
 	}
 	if !isSvc {
--- a/pkg/updater/guard_windows.go
+++ b/pkg/updater/guard_windows.go
@@ -2,12 +2,17 @@ package updater
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/alpacax/alpamon/v2/pkg/svcdef"
 	"github.com/rs/zerolog/log"
 	"golang.org/x/sys/windows/svc"
 	"golang.org/x/sys/windows/svc/mgr"
 )
+
+// maxAcceptableRestartDelay bounds how long the server may stay unreachable after
+// the self-update exit; a slower restart is healed to register's defaults, not trusted.
+const maxAcceptableRestartDelay = 60 * time.Second
 
 // recoveryConfigurer is the subset of *mgr.Service ensureRecoveryRestart needs,
 // an interface so the self-heal path is testable without a live SCM.
@@ -49,8 +54,8 @@ func ensureSelfRestartable() error {
 	return ensureRecoveryRestart(s)
 }
 
-// ensureRecoveryRestart passes when rc's first recovery action is a restart,
-// otherwise self-heals register's defaults and trusts only a confirming
+// ensureRecoveryRestart passes when rc's first recovery action is a prompt
+// restart, otherwise self-heals register's defaults and trusts only a confirming
 // re-query. Any SCM error aborts (fail-closed).
 func ensureRecoveryRestart(rc recoveryConfigurer) error {
 	actions, err := rc.RecoveryActions()
@@ -64,10 +69,10 @@ func ensureRecoveryRestart(rc recoveryConfigurer) error {
 	// register already applies exactly this config, but SetRecoveryActions replaces
 	// the whole array—record what's lost, since this log is the only way back.
 	log.Warn().Strs("replaced", describeActions(actions)).
-		Msg("First SCM recovery action is not a restart; restoring the defaults set by 'alpamon register'.")
+		Msg("First SCM recovery action is not a prompt restart; restoring the defaults set by 'alpamon register'.")
 	defaults := svcdef.DefaultRecoveryActions()
 	if err := rc.SetRecoveryActions(defaults, svcdef.RecoveryResetSeconds); err != nil {
-		return abortf("automatic restart is not configured and restoring it failed: %w; run 'alpamon register' to configure it", err)
+		return abortf("a prompt automatic restart is not configured and restoring it failed: %w; run 'alpamon register' to configure it", err)
 	}
 
 	actions, err = rc.RecoveryActions()
@@ -75,7 +80,7 @@ func ensureRecoveryRestart(rc recoveryConfigurer) error {
 		return abortf("failed to verify restored recovery actions: %w", err)
 	}
 	if !firstActionRestarts(actions) {
-		return abortf("automatic restart is still not configured after restoring defaults; run 'alpamon register' to configure it")
+		return abortf("a prompt automatic restart is still not configured after restoring defaults; run 'alpamon register' to configure it")
 	}
 	log.Info().Msg("SCM recovery actions restored and verified.")
 	return nil
@@ -109,9 +114,11 @@ func describeActions(actions []mgr.RecoveryAction) []string {
 	return out
 }
 
-// firstActionRestarts reports whether the first recovery action is a
-// restart—SCM applies element [N-1] on the Nth failure, and the self-update
-// exit is failure 1 in steady state.
+// firstActionRestarts reports whether the first recovery action restarts the
+// service soon enough to count as a guarantee—SCM applies element [N-1] on the
+// Nth failure, and the self-update exit is failure 1 in steady state.
 func firstActionRestarts(actions []mgr.RecoveryAction) bool {
-	return len(actions) > 0 && actions[0].Type == mgr.ServiceRestart
+	return len(actions) > 0 &&
+		actions[0].Type == mgr.ServiceRestart &&
+		actions[0].Delay <= maxAcceptableRestartDelay
 }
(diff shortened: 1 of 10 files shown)


pkg/updater/guard_windows_test.go
```diff
@@ -0,0 +1,30 @@
+package updater
+
+import (
+	"testing"
+
+	"golang.org/x/sys/windows/svc/mgr"
+)
```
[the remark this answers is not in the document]
jisung-02: Not applicable here: `guard_windows_test.go` is already constrained to GOOS=windows by Go's implicit filename convention. The `_windows` suffix (before `_test`) scopes the file to Windows, exactly like the sibling `format_windows_test.go`, `replace_windows.go`, and `cleanup_windows.go`, none of which carry an explicit build tag.

Verified: on darwin, `go list -test ./pkg/updater` excludes this file (only `updater_test.go` compiles) and `go vet ./pkg/updater` passes; under `GOOS=windows` the file is included and the Windows CI job (`build-and-test-windows`) runs it. Adding an explicit `//go:build windows` tag would be inconsistent with the package's existing `_windows` files, so no change is made.

pkg/updater/guard_windows.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil
+	}
+
+	m, err := mgr.Connect()
+	if err != nil {
+		return abortf("failed to connect to service manager: %w", err)
+	}
+	defer func() { _ = m.Disconnect() }()
+
+	s, err := m.OpenService(scmServiceName)
```
geunwoonoh: **Warning:** `mgr.Connect()` requests `SC_MANAGER_ALL_ACCESS` and `m.OpenService()` requests `SERVICE_ALL_ACCESS` (verified in x/sys v0.45.0, `mgr.go:43` and `mgr.go:195`). Combined with the fail-closed abort, a deployment where the service runs under a hardened non-LocalSystem logon account gets **self-update permanently blocked**: `OpenService` fails with ACCESS_DENIED, the guard aborts, and an upgrade path that worked before this PR never runs again.

The default install is fine — `register` never sets `ServiceStartName`, so the service is LocalSystem. But the query path only needs `SERVICE_QUERY_CONFIG`. Consider opening with least privilege and escalating only when healing is actually required:

```go
h, err := windows.OpenSCManager(nil, nil, windows.SC_MANAGER_CONNECT)
sh, err := windows.OpenService(h, namePtr, windows.SERVICE_QUERY_CONFIG)
```

pkg/updater/guard_windows.go
```diff
@@ -0,0 +1,100 @@
...
+
+func abortf(format string, args ...any) error {
+	return fmt.Errorf("self-update aborted: "+format, args...)
+}
+
+// firstActionRestarts reports whether the first recovery action is a
+// restart—SCM applies element [N-1] on the Nth failure, and the self-update
+// exit is failure 1 in steady state.
+func firstActionRestarts(actions []mgr.RecoveryAction) bool {
+	return len(actions) > 0 && actions[0].Type == mgr.ServiceRestart
```
geunwoonoh: **Warning:** the check ignores `Delay`. `[{Type: ServiceRestart, Delay: 6 * time.Hour}]` passes the gate, and the server then stays unreachable for six hours — exactly the outcome this guard exists to prevent. If the contract is "a relaunch is guaranteed", the delay needs an upper bound.

Requires a companion constant, e.g.:

```go
// maxAcceptableRestartDelay bounds how long the server may stay unreachable
// after the self-update exit. A slower restart is not a guarantee.
const maxAcceptableRestartDelay = 60 * time.Second
```

```suggestion
	return len(actions) > 0 &&
		actions[0].Type == mgr.ServiceRestart &&
		actions[0].Delay <= maxAcceptableRestartDelay
```

pkg/updater/guard_windows.go
```diff
@@ -0,0 +1,100 @@
...
+	actions, err := rc.RecoveryActions()
+	if err != nil {
+		return abortf("failed to query recovery actions: %w", err)
+	}
+	if firstActionRestarts(actions) {
+		return nil
+	}
+
+	// Not a new policy: register already applies exactly this configuration.
+	log.Warn().Msg("SCM recovery actions missing; restoring the defaults set by 'alpamon register'.")
```
geunwoonoh: **Warning:** two issues on this line.

"missing" is only accurate when there are no actions at all. A configuration like `[RunCommand(alerting script), Restart, Restart]` also has a non-restart first action, so it lands in this branch with a message that misdescribes the state — and `SetRecoveryActions` replaces the whole array, so that `RunCommand` disappears. The PR description acknowledges the overwrite as a deliberate tradeoff, but nothing is logged about *what* was overwritten, so it can't be reconstructed afterwards.

```suggestion
	log.Warn().Interface("previous", actions).
		Msg("First SCM recovery action is not a restart; restoring the defaults set by 'alpamon register'.")
```

pkg/updater/guard_windows_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	if i < len(f.queries) {
+		return f.queries[i], nil
+	}
+	return nil, nil
+}
+
+func (f *fakeRecovery) SetRecoveryActions(actions []mgr.RecoveryAction, resetPeriod uint32) error {
+	f.setCalled = true
+	f.setReset = resetPeriod
```
geunwoonoh, line 97: **Warning:** the fake discards the `actions` argument, so no test asserts the values the guard actually writes. The re-query result is supplied independently via `queries`, so if `defaults` in `ensureRecoveryRestart` regressed to `[{Type: mgr.NoAction}, ...]` or lost its `Delay`, **every test here would still pass**. The guard's primary output is untested.

Add a `setActions []mgr.RecoveryAction` field to `fakeRecovery`, then assert in the heal cases that `setActions[0].Type == mgr.ServiceRestart` and `setActions[0].Delay == recoveryRestartDelay`.

```suggestion
	f.setActions = actions
	f.setReset = resetPeriod
```

pkg/updater/guard_windows_test.go
```diff
@@ -0,0 +1,126 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			f := tt.fake
+			err := ensureRecoveryRestart(&f)
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("ensureRecoveryRestart() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if f.setCalled != tt.wantSet {
+				t.Errorf("SetRecoveryActions called = %v, want %v", f.setCalled, tt.wantSet)
+			}
+			if tt.wantReset != 0 && f.setReset != tt.wantReset {
```
geunwoonoh: **Suggestion:** the `tt.wantReset != 0` guard means the reset period is only asserted in 1 of the 4 cases where healing occurs — a zero `wantReset` is indistinguishable from "don't check this". Use `wantReset *uint32`, or add a separate `checkReset bool`, so 0 stays a valid expected value.

pkg/updater/guard_windows.go
```diff
@@ -0,0 +1,100 @@
...
+	"time"
+
+	"github.com/rs/zerolog/log"
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/mgr"
+)
+
+// scmServiceName duplicates register's serviceName and svc_windows.go's svcName;
+// pkg can't import cmd without inverting layering.
+const scmServiceName = "alpamon"
```
geunwoonoh: **Suggestion:** the PR description justifies the duplication with "`pkg` can't import `cmd` without inverting layering", which is true — but the fix runs the other way. `cmd -> pkg` is the correct direction, so putting the constant in `pkg` and importing it from `cmd` removes the duplication entirely instead of documenting it.

The service name now lives in three places:

- `cmd/alpamon/command/svc_windows.go:12` — `svcName`
- `cmd/alpamon/command/register/service_windows.go:21` — `serviceName`
- `pkg/updater/guard_windows.go:14` — `scmServiceName` (added here)

plus `recoveryResetSeconds`, `recoveryRestartDelay`, and the three-element `[]mgr.RecoveryAction` literal duplicated between `register` and this file. A small `pkg` package collapses all of it:

```go
package svcdef

const ServiceName = "alpamon"
const RecoveryResetSeconds = 60
const RecoveryRestartDelay = 5 * time.Second

// DefaultRecoveryActions is what register writes at install time and what the
// self-update guard restores.
func DefaultRecoveryActions() []mgr.RecoveryAction { ... }
```

That also collapses the test gap flagged in `guard_windows_test.go` into a single assertion (`setActions` vs `svcdef.DefaultRecoveryActions()`), and retires the "keep these in sync" comments rather than adding a third one.

geunwoonoh (commented):
## PR Review: #374 - feat(updater): gate Windows self-update on a verified SCM restart path

**Stack**: golang
**Changes**: +277 -3 in 6 files

**Local verification** (against `f16aeed`):

```
GOOS=windows go build ./...             -> BUILD_WINDOWS_OK
GOOS=windows go vet ./pkg/updater/...   -> VET_WINDOWS_OK
GOOS=windows go test -c ./pkg/updater/  -> WINDOWS_TEST_COMPILE_OK
gofmt -l pkg/updater cmd/.../register   -> GOFMT_CLEAN
go test ./pkg/updater/...               -> ok (0.463s)
```

---

### Summary

Adds a preflight guard (`ensureSelfRestartable`) that inspects SCM Recovery Actions before Windows self-update touches the binary, restores `register`'s defaults when the first action is not a restart, and confirms the change with a re-query. Anything it cannot confirm aborts the update (fail-closed). Concurrent `SelfUpdate` runs are serialized with an `atomic.Bool` CAS.

The design holds up: verify before replacing, fail closed, and isolate the SCM dependency behind a testable seam. Placing the guard *before* the download is also the correct ordering.

---

### Critical issues (0)

Nothing that breaks the happy path.

---

### Warnings (4)

**1. `pkg/updater/guard_windows.go:42,48` — over-broad SCM access rights, combined with a fail-closed abort, can block upgrades permanently**

`mgr.Connect()` requests `SC_MANAGER_ALL_ACCESS` and `m.OpenService()` requests `SERVICE_ALL_ACCESS` (verified in x/sys v0.45.0, `mgr.go:43` and `mgr.go:195`). The default install is unaffected — `register` never sets `ServiceStartName`, so the service runs as LocalSystem. But on a deployment where an operator hardened the service logon account, `OpenService` fails with ACCESS_DENIED, the guard aborts, and **self-update that worked before this PR is now permanently blocked**. The query path only needs `SERVICE_QUERY_CONFIG`.

**2. `pkg/updater/guard_windows.go:98` — `firstActionRestarts` ignores `Delay`**

`[{Type: ServiceRestart, Delay: 6 * time.Hour}]` passes the gate, and the server stays unreachable for six hours — exactly the outcome the guard exists to prevent. If the contract is "a relaunch is guaranteed", the delay needs an upper bound.

**3. `pkg/updater/guard_windows.go:70` — inaccurate log message, and an operator's custom action is overwritten silently**

`"SCM recovery actions missing"` is only accurate when there are no actions at all. A configuration like `[RunCommand(alerting script), Restart, Restart]` also lands in this branch, and `SetRecoveryActions` replaces the whole array, so that `RunCommand` disappears. The PR description acknowledges the overwrite as a deliberate tradeoff, but nothing is logged about *what* was overwritten, so it cannot be reconstructed afterwards.

**4. `pkg/updater/guard_windows_test.go:58-62` — the fake discards `actions`, leaving the healed values untested**

`fakeRecovery.SetRecoveryActions` records only `setCalled`/`setReset` and throws away `actions`. The re-query result is supplied independently via `queries`, so if `defaults` regressed to `[{Type: mgr.NoAction}, ...]` or lost its `Delay`, **every test would still pass**. The guard's primary output is a blind spot.

Related: the `tt.wantReset != 0` guard at `guard_windows_test.go:121` means the reset period is asserted in only 1 of the 4 cases where healing occurs.

---

### Suggestions (7)

**1. The duplicated constants can actually be removed** — the PR justifies duplication with "`pkg` can't import `cmd` without inverting layering", which is true, but the fix runs the other way: `cmd -> pkg` is the correct direction. The service name now lives in three places (`svc_windows.go:12`, `register/service_windows.go:21`, `guard_windows.go:14`), plus `recoveryResetSeconds`, `recoveryRestartDelay`, and the three-element `[]mgr.RecoveryAction` literal duplicated between `register` and the guard. A small `pkg` package collapses all of it, and retires the "keep these in sync" comments rather than adding a third one.

**2. `pkg/updater/updater.go:61` — expose a sentinel error.** Callers can't distinguish contention from a real failure; `system.go:201` renders it as `"Self-update failed: ..."` with exit code 1, even though rejecting a duplicate request is correct behavior.

**3. `pkg/updater/updater.go:63` — the flag is released on the success path too.** The actual restart happens ~1s later (`system.go:204`); a second upgrade command in that window re-runs the whole update and, on Windows, fails in `replaceBinary` with a misleading `"failed to move running binary aside"` after wasting a 100 MB download.

**4. `pkg/updater/guard_windows.go:34-37` — service-mode detection diverges from `runningAsWindowsService()`** (`svc_windows.go:60-67`), which treats the same error as "assume interactive". The stricter choice here is defensible, but the asymmetry deserves a comment so it doesn't read as a bug.

**5. `pkg/updater/guard_windows.go:29-32` — the documented limitation becomes permanent when `resetPeriod` is `INFINITE`.** The failure count never resets, so after two self-updates the third lands on `NoAction` and the service stays dead for good.

**6. `pkg/updater/updater.go:38-39` — note that the CAS is process-local.** It covers the worker-pool dispatch this PR targets; a second alpamon process on the same host still races on `.new`/`.old`.

**7. `pkg/updater/guard_windows.go:91` — `abortf` is the only function in the file without a doc comment.** Nit.

---

### Highlights

- **The guard runs before the download** (`updater.go:69-72`) — no wasted 100 MB fetch and checksum verification on a run that will abort. The ordering is right.
- **`recoveryConfigurer`** extracts the minimal subset of `*mgr.Service` needed, so the entire self-heal path is testable without a live SCM. That matches "mock interfaces, not implementations", and defining the interface on the consumer side is idiomatic Go.
- **Success is only granted by a confirming re-query** (`guard_windows.go:80-86`) — a nil return from `SetRecoveryActions` isn't trusted on its own. The fail-closed posture is applied consistently.
- **Build constraints are handled correctly.** `guard_unix.go` carries an explicit `//go:build !windows` because `_unix` is not a GOOS suffix, while `guard_windows.go` / `guard_windows_test.go` rely on the filename convention — consistent with `cleanup_*`, `replace_*`, and `format_windows_test.go`. The rebuttal to Copilot's comment on this is correct.
- **Table tests copy the fixture** (`f := tt.fake`), so cases can't leak state into each other.
- **Error strings follow Go convention**: lowercase, no trailing punctuation, `%w` wrapping throughout.

**Performance**: no regression. The guard adds one SCM connect/query per self-update, and self-update is a rare event; the `atomic.Bool` CAS is negligible.

**Dead / duplicated code**: none unused. The `SelfUpdateFunc` move (`updater.go:42-43`) is a pure reordering; it is genuinely used at `system.go:40,69`.

---

**Verdict**: COMMENTED

No critical issues and nothing that should block the merge. That said, warnings 1 (access rights) and 2 (delay) are real holes in the "guaranteed restart" contract the guard advertises, and warning 4 means future regressions in the healed values won't be caught. I'd address those three before merging. Suggestion 1 is worth revisiting since the PR description treats the duplication as unavoidable.

**Checklist**:
- [x] Code follows stack conventions — gofmt/vet clean; error wrapping, naming, and build tags all idiomatic
- [x] Tests included and passing — but the healed values are unasserted (warning 4)
- [x] No security vulnerabilities — none introduced; the SCM handle is more privileged than needed (warning 1)
- [x] Documentation updated if needed — PR description and code comments cover the limitations

pkg/updater/updater.go
```diff
@@ -45,16 +54,23 @@ func (o Options) baseURL() string {
...
 }
 
-// SelfUpdateFunc is the signature of SelfUpdate, held as a field so tests can inject a fake.
-type SelfUpdateFunc func(ctx context.Context, latestVersion string, opts Options) error
-
 // SelfUpdate downloads the latest binary from GitHub Releases,
 // verifies its checksum, and replaces the current binary.
 func SelfUpdate(ctx context.Context, latestVersion string, opts Options) error {
+	if !selfUpdateInFlight.CompareAndSwap(false, true) {
+		return errors.New("self-update already in progress")
```
geunwoonoh: **Suggestion:** callers can't distinguish contention from a real failure. `pkg/executor/handlers/system/system.go:201` turns this into `"Self-update failed: self-update already in progress"` with exit code 1 — but rejecting a duplicate request is correct behavior, not a failure, so the console reports an error for something that worked as designed. A package-level sentinel (the usual Go convention) lets the handler special-case it:

```go
// ErrSelfUpdateInProgress is returned when a self-update is already running.
var ErrSelfUpdateInProgress = errors.New("self-update already in progress")
```

pkg/updater/updater.go
```diff
@@ -45,16 +54,23 @@ func (o Options) baseURL() string {
...
+	}
+	defer selfUpdateInFlight.Store(false)
```
geunwoonoh: **Suggestion:** this releases the flag on the success path too, but the actual restart happens later — `system.go:204` schedules it 1 second out. A second upgrade command arriving in that window re-runs the entire self-update, because `version.Version != latestVersion` is still true in-process.

On Windows the second run then fails inside `replaceBinary`: `CleanupStaleOld()` can't remove `.old` (it's the running image), and `moveFileEx(currentPath, oldPath, MOVEFILE_REPLACE_EXISTING)` hits a sharing violation, surfacing `"failed to move running binary aside"` even though the replacement already succeeded — after a wasted 100 MB download.

A one-shot latch (don't clear on success) is more accurate, since the process is about to exit anyway. The tradeoff is that it also blocks a legitimate retry when restart scheduling fails at `system.go:207`, so clearing it explicitly on that branch would be the complete version.

pkg/updater/guard_windows.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// ensureSelfRestartable aborts the self-update unless a relaunch is guaranteed
+// afterward—otherwise the replaced binary stays dead and a remote-only server
+// goes unreachable. Preflight only: config drift or a failure streak past the
+// first action can still leave the service stopped.
+func ensureSelfRestartable() error {
+	isSvc, err := svc.IsWindowsService()
+	if err != nil {
+		return abortf("failed to determine service mode: %w", err)
```
geunwoonoh, line 38: **Suggestion:** this diverges from `runningAsWindowsService()` (`cmd/alpamon/command/svc_windows.go:60-67`), which treats the same `IsWindowsService()` error as "assume interactive" and returns `false`. Here the error aborts.

The consequence: on a box where that call errors, self-update is blocked even though `restartAgent`'s interactive path would have relaunched it via `exec.Command`. The stricter choice is defensible — misclassifying a real service as interactive would spawn a child that fights the SCM lifecycle — but anyone reading the two functions side by side will assume one of them is a bug. One line of comment marking the asymmetry as deliberate would settle it.

pkg/updater/guard_windows.go
```diff
@@ -0,0 +1,100 @@
...
+// an interface so the self-heal path is testable without a live SCM.
+type recoveryConfigurer interface {
+	RecoveryActions() ([]mgr.RecoveryAction, error)
+	SetRecoveryActions(actions []mgr.RecoveryAction, resetPeriod uint32) error
+}
+
+// ensureSelfRestartable aborts the self-update unless a relaunch is guaranteed
+// afterward—otherwise the replaced binary stays dead and a remote-only server
+// goes unreachable. Preflight only: config drift or a failure streak past the
+// first action can still leave the service stopped.
```
geunwoonoh: **Suggestion:** documenting "a failure streak past the first action" as a known limitation is good, but there's a case where it stops being transient. If the service's `resetPeriod` is `INFINITE`, the failure count never resets, so after two self-updates the third lands on `defaults[2]` (`NoAction`) and the service stays dead **permanently**, not just until the streak ages out.

`mgr.RecoveryActions()` doesn't return the reset period, so checking it needs a raw `QueryServiceConfig2`. The cheaper mitigation is to make the third healed action a `ServiceRestart` instead of `NoAction` — though that conflicts with why `NoAction` is there (bounding a restart loop on a genuinely broken binary), so it's a real tradeoff. At minimum, please name this case in the comment.

pkg/updater/updater.go
```diff
@@ -33,6 +35,13 @@ const (
 
 var versionRe = regexp.MustCompile(`^v\d+\.\d+\.\d+(-[\w.]+)?$`)
 
+// selfUpdateInFlight serializes SelfUpdate: concurrent runs race on the same
+// staging paths and can delete each other's files mid-replace, killing the agent.
```
geunwoonoh: **Suggestion:** worth stating that this is process-local. As written, "concurrent runs race on the same staging paths" reads as though the race is closed in general. It covers the worker-pool dispatch this PR targets, but a second alpamon process on the same host still races on `.new`/`.old`. One clause would prevent over-trusting the guard later.

pkg/updater/guard_windows.go
```diff
@@ -0,0 +1,100 @@
...
+		return abortf("failed to verify restored recovery actions: %w", err)
+	}
+	if !firstActionRestarts(actions) {
+		return abortf("automatic restart is still not configured after restoring defaults; run 'alpamon register' to configure it")
+	}
+	log.Info().Msg("SCM recovery actions restored and verified.")
+	return nil
+}
+
+func abortf(format string, args ...any) error {
```
geunwoonoh, line 142: **Nit:** `abortf` is the only function in this file without a doc comment.

pkg/updater/guard_windows.go
jisung-02: Fixed in b062ad2. The guard now opens the SCM with `SC_MANAGER_CONNECT` and the service with `SERVICE_QUERY_CONFIG` via the raw `windows` API (the `svc/mgr` wrapper hardcodes `*_ALL_ACCESS` with no access-mask option), and escalates to a dedicated `SERVICE_CHANGE_CONFIG | SERVICE_START` handle only on the heal path — so a hardened deployment is never blocked on the common read-only path. Verified end-to-end on Windows Server 2022 (LocalSystem): the query-only read passes, and the write heal restored the defaults from a 6h-delay restart, confirming `SERVICE_START` suffices for the `SC_ACTION_RESTART` policy. A hardened non-LocalSystem logon account is not yet covered.

pkg/updater/guard_windows.go
jisung-02: Fixed in 53437c6. Added `maxAcceptableRestartDelay` (60s); `firstActionRestarts` now also requires `actions[0].Delay <= maxAcceptableRestartDelay`. A restart scheduled beyond the bound is treated like a missing restart and healed to `register`'s defaults rather than aborting, so the server can't be left unreachable for hours. E2E-verified: a `restart/21600000` (6h) first action was rejected as non-prompt and healed back to `Restart 5s` x2.

pkg/updater/guard_windows.go
jisung-02: Fixed in 059db34. The message now reads `First SCM recovery action is not a prompt restart; restoring the defaults set by 'alpamon register'.`, and the previous actions are logged via `.Strs("replaced", describeActions(actions))` before the overwrite, so an operator's replaced config can be reconstructed afterward. `describeActions` renders `mgr.RecoveryAction.Type` (a bare `int`) as a readable name, since a raw dump is meaningless to the reader.

pkg/updater/guard_windows_test.go
jisung-02: Fixed in 2172756. `fakeRecovery` now records the `actions` argument (`setActions`), and every heal case asserts it against `svcdef.DefaultRecoveryActions()` plus the reset period — so a regression to `NoAction` or a dropped `Delay` is now caught.

pkg/updater/guard_windows_test.go
jisung-02: Addressed in 2172756. Rather than `wantReset *uint32`, I dropped the per-case field entirely: the guard always writes `svcdef.RecoveryResetSeconds`, so the expectation is invariant and is now asserted in all four heal cases (gated on `wantSet`), not just one.

pkg/updater/guard_windows.go
jisung-02: Done in 5b7f1e8. The service name, reset/delay constants, and default recovery actions now live in a single `pkg/svcdef` package (`cmd -> pkg`, the correct direction), imported by `register`, the SCM handler, and the guard. The three duplicated copies and their "keep in sync" comments are gone; the heal test now asserts against `svcdef.DefaultRecoveryActions()` as a single source of truth.

pkg/updater/updater.go
jisung-02: Done in ef7332e. Exposed `ErrSelfUpdateInProgress`; the system handler special-cases it (`errors.Is`) and returns exit 0 with a benign "already in progress" message instead of rendering it as `Self-update failed` with exit 1.

pkg/updater/updater.go
jisung-02: Done in ccaa258. The latch is now released only on failure; on success it stays held (the process is about to restart, which resets it), closing the pre-restart window where a second upgrade could re-run and fail in `replaceBinary`. The system handler calls `ReleaseSelfUpdateLatch()` on the one success branch where the restart can't be scheduled, so a legitimate retry isn't blocked.

pkg/updater/guard_windows.go
jisung-02: Done in 63ab2e3. Added a comment marking the asymmetry with `runningAsWindowsService()` as deliberate: an unknown service mode aborts here, because guessing interactive on a real service would skip the restart check and leave the replaced binary with nothing to relaunch it.

pkg/updater/guard_windows.go
jisung-02: Done in 7b583cc. The comment now names the permanent case: with an INFINITE reset period the failure count never ages out, so the third self-update lands on `NoAction` and the service stays dead for good — undetectable in the guard since `mgr.RecoveryActions()` omits the reset period.

pkg/updater/updater.go
jisung-02: Done in ccaa258. The comment now states the latch is process-local and does not guard against a second `alpamon` process on the same host racing on `.new`/`.old`.

pkg/updater/guard_windows.go
jisung-02: Done in 7b583cc. `abortf` now has a doc comment.

geunwoonoh (approved):

eunyoung14 (approved):
Second-round review: verified every one of the 12 review-thread fixes against the code at head, plus a fresh pass over the full diff (including a `GOOS=windows` build/vet/test-compile of the touched packages). Approving.

**All claimed fixes verified:**

- **Least-privilege SCM handles (b062ad2)**: the read path opens `SC_MANAGER_CONNECT` + `SERVICE_QUERY_CONFIG` via the raw `windows` API, and `RecoveryActions()` → `QueryServiceConfig2` indeed needs only QUERY_CONFIG. The write path opens a separate `SERVICE_CHANGE_CONFIG | SERVICE_START` handle (START is correctly required for `SC_ACTION_RESTART`), wrapped in `mgr.Service` with a single `Close()` on each path — no leak or double-close on any error path I could trace.
- **Delay bound (53437c6)**: `firstActionRestarts` now checks both `Type` and `Delay <= maxAcceptableRestartDelay`, with table tests at the bound, past it, and the 6h case.
- **Latch ownership (ccaa258 + ef7332e)**: traced every return path — CAS-reject returns before the release defer is registered (no stolen release), every failure path releases, success holds through the pending restart, and the system handler releases only when it owns the failure (restart scheduling failed). No double-release, no missed release. The one remaining case — update succeeded but the restart never fires, latch held until process restart — is documented and reasonable.
- **svcdef extraction (5b7f1e8)**: register, the SCM handler, and the guard all import it; grep at head finds no stale duplicate constants.
- Test-fake assertion fixes (2172756), log message (059db34), `IsWindowsService` asymmetry comment (63ab2e3), INFINITE-reset documentation (7b583cc), process-local-latch comment, `abortf` doc — all present as described.
- The Copilot build-constraint thread was correctly disputed: the `_windows` filename suffix is applied after stripping `_test`, so `guard_windows_test.go` is implicitly GOOS=windows, and the Windows CI job actually runs it.

**On the guard's premise** (checked deliberately, since everything hangs on it): the service-mode restart path exits via `os.Exit(1)` without reporting `SERVICE_STOPPED`, so SCM sees an abnormal termination and runs recovery actions regardless of `FailureActionsOnNonCrashFailures` — the guard is verifying a mechanism that genuinely fires on this exit, and the real-server E2E in the description confirms it end to end. The comment in `restartAgent` explaining exactly this is appreciated; please keep it in sync if the exit strategy ever changes, since the guard's contract depends on the non-clean exit.

**Two adjacent notes** (pre-existing, non-blocking — flagging for follow-up tickets rather than this PR):

- The binary replace path is correct about Windows file locking: it never overwrites the running exe, it renames it aside (allowed while running) and moves the staged sibling into place on the same volume, with a rollback path and a delete-on-reboot backstop. No sharing-violation failure mode there.
- Update authenticity currently rests on HTTPS + a SHA256 checksum file fetched from the same GitHub Releases origin as the artifact — the `TODO(security)` in the code already says this plainly. Since this PR makes unattended self-update more reliable (and therefore more used), it strengthens the case for prioritizing the signature-verification follow-up (cosign/minisign key pinned in the binary). Related: the trigger condition is `version.Version != latestVersion` against `releases/latest`, so a rolled-back or re-pointed "latest" release would be installed even if older — a monotonic version check would close that at the same time.

Nice work on the fail-closed structure, the `describeActions` audit log before the destructive heal, and the `recoveryConfigurer` seam that made the heal path actually testable.

PR #377, opened 2026-07-24
fix: report correct broadcast address for 16-byte net.IP

### LINKED ISSUE #372: calculateBroadcastAddress reports wrong broadcast on all platforms (16-byte net.IP indexing)
## Summary

`calculateBroadcastAddress` in `pkg/runner/commit.go:670` indexes the raw `net.IP` (`ip[i]`) instead of its 4-byte form (`ip.To4()[i]`). When `ip` is in the 16-byte IPv4-in-IPv6 representation, `ip[0..3]` is the `::ffff:` prefix (`0,0,0,0`), not the address — the real octets live in `ip[12..15]`. As a result the computed broadcast address is wrong.

This surfaced during the review of #371 (found in a file that PR touches, but it is a pre-existing bug unrelated to that refactor, so it was deliberately kept out to preserve the PR's behavior-preserving guarantee).

## Affected code

```go
func calculateBroadcastAddress(ip net.IP, mask net.IPMask) string {
	// only ipv4
	if ip.To4() == nil || len(mask) != net.IPv4len {
		return ""
	}

	broadcast := make(net.IP, len(ip.To4()))
	for i := 0; i < len(ip.To4()); i++ {
		broadcast[i] = ip[i] | ^mask[i]   // <- ip[i], not ip.To4()[i]
	}

	return broadcast.String()
}
```

## Impact

All supported platforms are affected, not just darwin. Go's stdlib constructs IPv4 interface addresses via `net.IPv4()` on every platform (`interface_bsd.go`, `interface_linux.go:180`, `interface_windows.go:122`), and `net.IPv4()` always returns the 16-byte IPv4-in-IPv6 form. So `iface.Addrs()` yields 16-byte IPs everywhere, and the broadcast address reported to Alpacon during the system-info commit is always the bitwise inverse of the mask (`0.0.0.255`, `0.0.255.255`, ...) regardless of OS.

(This section originally claimed Linux was unaffected because "netlink yields 4-byte IPs" — that was wrong. The netlink path calls `net.IPv4(a.Value[0], ...)`, which returns the 16-byte form. Verified by measurement below.)

Measured by extracting the function verbatim:

macOS host:

```
4-byte  IP len=4  -> "192.168.1.255"   (correct)
16-byte IP len=16 -> "0.0.0.255"       (wrong)
HOST en0  ip=10.1.0.28 len(ip)=16 mask=ffffff00 -> broadcast="0.0.0.255"       (should be 10.1.0.255)
HOST lo0  ip=127.0.0.1 len(ip)=16 mask=ff000000 -> broadcast="0.255.255.255"
```

Linux (golang:1.25 container):

```
HOST eth0 ip=172.17.0.2 len(ip)=16 mask=ffff0000 -> broadcast="0.0.255.255"    (should be 172.17.255.255)
HOST lo   ip=127.0.0.1  len(ip)=16 mask=ff000000 -> broadcast="0.255.255.255"  (should be 127.255.255.255)
```

Note the mask guard does not help: masks are 4-byte (`CIDRMask(n, 8*IPv4len)`) on all platforms, so `len(mask) != net.IPv4len` passes and the wrong value is produced silently. The existing `TestGetNetworkAddresses` only asserts `NotEmpty(addr.Broadcast)`, so it cannot catch this.

There is also a minor inefficiency: `ip.To4()` is called on every loop iteration (and again in the length expression).

## Suggested fix

Hoist `ip.To4()` once and index the 4-byte form:

```go
v4 := ip.To4()
broadcast := make(net.IP, len(v4))
for i := range v4 {
	broadcast[i] = v4[i] | ^mask[i]
}
```

## Notes

This is a behavior change, so it should land as its own PR rather than being folded into #371. A table test over both the 4-byte and 16-byte input forms would guard against regression.

jisung-02:
## Background

`calculateBroadcastAddress` (`pkg/runner/commit.go:670`) used `ip.To4()` to check for IPv4 and to size its loop, but then indexed the original `ip` slice when reading the octets. `net.IP` stores the same IPv4 address in either a 4-byte or a 16-byte IPv4-in-IPv6 form, and in the latter `ip[0..3]` is the `::ffff:` prefix, not the address. The real octets live in `ip[12..15]`.

As a result the computed broadcast address was always the bitwise inverse of the mask. That value flows through `getNetworkAddresses` (`pkg/runner/commit.go:644`) into the `broadcast` field of the system-info commit sent to Alpacon.

Issue #372 was originally filed as darwin-only, but measurement showed it affects every supported platform. Go's stdlib builds interface addresses with `net.IPv4()` on all platforms (`interface_bsd.go`, `interface_linux.go:180`, `interface_windows.go:122`), and `net.IPv4()` always returns the 16-byte form. The issue title and body have been corrected accordingly.

Measured by extracting the pre-fix function verbatim:

| Environment | Interface | Input | Before | Correct |
|---|---|---|---|---|
| macOS | en0 | 10.1.0.24/24 | 0.0.0.255 | 10.1.0.255 |
| macOS | lo0 | 127.0.0.1/8 | 0.255.255.255 | 127.255.255.255 |
| Linux (golang:1.25) | eth0 | 172.17.0.2/16 | 0.0.255.255 | 172.17.255.255 |

Windows is affected by the same stdlib path, but this was confirmed by reading `interface_windows.go:122` rather than by measuring on a Windows host, so it is not listed in the table.

## Changes

Hoist `ip.To4()` into `v4` once and index that instead of the raw `ip`. Both the 4-byte and the 16-byte representation now produce the same result, and the repeated `To4()` call in the loop condition goes away.

The work is split into two commits so that each one is green on its own. The fix comes first, because landing the test first would leave a red commit.

- `9437311 fix(runner)`: the behavior fix in `pkg/runner/commit.go`
- `ad2ec1d test(runner)`: the regression test in `pkg/runner/commit_test.go`

The existing `TestGetNetworkAddresses` only asserted the field was non-empty via `assert.NotEmpty`, so it stayed green against the wrong `0.0.0.255`-style values and could not have caught this. A table test against the pure function was added instead (`pkg/runner/commit_test.go:158`). It pins the 16-byte and 4-byte inputs to the same expected broadcast and covers `/8`, `/16`, `/24`, `/32` masks plus the guard cases (IPv6 address, IPv6-length mask, nil mask). The `/8` and `/16` cases pin the addresses actually measured in the table above.

## Safety

The JSON shape and field set are unchanged; only the `broadcast` value is corrected, so there is no wire format change. The `Address` struct (`pkg/runner/commit_types.go`) was not touched.

Indexing is in bounds: `mask` is checked with `len(mask) != net.IPv4len` before the loop, and `v4` is always 4 bytes when non-nil, so both slices indexed in the loop are the same length.

One thing this PR cannot fix in code: broadcast values already stored in Alpacon for existing servers remain wrong until each server upgrades to this version and re-commits its system info. The bad values are identifiable in the data because they are always the inverted netmask (`0.0.0.255`, `0.0.255.255`, `0.255.255.255`, and so on).

Where the wrong value actually surfaced, traced through the downstream repos:

- alpacon-server stores it as `proc.InterfaceAddress.broadcast` (`proc/models.py`) and serves it from `/api/proc/addresses/` as well as nested under `/api/proc/interfaces/` via `InterfaceSerializer.addresses` (`proc/api/serializers.py`, `proc/api/urls.py`). Any direct API consumer has been reading the wrong value.
- `broadcast` is in `InterfaceAddressViewSet.filterset_fields` (`proc/api/views.py`), so filtering by a real broadcast address returned no rows, since every stored value was of the inverted-netmask form. This is the most concrete practical impact.
- The Django admin lists the column (`proc/admin.py`, `InterfaceAddressAdmin.list_display`), so it was visible to staff there.
- alpacon-web types the field (`src/types/server.ts`, `InfraServerNetworkIF.addresses[].broadcast`) and fetches it through `useInterfacesQuery` (`src/queries/Proc.ts`), but the only component consuming that type, `src/components/servers/detail/lists/NetworkInterfaceList.tsx`, renders just `address.address` and `netmask2CIDR(address.mask)`. Broadcast is not shown in the console UI, so there was no user-visible symptom to notice.
- alpacon-cli has no reference to the field.

The other `broadcast` occurrences in alpacon-web belong to the DHCP and IPAM subnet forms, which are user-entered values unrelated to this agent-reported field. This downstream trace is based on GitHub code search plus reading `NetworkInterfaceList.tsx` in full, so treat the "not rendered anywhere" conclusion as strong but not exhaustively proven.

## Tests

- `go test ./... -p 1` passes, 0 failures
- `go test ./pkg/runner -run TestCalculateBroadcastAddress -v` passes all 8 subtests
- The 4 sixteen-byte cases were confirmed to fail against the pre-fix code before the fix was applied
- `go vet ./pkg/runner` and `gofmt -l pkg/runner` clean
- `grep -rn 'ip\[i\]\|\.To4()\['` over the repo confirms no other occurrence of the same pattern
- gopls `modernize` analyzer reports 0 diagnostics for `pkg/runner`

## Checklist

- [x] Regression test added, verified failing before the fix and passing after
- [x] Both the 4-byte and 16-byte `net.IP` representations covered
- [x] No wire format or JSON field change
- [x] Full test suite, vet, and gofmt pass
- [x] Platform scope error in issue #372 corrected

Closes #372

### DIFF
--- a/pkg/runner/commit.go
+++ b/pkg/runner/commit.go
@@ -669,13 +669,14 @@ func getFlags(iface net.Interface) int {
 
 func calculateBroadcastAddress(ip net.IP, mask net.IPMask) string {
 	// only ipv4
-	if ip.To4() == nil || len(mask) != net.IPv4len {
+	v4 := ip.To4()
+	if v4 == nil || len(mask) != net.IPv4len {
 		return ""
 	}
 
-	broadcast := make(net.IP, len(ip.To4()))
-	for i := 0; i < len(ip.To4()); i++ {
-		broadcast[i] = ip[i] | ^mask[i]
+	broadcast := make(net.IP, net.IPv4len)
+	for i := range v4 {
+		broadcast[i] = v4[i] | ^mask[i]
 	}
 
 	return broadcast.String()
--- a/pkg/runner/commit.go
+++ b/pkg/runner/commit.go
@@ -636,11 +636,12 @@ func getNetworkAddresses() ([]Address, error) {
 				ip = v.IP
 				mask = ip.DefaultMask()
 			}
-			if ip == nil || ip.To4() == nil {
+			v4 := ip.To4()
+			if v4 == nil {
 				continue
 			}
 			addresses = append(addresses, Address{
-				Address:       ip.To4().String(),
+				Address:       v4.String(),
 				Broadcast:     calculateBroadcastAddress(ip, mask),
 				InterfaceName: iface.Name,
 				Mask:          net.IP(mask).String(),
@@ -667,8 +668,10 @@ func getFlags(iface net.Interface) int {
 	return flags
 }
 
+// calculateBroadcastAddress returns the IPv4 directed broadcast for ip/mask.
+// It indexes ip.To4(), not ip: net.Interfaces() yields the 16-byte
+// IPv4-in-IPv6 form, where ip[0:4] is the ::ffff: prefix, not the address.
 func calculateBroadcastAddress(ip net.IP, mask net.IPMask) string {
-	// only ipv4
 	v4 := ip.To4()
 	if v4 == nil || len(mask) != net.IPv4len {
 		return ""
--- a/pkg/runner/commit_test.go
+++ b/pkg/runner/commit_test.go
@@ -152,6 +152,12 @@ func TestGetNetworkAddresses(t *testing.T) {
 		assert.NotEmpty(t, addr.Broadcast, "Broadcast address should not be empty.")
 		assert.NotEmpty(t, addr.InterfaceName, "Interface name should not be empty.")
 		assert.NotEmpty(t, addr.Mask, "Mask should not be empty.")
+
+		mask := net.IPMask(net.ParseIP(addr.Mask).To4())
+		ip := net.ParseIP(addr.Address)
+		bcast := net.ParseIP(addr.Broadcast)
+		assert.Equal(t, ip.Mask(mask), bcast.Mask(mask),
+			"Broadcast should be on the same network as the address.")
 	}
 }
 
@@ -166,10 +172,11 @@ func TestCalculateBroadcastAddress(t *testing.T) {
 		{name: "16-byte IP /24", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(24, 32), want: "192.168.1.255"},
 		{name: "4-byte IP /24", ip: net.IPv4(192, 168, 1, 10).To4(), mask: net.CIDRMask(24, 32), want: "192.168.1.255"},
 		{name: "16-byte IP /8", ip: net.IPv4(127, 0, 0, 1), mask: net.CIDRMask(8, 32), want: "127.255.255.255"},
+		{name: "4-byte IP /8", ip: net.IPv4(127, 0, 0, 1).To4(), mask: net.CIDRMask(8, 32), want: "127.255.255.255"},
 		{name: "16-byte IP /16", ip: net.IPv4(172, 17, 0, 2), mask: net.CIDRMask(16, 32), want: "172.17.255.255"},
 		{name: "16-byte IP /32", ip: net.IPv4(10, 1, 0, 28), mask: net.CIDRMask(32, 32), want: "10.1.0.28"},
 		{name: "IPv6 address", ip: net.ParseIP("2001:db8::1"), mask: net.CIDRMask(24, 32), want: ""},
-		{name: "IPv6 mask", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(64, 128), want: ""},
+		{name: "16-byte mask", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(64, 128), want: ""},
 		{name: "nil mask", ip: net.IPv4(192, 168, 1, 10), mask: nil, want: ""},
 	}
 
--- a/pkg/runner/commit_test.go
+++ b/pkg/runner/commit_test.go
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"encoding/json"
+	"net"
 	"runtime"
 	"slices"
 	"strings"
@@ -151,6 +152,31 @@ func TestGetNetworkAddresses(t *testing.T) {
 		assert.NotEmpty(t, addr.Broadcast, "Broadcast address should not be empty.")
 		assert.NotEmpty(t, addr.InterfaceName, "Interface name should not be empty.")
 		assert.NotEmpty(t, addr.Mask, "Mask should not be empty.")
+	}
+}
+
+func TestCalculateBroadcastAddress(t *testing.T) {
+	tests := []struct {
+		name string
+		ip   net.IP
+		mask net.IPMask
+		want string
+	}{
+		// net.Interfaces() yields 16-byte IPv4-in-IPv6 IPs on all platforms
+		{name: "16-byte IP /24", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(24, 32), want: "192.168.1.255"},
+		{name: "4-byte IP /24", ip: net.IPv4(192, 168, 1, 10).To4(), mask: net.CIDRMask(24, 32), want: "192.168.1.255"},
+		{name: "16-byte IP /8", ip: net.IPv4(127, 0, 0, 1), mask: net.CIDRMask(8, 32), want: "127.255.255.255"},
+		{name: "16-byte IP /16", ip: net.IPv4(172, 17, 0, 2), mask: net.CIDRMask(16, 32), want: "172.17.255.255"},
+		{name: "16-byte IP /32", ip: net.IPv4(10, 1, 0, 28), mask: net.CIDRMask(32, 32), want: "10.1.0.28"},
+		{name: "IPv6 address", ip: net.ParseIP("2001:db8::1"), mask: net.CIDRMask(24, 32), want: ""},
+		{name: "IPv6 mask", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(64, 128), want: ""},
+		{name: "nil mask", ip: net.IPv4(192, 168, 1, 10), mask: nil, want: ""},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.want, calculateBroadcastAddress(tc.ip, tc.mask))
+		})
 	}
 }
 


pkg/runner/commit.go
```diff
@@ -669,13 +669,14 @@ func getFlags(iface net.Interface) int {
 
 func calculateBroadcastAddress(ip net.IP, mask net.IPMask) string {
 	// only ipv4
```
geunwoonoh: Suggestion: keep the regression-preventing knowledge next to the code, not only in the test.

`// only ipv4` states *what* but not *why `v4` must be indexed*. The non-obvious fact that caused this bug — that `net.Interfaces()` hands back the 16-byte form — currently lives only in the comment at `commit_test.go:165`. Whoever touches this function next may not read the test.

```suggestion
// calculateBroadcastAddress returns the IPv4 directed broadcast for ip/mask.
// It indexes ip.To4(), not ip: net.Interfaces() yields the 16-byte
// IPv4-in-IPv6 form, where ip[0:4] is the ::ffff: prefix, not the address.
func calculateBroadcastAddress(ip net.IP, mask net.IPMask) string {
```

(The doc comment subsumes `// only ipv4`, so it is dropped.)

pkg/runner/commit_test.go
```diff
@@ -154,6 +155,31 @@ func TestGetNetworkAddresses(t *testing.T) {
...
+	tests := []struct {
+		name string
+		ip   net.IP
+		mask net.IPMask
+		want string
+	}{
+		// net.Interfaces() yields 16-byte IPv4-in-IPv6 IPs on all platforms
+		{name: "16-byte IP /24", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(24, 32), want: "192.168.1.255"},
+		{name: "4-byte IP /24", ip: net.IPv4(192, 168, 1, 10).To4(), mask: net.CIDRMask(24, 32), want: "192.168.1.255"},
+		{name: "16-byte IP /8", ip: net.IPv4(127, 0, 0, 1), mask: net.CIDRMask(8, 32), want: "127.255.255.255"},
```
geunwoonoh, line 174: Suggestion: 4-byte/16-byte parity is pinned for `/24` only.

The bug itself was representation-dependent, but `/24` is the single mask with both forms covered. Adding one 4-byte pair for a non-`/24` mask is a one-line change:

```suggestion
		{name: "16-byte IP /8", ip: net.IPv4(127, 0, 0, 1), mask: net.CIDRMask(8, 32), want: "127.255.255.255"},
		{name: "4-byte IP /8", ip: net.IPv4(127, 0, 0, 1).To4(), mask: net.CIDRMask(8, 32), want: "127.255.255.255"},
```

Restructuring the table to iterate every case in both representations is also possible, but the guard cases (IPv6 address) do not fit that axis and it ends up more complex — the one-line addition is the better trade.

pkg/runner/commit_test.go
```diff
@@ -154,6 +155,31 @@ func TestGetNetworkAddresses(t *testing.T) {
...
+		{name: "16-byte IP /16", ip: net.IPv4(172, 17, 0, 2), mask: net.CIDRMask(16, 32), want: "172.17.255.255"},
+		{name: "16-byte IP /32", ip: net.IPv4(10, 1, 0, 28), mask: net.CIDRMask(32, 32), want: "10.1.0.28"},
+		{name: "IPv6 address", ip: net.ParseIP("2001:db8::1"), mask: net.CIDRMask(24, 32), want: ""},
+		{name: "IPv6 mask", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(64, 128), want: ""},
```
geunwoonoh: Suggestion (nit): the case name is slightly off from what it verifies.

What is being exercised is not "an IPv6 mask" but the guard "a 16-byte-length mask is rejected" (`len(mask) != net.IPv4len`).

```suggestion
		{name: "16-byte mask", ip: net.IPv4(192, 168, 1, 10), mask: net.CIDRMask(64, 128), want: ""},
```

geunwoonoh (dismissed):
## Review summary

This PR fixes a real bug with an accurate root-cause diagnosis and a minimal (5-line) change. `calculateBroadcastAddress` used `ip.To4()` to decide IPv4-ness and to size the loop, but then indexed the raw `ip` slice — and in the 16-byte IPv4-in-IPv6 form that `net.Interfaces()` returns, `ip[0:4]` is the `::ffff:` prefix rather than the address, so the result was always the bitwise inverse of the netmask.

**No critical issues and no blocking warnings.** Everything below is optional.

---

### Verified locally

```
go test ./pkg/runner -run 'TestCalculateBroadcastAddress|TestGetNetworkAddresses' -v
  --- PASS: TestGetNetworkAddresses (0.00s)
  --- PASS: TestCalculateBroadcastAddress (0.00s)   # all 8 subtests PASS
gofmt -l pkg/runner  -> no output
go vet ./pkg/runner  -> clean
```

**Bounds re-check**: `To4()` returns a 4-byte slice for both the 4-byte and the v4-mapped 16-byte form, and the `len(mask) != net.IPv4len` guard precedes the loop, so both `v4[i]` and `mask[i]` are in range. No out-of-range risk.

**Duplicate-logic sweep**: `grep -rnE "ip\[[0-9i]|IP\[[0-9i]"` plus a full `net.IP` scan over `pkg/`, `internal/`, `cmd/` confirms **no other occurrence of the raw-`net.IP`-indexing pattern** anywhere in the repo, and no second implementation of broadcast computation. (The `cond.Broadcast()` in `pkg/scheduler/reporter.go:73` is an unrelated sync call.)

---

### Critical issues (0)

None.

### Warnings (0)

None.

---

### Suggestions (5)

Three are left as inline comments. The two below fall outside the diff hunks, so they could not be anchored inline.

**S1. `pkg/runner/commit.go:639,643` — the exact repeated-`To4()` pattern this PR removed still lives four lines above**

Inside `calculateBroadcastAddress` the call was hoisted; at the call site it was not:

```go
if ip == nil || ip.To4() == nil {   // 639
    continue
}
addresses = append(addresses, Address{
    Address:       ip.To4().String(),   // 643 — second call
```

Two extra observations: `ip.To4().String()` is equivalent to `ip.String()` (`net.IP.String()` already prints dotted-quad for a v4-mapped address), and `ip == nil` is redundant because `nil.To4()` returns nil. All three can be cleaned up at once:

```go
v4 := ip.To4()
if v4 == nil {
    continue
}
addresses = append(addresses, Address{
    Address:       v4.String(),
    Broadcast:     calculateBroadcastAddress(ip, mask),
    InterfaceName: iface.Name,
    Mask:          net.IP(mask).String(),
})
```

Passing `ip` (not `v4`) into `calculateBroadcastAddress` is the right call — keeping it a self-validating pure function is worth the one redundant `To4()`.

**S3. `pkg/runner/commit_test.go:152` — the reason this bug survived is still in place**

As the PR description notes, `assert.NotEmpty(t, addr.Broadcast)` stays green against `0.0.0.255`. Since CI runs the suite on Linux, Windows, and macOS (`build-and-test.yml:39,152,240`), asserting a real invariant here would catch an integration-level regression on all three platforms:

```go
mask := net.IPMask(net.ParseIP(addr.Mask).To4())
ip := net.ParseIP(addr.Address)
bcast := net.ParseIP(addr.Broadcast)
assert.Equal(t, ip.Mask(mask), bcast.Mask(mask),
    "broadcast should be on the same network as the address")
```

Any inverted-netmask value fails this assertion immediately.

---

### Performance

**Improvement**: `To4()` calls drop from `2N+1` (N=4) to `1` per invocation. `To4()` is not free — it includes an `isZeros(ip[0:10])` comparison.

**No regression**: allocations are unchanged — one 4-byte `make` plus one `String()`, same as before. `make(net.IP, net.IPv4len)` is equivalent to `make(net.IP, len(v4))`; the constant states the intent more clearly.

**Scale**: this runs on the periodic system-info commit path, bounded by the number of interface addresses. It is not a measurable hot path, so there is no reason to optimize further — `sync.Pool` and friends would be clear over-engineering here.

### Unused / duplicated code

None. `calculateBroadcastAddress` has exactly one call site (`commit.go:644`), and no dead code or redundant helper was introduced. The change is a net reduction: the repeated call in the loop condition is gone.

### Go conventions

| Item | Result |
|---|---|
| `for i := range v4` | Go 1.22+ idiom, consistent with the Go 1.25 modernize sweep landed just before this branch (`8245287`, `8b5dfb7`) |
| Table-driven test + testify | Matches the Go stack rules |
| `tc` loop variable | Matches the dominant convention in `pkg/runner` (`ftp_test.go`, `tunnel_daemon_test.go`, `commit_windows_test.go`) |
| No `tc := tc` rebinding | Correct for Go 1.22+ |
| `v4` naming | Idiomatic, appropriately short for its scope |
| Error wrapping / context propagation | N/A (pure function) |
| gofmt / vet | Clean |

### Note (not a defect)

`/31` (RFC 3021) and `/32` have no semantic broadcast address. This function follows the arithmetic definition `ip | ^mask`, so it returns the host address itself for `/32`, and the test pins that as intended behavior (`commit_test.go:170`). That matches the pre-fix semantics and the standard definition, so no change is warranted. Worth being aware of downstream: consumers filtering on `broadcast` (`InterfaceAddressViewSet.filterset_fields`) will see the host IP for `/32` interfaces.

---

### Highlights

- **The root cause is diagnosed correctly.** The fix targets the mismatch between the `To4()` guard and the `ip` indexing, not the `0.0.0.255` symptom.
- **The test was verified red before the fix.** The description states the four 16-byte cases fail against the pre-fix code, which is what makes this a genuine regression test rather than a restatement of current behavior.
- **The commit split is well reasoned.** Fix first, test second, with the rationale stated: landing the test first would leave a red commit. Each commit is green on its own.
- **Guard-path coverage is good.** IPv6 address, 16-byte mask, and nil mask pin all three early-return branches.
- **The platform scope error was corrected.** Issue #372 was filed as darwin-only; measurement plus reading the stdlib sources established that every supported platform is affected, and the issue was updated. Being explicit that Windows was confirmed by source reading rather than measurement is the honest framing.
- **Downstream impact is traced thoroughly.** alpacon-server, alpacon-web, and alpacon-cli were all checked, landing on "filtering is the concrete impact, no user-visible UI symptom", and the description is upfront that already-stored values stay wrong until each server re-commits.

---

**Verdict**: **APPROVED**

The five suggestions can all be handled separately after merge. S1 (repeated `To4()` at the call site) and S3 (`TestGetNetworkAddresses` invariant) fit this PR's intent most closely and would be nice to fold in.

**Checklist**:
- [x] Follows Go conventions (gofmt/vet clean, Go 1.25 idioms, repo test naming)
- [x] Tests included and passing (8/8 subtests, verified failing pre-fix)
- [x] No security vulnerabilities (no hardcoded secrets, injection, or sensitive logging)
- [x] No duplicated or unused code (confirmed by a full-repo sweep)
- [x] No performance regression (`To4()` calls `2N+1` -> `1`, allocations unchanged)
- [x] No wire format change (`commit_types.go:130` untouched)
- [ ] Integration-level regression guard (S3 — optional)

pkg/runner/commit.go
jisung-02: Done in 1f89e77. Added the doc comment recording that To4() (not raw ip) must be indexed because net.Interfaces() returns the 16-byte v4-in-v6 form where ip[0:4] is the ::ffff: prefix, and dropped the now-subsumed // only ipv4 line.

pkg/runner/commit_test.go
jisung-02: Done in ae8f58d. Added the "4-byte IP /8" case so both representations are pinned on a non-/24 mask, keeping the table flat rather than restructuring it as we discussed.

pkg/runner/commit_test.go
jisung-02: Done in ae8f58d. Renamed to "16-byte mask" since the case exercises the len(mask) != net.IPv4len guard, not an IPv6 address.

jisung-02:
Thanks for the review. Applied all five suggestions across two commits:

S1 (call-site To4 duplication) and the commit.go doc comment: 1f89e77. Hoisted ip.To4() into v4, dropped the redundant ip == nil check, and switched Address to v4.String(). Kept passing ip into calculateBroadcastAddress so it stays a self-validating pure function.

S3 (TestGetNetworkAddresses invariant) plus the two commit_test.go suggestions (4-byte /8 parity, "16-byte mask" rename): ae8f58d. The invariant asserts ip.Mask(mask) == bcast.Mask(mask), which any inverted-netmask value fails immediately.

gofmt/vet clean, and go test ./pkg/runner -run TestCalculateBroadcastAddress|TestGetNetworkAddresses passes (9 subtests).

geunwoonoh (approved):

PR #380, opened 2026-07-24
fix(runner): cap supplementary groups to the darwin setgroups limit

### LINKED ISSUE #373: macOS: websh/exec sessions may close immediately under a root daemon due to setgroups EINVAL
## Summary

On macOS hosts, when alpamon runs as a root daemon, websh (and other privilege-demoted) sessions appear to close immediately. This looks like it may stem from `setgroups(2)` returning `EINVAL` during privilege demotion, because the target account's supplementary group count can exceed the macOS `NGROUPS_MAX` limit. Filing this for discussion and confirmation.

## What we observed

On a dev macOS host, opening a websh session ends the session right after it connects, and the server returns `400 websh_session_closed` (the follow-up resize PATCH lands on an already-closed session). The local alpamon log shows:

```
fork/exec /bin/zsh: invalid argument
```

This seems to reproduce not only for `root`, but also for regular admin accounts whose supplementary group count is above 16.

## Likely cause

During demotion, `Demote` collects the target account's full supplementary group list into `syscall.Credential.Groups`. On macOS, `NGROUPS_MAX` is 16, while some accounts (e.g. `root`) belong to more groups than that.

- `pkg/utils/privilege_unix.go:63` — `usr.GroupIds()` reads all of the account's groups
- `pkg/utils/privilege_unix.go:68-78` — every group is appended to `groups`
- `pkg/utils/privilege_unix.go:89-92` — passed as `syscall.Credential{Uid, Gid, Groups: groups}`

The child applies these credentials just before `execve` (roughly `setgroups()` → `setgid()` → `setuid()` → `execve()`). If the group count exceeds `NGROUPS_MAX`, `setgroups(2)` returns `EINVAL`, the child never reaches `exec`, and this surfaces as `fork/exec ...: invalid argument`.

Observed values on macOS:

```
$ getconf NGROUPS_MAX
16
$ id root        # 23 groups: wheel, daemon, kmem, sys, tty, operator, ... com.apple.sharepoint.group.1
```

Linux uses a much larger `NGROUPS_MAX` (65536), so this looks specific to macOS.

## Trigger condition

`Demote` returns early when `os.Getuid() != 0`, so this only seems to apply when alpamon runs as root (the default for a daemon install). Running non-root does not call `setgroups`, and we did not see the issue there.

## Scope

This likely affects every path that goes through `Demote`, not just websh:

- `pkg/runner/pty_unix.go` — websh sessions
- openftp / file handlers / executor command execution

On macOS, accounts with more than 16 supplementary groups (including `root`, and commonly regular admin accounts) may fail across websh / exec / file transfer.

## Possible direction

One option would be to cap the group list passed in `Credential.Groups` to `NGROUPS_MAX` (16) on darwin, keeping the primary gid first. macOS resolves group membership dynamically via opendirectoryd, so truncating the list handed to `setgroups` is a common practice and should not change effective permissions. Linux behavior could stay as-is, with a regression test covering the darwin truncation. Happy to adjust the approach based on feedback.

## Side observations (may warrant separate issues)

1. The remote log reporter (`pkg/logger/logger.go`, `ZerologEntry`) seems to drop zerolog's `error` field, so the underlying `EINVAL` was not visible in the server log UI, which made this slower to diagnose.
2. proxy-server's `SessionClosed` event `reason` is always `"session_closed"`, making it hard to distinguish close paths.
3. proxy `ptyConnectTimeout` is 15s (the session closes if the user channel connects but the pty never arrives).

## Steps to reproduce

1. Install/register alpamon as a root daemon on a macOS host.
2. Open a websh session to that host as `root` (or an account with more than 16 supplementary groups).
3. The session ends right away with `400 websh_session_closed`, and the alpamon log shows `fork/exec /bin/zsh: invalid argument`.

jisung-02:
## Background

On macOS, `setgroups(2)` rejects a supplementary group list longer than the kernel limit (`kern.ngroups`, 16 by default) with `EINVAL`. When alpamon runs as a root daemon and demotes a websh/exec/file session to a login account that belongs to more than 16 groups (root and admin accounts commonly do), the `setgroups` call fails just before `execve`. It surfaces as `fork/exec ...: invalid argument` and the session closes immediately. Linux keeps a much larger limit and is unaffected, so this only reproduces on macOS.

## Changes

Route every privilege-demotion path through a single `utils.ResolveGroups`.

- `ResolveGroups` places the primary gid first so it survives truncation, reports whether the requested gid was in the user's group list (for `ValidateGroup`), and caps the list to the platform limit.
- On darwin the cap comes from the `kern.ngroups` sysctl, the limit the kernel reports and enforces, falling back to a constant 16 if the read fails. On Linux the cap is 0 (uncapped). The OID is read-only and derived from the compile-time `NGROUPS_MAX`, so the value cannot change while the process runs and is read once via `sync.OnceValue`; demotion runs on every command execution, not once per session.
- The per-platform cap lives in build-tagged files `pkg/utils/ngroups_darwin.go` and `pkg/utils/ngroups_other.go`, because `unix.SysctlUint32` is BSD/darwin-only and would not compile on Linux under a single-file `runtime.GOOS` switch.
- `Demote` and the websh PTY path in `pkg/runner/pty_unix.go` previously built `Credential.Groups` with separate code; the PTY path used `ConvertGroupIds`, which had no cap and was the actual websh failure. Both now share `ResolveGroups`, and the unused `ConvertGroupIds` is removed.
- Truncation logs the requested count and the cap at debug level, so an operator debugging a permission failure can tell that groups were dropped. A gid that fails to parse is wrapped as `invalid supplementary group id`, since `Demote` does not wrap it and the bare `strconv.ParseUint` error gave no indication it concerned supplementary groups.

## Safety

This touches the credential path used for privilege demotion, so a few properties matter.

- Truncation only removes trailing entries; it never adds a group, so it cannot escalate privileges.
- The primary gid is prepended before truncation, so it can never be dropped from the list.
- `groupInList` is computed over the full list before truncation, so `ValidateGroup` does not falsely reject a legitimate group.
- macOS resolves group membership dynamically via `opendirectoryd`, so truncating the `setgroups` list does not change effective permissions on darwin.
- The PTY path now propagates a parse error instead of silently skipping an unparsable gid (fail-closed). These ids come from local OS lookups via `usr.GroupIds()`, not from attacker-controlled input.

## Testing

- `go test ./pkg/utils/` passes. `TestResolveGroups` is table-driven and covers primary-gid ordering and dedup, the primary gid being prepended when the user is not a member, truncation to the cap, the exactly-at-cap and cap-of-one boundaries, `math.MaxUint32`, and the four rejected inputs (`-1`, `4294967296`, `""`, a non-numeric string).
- `TestResolveGroups_AppliesPlatformCap` exercises the exported `ResolveGroups` so the platform cap itself is covered, not only the unexported helper with an explicit `maxGroups`.
- `TestMaxSupplementaryGroups_ReadsSysctl` reads `kern.ngroups` directly and asserts it against `maxSupplementaryGroups()`, so a misspelled OID cannot hide behind the fallback constant. The `!darwin` counterpart asserts the uncapped 0.
- `go build` verified under `GOOS=darwin`, `GOOS=linux`, and `GOOS=windows` to confirm the build-tag split resolves to exactly one `maxSupplementaryGroups` on each platform.

Manual end-to-end verification on macOS (arm64, macOS 26.5): built this branch and ran it as a root launchd daemon (`com.alpacax.alpamon`), then opened a Websh session from the Alpacon console. The demoted PTY session now starts and stays open as an interactive shell. Before this change, the same setup closed the session immediately because `setgroups(2)` returned `EINVAL` when demoting to an account with more than 16 groups (surfacing as `fork/exec ...: invalid argument`).

## Follow-up

`syscall.Credential` is still assembled independently by each caller, and `pkg/runner/tunnel_linux.go` builds its group list with its own `parseGroupIDs`, which applies no cap. That is Linux-only and unrelated to this fix, tracked separately in #386.

Closes #373


pkg/utils/ngroups_darwin.go
```diff
@@ -0,0 +1,23 @@
+package utils
+
+import "golang.org/x/sys/unix"
```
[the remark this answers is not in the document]
jisung-02: `_darwin.go` is an implicit GOOS build constraint in the Go toolchain, so an explicit `//go:build darwin` line is redundant here. Verified by cross-compiling: `GOOS=linux go list -f '{{.GoFiles}}' ./pkg/utils` lists only `ngroups_other.go`, and `go build ./pkg/utils` succeeds for `GOOS=linux`, `GOOS=windows`, and `GOOS=darwin`.

The current implementation is appropriate because:
- The `_GOOS` filename suffix already excludes the file on every non-darwin target, so `unix.SysctlUint32` is never compiled off darwin.
- `ngroups_other.go` carries `//go:build !darwin` because its name has no platform suffix — it needs the explicit tag, `ngroups_darwin.go` does not. No duplicate definition of `maxSupplementaryGroups` is possible.

pkg/utils/ngroups_darwin.go
```diff
@@ -0,0 +1,23 @@
...
+// the list must be truncated. macOS resolves group membership dynamically via
+// opendirectoryd, so truncating the setgroups list does not change effective
+// permissions.
+const darwinNGroupsMax = 16
+
+// maxSupplementaryGroups reads kern.ngroups, the runtime-tunable limit the
+// kernel actually enforces in setgroups(2), falling back to darwinNGroupsMax
+// if the sysctl cannot be read.
+func maxSupplementaryGroups() int {
+	if n, err := unix.SysctlUint32("kern.ngroups"); err == nil && n > 0 {
```
geunwoonoh, line 25: **Warning:** the sysctl path is untestable as written, and silently degrades.

This falls back to `darwinNGroupsMax` on *any* error, including a bad OID name. If `"kern.ngroups"` were ever typo'd, the code would still behave correctly via the constant — so the bug would be invisible, and the entire runtime-read path would be dead with no test to notice.

Neither `maxSupplementaryGroups()` nor the exported `ResolveGroups` has any coverage today; every test calls the unexported `resolveGroups` with an explicit `maxGroups`. Nothing asserts the cap is actually applied on darwin, which is the whole point of the PR.

CI already runs `macos-arm64`, so this is cheap:

```go
// pkg/utils/ngroups_darwin_test.go  (implicit darwin constraint)
func TestMaxSupplementaryGroups_ReadsSysctl(t *testing.T) {
	n, err := unix.SysctlUint32("kern.ngroups")
	require.NoError(t, err, "kern.ngroups OID must be readable; a typo here would hide behind the fallback")
	assert.Equal(t, int(n), maxSupplementaryGroups())
}
```

plus a `!darwin` counterpart asserting `maxSupplementaryGroups() == 0`.

pkg/utils/privilege_unix.go
```diff
@@ -12,6 +12,38 @@ import (
...
+			return nil, false, err
+		}
+		if uint32(gidUint) == gid {
+			groupInList = true
+			continue
+		}
+		groups = append(groups, uint32(gidUint))
+	}
+	if maxGroups > 0 && len(groups) > maxGroups {
+		groups = groups[:maxGroups]
```
geunwoonoh, line 44: **Warning:** truncation is silent.

When the cap drops groups there is no log line at any level. An operator debugging "why can't this Websh session read `/X`" on macOS gets no signal that 5 groups were discarded. For a product whose stated promise is that every action is supervised and audited, a demotion path that silently narrows a credential should say so.

`log` is already imported in this file.

```suggestion
		log.Debug().Int("requested", len(groups)).Int("cap", maxGroups).
			Msg("Supplementary group list truncated to the platform setgroups limit.")
		groups = groups[:maxGroups]
```

pkg/utils/ngroups_darwin.go
```diff
@@ -0,0 +1,23 @@
...
+// kern.ngroups sysctl cannot be read. setgroups(2) returns EINVAL when the
+// supplementary group list exceeds the kernel limit, which aborts privilege
+// demotion just before execve and surfaces as "fork/exec ...: invalid
+// argument". Accounts such as root belong to more than 16 groups on macOS, so
+// the list must be truncated. macOS resolves group membership dynamically via
+// opendirectoryd, so truncating the setgroups list does not change effective
+// permissions.
+const darwinNGroupsMax = 16
+
+// maxSupplementaryGroups reads kern.ngroups, the runtime-tunable limit the
```
geunwoonoh: **Suggestion:** this doc comment is factually wrong, and correcting it unblocks a free optimization.

`kern.ngroups` is not "runtime-tunable":

```
$ sysctl -w kern.ngroups=15
sysctl: oid 'kern.ngroups' is read only
```

It is derived from the compile-time kernel `NGROUPS_MAX`. Suggest rewording to something like "the limit the kernel reports and enforces".

Because it cannot change during the process lifetime, caching is unambiguously safe. `ResolveGroups` sits on the per-command path (`pkg/executor/executor.go:489` demotes on *every* command execution, not just per session), so today this is a `sysctlbyname` per exec:

```go
var maxSupplementaryGroups = sync.OnceValue(func() int {
	if n, err := unix.SysctlUint32("kern.ngroups"); err == nil && n > 0 {
		return int(n)
	}
	return darwinNGroupsMax
})
```

Honest sizing: the win is roughly a microsecond against a `fork`/`execve`. This is about removing a pointless repeated syscall, not measurable throughput — worth taking mainly because the read-only finding makes it cost-free.

pkg/runner/pty_unix.go
```diff
@@ -20,11 +20,15 @@ func (pc *PtyClient) setPtyCmdSysProcAttrAndEnv(uid, gid int, groupIds []string,
 		if err != nil {
 			return err
 		}
+		groups, _, err := utils.ResolveGroups(u32gid, groupIds)
```
geunwoonoh, line 23: **Suggestion:** the duplication this PR targets is only half removed.

Group *resolution* is now shared, which is the important half. But `syscall.SysProcAttr{Credential: &syscall.Credential{Uid, Gid, Groups}}` is still assembled independently here (lines 27-33) and at `privilege_unix.go:112-118`. One helper collapses both and makes it structurally impossible for the two demotion paths to drift again:

```go
// pkg/utils/privilege_unix.go
func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (*syscall.SysProcAttr, bool, error)
```

`Demote` then returns `result.SysProcAttr` from it, and this function drops to a single call.

Arguably scope creep for a bug fix, so reasonable to defer — but it is the remaining half of the stated goal.

pkg/utils/privilege_unix.go
```diff
@@ -12,6 +12,38 @@ import (
...
+// The primary gid is placed first so it survives truncation, and the requested
+// gid's membership is reported via groupInList for ValidateGroup. When
+// maxGroups > 0 the list is capped to that many entries.
+func resolveGroups(gid uint32, groupIds []string, maxGroups int) (groups []uint32, groupInList bool, err error) {
+	groups = make([]uint32, 0, len(groupIds)+1)
+	groups = append(groups, gid)
+	for _, gidStr := range groupIds {
+		gidUint, err := strconv.ParseUint(gidStr, 10, 32)
+		if err != nil {
+			return nil, false, err
```
geunwoonoh: **Suggestion:** wrap with domain context.

This returns the bare `strconv.ParseUint` error. The PTY caller wraps it ("failed to set process credentials"), but `Demote` does not, so `pkg/executor/executor.go:489` surfaces `strconv.ParseUint: parsing "x": invalid syntax` with no indication it concerns supplementary groups.

In fairness, the surrounding `Demote` body returns bare strconv errors too, so this is a pre-existing pattern rather than something this PR introduced. `fmt` is already imported.

```suggestion
			return nil, false, fmt.Errorf("invalid supplementary group id: %w", err)
```

pkg/utils/privilege_unix_test.go
```diff
@@ -0,0 +1,42 @@
...
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestResolveGroups_PrimaryGidFirstNoCap(t *testing.T) {
```
geunwoonoh: **Suggestion:** not table-driven, and it drops edge cases the deleted tests covered.

The golang stack rules call for table-driven tests; the deleted `utils_test.go` was table-driven and these four replacements are not.

More substantively: the old table covered `-1`, `4294967296`, `""`, and `math.MaxUint32`. Those inputs changed semantics in this PR — from *silently skipped* to *hard error* — and nothing now asserts the new behavior. That is exactly the class of change that deserves a test.

Restoring them as rows in one table costs a few lines:

```go
tests := []struct {
	name      string
	gid       uint32
	groupIds  []string
	maxGroups int
	want      []uint32
	wantInList, wantErr bool
}{
	{name: "negative gid rejected",  gid: 20, groupIds: []string{"1000", "-1"},         wantErr: true},
	{name: "overflow gid rejected",  gid: 20, groupIds: []string{"4294967296"},         wantErr: true},
	{name: "empty string rejected",  gid: 20, groupIds: []string{""},                   wantErr: true},
	{name: "max uint32 accepted",    gid: 20, groupIds: []string{"4294967295"},         want: []uint32{20, math.MaxUint32}},
	{name: "exactly at cap",         gid: 99, groupIds: []string{"1", "2"}, maxGroups: 3, want: []uint32{99, 1, 2}},
	{name: "cap of 1 keeps primary", gid: 99, groupIds: []string{"1", "2"}, maxGroups: 1, want: []uint32{99}},
	// ... existing four cases
}
```

The `exactly at cap` and `cap of 1` rows are worth adding on their own — the current suite only exercises the over-cap branch, so an off-by-one in the `>` comparison would pass.

geunwoonoh (dismissed):
## PR Review

**Stack**: golang | **Changes**: +114 -105 in 7 files | **CI**: all 24 checks green (incl. `golangci-lint`, `verify-tidy`, macos-arm64, windows-amd64)

---

### Summary

Fixes Websh/exec/file session failures on macOS caused by `setgroups(2)` returning `EINVAL` when a demoted account belongs to more than `kern.ngroups` (16) groups. Consolidates supplementary-group construction into a single `utils.ResolveGroups`, prepends the primary gid so it survives truncation, applies a build-tag-split platform cap (darwin: `kern.ngroups`; others: uncapped), and removes the now-superseded `ConvertGroupIds`.

The diagnosis is correct and the fix is minimal and well-targeted.

### Verification run locally

| Check | Result |
|---|---|
| `go test ./pkg/utils/ -run TestResolveGroups -v` | 4/4 PASS |
| `go test ./pkg/utils/...` | ok |
| `GOOS={darwin,linux,windows} go build ./...` | all OK (after `go generate ./pkg/db/ent`) |
| `go list -f '{{.GoFiles}}' ./pkg/utils` per GOOS | darwin&rarr;`ngroups_darwin.go`, linux/windows&rarr;`ngroups_other.go`. Exactly one definition per platform. |
| `gofmt -l pkg/utils pkg/runner` | clean |
| `go vet ./pkg/utils/ ./pkg/runner/` | clean |
| `grep -rn ConvertGroupIds` | 0 references &mdash; removal is complete |
| `strconv` import in `utils.go` | still used at `utils.go:177,190` &mdash; correctly retained |

Two claims in the PR body were checked because they carry the security argument, and both hold:

- **Fail-closed parse is safe.** `groupIds` originates from `usr.GroupIds()` at `pkg/runner/pty.go:512`, a local OS lookup &mdash; not the WebSocket payload. Hard-failing on an unparsable gid cannot be triggered by the console.
- **No privilege escalation from prepending the gid.** With `ValidateGroup: false` (`terminal.go:150` openftp, `file.go:155` upload), the server-supplied `groupname`'s gid is now added to the supplementary list where it previously wasn't. This grants nothing new: `Credential.Gid` already sets real/effective/saved gid to that group via `setgid(2)`, so the group was fully held either way.

### Critical issues (0)

None.

### Warnings (2)

**1. `pkg/utils/ngroups_darwin.go:19` &mdash; the sysctl path is untestable as written, and silently degrades.**

`maxSupplementaryGroups()` falls back to `darwinNGroupsMax` on *any* error, including a bad OID name. If `"kern.ngroups"` were ever typo'd, the code would still behave correctly via the constant &mdash; so the bug would be invisible, and the entire runtime-read path would be dead with no test to notice. Neither `maxSupplementaryGroups()` nor the exported `ResolveGroups` has any coverage; every test calls the unexported `resolveGroups` with an explicit `maxGroups`. Nothing asserts the cap is actually applied on darwin, which is the whole point of the PR.

CI already runs `macos-arm64`, so this is cheap:

```go
// pkg/utils/ngroups_darwin_test.go  (implicit darwin constraint)
func TestMaxSupplementaryGroups_ReadsSysctl(t *testing.T) {
	n, err := unix.SysctlUint32("kern.ngroups")
	require.NoError(t, err, "kern.ngroups OID must be readable; a typo here would hide behind the fallback")
	assert.Equal(t, int(n), maxSupplementaryGroups())
}
```

plus a `!darwin` counterpart asserting `maxSupplementaryGroups() == 0`.

**2. `pkg/utils/privilege_unix.go:41-43` &mdash; truncation is silent.**

When the cap drops groups, there is no log line at any level. An operator debugging "why can't this Websh session read `/X`" on macOS gets no signal that 5 groups were discarded. For a product whose stated promise is that every action is supervised and audited, a demotion path that silently narrows a credential should say so. The package already imports zerolog.

### Suggestions (4)

**1. `pkg/utils/ngroups_darwin.go:15` &mdash; the doc comment is factually wrong, and that unblocks a free optimization.**

The comment calls `kern.ngroups` "the runtime-tunable limit." It isn't:

```
$ sysctl -w kern.ngroups=15
sysctl: oid 'kern.ngroups' is read only
```

It's derived from the compile-time kernel `NGROUPS_MAX`. Fix the wording ("the limit the kernel reports and enforces"). Because it cannot change during the process lifetime, caching is unambiguously safe &mdash; and `ResolveGroups` is on the per-command path (`pkg/executor/executor.go:489` demotes on *every* command execution, not just per session), so it currently does a `sysctlbyname` per exec:

```go
var maxSupplementaryGroups = sync.OnceValue(func() int {
	if n, err := unix.SysctlUint32("kern.ngroups"); err == nil && n > 0 {
		return int(n)
	}
	return darwinNGroupsMax
})
```

Honest sizing: the win is roughly a microsecond against a `fork`/`execve`, so this is about removing a pointless repeated syscall, not about measurable throughput. Worth taking mainly because the read-only finding makes it cost-free.

**2. `pkg/runner/pty_unix.go:27-33` &mdash; the duplication this PR targets is only half removed.**

Group *resolution* is now shared, but `syscall.SysProcAttr{Credential: &syscall.Credential{Uid, Gid, Groups}}` is still assembled independently there and at `privilege_unix.go:112-118`. One helper collapses both and makes it structurally impossible for the two demotion paths to drift again:

```go
// pkg/utils/privilege_unix.go
func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (*syscall.SysProcAttr, bool, error)
```

`Demote` then returns `result.SysProcAttr` from it, and `pty_unix.go` drops to a single call. Arguably scope creep for a bug fix &mdash; reasonable to defer, but it is the remaining half of the stated goal.

**3. `pkg/utils/privilege_unix_test.go` &mdash; not table-driven, and it drops edge cases the deleted tests covered.**

The golang stack rules call for table-driven tests; the deleted `utils_test.go` was table-driven and the four replacements are not. More substantively, the old table covered `-1`, `4294967296`, `""`, and `math.MaxUint32`. Those inputs changed semantics in this PR &mdash; from *silently skipped* to *hard error* &mdash; and nothing now asserts the new behavior. That is the exact class of change that deserves a test. Restoring them as rows in one table costs a few lines:

```go
tests := []struct {
	name      string
	gid       uint32
	groupIds  []string
	maxGroups int
	want      []uint32
	wantInList, wantErr bool
}{
	{name: "negative gid rejected",  gid: 20, groupIds: []string{"1000", "-1"},         wantErr: true},
	{name: "overflow gid rejected",  gid: 20, groupIds: []string{"4294967296"},         wantErr: true},
	{name: "empty string rejected",  gid: 20, groupIds: []string{""},                   wantErr: true},
	{name: "max uint32 accepted",    gid: 20, groupIds: []string{"4294967295"},         want: []uint32{20, math.MaxUint32}},
	{name: "exactly at cap",         gid: 99, groupIds: []string{"1", "2"}, maxGroups: 3, want: []uint32{99, 1, 2}},
	{name: "cap of 1 keeps primary", gid: 99, groupIds: []string{"1", "2"}, maxGroups: 1, want: []uint32{99}},
	// ... existing four cases
}
```

The `exactly at cap` and `cap of 1` rows are worth adding on their own &mdash; the current suite only exercises the over-cap branch, so an off-by-one in the `>` comparison would pass.

**4. `pkg/utils/privilege_unix.go:33` &mdash; error lacks domain context.**

`resolveGroups` returns the bare `strconv.ParseUint` error. The PTY caller wraps it ("failed to set process credentials"), but `Demote` does not, so `pkg/executor/executor.go:489` surfaces `strconv.ParseUint: parsing "x": invalid syntax` with no indication it concerns supplementary groups. In fairness, the surrounding `Demote` body returns bare strconv errors too, so this is a pre-existing pattern rather than something the PR introduced.

### Minor notes (non-actionable)

- `pkg/utils/ngroups_other.go` (`//go:build !darwin`) compiles `maxSupplementaryGroups` on Windows, where nothing calls it &mdash; `privilege_windows.go` has no group handling. Harmless: `golangci-lint` runs only under GOOS=linux, so `unused` never sees it, and CI confirms the Windows build is clean. Leaving it as-is is the right call; `!darwin && !windows` would be more precise but breaks the moment Windows demotion lands.
- Deleting `utils_test.go` leaves `pkg/utils/utils.go` (`Quote`, `GetEnvOrDefault`, ...) with no test file. Not a regression from this PR &mdash; that file only ever contained `TestConvertGroupIds`, and those functions were untested before. Worth a separate ticket, not a change here.
- Commit scope is `runner`, but the substance lives in `pkg/utils`. Defensible since the user-visible symptom is a Websh session failure. Not worth a rewrite.

### Highlights

- **The build-tag split is correct, and Copilot's objection is a false positive.** `_darwin.go` is an implicit GOOS constraint; the explicit `//go:build darwin` Copilot asked for would be redundant. Confirmed independently via `go list -f '{{.GoFiles}}'` across all three GOOS values. @jisung-02's rebuttal on that thread is accurate and well-evidenced.
- **`groupInList` is computed before truncation.** This is the subtle one, and getting it backwards would have made `ValidateGroup` reject legitimate group membership on macOS whenever the truncated entry happened to be the requested group. Deliberately handled and directly tested.
- **Primary-gid-first is the right invariant** for making truncation safe, and it is stated in the doc comment *and* pinned by `TestResolveGroups_TruncatesToMaxKeepingPrimaryGid`.
- **The PR body is a model of what a security-adjacent fix should look like** &mdash; an explicit "Safety" section reasoning about escalation, plus real end-to-end verification on macOS 26.5 rather than test output alone.

---

**Verdict**: **APPROVED** &mdash; no blocking issues. Both warnings are hardening (observability + closing the untested sysctl path), not correctness.

**Checklist**:
- [x] Code follows stack conventions &mdash; except table-driven tests (Suggestion 3)
- [x] Tests included and passing &mdash; coverage gaps noted in Warning 1 / Suggestion 3
- [x] No security vulnerabilities &mdash; escalation and fail-closed claims independently verified
- [x] Documentation updated if needed &mdash; one inaccurate code comment (Suggestion 1); no user-facing docs required

pkg/utils/ngroups_darwin.go
jisung-02: Fixed in 8c4d976. Added `pkg/utils/ngroups_darwin_test.go` with the assertion you suggested, so a misspelled OID fails the test instead of silently falling back to the constant, plus the `!darwin` counterpart in `ngroups_other_test.go` pinning the uncapped 0.

Also added `TestResolveGroups_AppliesPlatformCap`, which drives the exported `ResolveGroups` with 64 groups and asserts the resulting length against `maxSupplementaryGroups()`. That closes the gap you pointed at: every other test called the unexported helper with an explicit `maxGroups`, so nothing asserted the cap was applied at all.

pkg/utils/privilege_unix.go
jisung-02: Fixed in f07af42, taking your suggestion as written. Debug level rather than anything higher because `ResolveGroups` sits on the per-command path at `executor.go:489`, so a warn would fire on every exec; on macOS the truncation is the expected outcome rather than an anomaly, since `opendirectoryd` resolves membership dynamically.

Confirmed it actually fires — from the test run:

```
{"level":"debug","requested":65,"cap":16,"message":"Supplementary group list truncated to the platform setgroups limit."}
```

pkg/utils/ngroups_darwin.go
jisung-02: Both fixed in 78cd7ea. Reproduced the read-only property before changing anything:

```
$ sysctl kern.ngroups
kern.ngroups: 16
$ sysctl -w kern.ngroups=15
sysctl: oid 'kern.ngroups' is read only
```

Comment now says "the limit the kernel reports and enforces" and states why the value is stable, and the read is wrapped in `sync.OnceValue`.

One deviation: `ngroups_other.go` stays a plain function. Wrapping a constant 0 in `OnceValue` would add synchronization for nothing, and callers are identical either way, so the asymmetry in the declaration seemed the lesser cost. Happy to make them symmetric if you would rather.

Also updated the PR body, which repeated the same "runtime-tunable" claim.

pkg/utils/privilege_unix.go
jisung-02: Fixed in f07af42, as suggested: `fmt.Errorf("invalid supplementary group id: %w", err)`. `%w` keeps the original `strconv` error unwrappable for `errors.Is`/`As`, and the PTY caller wrapping it again only nests the message rather than losing anything.

pkg/utils/privilege_unix_test.go
jisung-02: Fixed in 8c4d976. Folded the four tests into one table and restored the edge cases from the deleted `ConvertGroupIds` table: `-1`, `4294967296`, `""` and `math.MaxUint32`. You were right that this was the gap that mattered most, since those inputs went from silently skipped to a hard error in this PR with nothing asserting the new behavior.

Added the `exactly at cap` and `cap of one` rows too. Also strengthened the not-a-member case, which previously only checked `groups[0]`, to compare the whole slice.

pkg/runner/pty_unix.go
jisung-02: Agreed, and split out as #386 rather than done here.

While scoping it I found the drift is wider than the two sites: `pkg/runner/tunnel_linux.go:181` `getCodeServerCredential` assembles a third `Credential` using its own `parseGroupIDs` at `:189`, which does not prepend the primary gid, never calls `maxSupplementaryGroups()`, skips unparsable gids with `continue` (the behavior this PR just changed to a hard error), and swallows a `usr.GroupIds()` failure by returning `nil`. It is `_linux.go`, so it is unaffected by the macOS symptom, but it is a third path in the same sense.

That makes the consolidation touch code-server session privileges, which needs its own test coverage, so it seemed clearly outside a bug fix. Details in #386.

geunwoonoh (approved):

PR #381, opened 2026-07-24
Harden Unix process-tree cleanup against a Start/cancel race

jisung-02:
## Background

While looking into an intermittent CI failure in the `executor` package, a couple of small timing-related rough edges surfaced around the Unix process-tree cleanup path. None of them affect the happy path, but they are worth smoothing out so the behavior is predictable under load and during shutdown.

The CI failure itself came from a test helper occasionally reading a half-written pid file. Alongside that, the race detector pointed at unsynchronized access to the cleanup state, and review found that the fix for the resulting early-cancel window was a no-op on darwin.

## Changes

Guard the process-group state against a Start/cancel race. `commandCleanup.pgid` was written by `afterStart` on the main goroutine and read by `cancel` from the context-watcher goroutine that `os/exec` invokes via `cmd.Cancel`, with no synchronization between them. Both fields are now serialized behind a mutex through small `recordPgid`/`takeForCancel` helpers, keeping the lock scoped to the field access so the `SIGKILL` itself stays lock-free. This mirrors the locking discipline already used in the Windows sibling.

Close the narrow early-cancel window the synchronization exposes. If a cancel fires before the group is recorded, it can only reach the leader; `afterStart` now redoes the kill once the group is known.

Record the process group by construction rather than by lookup. `afterStart` used to ask `syscall.Getpgid` for the group, which is unreliable exactly when the redo needs it: on a signalled-but-unreaped leader it succeeds on Linux but returns `ESRCH` on darwin, where XNU's `proc_find` skips `SZOMB`. The redo then recorded `pgid` as 0, targeted the already-killed leader instead of the group, and let backgrounded descendants survive—silently, since `kill` on the zombie returns `nil`. Since `configureProcessTreeCleanup` requests `setsid` or `setpgid(0,0)` itself, PGID == PID is guaranteed by construction, so that is recorded as `leadsGroup` and used directly. The `Getpgid == pid` guard remains for the only case it protects: a caller that pre-set `Pgid` to an existing group. This also keeps the unconditional deferred cancel from targeting a reaped pid, and removes one syscall per command execution.

Report the real termination status when the redo meets an already-gone group. Previously the redo could surface `os.ErrProcessDone` as an `afterStart` failure, which led `Execute` to report exit 1 with `os: process already finished` under a non-timeout cancellation such as a graceful shutdown. An already-gone group is treated as the benign outcome it is—consistent with the unconditional deferred cancel—so the normal `Wait` path reports the actual status; only a genuine kill error still surfaces.

Make the test pid file publish atomically. The affected test scripts wrote the backgrounded child pid with `echo $! > "$1"`, which truncate-opens the file before writing, leaving a brief window where a reader sees it empty. They now write to a temp file and rename it into place; a same-directory `rename(2)` is atomic, so the reader sees the file either absent or fully written.

## Safety

The changes are confined to the Unix cleanup path and its tests. The `Getpgid == pid` guard that keeps `-pgid` from targeting an unrelated group is unchanged in the case it applies to, and the redo remains idempotent alongside the existing deferred cancel. The pid file change is test-only.

## Testing

- `go test -race ./pkg/executor` across repeated runs, including the process-group and timeout cleanup tests, with no race reports.
- The original flaky read no longer reproduces without the race detector over many iterations.
- `TestCommandCleanup_AfterStartRedoAfterRacedCancel` asserts the redo reaches the whole group: it waits for the leader to reach the signalled-but-unreaped state, checks the backgrounded grandchild survived the raced cancel, and then requires the redo to kill it. Verified on darwin that the test fails when the `leadsGroup` branch is disabled and passes with it.
- `TestCommandCleanup_AfterStartRedoSwallowsProcessDone` pins the `os.ErrProcessDone` branch, which `runCommand` cannot reach because `afterStart` always precedes `Wait`.
- `gofmt -l`, `go vet` on linux and darwin, and a Windows cross-build all pass.

## Checklist

- [x] `go vet` clean
- [x] Tests pass, including `-race`
- [x] Windows cross-build verified
- [x] No changes outside the Unix cleanup path and its tests

### DIFF
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -352,8 +352,8 @@ func (e *Executor) runCommand(cmd *exec.Cmd, pidHook func(pid int), cw *chunkWri
 		}
 	}()
 	if err := cleanup.afterStart(cmd); err != nil {
-		// Windows afterStart may have already canceled on a Start/cancel race; the deferred cancel
-		// re-runs harmlessly (idempotent). Unix afterStart never errors here.
+		// afterStart may re-run cancel on a Start/cancel race (both platforms) and surface its error
+		// here; the deferred cancel then re-runs harmlessly (idempotent).
 		_ = cmd.Wait()
 		return finish(err)
 	}
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -352,7 +352,7 @@ func (e *Executor) runCommand(cmd *exec.Cmd, pidHook func(pid int), cw *chunkWri
 		}
 	}()
 	if err := cleanup.afterStart(cmd); err != nil {
-		// afterStart may re-run cancel on a Start/cancel race (both platforms) and surface its error
+		// afterStart may re-run cancel on a Start/cancel race (unix-like and windows alike) and surface its error
 		// here; the deferred cancel then re-runs harmlessly (idempotent).
 		_ = cmd.Wait()
 		return finish(err)
--- a/pkg/executor/process_tree_unix.go
+++ b/pkg/executor/process_tree_unix.go
@@ -6,13 +6,16 @@ import (
 	"errors"
 	"os"
 	"os/exec"
+	"sync"
 	"syscall"
 )
 
 // commandCleanup mirrors the type in process_tree_windows.go; the commandCleaner assertion below pins
 // the shared afterStart/cancel/close method set across build tags. State differs per platform.
 type commandCleanup struct {
-	pgid int // process group to SIGKILL; 0 when the child does not lead its own group
+	mu       sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
+	pgid     int        // process group to SIGKILL; 0 when the child does not lead its own group
+	canceled bool       // a cancel already fired; afterStart re-runs it once the group is recorded
 }
 
 var _ commandCleaner = (*commandCleanup)(nil)
@@ -41,10 +44,22 @@ func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
 	if cmd.Process == nil {
 		return os.ErrProcessDone
 	}
-	if pgid, err := syscall.Getpgid(cmd.Process.Pid); err == nil && pgid == cmd.Process.Pid {
-		c.pgid = pgid
+	pgid := 0
+	if p, err := syscall.Getpgid(cmd.Process.Pid); err == nil && p == cmd.Process.Pid {
+		pgid = p
+	}
+	// A cancel that fired between Start and here read pgid==0 and hit only the leader; redo it with the group recorded.
+	if canceled := c.recordPgid(pgid); canceled {
+		return c.cancel(cmd)
 	}
 	return nil
+}
+
+func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.pgid = pgid
+	return c.canceled
 }
 
 func (c *commandCleanup) cancel(cmd *exec.Cmd) error {
@@ -52,8 +67,8 @@ func (c *commandCleanup) cancel(cmd *exec.Cmd) error {
 		return os.ErrProcessDone
 	}
 	pid := cmd.Process.Pid
-	if c.pgid != 0 {
-		pid = -c.pgid
+	if pgid := c.markCanceled(); pgid != 0 {
+		pid = -pgid
 	}
 	if err := syscall.Kill(pid, syscall.SIGKILL); err != nil {
 		if errors.Is(err, syscall.ESRCH) {
@@ -62,6 +77,15 @@ func (c *commandCleanup) cancel(cmd *exec.Cmd) error {
 		return err
 	}
 	return nil
+}
+
+// markCanceled records that a cancel fired and returns the recorded group; afterStart uses the flag to
+// redo a kill that raced ahead of the group being recorded.
+func (c *commandCleanup) markCanceled() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.canceled = true
+	return c.pgid
 }
 
 func (c *commandCleanup) close() error {
--- a/pkg/executor/process_tree_unix.go
+++ b/pkg/executor/process_tree_unix.go
@@ -48,9 +48,13 @@ func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
 	if p, err := syscall.Getpgid(cmd.Process.Pid); err == nil && p == cmd.Process.Pid {
 		pgid = p
 	}
-	// A cancel that fired between Start and here read pgid==0 and hit only the leader; redo it with the group recorded.
+	// A cancel that fired between Start and here read pgid==0 and hit only the leader; redo it with the
+	// group recorded. An already-gone group (ErrProcessDone) is the expected benign outcome, so let the
+	// normal Wait path report the real termination status rather than surfacing it as an afterStart failure.
 	if canceled := c.recordPgid(pgid); canceled {
-		return c.cancel(cmd)
+		if err := c.cancel(cmd); err != nil && !errors.Is(err, os.ErrProcessDone) {
+			return err
+		}
 	}
 	return nil
 }
--- a/pkg/executor/process_tree_unix.go
+++ b/pkg/executor/process_tree_unix.go
@@ -59,13 +59,6 @@ func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
 	return nil
 }
 
-func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	c.pgid = pgid
-	return c.canceled
-}
-
 func (c *commandCleanup) cancel(cmd *exec.Cmd) error {
 	if cmd.Process == nil {
 		return os.ErrProcessDone
@@ -83,6 +76,17 @@ func (c *commandCleanup) cancel(cmd *exec.Cmd) error {
 	return nil
 }
 
+func (c *commandCleanup) close() error {
+	return nil
+}
+
+func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.pgid = pgid
+	return c.canceled
+}
+
 // markCanceled records that a cancel fired and returns the recorded group; afterStart uses the flag to
 // redo a kill that raced ahead of the group being recorded.
 func (c *commandCleanup) markCanceled() int {
@@ -90,8 +94,4 @@ func (c *commandCleanup) markCanceled() int {
 	defer c.mu.Unlock()
 	c.canceled = true
 	return c.pgid
-}
-
-func (c *commandCleanup) close() error {
-	return nil
 }
--- a/pkg/executor/process_tree_unix.go
+++ b/pkg/executor/process_tree_unix.go
@@ -13,10 +13,11 @@ import (
 // commandCleanup mirrors the type in process_tree_windows.go; the commandCleaner assertion below pins
 // the shared afterStart/cancel/close method set across build tags. State differs per platform.
 type commandCleanup struct {
-	mu         sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
-	leadsGroup bool       // configure requested setsid/setpgid(0,0), so PGID == PID by construction
-	pgid       int        // process group to SIGKILL; 0 when the child does not lead its own group
-	canceled   bool       // a cancel already fired; afterStart re-runs it once the group is recorded
+	leadsGroup bool // immutable after configure: setsid/setpgid(0,0) was requested, so PGID == PID
+
+	mu       sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
+	pgid     int        // process group to SIGKILL; 0 when the child does not lead its own group
+	canceled bool       // a cancel already fired; afterStart re-runs it once the group is recorded
 }
 
 var _ commandCleaner = (*commandCleanup)(nil)
--- a/pkg/executor/process_tree_unix.go
+++ b/pkg/executor/process_tree_unix.go
@@ -95,8 +95,9 @@ func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
 	return c.canceled
 }
 
-// Mirrors the Windows sibling: records that a cancel fired and hands back the group, so cancel's SIGKILL
-// runs lock-free. afterStart uses the flag to redo a kill that raced ahead of the group being recorded.
+// takeForCancel records that a cancel fired and hands back the group, so cancel's SIGKILL runs lock-free;
+// it mirrors the Windows sibling of the same name. afterStart uses the flag to redo a kill that raced
+// ahead of the group being recorded.
 func (c *commandCleanup) takeForCancel() int {
 	c.mu.Lock()
 	defer c.mu.Unlock()
--- a/pkg/executor/process_tree_unix.go
+++ b/pkg/executor/process_tree_unix.go
@@ -13,9 +13,10 @@ import (
 // commandCleanup mirrors the type in process_tree_windows.go; the commandCleaner assertion below pins
 // the shared afterStart/cancel/close method set across build tags. State differs per platform.
 type commandCleanup struct {
-	mu       sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
-	pgid     int        // process group to SIGKILL; 0 when the child does not lead its own group
-	canceled bool       // a cancel already fired; afterStart re-runs it once the group is recorded
+	mu         sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
+	leadsGroup bool       // configure requested setsid/setpgid(0,0), so PGID == PID by construction
+	pgid       int        // process group to SIGKILL; 0 when the child does not lead its own group
+	canceled   bool       // a cancel already fired; afterStart re-runs it once the group is recorded
 }
 
 var _ commandCleaner = (*commandCleanup)(nil)
@@ -34,18 +35,23 @@ func configureProcessTreeCleanup(cmd *exec.Cmd, sessionLeader bool) (*commandCle
 		cmd.SysProcAttr.Setpgid = true
 	}
 
-	return &commandCleanup{}, nil
+	sysAttr := cmd.SysProcAttr
+	return &commandCleanup{leadsGroup: sysAttr.Setsid || (sysAttr.Setpgid && sysAttr.Pgid == 0)}, nil
 }
 
-// afterStart captures the process group while the leader is alive: getpgid fails with ESRCH once Wait
-// reaps it, yet -pgid stays killable while a descendant holds the group open (the post-reap leak path).
-// Recorded only when the child leads its own group (PGID == PID), so -pgid can't hit an unrelated group.
+// afterStart records the process group cancel should target: -pgid stays killable while a descendant holds
+// the group open even after Wait reaps the leader (the post-reap leak path). Recorded only when the child
+// leads its own group (PGID == PID), so -pgid can't hit an unrelated group.
 func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
 	if cmd.Process == nil {
 		return os.ErrProcessDone
 	}
 	pgid := 0
-	if p, err := syscall.Getpgid(cmd.Process.Pid); err == nil && p == cmd.Process.Pid {
+	if c.leadsGroup {
+		// Don't ask getpgid: on darwin it returns ESRCH for a signalled-but-unreaped leader, which would
+		// drop the group in exactly the raced cancel handled below.
+		pgid = cmd.Process.Pid
+	} else if p, err := syscall.Getpgid(cmd.Process.Pid); err == nil && p == cmd.Process.Pid {
 		pgid = p
 	}
 	// A cancel that fired between Start and here read pgid==0 and hit only the leader; redo it with the
--- a/pkg/executor/process_tree_unix.go
+++ b/pkg/executor/process_tree_unix.go
@@ -70,7 +70,7 @@ func (c *commandCleanup) cancel(cmd *exec.Cmd) error {
 		return os.ErrProcessDone
 	}
 	pid := cmd.Process.Pid
-	if pgid := c.markCanceled(); pgid != 0 {
+	if pgid := c.takeForCancel(); pgid != 0 {
 		pid = -pgid
 	}
 	if err := syscall.Kill(pid, syscall.SIGKILL); err != nil {
@@ -86,6 +86,8 @@ func (c *commandCleanup) close() error {
 	return nil
 }
 
+// recordPgid stores the group cancel should target and reports whether a cancel already fired, which is
+// what tells afterStart to redo the kill.
 func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
@@ -93,9 +95,9 @@ func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
 	return c.canceled
 }
 
-// markCanceled records that a cancel fired and returns the recorded group; afterStart uses the flag to
-// redo a kill that raced ahead of the group being recorded.
-func (c *commandCleanup) markCanceled() int {
+// Mirrors the Windows sibling: records that a cancel fired and hands back the group, so cancel's SIGKILL
+// runs lock-free. afterStart uses the flag to redo a kill that raced ahead of the group being recorded.
+func (c *commandCleanup) takeForCancel() int {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 	c.canceled = true
--- a/pkg/executor/process_tree_unix_test.go
+++ b/pkg/executor/process_tree_unix_test.go
@@ -85,7 +85,7 @@ func TestCommandCleanup_CancelNilProcessReturnsProcessDone(t *testing.T) {
 // the whole group, so a backgrounded grandchild holding the pipe open dies too.
 func TestCommandCleanup_CancelKillsProcessGroup(t *testing.T) {
 	pidFile := filepath.Join(t.TempDir(), "child.pid")
-	cmd := exec.Command("/bin/sh", "-c", "sleep 600 & echo $! > \"$1\"; wait", "sh", pidFile)
+	cmd := exec.Command("/bin/sh", "-c", `sleep 600 & echo $! > "$1.tmp"; mv "$1.tmp" "$1"; wait`, "sh", pidFile)
 	cleanup, err := configureProcessTreeCleanup(cmd, false)
 	if err != nil {
 		t.Fatalf("configureProcessTreeCleanup: %v", err)
@@ -127,7 +127,7 @@ func TestExecutor_TimeoutCleansProcessTreeWhenChildKeepsPipeOpen(t *testing.T) {
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			pidFile := filepath.Join(t.TempDir(), "child.pid")
-			script := "sleep 600 & echo $! > \"$1\"; wait"
+			script := `sleep 600 & echo $! > "$1.tmp"; mv "$1.tmp" "$1"; wait`
 
 			type result struct {
 				exitCode int
@@ -178,7 +178,7 @@ func TestExecutor_TimeoutCleansProcessTreeWhenChildKeepsPipeOpen(t *testing.T) {
 // timeout means no Cancel, and Wait returns ExitError not ErrWaitDelay, so only the unconditional cancel covers it.
 func TestExecutor_CleansDescendantWhenCommandExitsNonZero(t *testing.T) {
 	pidFile := filepath.Join(t.TempDir(), "child.pid")
-	script := "sleep 600 & echo $! > \"$1\"; exit 3"
+	script := `sleep 600 & echo $! > "$1.tmp"; mv "$1.tmp" "$1"; exit 3`
 
 	type result struct {
 		exitCode int
--- a/pkg/executor/process_tree_unix_test.go
+++ b/pkg/executor/process_tree_unix_test.go
@@ -81,10 +81,11 @@ func TestCommandCleanup_CancelNilProcessReturnsProcessDone(t *testing.T) {
 	}
 }
 
-// When a cancel races ahead of the process group being recorded, afterStart redoes the kill. An
-// already-gone group is the benign expected outcome, so the redo must not surface os.ErrProcessDone
-// as an afterStart failure—otherwise Execute reports exit 1 instead of the real termination status.
-func TestCommandCleanup_AfterStartRedoSwallowsProcessDone(t *testing.T) {
+// When a cancel races ahead of the process group being recorded, afterStart redoes the kill. runCommand
+// always runs afterStart before Wait, so the redo targets a still-unreaped process—never a reaped pid the
+// OS could have handed to an unrelated process. Exercise that exact order: the redo must complete without
+// surfacing an error, so Execute reports the real termination status rather than exit 1. Reap only after.
+func TestCommandCleanup_AfterStartRedoAfterRacedCancel(t *testing.T) {
 	cmd := exec.Command("/bin/sh", "-c", "exit 0")
 	cleanup, err := configureProcessTreeCleanup(cmd, false)
 	if err != nil {
@@ -93,13 +94,13 @@ func TestCommandCleanup_AfterStartRedoSwallowsProcessDone(t *testing.T) {
 	if err := cmd.Start(); err != nil {
 		t.Fatalf("cmd.Start: %v", err)
 	}
-	// Simulate a cancel that fired between Start and afterStart, then reap so the redo hits ESRCH.
+	// Cancel before the group is recorded: pgid is still 0, so only the leader is targeted.
 	_ = cleanup.cancel(cmd)
-	_ = cmd.Wait()
-
+	// Redo while the process is still unreaped, matching runCommand's afterStart-before-Wait order.
 	if err := cleanup.afterStart(cmd); err != nil {
-		t.Fatalf("afterStart redo surfaced %v, want nil for an already-gone group", err)
+		t.Fatalf("afterStart redo surfaced %v, want nil after a raced cancel", err)
 	}
+	_ = cmd.Wait()
 }
 
 // White-box counterpart to the Windows job-object test: configure -> Start -> cancel must SIGKILL
--- a/pkg/executor/process_tree_unix_test.go
+++ b/pkg/executor/process_tree_unix_test.go
@@ -81,6 +81,27 @@ func TestCommandCleanup_CancelNilProcessReturnsProcessDone(t *testing.T) {
 	}
 }
 
+// When a cancel races ahead of the process group being recorded, afterStart redoes the kill. An
+// already-gone group is the benign expected outcome, so the redo must not surface os.ErrProcessDone
+// as an afterStart failure—otherwise Execute reports exit 1 instead of the real termination status.
+func TestCommandCleanup_AfterStartRedoSwallowsProcessDone(t *testing.T) {
+	cmd := exec.Command("/bin/sh", "-c", "exit 0")
+	cleanup, err := configureProcessTreeCleanup(cmd, false)
+	if err != nil {
+		t.Fatalf("configureProcessTreeCleanup: %v", err)
+	}
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("cmd.Start: %v", err)
+	}
+	// Simulate a cancel that fired between Start and afterStart, then reap so the redo hits ESRCH.
+	_ = cleanup.cancel(cmd)
+	_ = cmd.Wait()
+
+	if err := cleanup.afterStart(cmd); err != nil {
+		t.Fatalf("afterStart redo surfaced %v, want nil for an already-gone group", err)
+	}
+}
+
 // White-box counterpart to the Windows job-object test: configure -> Start -> cancel must SIGKILL
 // the whole group, so a backgrounded grandchild holding the pipe open dies too.
 func TestCommandCleanup_CancelKillsProcessGroup(t *testing.T) {
--- a/pkg/executor/process_tree_unix_test.go
+++ b/pkg/executor/process_tree_unix_test.go
@@ -94,13 +94,16 @@ func TestCommandCleanup_AfterStartRedoAfterRacedCancel(t *testing.T) {
 	if err := cmd.Start(); err != nil {
 		t.Fatalf("cmd.Start: %v", err)
 	}
+	// Reap last—after afterStart, and even when an assertion below fails early.
+	t.Cleanup(func() { _ = cmd.Wait() })
 	// Cancel before the group is recorded: pgid is still 0, so only the leader is targeted.
-	_ = cleanup.cancel(cmd)
+	if err := cleanup.cancel(cmd); err != nil && !errors.Is(err, os.ErrProcessDone) {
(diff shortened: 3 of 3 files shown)


pkg/executor/process_tree_unix_test.go
```diff
@@ -81,11 +81,32 @@ func TestCommandCleanup_CancelNilProcessReturnsProcessDone(t *testing.T) {
...
+		t.Fatalf("cmd.Start: %v", err)
+	}
+	// Simulate a cancel that fired between Start and afterStart, then reap so the redo hits ESRCH.
+	_ = cleanup.cancel(cmd)
+	_ = cmd.Wait()
+
+	if err := cleanup.afterStart(cmd); err != nil {
+		t.Fatalf("afterStart redo surfaced %v, want nil for an already-gone group", err)
+	}
+}
```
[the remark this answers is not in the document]
jisung-02: Good catch, fixed. The redo now runs before `cmd.Wait()`, matching `runCommand`'s actual sequence (afterStart at executor.go:354, Wait at :366), so it only ever targets a still-unreaped pid the OS cannot have reassigned.

Worth noting for the record: the ErrProcessDone the redo swallows is only reachable once the process is fully reaped (post-reap `Getpgid` returns ESRCH; an unreaped zombie yields nil/EPERM on both Linux and macOS, never ESRCH), which is precisely the reused-pid window you flagged. Since that path can't be reproduced safely, the test now asserts the production-faithful ordering stays clean under a raced cancel instead. The `cancel` -> ErrProcessDone mapping itself stays covered by TestCommandCleanup_CancelNilProcessReturnsProcessDone.

pkg/executor/process_tree_unix_test.go
```diff
@@ -81,11 +81,33 @@ func TestCommandCleanup_CancelNilProcessReturnsProcessDone(t *testing.T) {
...
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("cmd.Start: %v", err)
+	}
+	// Cancel before the group is recorded: pgid is still 0, so only the leader is targeted.
+	_ = cleanup.cancel(cmd)
+	// Redo while the process is still unreaped, matching runCommand's afterStart-before-Wait order.
+	if err := cleanup.afterStart(cmd); err != nil {
+		t.Fatalf("afterStart redo surfaced %v, want nil after a raced cancel", err)
+	}
+	_ = cmd.Wait()
```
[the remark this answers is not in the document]
jisung-02: Fixed in a632ccb.

Both points were valid:
- The started leader is now reaped from a `t.Cleanup` registered right after `cmd.Start()`, so an early `t.Fatalf` can no longer skip it. Registering it there also keeps the reap after `afterStart`, which is the ordering the test exists to exercise.
- The early `cleanup.cancel(cmd)` error is now asserted, tolerating only `os.ErrProcessDone` (a leader that already exited unreaped is legitimate); anything else fails the test.

The same early-exit leak existed in `TestCommandCleanup_CancelKillsProcessGroup`, where a fatal before the in-body `cancel` left a live `sh` plus its `sleep 600` grandchild rather than just a zombie. Its `go func() { _ = cmd.Wait() }()` was replaced by a `t.Cleanup` that cancels and reaps; the in-body cancel makes it a no-op on the happy path.

pkg/executor/process_tree_unix.go
```diff
@@ -41,19 +44,35 @@ func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
 	if cmd.Process == nil {
 		return os.ErrProcessDone
 	}
-	if pgid, err := syscall.Getpgid(cmd.Process.Pid); err == nil && pgid == cmd.Process.Pid {
-		c.pgid = pgid
+	pgid := 0
+	if p, err := syscall.Getpgid(cmd.Process.Pid); err == nil && p == cmd.Process.Pid {
```
geunwoonoh: **Critical:** the redo is a no-op on darwin/BSD, so leaked descendants survive exactly the race this PR closes.

`syscall.Getpgid` on a signalled-but-unreaped leader succeeds on Linux (`find_task_by_vpid` still finds zombie tasks) but returns `ESRCH` on XNU (`proc_find` skips `SZOMB`). In the raced path on darwin:

- `Getpgid` fails, so `pgid` is recorded as **0**
- the redo calls `cancel`, which with `pgid == 0` targets `cmd.Process.Pid` — the leader the raced cancel already killed
- `kill` on that zombie returns `nil`, so `afterStart` returns `nil` and everything looks fine
- the backgrounded grandchild survives

Measured with a verbatim replica of this file, using the repo's own zombie-aware `processGone` check:

| | `Getpgid` on the killed-but-unreaped leader | redo kill target | grandchild after redo |
|---|---|---|---|
| linux (`golang:1.25`) | `(711, nil)` | `-711` (group) | dead |
| darwin (arm64) | `(-1, ESRCH)` | `40737` (leader only) | **alive** |

darwin is a shipped target (`.goreleaser.yaml` `alpamon-darwin`) with its own `macos-14` CI job.

Secondary consequence: with `pgid` stuck at 0, the unconditional deferred cancel at `executor.go:350` also targets `cmd.Process.Pid` — now a **reaped** pid — which is precisely the PID-reuse hazard the new test's comment says the design avoids.

### Suggested fix

Stop asking the kernel something we already know. `configureProcessTreeCleanup` guarantees `PGID == PID` by construction (`Setsid`, or `Setpgid` with `Pgid == 0`, which Go's fork path turns into `setpgid(0, 0)`). Record that at configure time:

```go
type commandCleanup struct {
	mu         sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
	leadsGroup bool       // configure requested setsid/setpgid(0,0), so PGID == PID by construction
	pgid       int        // process group to SIGKILL; 0 when the child does not lead its own group
	canceled   bool       // a cancel already fired; afterStart re-runs it once the group is recorded
}
```

```go
	// end of configureProcessTreeCleanup
	sysAttr := cmd.SysProcAttr
	return &commandCleanup{leadsGroup: sysAttr.Setsid || (sysAttr.Setpgid && sysAttr.Pgid == 0)}, nil
```

```go
	// in afterStart
	pgid := 0
	if c.leadsGroup {
		// Guaranteed by the setsid/setpgid(0,0) requested at configure time. Don't ask Getpgid:
		// on darwin it returns ESRCH for a signalled-but-unreaped leader, which would drop the
		// group in exactly the raced-cancel case handled below.
		pgid = cmd.Process.Pid
	} else if p, err := syscall.Getpgid(cmd.Process.Pid); err == nil && p == cmd.Process.Pid {
		pgid = p
	}
```

I tested this patch in the probe: darwin then records the group and the grandchild dies, Linux is unchanged. The `Getpgid == pid` guard is preserved for the case it actually protects — a caller that pre-set `Setpgid: true, Pgid: <existing>` keeps `leadsGroup == false`. Bonus: it removes one `Getpgid` syscall per command execution on the happy path.

pkg/executor/process_tree_unix_test.go
```diff
@@ -81,18 +81,49 @@ func TestCommandCleanup_CancelNilProcessReturnsProcessDone(t *testing.T) {
 	}
 }
 
+// When a cancel races ahead of the process group being recorded, afterStart redoes the kill. runCommand
+// always runs afterStart before Wait, so the redo targets a still-unreaped process—never a reaped pid the
+// OS could have handed to an unrelated process. Exercise that exact order: the redo must complete without
+// surfacing an error, so Execute reports the real termination status rather than exit 1. Reap only after.
+func TestCommandCleanup_AfterStartRedoAfterRacedCancel(t *testing.T) {
```
geunwoonoh, line 109: **Warning:** this test asserts that the redo did not error, not that it worked.

`/bin/sh -c "exit 0"` has no descendants, and the only assertion is `afterStart` returning `nil` — which cannot distinguish "redo killed the group" from "redo killed nothing". The Windows counterpart it is modeled on does assert the effect (`process_tree_windows_test.go:124-130`: *"process %d still running after afterStart re-cancel"*). That missing assertion is why the darwin hole above got through: this test passes on macOS CI today.

Reusing the sibling test's grandchild shape makes it deterministic — the leader stays unreaped, so nothing depends on timing:

```go
	pidFile := filepath.Join(t.TempDir(), "child.pid")
	cmd := exec.Command("/bin/sh", "-c", bgChildScript("wait"), "sh", pidFile)
	// ... configureProcessTreeCleanup, Start, t.Cleanup(reap)
	childPID := readExecutorTimeoutChildPID(t, pidFile)
	// Raced cancel: pgid is still 0, so only the leader is targeted and the grandchild survives.
	_ = cleanup.cancel(cmd)
	if err := cleanup.afterStart(cmd); err != nil {
		t.Fatalf("afterStart redo surfaced %v, want nil after a raced cancel", err)
	}
	if !waitForExecutorTimeoutChildExit(childPID, 3*time.Second) {
		t.Fatalf("grandchild %d survived the afterStart redo", childPID)
	}
```

pkg/executor/process_tree_unix.go
```diff
@@ -41,19 +44,35 @@ func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
...
+		pgid = p
+	}
+	// A cancel that fired between Start and here read pgid==0 and hit only the leader; redo it with the
```
geunwoonoh, line 58: **Warning:** this comment (and the `afterStart` doc at lines 40-42) states a Linux-only behavior as a `!windows` guarantee.

`"getpgid fails with ESRCH once Wait reaps it"` is wrong on darwin, where it fails as soon as the leader becomes a zombie — reaping is not the boundary. Same for `"redo it with the group recorded"` here: on darwin the group is *not* recorded in the raced path.

These read as verified invariants, so they should either name Linux explicitly or become true on both platforms via the fix above (preferred).

pkg/executor/process_tree_unix.go
```diff
@@ -41,19 +44,35 @@ func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
...
+	// group recorded. An already-gone group (ErrProcessDone) is the expected benign outcome, so let the
+	// normal Wait path report the real termination status rather than surfacing it as an afterStart failure.
+	if canceled := c.recordPgid(pgid); canceled {
+		if err := c.cancel(cmd); err != nil && !errors.Is(err, os.ErrProcessDone) {
```
geunwoonoh, line 62: **Warning:** this `os.ErrProcessDone` swallow — the point of commit 841a47b — is unreachable via `runCommand` and untested.

I measured both paths: on Linux the zombie leader keeps the group non-empty, so `kill(-pgid)` returns `nil`; on darwin the redo targets the zombie's positive pid and `kill` also returns `nil`. Since `afterStart` always runs before `Wait`, nothing reaches `ESRCH` here today — so `TestCommandCleanup_AfterStartRedoAfterRacedCancel` never exercises the branch it was named for.

Not harmful, and it becomes genuinely reachable once the `leadsGroup` fix lands (darwin's post-`Wait` `kill(-pgid)` on an emptied group does return `ESRCH`). But either add a direct unit test that sets `pgid` on an already-reaped `cmd`, or say in the comment that it is defensive.

pkg/executor/process_tree_unix_test.go
```diff
@@ -81,18 +81,49 @@ func TestCommandCleanup_CancelNilProcessReturnsProcessDone(t *testing.T) {
...
+		t.Fatalf("afterStart redo surfaced %v, want nil after a raced cancel", err)
+	}
+}
+
 // White-box counterpart to the Windows job-object test: configure -> Start -> cancel must SIGKILL
 // the whole group, so a backgrounded grandchild holding the pipe open dies too.
 func TestCommandCleanup_CancelKillsProcessGroup(t *testing.T) {
 	pidFile := filepath.Join(t.TempDir(), "child.pid")
-	cmd := exec.Command("/bin/sh", "-c", "sleep 600 & echo $! > \"$1\"; wait", "sh", pidFile)
+	cmd := exec.Command("/bin/sh", "-c", `sleep 600 & echo $! > "$1.tmp"; mv "$1.tmp" "$1"; wait`, "sh", pidFile)
```
geunwoonoh: **Suggestion:** the background-child script is now duplicated three times (here, line 160, line 211), tmp+rename dance included. One helper keeps the atomicity rationale in a single place:

```go
// bgChildScript backgrounds a long sleeper and publishes its pid atomically: echo truncate-opens the
// target, so write to a temp file and rename it in (a same-directory rename(2) is atomic).
func bgChildScript(tail string) string {
	return `sleep 600 & echo $! > "$1.tmp"; mv "$1.tmp" "$1"; ` + tail
}
```

Call sites become `bgChildScript("wait")` and `bgChildScript("exit 3")`.

```suggestion
	cmd := exec.Command("/bin/sh", "-c", bgChildScript("wait"), "sh", pidFile)
```

pkg/executor/process_tree_unix.go
```diff
@@ -41,19 +44,35 @@ func (c *commandCleanup) afterStart(cmd *exec.Cmd) error {
...
+			return err
+		}
 	}
 	return nil
 }
 
+func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
```
geunwoonoh: **Suggestion:** `recordPgid` has no doc comment, while its sibling `markCanceled` (line 86) and the Windows `assignForStart` both do. It is the half that needs one more: the name says "record", but the return value is what drives the redo.

```suggestion
// recordPgid stores the group cancel should target and reports whether a cancel already fired, which is
// what tells afterStart to redo the kill.
func (c *commandCleanup) recordPgid(pgid int) (canceled bool) {
```

pkg/executor/process_tree_unix.go
```diff
@@ -64,6 +83,15 @@ func (c *commandCleanup) cancel(cmd *exec.Cmd) error {
 	return nil
 }
 
+// markCanceled records that a cancel fired and returns the recorded group; afterStart uses the flag to
+// redo a kill that raced ahead of the group being recorded.
+func (c *commandCleanup) markCanceled() int {
```
geunwoonoh: **Suggestion:** `markCanceled` mixes command and query — it marks *and* returns the pgid. The Windows sibling names this exact shape `takeForCancel`, and this PR explicitly positions itself as mirroring that locking discipline, so matching the name would let the two files diff cleanly against each other. (Renaming also touches the call site at line 74.)

geunwoonoh (changes requested):
## Review summary

Three things bundled here: (1) serialize `commandCleanup.pgid`/`canceled` behind a mutex to kill a genuine data race between `afterStart` (main goroutine) and `cmd.Cancel`'s context watcher; (2) redo the `SIGKILL` from `afterStart` when a cancel raced ahead of the group being recorded, treating an already-gone group as benign so `Execute` reports the real exit status; (3) make the test pid file publish atomically via same-directory rename.

The race diagnosis is correct and not speculative — `os/exec` starts `watchCtx` at the end of `Start`, so `Cancel` really can fire concurrently with `afterStart`. The mutex discipline (lock scoped to field access, `SIGKILL` outside it) is right, and it does mirror the Windows sibling.

**But the redo — the actual behavioral fix — is a no-op on darwin.** I verified this empirically on both platforms.

## Verification I ran

```
gofmt -l pkg/executor/          -> clean
go vet ./pkg/executor           -> clean (GOOS=linux/darwin/windows)
go build ./pkg/executor/...     -> OK   (GOOS=linux/darwin/windows)
go test -race -count=2 ./pkg/executor (process-tree tests)  -> ok  7.490s
```

Then I replicated `process_tree_unix.go` verbatim in a standalone probe and ran the exact raced-cancel scenario the redo exists for (leader + backgrounded `sleep 600` grandchild, cancel before the group is recorded, then `afterStart`), using this repo's own zombie-aware `processGone` liveness check:

| | `Getpgid` on the killed-but-unreaped leader | redo kill target | grandchild after redo |
|---|---|---|---|
| **linux** (`golang:1.25` container) | `(711, nil)` OK | `-711` (group) | **dead** OK |
| **darwin** (local, arm64) | `(-1, ESRCH)` FAIL | `40737` (leader only) | **alive** FAIL |

## Critical issues (1)

**1. `pkg/executor/process_tree_unix.go:48` — the redo silently drops the process group on darwin/BSD, so leaked descendants survive exactly the race this PR closes.**

`syscall.Getpgid` on a signalled-but-unreaped leader succeeds on Linux (`find_task_by_vpid` still finds zombie tasks) but returns `ESRCH` on XNU (`proc_find` skips `SZOMB`). So in the raced path on darwin: `Getpgid` fails, `pgid` is recorded as **0**, the redo targets `cmd.Process.Pid` (the leader the raced cancel already killed), `kill` on that zombie returns `nil` so `afterStart` returns `nil` and everything looks fine — and the backgrounded grandchild survives.

The PR description's central claim, *"`Getpgid` on a signalled-but-unreaped leader still returns the correct group, so the redo reliably reaches the surviving members"*, holds only on Linux. darwin is a shipped target (`.goreleaser.yaml` `alpamon-darwin`, amd64+arm64) with its own `macos-14` CI job, so this is not a hypothetical platform.

Secondary consequence: with `pgid` stuck at 0, the unconditional deferred cancel at `pkg/executor/executor.go:350` also targets `cmd.Process.Pid` — now a **reaped** pid — which is precisely the PID-reuse hazard the new test's comment says the design avoids.

See the inline comment for the suggested fix (record `leadsGroup` at configure time instead of asking `Getpgid`). I tested that patch in the probe: darwin then records the group and the grandchild dies; Linux is unchanged. It also removes one syscall per command execution on the happy path.

## Warnings (3)

**2. `process_tree_unix_test.go:88` — the new test asserts the redo didn't error, not that it worked.** Its Windows counterpart (`process_tree_windows_test.go:124-130`) does assert the effect. This missing assertion is why #1 got through. Inline comment has a deterministic replacement.

**3. `process_tree_unix.go:40-42, 51-53` — comments assert Linux-only behavior as a `!windows` guarantee.** `"getpgid fails with ESRCH once Wait reaps it"` is wrong on darwin, where it fails as soon as the leader is a zombie — reaping isn't the boundary.

**4. `process_tree_unix.go:55` — the `os.ErrProcessDone` swallow (commit 841a47b) is unreachable via `runCommand` and untested.** I measured both paths: on Linux the zombie leader keeps the group non-empty so `kill(-pgid)` returns `nil`; on darwin the redo targets the zombie's positive pid and `kill` also returns `nil`. Since `afterStart` always runs before `Wait`, no path through `runCommand` reaches `ESRCH` today, so the test named for this fix never exercises the branch it was added for.

## Suggestions (3)

**5. `process_tree_unix_test.go:113, 160, 211`** — the background-child script is now duplicated three times, tmp+rename dance included. Extract a `bgChildScript(tail string)` helper so the atomicity rationale lives in one place.

**6. `process_tree_unix.go:62`** — `recordPgid` has no doc comment while its sibling `markCanceled` (line 86) and the Windows `assignForStart` both do. It's the half that needs it more: the name says "record" but the return value is what drives the redo.

**7. `process_tree_unix.go:88`** — `markCanceled` mixes command and query. The Windows sibling names this shape `takeForCancel`, and this PR explicitly positions itself as mirroring that discipline.

## Performance

No concerns. The mutex is per-command, contended by at most two goroutines, held for two field writes, with `SIGKILL` deliberately outside it — the right call. The redo adds at most one extra `kill(2)`, and only in the raced path. Applying #1 removes a `Getpgid` syscall per command, a small net win.

## Highlights

- The race is real and correctly diagnosed — `watchCtx` genuinely starts inside `Start`.
- Lock scoped to field access with the slow syscall outside; consistent with the Windows sibling's `takeForCancel`/`assignForStart` pattern.
- Replacing `go func() { _ = cmd.Wait() }()` with `t.Cleanup` is a real improvement: no leaked goroutine, no race between the reap and the pgid the assertions depend on, and it still reaps when an assertion fails early.
- The pid-file fix is correct and correctly reasoned — same-directory `rename(2)`, and the path is passed as `$1` rather than interpolated into the script, so there is no shell-injection surface.
- Fixing the `ErrProcessDone`-as-failure reporting is a genuine catch; surfacing exit 1 with `os: process already finished` on graceful shutdown would have been a confusing operator-facing bug.
- Comments explain *why* rather than *what*, and the `Getpgid == pid` guard against targeting an unrelated group was preserved.

---

**Verdict**: CHANGES REQUESTED — finding #1 only. Everything else is a warning or a nit.

**Checklist**
- [x] Code follows stack conventions (gofmt/vet clean; minor doc-comment and naming nits, #6/#7)
- [ ] Tests included and passing — they pass, but the new test does not assert the behavior it is named for (#2), and the branch from 841a47b is uncovered (#4)
- [x] No security vulnerabilities (no injection surface; test-only shell changes)
- [ ] Documentation — the code comments and PR description overstate a Linux-only guarantee (#3)

pkg/executor/process_tree_unix.go
jisung-02: Confirmed and fixed in 3d99e56. I reproduced it locally on darwin before touching anything, with a probe replicating this file:

```
Getpgid on unreaped leader -> (-1, no such process)
redo kill target = 54954        (the leader, not the group)
redo kill err = <nil>
grandchild 54964 alive after redo: true
```

Applied your `leadsGroup` patch as suggested. Same probe afterwards: `redo kill target = -56042`, grandchild dead. I also checked the premise in the Go source rather than assuming it — `exec_bsd.go:118` and `exec_linux.go:393` both issue `RawSyscall(SYS_SETPGID, 0, uintptr(sys.Pgid), 0)`, so `Pgid == 0` really is `setpgid(0, 0)`.

The `Getpgid == pid` guard is kept for the caller-supplied-`Pgid` case exactly as you described. I updated the PR description too, since its central claim about `Getpgid` no longer describes what the code does.

pkg/executor/process_tree_unix_test.go
jisung-02: Fixed in f880205. The test now uses the grandchild shape and asserts the effect.

One correction worth recording: the snippet as written still passes against the unfixed code. Running `afterStart` straight after the raced cancel usually catches the leader before it has processed the `SIGKILL`, so `Getpgid` still resolves the group and the redo works even without the fix. I hit this on the first attempt — the new test passed on unpatched source.

So the test now waits for the signalled-but-unreaped window before the redo:

```go
func waitForLeaderStopped(pid int, timeout time.Duration) bool {
	return waitFor(timeout, func() bool {
		if processGone(pid) {
			return true
		}
		_, err := syscall.Getpgid(pid)
		return errors.Is(err, syscall.ESRCH)
	})
}
```

Two conditions because the platforms disagree about zombies in exactly the way you documented: `processGone` reads the `Z` state from `/proc` on linux, while on darwin `getpgid` is the one that stops seeing the leader. A fixed sleep would have worked too, but that trades the false pass for flakiness under load.

Verified in both directions on darwin: with the `leadsGroup` branch disabled the test fails with `grandchild 96329 survived the afterStart redo`, and passes with it. It also asserts the grandchild is still alive before the redo, so it fails loudly rather than silently degrading if the raced cancel ever starts reaching the whole group.

pkg/executor/process_tree_unix.go
jisung-02: Fixed in 3d99e56, via the second option you preferred. Since `afterStart` no longer calls `getpgid` on that path, the sentence about `ESRCH` after reaping is gone rather than qualified, and the redo comment is now true on both platforms.

The rationale moved into the `leadsGroup` branch so the darwin behavior stays on the record where someone would be tempted to reintroduce the lookup:

```go
// Don't ask getpgid: on darwin it returns ESRCH for a signalled-but-unreaped leader, which would
// drop the group in exactly the raced cancel handled below.
```

pkg/executor/process_tree_unix.go
jisung-02: Added a direct unit test in f880205 — `TestCommandCleanup_AfterStartRedoSwallowsProcessDone`. It reaps the cmd, marks a cancel as having fired, and calls `afterStart`, so the redo meets an already-gone group.

Your measurement matches what I saw, and the branch is genuinely reachable now that `leadsGroup` landed. To avoid the pid-reuse hazard the test itself would otherwise create, it verifies the group is actually empty and skips rather than signal one the OS may have reassigned:

```go
if !errors.Is(syscall.Kill(-pid, 0), syscall.ESRCH) {
	t.Skipf("pgid %d was reused, cannot exercise an already-gone group", pid)
}
```

Note this test does not detect the darwin regression — it passes with or without the `leadsGroup` fix, since either way the kill lands on something already gone. It pins the branch; the regression is covered by the redo test above.

pkg/executor/process_tree_unix_test.go
jisung-02: Extracted in 7a35f14, with your helper and comment as written. By then it was four copies, not three — the rewritten redo test needed the same script.

```go
// bgChildScript backgrounds a long sleeper and publishes its pid atomically: echo truncate-opens the
// target, so write to a temp file and rename it in (a same-directory rename(2) is atomic).
func bgChildScript(tail string) string {
	return `sleep 600 & echo $! > "$1.tmp"; mv "$1.tmp" "$1"; ` + tail
}
```

pkg/executor/process_tree_unix.go
jisung-02: Added in 7a35f14, using your wording — the return value driving the redo is indeed the part worth documenting.

pkg/executor/process_tree_unix.go
jisung-02: Renamed in 7a35f14, along with the call site in `cancel`. The doc comment now names the Windows counterpart explicitly so the correspondence is visible from either file.

While there I left the declaration order alone: cffd7f0 earlier in this branch deliberately grouped the `commandCleaner` methods ahead of the lock helpers, whereas the Windows file interleaves each helper after its caller. Happy to align that too if you would rather the two files match on layout as well as naming.

pkg/executor/process_tree_unix.go
```diff
@@ -6,13 +6,17 @@ import (
...
+	"sync"
 	"syscall"
 )
 
 // commandCleanup mirrors the type in process_tree_windows.go; the commandCleaner assertion below pins
 // the shared afterStart/cancel/close method set across build tags. State differs per platform.
 type commandCleanup struct {
-	pgid int // process group to SIGKILL; 0 when the child does not lead its own group
+	mu         sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
+	leadsGroup bool       // configure requested setsid/setpgid(0,0), so PGID == PID by construction
```
geunwoonoh: **Suggestion (non-blocking):** `leadsGroup` is not guarded by `mu`, but it sits directly beneath it, and Go convention reads the fields below a mutex as the ones it protects.

It is correct as written — written once in `configureProcessTreeCleanup` before `Start`, then read only in `afterStart`, so there is no race (confirmed under `-race`). Only the *grouping* is ambiguous. Either annotation removes the doubt for the next reader:

```suggestion
	leadsGroup bool       // immutable after configure: setsid/setpgid(0,0) was requested, so PGID == PID
```

Moving it above `mu` would work equally well.

pkg/executor/process_tree_unix.go
```diff
@@ -31,18 +35,32 @@ func configureProcessTreeCleanup(cmd *exec.Cmd, sessionLeader bool) (*commandCle
 		cmd.SysProcAttr.Setpgid = true
 	}
 
-	return &commandCleanup{}, nil
+	sysAttr := cmd.SysProcAttr
+	return &commandCleanup{leadsGroup: sysAttr.Setsid || (sysAttr.Setpgid && sysAttr.Pgid == 0)}, nil
```
geunwoonoh, line 40: **Suggestion (non-blocking):** this expression has no direct test, and the case that matters is the one where it must be `false`.

`TestConfigureProcessTreeCleanup_FlagMatrix` already walks exactly these branches, so it is the natural home for a one-line assertion per case. The valuable one is the fourth: a caller passing `Setpgid: true` with a nonzero `Pgid` must keep `leadsGroup == false`, because that is the only path where the `Getpgid == pid` fallback still does real work.

```go
	t.Run("preexisting_pgid_is_not_own_group_leader", func(t *testing.T) {
		cmd := &exec.Cmd{SysProcAttr: &syscall.SysProcAttr{Setpgid: true, Pgid: 1234}}
		cleanup, err := configureProcessTreeCleanup(cmd, false)
		if err != nil {
			t.Fatalf("configureProcessTreeCleanup: %v", err)
		}
		// Joining an existing group means PGID != PID, so afterStart must fall back to the getpgid guard.
		if cleanup.leadsGroup {
			t.Error("leadsGroup: got true, want false for a child joining an existing group")
		}
	})
```

Nothing in the tree sets `Pgid` today, so that fallback is now purely defensive — which is the argument for pinning it cheaply rather than leaving it to drift.

pkg/executor/executor.go
```diff
@@ -352,8 +352,8 @@ func (e *Executor) runCommand(cmd *exec.Cmd, pidHook func(pid int), cw *chunkWri
 		}
 	}()
 	if err := cleanup.afterStart(cmd); err != nil {
-		// Windows afterStart may have already canceled on a Start/cancel race; the deferred cancel
-		// re-runs harmlessly (idempotent). Unix afterStart never errors here.
+		// afterStart may re-run cancel on a Start/cancel race (unix-like and windows alike) and surface its error
```
geunwoonoh, line 355: **Nit (non-blocking):** "unix-like and windows alike" reads a little awkwardly. The previous "both platforms" was fine, and is shorter.

```suggestion
		// afterStart may re-run cancel on a Start/cancel race (both platforms) and surface its error
```

geunwoonoh (dismissed):
## Re-review: all findings resolved

All 7 findings from the previous round are fixed, each with a reply naming the commit. The critical one landed exactly as proposed, and two parts of the response went beyond what I asked for.

| # | Finding | Commit |
|---|---|---|
| 1 | Critical: redo is a no-op on darwin | `3d99e56` |
| 2 | Test asserts no-error, not effect | `f880205` |
| 3 | Comments assert a Linux-only guarantee | `3d99e56` |
| 4 | `ErrProcessDone` swallow untested | `f880205` |
| 5 | DRY: script duplicated | `7a35f14` |
| 6 | `recordPgid` missing doc comment | `7a35f14` |
| 7 | `markCanceled` -> `takeForCancel` | `7a35f14` |

## The critical fix, verified two ways

`leadsGroup` handles every branch of `configureProcessTreeCleanup` correctly, including the one case that must stay `false` (a caller passing `Setpgid: true` with a nonzero `Pgid`), which keeps the `Getpgid == pid` guard meaningful.

I did not just confirm the test passes — I mutated the fix away in a throwaway worktree, reverting only the `leadsGroup` shortcut back to the `Getpgid`-only capture, and re-ran on darwin:

```
--- FAIL: TestCommandCleanup_AfterStartRedoAfterRacedCancel (3.06s)
    process_tree_unix_test.go:116: grandchild 80286 survived the afterStart redo
```

That is the exact symptom from the original report, so the new test is a genuine regression guard on darwin rather than a test that happens to pass.

## Two places this improved on the suggestion

**`process_tree_unix_test.go:104-109`** — my proposed test would have passed vacuously if the raced cancel had taken the grandchild down for some unrelated reason. This version adds a `waitForLeaderStopped` barrier that forces the unreaped-zombie-leader state (precisely the state that defeated `getpgid`), then asserts the grandchild is *still alive* before the redo. A no-op redo and a working redo are now genuinely distinguishable.

**`waitForLeaderStopped`** — splits detection for the right reason: on linux `processGone` reads `Z` from `/proc`, while on darwin `getpgid` is what stops seeing the leader because `proc_find` skips `SZOMB`. That matches what I measured independently.

`TestCommandCleanup_AfterStartRedoSwallowsProcessDone` also guards pgid reuse with a pre-check and `t.Skipf` instead of blindly signalling a possibly-recycled group.

## Unprompted cleanups, all sound

- `cffd7f0`: `commandCleaner` methods grouped ahead of the lock helpers, so the file reads interface-first.
- `59250b7`: `waitFor(timeout, cond)` extracted; `readExecutorTimeoutChildPID` / `waitForExecutorTimeoutChildExit` renamed to `readChildPID` / `waitForProcessGone`. The old names were misleading — both were used well outside the timeout test. The rewritten poller also preserves the original's final post-sleep condition check, so no check is lost.

## Verification at `59250b7`

```
gofmt -l pkg/executor/                             -> clean
go vet ./pkg/executor       (darwin/linux/windows) -> clean
go build ./pkg/executor/... (darwin/linux/windows) -> OK
go test -race -count=3 (cleanup tests, darwin)     -> all PASS, no race reports
go test -race -count=2 ./pkg/executor (darwin)     -> ok  9.447s
mutation test (revert leadsGroup, darwin)          -> FAILS as expected
gh pr checks 381                                   -> 24/24 pass
```

CI green covers `macos-arm64`, `windows-amd64`, 14 linux distros, `golangci-lint`, `govulncheck` and CodeQL.

One gap in my own verification, stated plainly: I did not complete a local linux `-race` run — my container went unresponsive partway and produced no output. I am comfortable without it, since CI runs the full suite on 14 linux distros at this exact commit and the mutex logic under test is platform-independent and was exercised under `-race` on darwin.

## Suggestions (3, all non-blocking)

Left inline. None of them warrant another round; take or drop them as you see fit.

1. `process_tree_unix.go:17` — `leadsGroup` is unguarded but sits directly under `mu`, where Go convention reads fields below a mutex as guarded by it.
2. `process_tree_unix.go:39` — the `leadsGroup` expression has no direct test, and the `Pgid != 0` case is the one that matters.
3. `executor.go:355` — "unix-like and windows alike" reads a little awkwardly; the previous "both platforms" was fine.

## Highlights

- Every finding got a reply naming its commit, and the critical one was independently reproduced on darwin before being fixed rather than taken on faith.
- The redo test is strictly stronger than what I proposed, with an explicit guard against passing vacuously.
- Test helper naming and file organization were cleaned up beyond review scope, and correctly.
- The `Getpgid == pid` guard survived the refactor, so the "never signal an unrelated group" invariant is intact.
- Net effect: one `Getpgid` syscall removed per command execution on the happy path.

---

**Verdict**: APPROVED — the critical issue is fixed and proven fixed; the three remaining items are nits.

**Checklist**
- [x] Code follows stack conventions (gofmt / vet / golangci-lint clean)
- [x] Tests included and passing, and the new test demonstrably catches the bug it guards
- [x] Documentation — comments no longer overstate the platform guarantee

pkg/executor/process_tree_unix.go
jisung-02: Fixed in bf0f206, though by moving the field rather than annotating it.

```go
type commandCleanup struct {
	leadsGroup bool // immutable after configure: setsid/setpgid(0,0) was requested, so PGID == PID

	mu       sync.Mutex // afterStart (main goroutine) writes pgid while cmd.Cancel's context watcher reads it
	pgid     int        // process group to SIGKILL; 0 when the child does not lead its own group
	canceled bool       // a cancel already fired; afterStart re-runs it once the group is recorded
}
```

Your suggested comment is what I used, but on the moved field. The reasoning for moving instead: annotating in place documents an exception to the convention, whereas moving restores the convention itself. `process_tree_windows.go` already has everything under `mu` guarded by it, so after this both files can be read with the same rule and the blank line marks where the guarded block starts.

pkg/executor/process_tree_unix.go
jisung-02: Added in bf0f206, with your `preexisting_pgid_is_not_own_group_leader` case as written.

I also added a one-line `leadsGroup` assertion to the other four subtests, since your framing was "one assertion per case" and the branch-wide habit here is to fix every occurrence rather than only the flagged one. They were discarding the return value with `_`, so each now binds `cleanup`.

To check the new case is not vacuous, I dropped the guard from the expression (`leadsGroup: sysAttr.Setsid || sysAttr.Setpgid`) and re-ran:

```
--- FAIL: TestConfigureProcessTreeCleanup_FlagMatrix/preexisting_pgid_is_not_own_group_leader
    leadsGroup: got true, want false for a child joining an existing group
```

So the defensive path is genuinely pinned, which was your argument for adding it.

pkg/executor/executor.go
jisung-02: Not taking this one, since the earlier wording was changed on purpose rather than by accident.

`cffd7f0` in this branch replaced "both platforms" specifically because it reads as a pair of operating systems, while the split here is by the `!windows` build tag, which covers Linux, macOS and the BSDs. In a comment that exists to say "this applies on the non-Windows side too", "both platforms" quietly implies there are two, which is the misreading the current phrasing avoids.

I agree it is the clumsier of the two. If the awkwardness is worth another pass I would rather reword than shorten, e.g.:

```go
// afterStart may re-run cancel on a Start/cancel race (on every build tag) and surface its error
```

Happy to apply that or leave it as is - it is your nit, so your call.

geunwoonoh (approved):
## Re-review at `bf0f206`

Both code nits from the last round are fixed, and the third is settled — see below. Nothing outstanding.

Re-approving because the previous approval was auto-dismissed: the ruleset sets `dismiss_stale_reviews_on_push: true`, so pushing `bf0f206` invalidated it independently of content.

## Nit 1 — field grouping: fixed, better than suggested

The field moved above `mu` *and* kept the annotation, with a blank line separating it from the guarded group. I offered move *or* annotate; doing both makes the mutex's scope unambiguous at a glance.

## Nit 2 — coverage: fixed, and extended past what I asked

`preexisting_pgid_is_not_own_group_leader` landed as written, and the `leadsGroup` assertion was added to the other four subtests as well, rebinding `cleanup` where the return had been discarded with `_`. Fixing every occurrence rather than only the flagged one is the right instinct.

I mutation-tested the new subtest rather than trusting the green run. Dropping the `Pgid == 0` half of the guard:

```
--- FAIL: TestConfigureProcessTreeCleanup_FlagMatrix/preexisting_pgid_is_not_own_group_leader
    process_tree_unix_test.go:81: leadsGroup: got true, want false for a child joining an existing group
```

Only that subtest fails, so it pins exactly the branch it was added for.

## Nit 3 — withdrawn, your reading is better

You are right and I was wrong. This file is `//go:build !windows` — Linux, macOS *and* the BSDs — so "both platforms" does imply there are two, and this PR is the proof that the non-Windows side is not one thing: the whole critical finding was Linux and darwin diverging. Wording that flattens `!windows` into half of a pair is the exact assumption that hid the bug.

The current phrasing is the clumsier of the two and the accurate one, and accuracy wins in a comment whose job is to say the race applies off Windows too. No change wanted, and the `(on every build tag)` alternative is not needed either. Leaving the thread open for you to close, along with the other two.

## Verification at `bf0f206`

```
gofmt -l pkg/executor/                             -> clean
go vet ./pkg/executor       (darwin/linux/windows) -> clean
go build ./pkg/executor/... (darwin/linux/windows) -> OK
go test -race -count=2 (cleanup + flag matrix)     -> all PASS, no race reports
go test -race -count=2 ./pkg/executor              -> ok  9.314s
mutation test (drop Pgid == 0 guard)               -> FAILS as expected
gh pr checks 381                                   -> 24 pass / 0 non-pass
```

## Highlights across the whole review

- All 10 findings over three rounds got a reply naming the fixing commit, and the one declined item came with a reasoned technical case rather than silence or compliance. That is the part worth calling out.
- Nit 2 was generalized to all five subtests instead of patched at the flagged case.
- Both rounds of fixes hold up under mutation testing, not just under a green suite.

---

**Verdict**: APPROVED

**Checklist**
- [x] Code follows stack conventions (gofmt / vet / golangci-lint clean)
- [x] Tests included and passing; both new guards verified by mutation testing
- [x] Documentation — comments accurately scoped per build tag

Note on merging: the ruleset also enforces `required_review_thread_resolution`, and the three nit threads are still open. Two are fixed and this one is withdrawn, so all three are ready for you to close whenever you like.

pkg/executor/executor.go
geunwoonoh: Withdrawing this nit — your reading is better. `!windows` covers Linux, macOS and the BSDs, so "both platforms" does imply there are two, and this PR is the proof the non-Windows side is not one thing: the critical finding was Linux and darwin diverging. Flattening `!windows` into half of a pair is the exact assumption that hid the bug.

Current phrasing is the clumsier one and the accurate one; accuracy wins here. No change wanted, and `(on every build tag)` is not needed either. Leaving the thread for you to close.

jisung-02:
Closed in favour of #387, which is open on the same branch at the same head commit (`bf0f206`). No code changed in the move.

The review history here stays the authoritative record for these changes: three rounds of findings, the replies naming each fixing commit, and the darwin reproduction and mutation testing behind the critical fix. #387 links back to this thread rather than restating it.

PR #383, opened 2026-07-24
test(executor): fix flaky windows process-tree cancel tests

### LINKED ISSUE #382: Flaky test: TestCommandCleanup_CancelTerminatesMultiLevelTreeViaJob (windows-amd64)
# Flaky test: `TestCommandCleanup_CancelTerminatesMultiLevelTreeViaJob` (windows-amd64)

## Summary

`TestCommandCleanup_CancelTerminatesMultiLevelTreeViaJob` in `pkg/executor` intermittently
fails on the `windows-amd64` runner with:

```
process_tree_windows_test.go:99: child process 5452 still running after cancel
--- FAIL: TestCommandCleanup_CancelTerminatesMultiLevelTreeViaJob (0.06s)
FAIL	github.com/alpacax/alpamon/v2/pkg/executor	2.739s
```

A plain re-run of the same commit passes, confirming it is a timing flake, not a
regression. The production cleanup code is correct; the **test's assertion is
racing an inherently asynchronous Windows teardown**.

## Evidence

- Failing job: `Build and Test (windows-amd64)`, run `30079973803` / job `89439077268`
- Re-run of the identical commit: **passed** (job `89440285342`)
- Only `windows-amd64` is affected; the other 17 build-and-test targets pass.
- Not tied to any particular branch — the failing test lives in
  `pkg/executor/process_tree_windows_test.go` on `main` and the branch that
  observed it does not touch `pkg/executor`.

## Root cause

The test builds a two-level process tree and cancels it via the Job Object:

```go
cmd := exec.Command("cmd", "/c", "ping", "-n", "60", "127.0.0.1") // cmd.exe (root) → ping.exe (child)
...
childPID := waitForWindowsChild(t, rootPID) // polls up to 3s for the child to APPEAR
...
cleanup.cancel(cmd)
waitForCmd(t, cmd)                           // waits for cmd.Wait() — confirms the ROOT is reaped

if isWindowsPidAlive(rootPID)  { t.Fatalf("root process %d still running after cancel", rootPID) }
if isWindowsPidAlive(childPID) { t.Fatalf("child process %d still running after cancel", childPID) }
```

`cancel()` (`process_tree_windows.go:73`) tears the tree down through
`setKillOnJobClose(job)` + `CloseHandle(job)` (kill-on-job-close), a
`TerminateProcess` on the root handle, and a PID-based
`terminateWindowsProcessTree` fallback. **All three are asynchronous on
Windows**: the kernel signals the job members to terminate, but their teardown
and handle-signalling is not synchronous with the calls returning.

`waitForCmd` only guarantees the **root** (`cmd.exe`) has exited — `cmd.Wait()`
reaps the process the runtime launched, nothing more. The **child** (`ping.exe`)
is killed via the asynchronous job path, and its teardown can lag behind the
root's, especially on a loaded shared CI runner (the same run logs repeated
`Access is denied` fsync warnings, a sign of a contended host).

The core defect is an **asymmetry in the test itself**: it *polls* for the child
to appear (`waitForWindowsChild`, 3s deadline) but checks that the child has
*disappeared* with a single instantaneous `isWindowsPidAlive(childPID)`
(`process_tree_windows_test.go:99`). A single-shot liveness check cannot tolerate
the OS's asynchronous termination latency.

`isWindowsPidAlive` (`process_tree_windows_test.go:195`) is a point-in-time
`WaitForSingleObject(h, 0)` — it returns "alive" while the child is mid-teardown
but not yet signalled.

## Why this is a test bug, not a product bug

- `cancel()` does terminate the whole tree; the child *does* die — just not
  before the un-polled assertion runs. A re-run passes with no code change.
- The product contract ("cancel eventually terminates the whole tree") is
  satisfied; only the test's "terminated *synchronously by the time cancel
  returns + root is reaped*" expectation is wrong for Windows.

## Affected tests

Same single-shot pattern, same latent flake:

- `process_tree_windows_test.go:99` — `...MultiLevelTreeViaJob` (the observed failure)
- `process_tree_windows_test.go:96` — root check in the same test
- `process_tree_windows_test.go:40`, `:63` — `...CancelTerminatesViaJobAssignment`

## Proposed fix (test-only)

Replace the single-shot `isWindowsPidAlive(pid)` assertions with a bounded poll
that mirrors `waitForWindowsChild` — the disappearance check should be as patient
as the appearance check:

```go
// waitForWindowsPidGone fails only if pid is still alive after a bounded wait,
// tolerating the asynchronous job-object teardown.
func waitForWindowsPidGone(t *testing.T, pid uint32, what string) {
	t.Helper()
	deadline := time.Now().Add(3 * time.Second)
	for time.Now().Before(deadline) {
		if !isWindowsPidAlive(pid) {
			return
		}
		time.Sleep(50 * time.Millisecond)
	}
	t.Fatalf("%s process %d still running after cancel", what, pid)
}
```

and use it in both Job-Object cancel tests:

```go
waitForWindowsPidGone(t, rootPID, "root")
waitForWindowsPidGone(t, childPID, "child")
```

No change to `process_tree_windows.go` is warranted.

## Notes

- The 3s bound matches the existing `waitForWindowsChild` / `waitForCmd` deadlines.
- If the tail of teardown ever proves longer than 3s on CI, widen the deadline
  rather than reverting to a single-shot check.

---

Labels: `flaky-test`, `windows`, `test`

junho226:
## Summary

Fixes the intermittent `windows-amd64` failure of the Job Object cancel tests in
`pkg/executor`, e.g.:

```
process_tree_windows_test.go: child process 5452 still running after cancel
--- FAIL: TestCommandCleanup_CancelTerminatesMultiLevelTreeViaJob
```

Closes #382.

## Root cause

`cancel()` tears a process tree down through kill-on-job-close, `TerminateProcess`,
and a PID-tree walk — all **asynchronous** on Windows. The tests then asserted the
processes were gone with a single `isWindowsPidAlive(pid)` call right after
`waitForCmd`, but `waitForCmd` only proves the **root** was reaped; a job member
such as the child `ping.exe` can still be mid-teardown. The check raced the OS and
flaked under CI load.

The tell is an asymmetry in the test helpers: the child's *appearance* is polled
for up to 3s (`waitForWindowsChild`), while its *disappearance* was a single
instantaneous check.

## Change

- Add `waitForWindowsPidGone`, a bounded (3s) poll that mirrors
  `waitForWindowsChild`, and use it for every post-cancel liveness assertion in
  the four Job Object / PID-tree cancel tests.
- **Test-only.** `process_tree_windows.go` is unchanged — the production cleanup
  already terminates the whole tree; only the test's timing expectation was wrong.

## Testing

- `GOOS=windows go vet ./pkg/executor/` — clean.
- `GOOS=windows go test -c ./pkg/executor/` — compiles (the runner tests only
  build under Windows; the fix is exercised by CI's `windows-amd64` job).
- `gofmt` clean.

## Notes

The 3s deadline matches the existing `waitForWindowsChild` / `waitForCmd` bounds.
If teardown ever proves slower than 3s on CI, widen the deadline rather than
reverting to a single-shot check.

### DIFF
--- a/pkg/executor/process_tree_windows_test.go
+++ b/pkg/executor/process_tree_windows_test.go
@@ -194,7 +194,7 @@ func isWindowsPidAlive(pid uint32) bool {
 
 // waitForWindowsPidGone fails only if pid is still alive after a bounded wait.
 // cancel() tears the tree down through kill-on-job-close, TerminateProcess, and
-// a PID-tree walk — all asynchronous on Windows — and waitForCmd only proves the
+// a PID-tree walk—all asynchronous on Windows—and waitForCmd only proves the
 // root was reaped, not that every job member has finished terminating. Polling
 // for the pid to disappear mirrors waitForWindowsChild's poll for it to appear,
 // so the check tolerates that teardown latency instead of racing it.
--- a/pkg/executor/process_tree_windows_test.go
+++ b/pkg/executor/process_tree_windows_test.go
@@ -36,9 +36,7 @@ func TestCommandCleanup_CancelTerminatesViaJobAssignment(t *testing.T) {
 		t.Fatalf("cancel: %v", err)
 	}
 	waitForCmd(t, cmd)
-	if isWindowsPidAlive(pid) {
-		t.Fatalf("process %d still running after cancel", pid)
-	}
+	waitForWindowsPidGone(t, pid, "process")
 }
 
 func TestCommandCleanup_CancelFallsBackToPIDTreeWithoutJobAssignment(t *testing.T) {
@@ -59,9 +57,7 @@ func TestCommandCleanup_CancelFallsBackToPIDTreeWithoutJobAssignment(t *testing.
 		t.Fatalf("cancel: %v", err)
 	}
 	waitForCmd(t, cmd)
-	if isWindowsPidAlive(pid) {
-		t.Fatalf("process %d still running after cancel", pid)
-	}
+	waitForWindowsPidGone(t, pid, "process")
 }
 
 // The assigned job must take a descendant with it: cmd.exe -> ping, both die on cancel.
@@ -92,12 +88,8 @@ func TestCommandCleanup_CancelTerminatesMultiLevelTreeViaJob(t *testing.T) {
 	}
 	waitForCmd(t, cmd)
 
-	if isWindowsPidAlive(rootPID) {
-		t.Fatalf("root process %d still running after cancel", rootPID)
-	}
-	if isWindowsPidAlive(childPID) {
-		t.Fatalf("child process %d still running after cancel", childPID)
-	}
+	waitForWindowsPidGone(t, rootPID, "root process")
+	waitForWindowsPidGone(t, childPID, "child process")
 }
 
 // cancel racing ahead of afterStart: afterStart must notice canceled==true and re-run cancel.
@@ -125,9 +117,7 @@ func TestCommandCleanup_AfterStartReCancelsWhenAlreadyCanceled(t *testing.T) {
 		t.Fatalf("afterStart: %v", err)
 	}
 	waitForCmd(t, cmd)
-	if isWindowsPidAlive(pid) {
-		t.Fatalf("process %d still running after afterStart re-cancel", pid)
-	}
+	waitForWindowsPidGone(t, pid, "process")
 }
 
 // Black-box counterpart to the Unix test: cmd.exe runs ping, which inherits and holds stdout open.
@@ -200,6 +190,24 @@ func isWindowsPidAlive(pid uint32) bool {
 	defer windows.CloseHandle(h)
 	event, err := windows.WaitForSingleObject(h, 0)
 	return err == nil && event == uint32(windows.WAIT_TIMEOUT)
+}
+
+// waitForWindowsPidGone fails only if pid is still alive after a bounded wait.
+// cancel() tears the tree down through kill-on-job-close, TerminateProcess, and
+// a PID-tree walk — all asynchronous on Windows — and waitForCmd only proves the
+// root was reaped, not that every job member has finished terminating. Polling
+// for the pid to disappear mirrors waitForWindowsChild's poll for it to appear,
+// so the check tolerates that teardown latency instead of racing it.
+func waitForWindowsPidGone(t *testing.T, pid uint32, what string) {
+	t.Helper()
+	deadline := time.Now().Add(3 * time.Second)
+	for time.Now().Before(deadline) {
+		if !isWindowsPidAlive(pid) {
+			return
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	t.Fatalf("%s %d still running after cancel", what, pid)
 }
 
 func waitForWindowsChild(t *testing.T, parent uint32) uint32 {
--- a/pkg/executor/process_tree_windows_test.go
+++ b/pkg/executor/process_tree_windows_test.go
@@ -36,7 +36,7 @@ func TestCommandCleanup_CancelTerminatesViaJobAssignment(t *testing.T) {
 		t.Fatalf("cancel: %v", err)
 	}
 	waitForCmd(t, cmd)
-	waitForWindowsPidGone(t, pid, "process")
+	waitForWindowsPidGone(t, pid, "after cancel via job assignment")
 }
 
 func TestCommandCleanup_CancelFallsBackToPIDTreeWithoutJobAssignment(t *testing.T) {
@@ -57,7 +57,7 @@ func TestCommandCleanup_CancelFallsBackToPIDTreeWithoutJobAssignment(t *testing.
 		t.Fatalf("cancel: %v", err)
 	}
 	waitForCmd(t, cmd)
-	waitForWindowsPidGone(t, pid, "process")
+	waitForWindowsPidGone(t, pid, "after cancel via PID-tree fallback")
 }
 
 // The assigned job must take a descendant with it: cmd.exe -> ping, both die on cancel.
@@ -88,8 +88,8 @@ func TestCommandCleanup_CancelTerminatesMultiLevelTreeViaJob(t *testing.T) {
 	}
 	waitForCmd(t, cmd)
 
-	waitForWindowsPidGone(t, rootPID, "root process")
-	waitForWindowsPidGone(t, childPID, "child process")
+	waitForWindowsPidGone(t, rootPID, "root process after cancel")
+	waitForWindowsPidGone(t, childPID, "child process after cancel")
 }
 
 // cancel racing ahead of afterStart: afterStart must notice canceled==true and re-run cancel.
@@ -117,7 +117,7 @@ func TestCommandCleanup_AfterStartReCancelsWhenAlreadyCanceled(t *testing.T) {
 		t.Fatalf("afterStart: %v", err)
 	}
 	waitForCmd(t, cmd)
-	waitForWindowsPidGone(t, pid, "process")
+	waitForWindowsPidGone(t, pid, "after afterStart re-cancel")
 }
 
 // Black-box counterpart to the Unix test: cmd.exe runs ping, which inherits and holds stdout open.
@@ -171,13 +171,20 @@ func TestExecutor_TimeoutCleansProcessTreeWhenChildKeepsPipeOpen(t *testing.T) {
 	}
 }
 
+// One bound for every wait in this file: the same teardown latency is what all
+// of them are tolerating, so widening it is a single edit.
+const (
+	windowsPollDeadline = 3 * time.Second
+	windowsPollInterval = 50 * time.Millisecond
+)
+
 func waitForCmd(t *testing.T, cmd *exec.Cmd) {
 	t.Helper()
 	done := make(chan error, 1)
 	go func() { done <- cmd.Wait() }()
 	select {
 	case <-done:
-	case <-time.After(3 * time.Second):
+	case <-time.After(windowsPollDeadline):
 		t.Fatal("cmd.Wait did not return after cancel")
 	}
 }
@@ -193,27 +200,37 @@ func isWindowsPidAlive(pid uint32) bool {
 }
 
 // waitForWindowsPidGone fails only if pid is still alive after a bounded wait.
-// cancel() tears the tree down through kill-on-job-close, TerminateProcess, and
-// a PID-tree walk—all asynchronous on Windows—and waitForCmd only proves the
-// root was reaped, not that every job member has finished terminating. Polling
-// for the pid to disappear mirrors waitForWindowsChild's poll for it to appear,
-// so the check tolerates that teardown latency instead of racing it.
-func waitForWindowsPidGone(t *testing.T, pid uint32, what string) {
+// cancel() tears the tree down asynchronously (kill-on-job-close,
+// TerminateProcess, PID-tree walk), and waitForCmd only proves the root was
+// reaped—not that every job member has finished terminating.
+//
+// Only the child-pid assertion was ever racy (#382); where cmd.Wait already
+// reaped the pid the loop returns on iteration zero. Those call sites poll for
+// helper symmetry, deliberately relaxing "gone now" to "gone within
+// windowsPollDeadline".
+//
+// stage names the assertion so a bare CI line identifies which one failed.
+func waitForWindowsPidGone(t *testing.T, pid uint32, stage string) {
 	t.Helper()
-	deadline := time.Now().Add(3 * time.Second)
+	deadline := time.Now().Add(windowsPollDeadline)
 	for time.Now().Before(deadline) {
 		if !isWindowsPidAlive(pid) {
 			return
 		}
-		time.Sleep(50 * time.Millisecond)
+		time.Sleep(windowsPollInterval)
 	}
-	t.Fatalf("%s %d still running after cancel", what, pid)
+	// Re-check: the loop gives back the last interval to sleep, and a process
+	// that exits inside it is within the stated bound.
+	if !isWindowsPidAlive(pid) {
+		return
+	}
+	t.Fatalf("%s: process %d still running", stage, pid)
 }
 
 func waitForWindowsChild(t *testing.T, parent uint32) uint32 {
 	t.Helper()
-	deadline := time.Now().Add(3 * time.Second)
-	for time.Now().Before(deadline) {
+	deadline := time.Now().Add(windowsPollDeadline)
+	for {
 		children, err := snapshotWindowsChildProcesses()
 		if err != nil {
 			t.Fatalf("snapshotWindowsChildProcesses: %v", err)
@@ -221,7 +238,10 @@ func waitForWindowsChild(t *testing.T, parent uint32) uint32 {
 		if kids := children[parent]; len(kids) > 0 {
 			return kids[0]
 		}
-		time.Sleep(50 * time.Millisecond)
+		if !time.Now().Before(deadline) {
+			break
+		}
+		time.Sleep(windowsPollInterval)
 	}
 	t.Fatalf("child of process %d did not appear", parent)
 	return 0


pkg/executor/process_tree_windows_test.go
```diff
@@ -202,6 +192,24 @@ func isWindowsPidAlive(pid uint32) bool {
 	return err == nil && event == uint32(windows.WAIT_TIMEOUT)
 }
 
+// waitForWindowsPidGone fails only if pid is still alive after a bounded wait.
+// cancel() tears the tree down through kill-on-job-close, TerminateProcess, and
+// a PID-tree walk — all asynchronous on Windows — and waitForCmd only proves the
+// root was reaped, not that every job member has finished terminating. Polling
```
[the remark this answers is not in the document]
junho226: Fixed in 7f8ccfc. Removed the spaces around both em-dashes in the comment, per the CLAUDE.md convention.

pkg/executor/process_tree_windows_test.go
```diff
@@ -202,6 +192,24 @@ func isWindowsPidAlive(pid uint32) bool {
...
+func waitForWindowsPidGone(t *testing.T, pid uint32, what string) {
+	t.Helper()
+	deadline := time.Now().Add(3 * time.Second)
+	for time.Now().Before(deadline) {
+		if !isWindowsPidAlive(pid) {
+			return
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	t.Fatalf("%s %d still running after cancel", what, pid)
```
geunwoonoh: **Warning:** No re-check after the deadline expires—this diverges from the repo's own Unix counterpart.

The last observation happens at ~2.95s; the final 50ms of the nominal 3s budget is spent asleep, then the helper fails without checking again. A process that dies at t=2.97s fails a test whose stated bound is 3s. For a change whose whole purpose is "stop racing the OS," giving back a poll interval at the exact tail is the wrong interval to lose.

`waitForExecutorTimeoutChildExit` at `pkg/executor/process_tree_unix_test.go:236-245`—the direct Unix analogue—already handles this with a final `return processGone(pid)` after the loop.

```suggestion
	for time.Now().Before(deadline) {
		if !isWindowsPidAlive(pid) {
			return
		}
		time.Sleep(50 * time.Millisecond)
	}
	if !isWindowsPidAlive(pid) {
		return
	}
	t.Fatalf("%s %d still running after cancel", what, pid)
```

pkg/executor/process_tree_windows_test.go
```diff
@@ -125,9 +117,7 @@ func TestCommandCleanup_AfterStartReCancelsWhenAlreadyCanceled(t *testing.T) {
 		t.Fatalf("afterStart: %v", err)
 	}
 	waitForCmd(t, cmd)
-	if isWindowsPidAlive(pid) {
-		t.Fatalf("process %d still running after afterStart re-cancel", pid)
-	}
+	waitForWindowsPidGone(t, pid, "process")
```
geunwoonoh: **Warning:** Diagnostic regression—this call site lost the detail that made its failure identifiable.

The original message was `"process %d still running after afterStart re-cancel"`. The helper hardcodes `"after cancel"`, so `TestCommandCleanup_AfterStartReCancelsWhenAlreadyCanceled`—whose entire point is the *re-cancel* path—now emits text identical to the three plain-cancel tests. Three of four call sites produce the byte-identical string `process <pid> still running after cancel`, so a bare CI log line no longer says which assertion blew up.

Making the parameter carry the full context instead of a noun fragment restores it:

```go
func waitForWindowsPidGone(t *testing.T, pid uint32, context string) {
	...
	t.Fatalf("%s: process %d still running", context, pid)
}

// call sites: "after cancel", "root process after cancel",
//             "child process after cancel", "after afterStart re-cancel"
```

pkg/executor/process_tree_windows_test.go
```diff
@@ -202,6 +192,24 @@ func isWindowsPidAlive(pid uint32) bool {
...
 
+// waitForWindowsPidGone fails only if pid is still alive after a bounded wait.
+// cancel() tears the tree down through kill-on-job-close, TerminateProcess, and
+// a PID-tree walk—all asynchronous on Windows—and waitForCmd only proves the
+// root was reaped, not that every job member has finished terminating. Polling
+// for the pid to disappear mirrors waitForWindowsChild's poll for it to appear,
+// so the check tolerates that teardown latency instead of racing it.
+func waitForWindowsPidGone(t *testing.T, pid uint32, what string) {
+	t.Helper()
+	deadline := time.Now().Add(3 * time.Second)
```
geunwoonoh: **Suggestion:** Duplicated poll scaffolding and magic numbers.

`waitForWindowsPidGone` and `waitForWindowsChild` (line 215) are now structurally identical: same `3 * time.Second` deadline, same `50 * time.Millisecond` interval, differing only in predicate. `waitForCmd` (line 180) repeats the 3s a third time. That's three copies of the deadline and two of the interval in one file, with no name explaining that they are deliberately the same bound—which the PR description says they are.

Minimum fix, so the coupling is explicit and a future widening is one edit:

```go
const (
	windowsPollDeadline = 3 * time.Second
	windowsPollInterval = 50 * time.Millisecond
)
```

Collapsing both loops into a `pollUntil(deadline, interval, cond) bool` is possible but marginal at two call sites; the constants alone carry most of the value.

pkg/executor/process_tree_windows_test.go
```diff
@@ -36,9 +36,7 @@ func TestCommandCleanup_CancelTerminatesViaJobAssignment(t *testing.T) {
 		t.Fatalf("cancel: %v", err)
 	}
 	waitForCmd(t, cmd)
-	if isWindowsPidAlive(pid) {
-		t.Fatalf("process %d still running after cancel", pid)
-	}
+	waitForWindowsPidGone(t, pid, "process")
```
geunwoonoh: **Suggestion:** Worth knowing the actual blast radius—three of the four conversions are redundant rather than flake-fixing.

Only the `childPID` assertion at line 92 was ever racy, which matches the reported failure (`child process 5452 still running after cancel`) and issue #382.

At this site (and lines 60 and 120) the command is a bare `ping` with no descendants, and `waitForCmd` → `cmd.Wait()` on Windows returns only after `WaitForSingleObject` reports the process signaled. The root is therefore *provably* exited before the assertion runs, so `isWindowsPidAlive` returns false on poll iteration zero and the loop never spins. The poll cannot help here—and in the one scenario that would make these sites fail (PID reuse between `Wait()` and the check), polling doesn't help either; it just spends 3s arriving at the same failure.

Not wrong, and helper symmetry across the four tests has real value. But it does permanently relax three assertions from "gone now, guaranteed by `cmd.Wait()`" to "gone within 3s" in exchange for nothing. Either keep it and say so in the helper comment, or leave these three single-shot and convert only line 92. Preference: keep the symmetry, just be deliberate about it.

pkg/executor/process_tree_windows_test.go
```diff
@@ -202,6 +192,24 @@ func isWindowsPidAlive(pid uint32) bool {
 	return err == nil && event == uint32(windows.WAIT_TIMEOUT)
 }
 
+// waitForWindowsPidGone fails only if pid is still alive after a bounded wait.
+// cancel() tears the tree down through kill-on-job-close, TerminateProcess, and
+// a PID-tree walk—all asynchronous on Windows—and waitForCmd only proves the
+// root was reaped, not that every job member has finished terminating. Polling
+// for the pid to disappear mirrors waitForWindowsChild's poll for it to appear,
+// so the check tolerates that teardown latency instead of racing it.
```
geunwoonoh: **Suggestion:** Six lines of comment for a ten-line helper, largely restating the PR description.

It is WHY-focused per `CLAUDE.md` and encodes a root cause that took real work to find, so it earns its place—but the middle two lines could compress without losing the insight. Optional; feel free to leave as-is.

geunwoonoh (commented):
## Summary

Replaces four single-shot `isWindowsPidAlive(pid)` post-cancel assertions with `waitForWindowsPidGone`, a bounded 3s poll. Test-only; `process_tree_windows.go` is untouched.

The diagnosis is correct. `cancel()` (`process_tree_windows.go:73`) fires kill-on-job-close, `TerminateProcess`, and a Toolhelp32 PID-tree walk—all asynchronous—while `waitForCmd` only proves the *root* was reaped via `cmd.Wait()`. A job member mid-teardown can still be alive at that instant. The asymmetry identified here (appearance polled for 3s, disappearance checked once) is the right tell.

**Verified locally:**

- `gofmt -l pkg/executor/` — clean
- `GOOS=windows go vet ./pkg/executor/` — clean
- `GOOS=windows go test -c ./pkg/executor/` — builds
- `go vet ./pkg/executor/` (darwin) — clean

## Critical issues (0)

None.

## Warnings (2)

**1. `process_tree_windows_test.go:204-210` — no re-check after the deadline expires, diverging from this repo's own Unix counterpart**

```go
for time.Now().Before(deadline) {
    if !isWindowsPidAlive(pid) { return }
    time.Sleep(50 * time.Millisecond)
}
t.Fatalf(...)   // <- fails without looking one last time
```

The last observation happens at ~2.95s; the final 50ms of the nominal 3s budget is spent asleep, then it fails without checking again. A process that dies at t=2.97s fails a test whose stated bound is 3s. For a change whose entire purpose is "stop racing the OS," giving back a poll interval at the exact tail is the wrong interval to lose.

`waitForExecutorTimeoutChildExit` at `pkg/executor/process_tree_unix_test.go:236-245`—the direct Unix analogue—already does this correctly:

```go
for time.Now().Before(deadline) {
    if processGone(pid) { return true }
    time.Sleep(20 * time.Millisecond)
}
return processGone(pid)   // final check
```

**2. `process_tree_windows_test.go:120` — diagnostic regression in the re-cancel test**

The original message was `"process %d still running after afterStart re-cancel"`. The helper hardcodes `"after cancel"`, so `TestCommandCleanup_AfterStartReCancelsWhenAlreadyCanceled`—whose entire point is the *re-cancel* path—now emits text identical to the three plain-cancel tests. Three of four call sites produce the byte-identical string `process <pid> still running after cancel`, so a bare CI log line no longer says which assertion blew up.

## Suggestions (3)

**3. `process_tree_windows_test.go:203-208, 215, 180` — duplicated poll scaffolding and magic numbers**

`waitForWindowsPidGone` and `waitForWindowsChild` are now structurally identical: same `3 * time.Second` deadline, same `50 * time.Millisecond` interval, differing only in predicate. `waitForCmd:180` repeats the 3s a third time. Three copies of the deadline and two of the interval in one file, with no name explaining that they are deliberately the same bound (the PR description says they are).

Naming the constants makes the coupling explicit and a future widening a one-line edit.

**4. `process_tree_windows_test.go:39, 60, 120` — three of the four conversions are redundant, not flake-fixing**

Worth knowing the actual blast radius: only the `childPID` assertion at line 92 was ever racy, which matches the reported failure (`child process 5452 still running after cancel`) and issue #382.

For the other three sites the command is a bare `ping` with no descendants, and `waitForCmd` → `cmd.Wait()` on Windows returns only after `WaitForSingleObject` reports the process signaled. The root is therefore *provably* exited before the assertion runs, so `isWindowsPidAlive` returns false on poll iteration zero and the loop never spins. The poll cannot help there—and in the one scenario that would make those sites fail (PID reuse between `Wait()` and the check), polling doesn't help either; it just spends 3s arriving at the same failure.

This isn't wrong, and helper symmetry across the four tests has real value. But it does permanently relax three assertions from "gone now, guaranteed by `cmd.Wait()`" to "gone within 3s" in exchange for nothing. Either keep it and say so in the comment, or leave those three single-shot and convert only line 92. Preference: keep the symmetry, just be deliberate about it.

**5. `process_tree_windows_test.go:195-200` — comment length**

Six lines of comment for a ten-line helper, largely restating the PR description. It is WHY-focused per `CLAUDE.md` and encodes a root cause that took real work to find, so it earns its place—but the middle two lines could compress without losing the insight. Optional.

## Performance

No regression, small net win:

| Path | Before | After |
|---|---|---|
| Passing (normal) | 1 `OpenProcess`/`Wait`/`CloseHandle` | identical—loop exits on iteration 0 |
| Passing (was flaky) | FAIL → full CI rerun | ~50ms poll, passes |
| Genuinely broken | fails at 0s | fails at 3s (6s for the two-PID test) |

Worst case adds ~15s wall time across the four tests on an already-red build. Irrelevant against the cost of one flaky-rerun cycle. No allocations, no goroutines, no `t.Parallel()` interaction (CI runs `go test ./... -p 1`).

## Security

Nothing to flag. Test-only, no credentials, no external input, no logging of sensitive data.

## Go convention

Clean. `t.Helper()` present, `t.Fatalf` called from the test goroutine (safe), verb+noun helper naming, correct import grouping, no error swallowing. Only nit is the `what string` parameter name—vague, and folded into warning 2 above.

## Highlights

- Diagnosed via a structural tell (poll-for-appearance vs. single-shot-for-disappearance) rather than by guessing at a timeout.
- Resisted touching production code, and the PR description states explicitly why the production path is already correct.
- Bounded wait rather than unbounded, matching the existing 3s helper bounds.
- The PR description pre-registers the escalation path ("widen the deadline rather than reverting to a single-shot check")—the note a future maintainer will actually need.
- Comment follows the `CLAUDE.md` no-space em-dash rule (fixed in 7f8ccfc).

---

**Verdict**: COMMENTED—approve-with-nits. No blockers. Warning 1 is the one worth fixing before merge: it reintroduces a one-poll-interval race into a PR whose purpose is removing races, and the repo already has the correct pattern 20 lines away in the Unix test file.

**Checklist**:

- [x] Tests included and passing (test-only change; cross-compile + vet verified)
- [x] Documentation updated if needed (n/a)

pkg/executor/process_tree_windows_test.go
junho226: Fixed (2703d83). Added the re-check after the loop, matching `waitForExecutorTimeoutChildExit`'s trailing `return processGone(pid)` on the Unix side:

```go
	// Re-check: the loop gives back the last interval to sleep, and a process
	// that exits inside it is within the stated bound.
	if !isWindowsPidAlive(pid) {
		return
	}
	t.Fatalf("%s: process %d still running", stage, pid)
```

`waitForWindowsChild` gave back its last interval the same way — the last snapshot was at ~2.95s, so a child appearing at 2.97s failed a 3s wait. Restructured it to check-then-sleep so the deadline is honoured on the appear side too. You only flagged the gone side, but the two are the mirror of each other and I'd rather not leave the asymmetry behind.

pkg/executor/process_tree_windows_test.go
junho226: Fixed. The parameter now carries the full context and the helper formats it as a prefix:

```go
t.Fatalf("%s: process %d still running", stage, pid)
```

All five call sites are distinct now:

| test | stage |
|---|---|
| `CancelTerminatesViaJobAssignment` | `after cancel via job assignment` |
| `CancelFallsBackToPIDTreeWithoutJobAssignment` | `after cancel via PID-tree fallback` |
| `CancelTerminatesMultiLevelTreeViaJob` | `root process after cancel` / `child process after cancel` |
| `AfterStartReCancelsWhenAlreadyCanceled` | `after afterStart re-cancel` |

Your list had four strings for five sites — the two plain-cancel tests would still have collided. They exercise different teardown paths (job assignment vs the PID-walk fallback), so they get different strings rather than sharing one.

pkg/executor/process_tree_windows_test.go
junho226: Added, with a line saying why they are shared:

```go
// One bound for every wait in this file: the same teardown latency is what all
// of them are tolerating, so widening it is a single edit.
const (
	windowsPollDeadline = 3 * time.Second
	windowsPollInterval = 50 * time.Millisecond
)
```

Both loops and `waitForCmd`'s `time.After` now use them. Took your read on `pollUntil` — marginal at two call sites, and the predicates differ enough (one returns a pid, one returns nothing) that the signature would be the awkward part.

pkg/executor/process_tree_windows_test.go
junho226: Kept the symmetry and made the trade-off explicit in the helper, per your preference:

```go
// Only the child-pid assertion was ever racy (#382); where cmd.Wait already
// reaped the pid the loop returns on iteration zero. Those call sites poll for
// helper symmetry, deliberately relaxing "gone now" to "gone within
// windowsPollDeadline".
```

Thanks for actually measuring which of the four mattered — I had assumed all of them were load-bearing. Your point that polling doesn't help the PID-reuse case either is the part that settled it for me: there's no scenario where the extra three loops turn a failure into a pass, so the only thing they buy is one shape for four assertions. Worth stating rather than leaving a reader to rediscover it.

pkg/executor/process_tree_windows_test.go
junho226: Compressed the middle two lines — the teardown mechanisms are now a parenthetical instead of a sentence, and the "mirrors waitForWindowsChild" line is gone since the constants say that structurally now.

The block did not get shorter overall, because it absorbed the deliberateness note from the thread above. It covers two things now (why polling at all, and why three of the four sites poll anyway) instead of restating the PR description.

junho226:
@geunwoonoh Thanks for the review — all five threads are addressed in 2703d83.

| # | Finding | Resolution |
|---|---|---|
| 1 | No re-check after the deadline expires (Warning) | Trailing `isWindowsPidAlive` check, as `waitForExecutorTimeoutChildExit` does |
| 2 | Diagnostic regression at the re-cancel call site (Warning) | `stage` parameter carries full context; all five sites distinct |
| 3 | Duplicated poll scaffolding and magic numbers | `windowsPollDeadline` / `windowsPollInterval` |
| 4 | Three of four conversions are redundant | Kept the symmetry, stated it in the helper comment |
| 5 | Six lines of comment for a ten-line helper | Compressed; the block now also carries #4's note |

Two places I went further than the thread asked:

- **#1** also applies to `waitForWindowsChild`, which loses its last interval identically. Restructured to check-then-sleep so the 3s bound holds on the appear side too.
- **#2** needed five strings, not four — the two plain-cancel tests would still have collided, so they're split by teardown path (job assignment vs PID-walk fallback).

Verification: `gofmt` clean, `GOOS=windows go vet ./pkg/executor/` passes. The tests themselves are `//go:build windows`, so the actual run is on Windows CI.

geunwoonoh (approved):

PR #391, opened 2026-07-30
Consolidate syscall.Credential assembly into one demotion entry point

### LINKED ISSUE #386: Consolidate syscall.Credential assembly across the privilege demotion paths
## Background

PR #380 fixed the macOS `setgroups(2)` `EINVAL` failure and consolidated supplementary-group resolution into a single `utils.ResolveGroups`. Assembling that result into a `syscall.Credential` is still done independently by each caller, and one path still builds its group list on its own. The bug itself came from group handling being duplicated across demotion paths, so unless the assembly is consolidated too, the paths can drift apart the same way again.

## Current code locations

`syscall.Credential` is assembled in four places.

| Location | Group handling |
| --- | --- |
| `pkg/utils/privilege_unix.go:115` (`Demote`) | `utils.ResolveGroups` |
| `pkg/runner/pty_unix.go:28` (`setPtyCmdSysProcAttrAndEnv`) | `utils.ResolveGroups` |
| `pkg/runner/tunnel_linux.go:181` (`getCodeServerCredential`) | `parseGroupIDs` (own implementation) |
| `pkg/runner/tunnel_linux.go:33` (`getNobodyCredential`) | `Groups` not set |

Related functions:

- `pkg/utils/privilege_unix.go:19` `ResolveGroups`, and `:27` `resolveGroups` in the same file
- `maxSupplementaryGroups` in `pkg/utils/ngroups_darwin.go` and `pkg/utils/ngroups_other.go`
- `pkg/runner/tunnel_linux.go:189` `parseGroupIDs`
- `pkg/runner/tunnel_linux.go:40` `parseUserCredentials`

Call sites: `pkg/executor/executor.go:488` `demotePrivileges`, `pkg/executor/handlers/file/file.go:376` `demoteWithHomeDir`, `pkg/executor/handlers/terminal/terminal.go:156`, and `pkg/runner/pty.go:112`.

## How parseGroupIDs differs from resolveGroups

`pkg/runner/tunnel_linux.go:189` `parseGroupIDs` serves the same purpose as `resolveGroups` but behaves differently.

- It does not place the primary gid at the front of the list, so there is no guarantee the primary gid survives truncation.
- It never calls `maxSupplementaryGroups()`, so no platform cap is applied.
- It skips a gid that fails to parse with `continue`. PR #380 changed that same behavior in `resolveGroups` to return an error.
- It swallows a `usr.GroupIds()` failure by returning `nil`, so the caller never sees it.

The `_linux.go` suffix means this file only compiles on Linux, so it is unrelated to the macOS symptom PR #380 fixed. It does, however, remain a separate path from the perspective of keeping group handling in one place.

## Scope

This spans `pkg/utils` and `pkg/runner`, and it can change the privileges a code-server session runs with, so it needs test coverage. The PR #380 review also judged this to be beyond the scope of a bug fix and split it out as follow-up work.

## Reference

- PR #380 review comment: https://github.com/alpacax/alpamon/pull/380#discussion_r3662415919

jisung-02:
## Background

PR #380 consolidated supplementary-group resolution into `utils.ResolveGroups`, settling four rules: the primary gid goes first, the platform `setgroups(2)` cap applies, an unparsable gid is an error, and a `usr.GroupIds()` failure propagates. Only resolution was consolidated, though. Assembly stayed with the callers, so each one called `ResolveGroups` and then built `syscall.SysProcAttr{Credential: ...}` on its own.

As a result `parseGroupIDs` in `pkg/runner/tunnel_linux.go` never went through #380, and it handled all four rules differently.

| Rule | `resolveGroups` | `parseGroupIDs` |
|---|---|---|
| Primary gid placed first | Yes | No |
| `maxSupplementaryGroups()` cap | Applied | Not applied |
| Unparsable gid | Returns an error | Skipped with `continue` |
| `usr.GroupIds()` failure | Propagated | `return nil` |

That last one is the damaging case. A nil `Credential.Groups` makes Go call `setgroups(0, NULL)` (`$GOROOT/src/syscall/exec_linux.go:490-506`), so the code-server session came up with every supplementary group stripped and no error to show for it. The file is `_linux.go`, so none of this was reachable from the darwin symptom #380 fixed.

## Changes

`ResolveGroups` is gone, replaced by `DemotedSysProcAttr(uid, gid, groupIds)`, which resolves the group list and assembles the credential together. Leaving exactly one exported entry point is the point: as long as `ResolveGroups` existed, a new call site could reach for manual assembly again.

`Demote` returns the helper's result and `setPtyCmdSysProcAttrAndEnv` assigns it directly, which also drops `syscall` from `pty_unix.go`'s imports.

`getCodeServerCredential` deletes `parseGroupIDs` and assembles through `utils.DemotedSysProcAttr`. The `usr.GroupIds()` error is now propagated, and an unparsable gid fails the session start instead of being silently omitted.

One more gap surfaced in review: `resolveGroups` only dropped gids that matched the primary gid, so a gid repeated inside `groupIds` stayed in the list, consumed a slot of the `setgroups(2)` cap, and could push a distinct group out. It now dedupes the whole list before capping. On darwin, where the cap is `kern.ngroups` (16), a repeated gid no longer evicts a group the user actually needs. Dedup lives in `resolveGroups` rather than in `DemotedSysProcAttr` so it sits beside the cap it interacts with.

## Permissions

`utils.Demote` could not be substituted wholesale. It returns `nil, nil` for an empty groupname, which callers read as "run as the current user", while code-server accepts an empty groupname and must still demote to the user's primary group. Substituting it would have run code-server as root.

Only the uid/gid resolution was split out into `codeServerIDs`, with assembly shared. That split makes it testable without root, so `TestCodeServerIDs` actually runs in CI and pins the rule as a regression test.

Effective privileges are unchanged. The primary gid now appears in `Groups`, but `Credential.Gid` already grants the same gid via setgid.

## Safety

`getNobodyCredential` is deliberately excluded and keeps assembling its own `Credential`. There the `setgroups(0, NULL)` call is the point: the tunnel daemon should hold no supplementary groups at all, so routing it through the shared helper would widen what it can reach. The reason is noted in place so the exception reads as deliberate.

Production sites that assemble a `syscall.Credential{}` therefore drop from four to two: the shared entry point plus the documented nobody exception. Consolidating all four was never the goal. The nil `Groups` in `getNobodyCredential` is a feature, not a defect.

## Tests

`TestDemotedSysProcAttr` adds five cases: uid/gid placement, a requested group the user is not a member of still assembling while membership is reported as false, duplicate group ids being deduped, no supplementary groups leaving only the primary gid, and an unparsable gid failing the whole demotion. `TestResolveGroups_AppliesPlatformCap` now guards the exported surface, so it moved to `TestDemotedSysProcAttr_AppliesPlatformCap`. The existing ten `TestResolveGroups` cases covering the unexported `resolveGroups` are kept, and an eleventh pins dedup-before-capping: with a cap of three, `groupIds` of `80, 80, 12` under primary gid 99 used to yield `99, 80, 80` and drop 12 entirely.

`TestCodeServerIDs` adds four cases: the primary-group fallback for an empty groupname, groupname taking precedence, an unknown groupname erroring, and an unparsable uid erroring. `tunnel_linux_test.go` only compiles on Linux, so it was run in Docker (`golang:1.25`) to verify.

`go build` and `go vet` pass on darwin, linux, and windows. `GOOS=linux golangci-lint run` reports 0 issues and `go test ./pkg/utils/... ./pkg/runner/...` passes. To confirm the branch adds no dead code, `origin/main` was checked out in a separate worktree and the same linters (`unused`, `unparam`, `ineffassign`, `staticcheck`) were run across all three platforms; the finding lists matched down to the line numbers.

`getCodeServerCredential` and `Demote` themselves remain untested because they require root, unchanged from before. Demotion was not verified against a real production server.

Supersedes #390, which GitHub closed when its head branch was renamed. The commits are unchanged (`b54fbe9`), and the two Copilot review threads resolved there are not re-opened here.

Closes #386

### DIFF
--- a/pkg/runner/tunnel_linux.go
+++ b/pkg/runner/tunnel_linux.go
@@ -171,12 +171,13 @@ func getCodeServerCredential(username, groupname string) (*syscall.SysProcAttr, 
 		return nil, fmt.Errorf("failed to list groups of user %s: %w", username, err)
 	}
 
-	log.Debug().Msgf("Demoting code-server to user %s (group: %s).", username, groupname)
-
 	sysProcAttr, _, err := utils.DemotedSysProcAttr(uid, gid, groupIds)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to build credentials for user %s: %w", username, err)
 	}
+
+	log.Debug().Msgf("Demoting code-server to user %s (group: %s).", username, groupname)
+
 	return sysProcAttr, nil
 }
 
--- a/pkg/runner/tunnel_linux.go
+++ b/pkg/runner/tunnel_linux.go
@@ -181,9 +181,10 @@ func getCodeServerCredential(username, groupname string) (*syscall.SysProcAttr, 
 	return sysProcAttr, nil
 }
 
-// codeServerIDs cannot defer to utils.Demote: Demote treats an empty groupname
-// as "do not demote", which would leave the code-server session running as root.
-func codeServerIDs(usr *user.User, groupname string) (uint32, uint32, error) {
+// codeServerIDs resolves the uid and gid to demote code-server to: groupname when
+// set, otherwise the user's primary group. It cannot defer to utils.Demote, which
+// treats an empty groupname as "do not demote" and would leave the session as root.
+func codeServerIDs(usr *user.User, groupname string) (uid, gid uint32, err error) {
 	gidStr := usr.Gid
 	if groupname != "" {
 		group, err := user.LookupGroup(groupname)
--- a/pkg/runner/tunnel_linux.go
+++ b/pkg/runner/tunnel_linux.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"syscall"
 
+	"github.com/alpacax/alpamon/v2/pkg/utils"
 	"github.com/rs/zerolog/log"
 )
 
@@ -29,6 +30,8 @@ func getNobodyCredential() (*syscall.SysProcAttr, error) {
 		return nil, fmt.Errorf("failed to parse nobody credentials: %w", err)
 	}
 
+	// Groups stays nil so Go calls setgroups(0, NULL), dropping the daemon's
+	// supplementary groups; resolving them here would widen what it can reach.
 	return &syscall.SysProcAttr{
 		Credential: &syscall.Credential{
 			Uid: uid,
@@ -158,47 +161,36 @@ func getCodeServerCredential(username, groupname string) (*syscall.SysProcAttr, 
 		return nil, fmt.Errorf("user %s not found: %w", username, err)
 	}
 
-	// Use user's primary group if groupname is empty
-	gidStr := usr.Gid
-	if groupname != "" {
-		group, err := user.LookupGroup(groupname)
-		if err != nil {
-			return nil, fmt.Errorf("group %s not found: %w", groupname, err)
-		}
-		gidStr = group.Gid
-	}
-
-	uid, gid, err := parseUserCredentials(usr.Uid, gidStr)
+	uid, gid, err := codeServerIDs(usr, groupname)
 	if err != nil {
 		return nil, err
 	}
 
-	groups := parseGroupIDs(usr)
+	groupIds, err := usr.GroupIds()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list groups of user %s: %w", username, err)
+	}
 
 	log.Debug().Msgf("Demoting code-server to user %s (group: %s).", username, groupname)
 
-	return &syscall.SysProcAttr{
-		Credential: &syscall.Credential{
-			Uid:    uid,
-			Gid:    gid,
-			Groups: groups,
-		},
-	}, nil
+	sysProcAttr, _, err := utils.DemotedSysProcAttr(uid, gid, groupIds)
+	if err != nil {
+		return nil, err
+	}
+	return sysProcAttr, nil
 }
 
-func parseGroupIDs(usr *user.User) []uint32 {
-	groupIds, err := usr.GroupIds()
-	if err != nil {
-		return nil
+// codeServerIDs cannot defer to utils.Demote: Demote treats an empty groupname
+// as "do not demote", which would leave the code-server session running as root.
+func codeServerIDs(usr *user.User, groupname string) (uint32, uint32, error) {
+	gidStr := usr.Gid
+	if groupname != "" {
+		group, err := user.LookupGroup(groupname)
+		if err != nil {
+			return 0, 0, fmt.Errorf("group %s not found: %w", groupname, err)
+		}
+		gidStr = group.Gid
 	}
 
-	groups := make([]uint32, 0, len(groupIds))
-	for _, gidStr := range groupIds {
-		gid, err := strconv.ParseUint(gidStr, 10, 32)
-		if err != nil {
-			continue
-		}
-		groups = append(groups, uint32(gid))
-	}
-	return groups
+	return parseUserCredentials(usr.Uid, gidStr)
 }
--- a/pkg/runner/tunnel_linux.go
+++ b/pkg/runner/tunnel_linux.go
@@ -171,7 +171,7 @@ func getCodeServerCredential(username, groupname string) (*syscall.SysProcAttr, 
 		return nil, fmt.Errorf("failed to list groups of user %s: %w", username, err)
 	}
 
-	sysProcAttr, _, err := utils.DemotedSysProcAttr(uid, gid, groupIds)
+	sysProcAttr, err := utils.DemotedSysProcAttr(uid, gid, groupIds)
 	if err != nil {
 		return nil, fmt.Errorf("failed to build credentials for user %s: %w", username, err)
 	}
--- a/pkg/utils/privilege_unix.go
+++ b/pkg/utils/privilege_unix.go
@@ -32,21 +32,27 @@ func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (attr *syscall.SysPr
 
 // resolveGroups builds the supplementary group list for Credential.Groups.
 // The primary gid is placed first so it survives truncation, and the requested
-// gid's membership is reported via groupInList for ValidateGroup. When
+// gid's membership is reported via groupInList for ValidateGroup. Duplicates are
+// dropped before capping so a repeated gid cannot push a distinct one out. When
 // maxGroups > 0 the list is capped to that many entries.
 func resolveGroups(gid uint32, groupIds []string, maxGroups int) (groups []uint32, groupInList bool, err error) {
 	groups = make([]uint32, 0, len(groupIds)+1)
 	groups = append(groups, gid)
+	seen := map[uint32]struct{}{gid: {}}
 	for _, gidStr := range groupIds {
 		gidUint, err := strconv.ParseUint(gidStr, 10, 32)
 		if err != nil {
 			return nil, false, fmt.Errorf("invalid supplementary group id: %w", err)
 		}
-		if uint32(gidUint) == gid {
+		group := uint32(gidUint)
+		if group == gid {
 			groupInList = true
+		}
+		if _, dup := seen[group]; dup {
 			continue
 		}
-		groups = append(groups, uint32(gidUint))
+		seen[group] = struct{}{}
+		groups = append(groups, group)
 	}
 	if maxGroups > 0 && len(groups) > maxGroups {
 		log.Debug().Int("requested", len(groups)).Int("cap", maxGroups).
--- a/pkg/utils/privilege_unix.go
+++ b/pkg/utils/privilege_unix.go
@@ -12,12 +12,22 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-// ResolveGroups builds the supplementary group list for Credential.Groups,
-// applying the current platform's setgroups(2) cap. It is the single entry
-// point shared by every privilege-demotion path (Demote and the websh PTY),
-// so group handling stays identical across them.
-func ResolveGroups(gid uint32, groupIds []string) (groups []uint32, groupInList bool, err error) {
-	return resolveGroups(gid, groupIds, maxSupplementaryGroups())
+// DemotedSysProcAttr is the single entry point for demotions that keep the
+// user's supplementary groups (Demote, websh PTY, code-server): one place for
+// setgroups(2) capping, so those paths cannot drift. Demotions that must drop
+// the group list build their own Credential with Groups left nil.
+func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (attr *syscall.SysProcAttr, groupInList bool, err error) {
+	groups, groupInList, err := resolveGroups(gid, groupIds, maxSupplementaryGroups())
+	if err != nil {
+		return nil, false, err
+	}
+	return &syscall.SysProcAttr{
+		Credential: &syscall.Credential{
+			Uid:    uid,
+			Gid:    gid,
+			Groups: groups,
+		},
+	}, groupInList, nil
 }
 
 // resolveGroups builds the supplementary group list for Credential.Groups.
@@ -99,7 +109,7 @@ func Demote(username, groupname string, opts DemoteOptions) (*DemoteResult, erro
 		return nil, err
 	}
 
-	groups, groupInList, err := ResolveGroups(uint32(gid), groupIds)
+	sysProcAttr, groupInList, err := DemotedSysProcAttr(uint32(uid), uint32(gid), groupIds)
 	if err != nil {
 		return nil, err
 	}
@@ -111,13 +121,7 @@ func Demote(username, groupname string, opts DemoteOptions) (*DemoteResult, erro
 	log.Debug().Msgf("Demote permission to match user: %s, group: %s.", username, groupname)
 
 	return &DemoteResult{
-		SysProcAttr: &syscall.SysProcAttr{
-			Credential: &syscall.Credential{
-				Uid:    uint32(uid),
-				Gid:    uint32(gid),
-				Groups: groups,
-			},
-		},
-		User: usr,
+		SysProcAttr: sysProcAttr,
+		User:        usr,
 	}, nil
 }
--- a/pkg/utils/privilege_unix.go
+++ b/pkg/utils/privilege_unix.go
@@ -16,10 +16,15 @@ import (
 // user's supplementary groups (Demote, Websh PTY, code-server): one place for
 // setgroups(2) capping, so those paths cannot drift. Demotions that must drop
 // the group list build their own Credential with Groups left nil.
-//
-// groupInList reports whether gid was among groupIds, for callers that enforce
-// membership; it is not an error here, since only the caller knows if it should be.
-func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (attr *syscall.SysProcAttr, groupInList bool, err error) {
+func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (*syscall.SysProcAttr, error) {
+	attr, _, err := demotedSysProcAttr(uid, gid, groupIds)
+	return attr, err
+}
+
+// demotedSysProcAttr additionally reports whether gid was among groupIds. Only
+// Demote's ValidateGroup consumes that, so it stays off the exported signature
+// rather than making both other call sites discard it.
+func demotedSysProcAttr(uid, gid uint32, groupIds []string) (attr *syscall.SysProcAttr, groupInList bool, err error) {
 	groups, groupInList, err := resolveGroups(gid, groupIds, maxSupplementaryGroups())
 	if err != nil {
 		return nil, false, err
@@ -120,7 +125,7 @@ func Demote(username, groupname string, opts DemoteOptions) (*DemoteResult, erro
 		return nil, err
 	}
 
-	sysProcAttr, groupInList, err := DemotedSysProcAttr(uint32(uid), uint32(gid), groupIds)
+	sysProcAttr, groupInList, err := demotedSysProcAttr(uint32(uid), uint32(gid), groupIds)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/utils/privilege_unix.go
+++ b/pkg/utils/privilege_unix.go
@@ -16,6 +16,9 @@ import (
 // user's supplementary groups (Demote, Websh PTY, code-server): one place for
 // setgroups(2) capping, so those paths cannot drift. Demotions that must drop
 // the group list build their own Credential with Groups left nil.
+//
+// groupInList reports whether gid was among groupIds, for callers that enforce
+// membership; it is not an error here, since only the caller knows if it should be.
 func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (attr *syscall.SysProcAttr, groupInList bool, err error) {
 	groups, groupInList, err := resolveGroups(gid, groupIds, maxSupplementaryGroups())
 	if err != nil {
--- a/pkg/utils/privilege_unix.go
+++ b/pkg/utils/privilege_unix.go
@@ -55,7 +55,9 @@ func resolveGroups(gid uint32, groupIds []string, maxGroups int) (groups []uint3
 		groups = append(groups, group)
 	}
 	if maxGroups > 0 && len(groups) > maxGroups {
-		log.Debug().Int("requested", len(groups)).Int("cap", maxGroups).
+		// Warn, not Debug: the process silently loses file access it was granted,
+		// and Debug is off in production (config.go sets InfoLevel).
+		log.Warn().Int("requested", len(groups)).Int("cap", maxGroups).
 			Msg("Supplementary group list truncated to the platform setgroups limit.")
 		groups = groups[:maxGroups]
 	}
--- a/pkg/utils/privilege_unix.go
+++ b/pkg/utils/privilege_unix.go
@@ -13,7 +13,7 @@ import (
 )
 
 // DemotedSysProcAttr is the single entry point for demotions that keep the
-// user's supplementary groups (Demote, websh PTY, code-server): one place for
+// user's supplementary groups (Demote, Websh PTY, code-server): one place for
 // setgroups(2) capping, so those paths cannot drift. Demotions that must drop
 // the group list build their own Credential with Groups left nil.
 func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (attr *syscall.SysProcAttr, groupInList bool, err error) {
--- a/pkg/utils/privilege_unix.go
+++ b/pkg/utils/privilege_unix.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/user"
+	"slices"
 	"strconv"
 	"syscall"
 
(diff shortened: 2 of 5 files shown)


pkg/utils/privilege_unix_test.go
```diff
@@ -89,17 +95,76 @@ func TestResolveGroups(t *testing.T) {
...
 		groupIds = append(groupIds, strconv.Itoa(i))
 	}
 
-	groups, _, err := ResolveGroups(99, groupIds)
+	attr, _, err := DemotedSysProcAttr(501, 99, groupIds)
 	require.NoError(t, err)
 
 	// The exported entry point must apply maxSupplementaryGroups(); anything
 	// larger would make setgroups(2) fail with EINVAL on a capped platform.
+	groups := attr.Credential.Groups
```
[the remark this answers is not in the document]
jisung-02: Fixed in 54cd4d9. Added `require.NotNil(t, attr)` and `require.NotNil(t, attr.Credential)` after the `NoError` check, matching the guards `TestDemotedSysProcAttr` already had, so a nil attr with a nil error fails readably instead of panicking.

pkg/utils/privilege_unix.go
```diff
@@ -12,31 +12,47 @@ import (
...
-// ResolveGroups builds the supplementary group list for Credential.Groups,
-// applying the current platform's setgroups(2) cap. It is the single entry
-// point shared by every privilege-demotion path (Demote and the websh PTY),
-// so group handling stays identical across them.
-func ResolveGroups(gid uint32, groupIds []string) (groups []uint32, groupInList bool, err error) {
-	return resolveGroups(gid, groupIds, maxSupplementaryGroups())
+// DemotedSysProcAttr is the single entry point for demotions that keep the
+// user's supplementary groups (Demote, websh PTY, code-server): one place for
+// setgroups(2) capping, so those paths cannot drift. Demotions that must drop
+// the group list build their own Credential with Groups left nil.
```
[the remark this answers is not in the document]
jisung-02: Fixed in 54cd4d9. The doc comment now reads "Websh PTY", per the CLAUDE.md convention.

For the record, lowercase `websh` also appears in comments in `pkg/runner/auth_manager.go`, `pkg/runner/auth_manager_tracker_test.go`, and `pkg/executor/handlers/user/user_test.go`. Those predate this branch and some of them refer to the literal `Kind=websh` value rather than the product name, so they are left alone here rather than widening this diff.

geunwoonoh (commented):
## Review summary

Follow-up to #380. That PR consolidated group *resolution* into `utils.ResolveGroups` but left *assembly* with each caller, so `pkg/runner/tunnel_linux.go`'s `parseGroupIDs` never adopted any of the four settled rules — most damagingly, it returned `nil` on a `GroupIds()` failure, which makes Go call `setgroups(0, NULL)` and strips every supplementary group from the code-server session with no error surfaced.

This PR replaces `ResolveGroups` with `DemotedSysProcAttr(uid, gid, groupIds)`, which resolves *and* assembles, leaving one exported entry point so a new call site can't reach for manual assembly. `resolveGroups` additionally now dedupes the whole list before capping. `getNobodyCredential` is deliberately excluded with an in-code rationale.

### What I verified independently

| Check | Result |
|---|---|
| `go build ./...` on darwin / `GOOS=linux` / `GOOS=windows` | clean (after `go generate ./pkg/db/ent`) |
| `go vet ./pkg/utils/... ./pkg/runner/...` darwin + `GOOS=linux` | clean |
| `go test ./pkg/utils/... ./pkg/runner/... -p 1` | `ok utils 0.455s`, `ok runner 32.562s` |
| Linux-only `TestCodeServerIDs` in `golang:1.25` Docker | 4/4 PASS |
| `ResolveGroups` / `parseGroupIDs` residue | gone; only `TestResolveGroups` remains, testing unexported `resolveGroups` — correct |
| Production `syscall.Credential{}` sites | 2, as the PR body claims |

`golangci-lint` isn't installed locally, so the PR's "0 issues" claim is the one thing I could not reproduce.

## Critical issues (0)

None.

## Warnings (2)

**1. `pkg/runner/tunnel_linux.go:176` — the new hard-failure path loses the username, while the path right above it keeps it.**

This PR converts an unparsable gid in the group database from a silent skip into a session-start failure. That's the right call, but the diagnostic is now what operators depend on, and it's thinner than its neighbor: line 171 wraps with `failed to list groups of user %s`, line 178 returns bare. `pkg/runner/editor.go:345` passes the error through unchanged, so the operator sees:

```
failed to get user credentials: invalid supplementary group id: strconv.ParseUint: parsing "x": invalid syntax
```

No username, no indication that `/etc/group` is the place to look.

**2. `pkg/utils/privilege_unix.go:58` — group truncation is invisible in production.** `pkg/config/config.go:112-114` sets `InfoLevel` unless debug mode is on, so on darwin a user with >16 groups silently loses supplementary groups — lost file access with zero signal. Pre-existing from #380, but this PR now owns the function, and its doc comment names the `setgroups(2)` cap as the reason the entry point exists. `log.Warn()` costs nothing and this is a degradation, not a routine event.

## Suggestions (6)

1. **`groupInList` is discarded at 2 of 3 call sites** (`pty_unix.go:22`, `tunnel_linux.go:176`) — a naked bool in the middle of an exported 3-value signature that only `Demote` reads.
2. **`pkg/utils/privilege_unix.go:41` — the `seen` map buys nothing at these list sizes.** `slices.Contains(groups, group)` reproduces the behavior exactly and drops three lines of bookkeeping. Benchmarked; details inline.
3. **`pkg/utils/privilege_unix.go:15-19` — the godoc doesn't document `groupInList`,** which is the non-obvious part of the exported signature.
4. **`pkg/runner/tunnel_linux.go:183` — the doc comment says only what the function *isn't*,** and two same-typed unnamed `uint32` returns are ambiguous at the call site.
5. **`pkg/utils/privilege_unix_test.go:108-129` — cases 1-4 of `TestDemotedSysProcAttr` re-pin `resolveGroups` semantics** already covered at `:24-50`.
6. **`pkg/runner/tunnel_linux.go:174` — the "Demoting code-server to user %s" log fires before assembly can fail.**

Full reasoning for each is in the inline comments.

## Highlights

- **The diagnosis is verifiable, not asserted.** The four-rule divergence table plus the `$GOROOT/src/syscall/exec_linux.go:490-506` citation for why nil `Groups` becomes `setgroups(0, NULL)` — and the note that the file is `_linux.go`, so the darwin symptom #380 chased could never have reached it.
- **`getNobodyCredential` excluded on purpose, documented in place** (`tunnel_linux.go:33-34`). Nil `Groups` there is the feature. That comment is what stops a future "consolidate the last one" pass from quietly widening the tunnel daemon's reach — the exception is now self-defending rather than relying on a reviewer remembering.
- **`resolveGroups` keeps `maxGroups` as a parameter while the wrapper injects `maxSupplementaryGroups()`.** Cap behavior gets pinned deterministically instead of through a platform-dependent test.
- **`codeServerIDs` was extracted for testability, not tidiness.** The empty-groupname → primary-group rule is the one that would run code-server as root if it regressed, and it now has a root-free test with the intent written into it (`tunnel_linux_test.go:11`).
- **Dedup landed in `resolveGroups`, beside the cap it interacts with,** rather than in the wrapper. The two behaviors are coupled and now sit together.
- **Zero added cost from centralizing the sysctl** — `maxSupplementaryGroups()` is already `sync.OnceValue`-memoized (`ngroups_darwin.go:24`), so moving it behind the per-demotion entry point is free. I checked this specifically because it looked like a per-call syscall.
- **Error handling is now genuinely consistent across demotion paths.** `pty.go:512-515` already propagated the `GroupIds()` failure; `tunnel_linux.go:169-172` now matches it, which is exactly the drift the PR set out to close.
- Both Copilot comments from #390 (`Websh` capitalization, nil-guard in the cap test) are resolved in `54cd4d9`, and the reply also catalogued the pre-existing lowercase `websh` occurrences it deliberately left alone.

---

**Verdict**: COMMENTED

Nothing blocks merge. Warning 1 is a three-line change and I'd do it before merging, since this PR is precisely what makes that error path reachable. Warning 2 and all six suggestions are fine as follow-ups.

**Checklist**

- [x] Code follows stack conventions — two godoc nits (suggestions 3, 4)
- [x] Tests included and passing — verified on darwin and in Linux Docker
- [x] No security vulnerabilities — the `Groups` set is unchanged in effect; `Credential.Gid` already grants the prepended gid via setgid, and the `getNobodyCredential` exception is preserved
- [x] Documentation updated if needed — no user-facing docs affected

pkg/runner/tunnel_linux.go
```diff
@@ -158,47 +161,36 @@ func getCodeServerCredential(username, groupname string) (*syscall.SysProcAttr,
...
-	}, nil
-}
-
-func parseGroupIDs(usr *user.User) []uint32 {
-	groupIds, err := usr.GroupIds()
+	sysProcAttr, _, err := utils.DemotedSysProcAttr(uid, gid, groupIds)
 	if err != nil {
-		return nil
+		return nil, err
 	}
```
geunwoonoh: **Warning: the new hard-failure path loses the username, while the path right above it keeps it.**

This PR converts an unparsable gid in the group database from a silent skip into a session-start failure. That's the right call, but the diagnostic is now what operators depend on, and it's thinner than its neighbour: line 171 wraps with `failed to list groups of user %s`, line 178 returns bare. `pkg/runner/editor.go:345` passes the error through unchanged, so the operator sees:

```
failed to get user credentials: invalid supplementary group id: strconv.ParseUint: parsing "x": invalid syntax
```

No username, no indication that `/etc/group` is the place to look.

```suggestion
	sysProcAttr, _, err := utils.DemotedSysProcAttr(uid, gid, groupIds)
	if err != nil {
		return nil, fmt.Errorf("failed to build credentials for user %s: %w", username, err)
	}
```

pkg/utils/privilege_unix.go
```diff
@@ -12,31 +12,47 @@ import (
...
+		}
+		if _, dup := seen[group]; dup {
 			continue
 		}
-		groups = append(groups, uint32(gidUint))
+		seen[group] = struct{}{}
+		groups = append(groups, group)
 	}
 	if maxGroups > 0 && len(groups) > maxGroups {
 		log.Debug().Int("requested", len(groups)).Int("cap", maxGroups).
```
geunwoonoh: **Warning: group truncation is invisible in production.**

`pkg/config/config.go:112-114` sets `InfoLevel` unless debug mode is on, so on darwin a user with more than 16 groups silently loses supplementary groups — lost file access with zero signal.

Pre-existing from #380, but this PR now owns the function, and its doc comment names the `setgroups(2)` cap as the reason the entry point exists. `log.Warn()` costs nothing and this is a degradation, not a routine event.

```suggestion
		log.Warn().Int("requested", len(groups)).Int("cap", maxGroups).
```

pkg/utils/privilege_unix.go
```diff
@@ -12,31 +12,47 @@ import (
...
-// ResolveGroups builds the supplementary group list for Credential.Groups,
-// applying the current platform's setgroups(2) cap. It is the single entry
-// point shared by every privilege-demotion path (Demote and the websh PTY),
-// so group handling stays identical across them.
-func ResolveGroups(gid uint32, groupIds []string) (groups []uint32, groupInList bool, err error) {
-	return resolveGroups(gid, groupIds, maxSupplementaryGroups())
+// DemotedSysProcAttr is the single entry point for demotions that keep the
+// user's supplementary groups (Demote, Websh PTY, code-server): one place for
+// setgroups(2) capping, so those paths cannot drift. Demotions that must drop
+// the group list build their own Credential with Groups left nil.
```
geunwoonoh, line 19: Suggestion: the godoc doesn't document `groupInList`, which is the non-obvious part of the signature.

`resolveGroups` documents it, but the exported wrapper is the godoc-visible surface — a caller reading `go doc` learns why the entry point exists but not what the middle return value means. One sentence would close it.

pkg/utils/privilege_unix.go
```diff
@@ -12,31 +12,47 @@ import (
...
+func DemotedSysProcAttr(uid, gid uint32, groupIds []string) (attr *syscall.SysProcAttr, groupInList bool, err error) {
```
geunwoonoh: Suggestion: `groupInList` is discarded at 2 of 3 call sites (`pty_unix.go:22`, `tunnel_linux.go:176`) — a naked bool in the middle of an exported 3-value signature that only `Demote` reads.

The cleanest removal is to have `Demote` check membership directly, since `grp.Gid` and the `groupIds` entries both come from `os/user`:

```go
if opts.ValidateGroup && !slices.Contains(groupIds, grp.Gid) {
```

Caveat worth weighing: that's a string compare, so it's less robust than the current numeric one against non-normalised gid strings. If you'd rather keep the numeric check, a small result struct would also work — or leave the signature and just document the return.

pkg/utils/privilege_unix.go
```diff
@@ -12,31 +12,47 @@ import (
...
+		group := uint32(gidUint)
+		if group == gid {
 			groupInList = true
+		}
+		if _, dup := seen[group]; dup {
 			continue
 		}
-		groups = append(groups, uint32(gidUint))
+		seen[group] = struct{}{}
+		groups = append(groups, group)
```
geunwoonoh: Suggestion: the `seen` map buys nothing at these list sizes.

`gid` is already `groups[0]`, so `slices.Contains(groups, group)` reproduces the current behaviour exactly — including the primary-gid case — and drops the map plus its declaration.

I benchmarked both (M4 Pro, `-benchtime=200000x`):

```
BenchmarkResolveGroupsMap8-14       301.5 ns/op    232 B/op   4 allocs/op
BenchmarkResolveGroupsSlice8-14      58.3 ns/op     48 B/op   1 allocs/op
BenchmarkResolveGroupsMap32-14     1049   ns/op   1256 B/op   8 allocs/op
BenchmarkResolveGroupsSlice32-14    313.6 ns/op    144 B/op   1 allocs/op
```

Be clear about what that means: 3-5x and 4-8 allocs down to 1, but the absolute saving is ~250ns and every demotion is immediately followed by a fork+exec costing milliseconds. Even though `Demote` runs on every command execution (`pkg/executor/executor.go:489`), **this is not a performance fix.** Take it for the three lines of bookkeeping it removes from a six-line loop, not for the nanoseconds.

Needs `"slices"` added to the import block, which the suggestion below can't include:

```suggestion
	for _, gidStr := range groupIds {
		gidUint, err := strconv.ParseUint(gidStr, 10, 32)
		if err != nil {
			return nil, false, fmt.Errorf("invalid supplementary group id: %w", err)
		}
		group := uint32(gidUint)
		if group == gid {
			groupInList = true
		}
		if slices.Contains(groups, group) {
			continue
		}
		groups = append(groups, group)
```

pkg/utils/privilege_unix_test.go
```diff
@@ -89,17 +95,78 @@ func TestResolveGroups(t *testing.T) {
...
+		{
+			name: "duplicate group ids deduped", uid: 501, gid: 99,
+			groupIds:   []string{"80", "80", "12"},
+			wantGroups: []uint32{99, 80, 12},
+		},
+		{
+			name: "no supplementary groups keeps only the primary gid", uid: 0, gid: 0,
+			groupIds:   nil,
+			wantGroups: []uint32{0},
+		},
```
geunwoonoh: Suggestion: cases 1-4 here re-pin `resolveGroups` semantics.

Hoisting, dedup, non-member reporting, and the empty list are all already covered by `TestResolveGroups` at `:24-50`. The wrapper's own contract is narrower: uid/gid land on `Credential`, `groupInList` passes through, an error yields a nil attr, and the platform cap is applied (already its own test at `:156`). Two cases cover that.

Table cases are cheap, so this isn't worth a change by itself — it's just where to trim if the table grows.

pkg/runner/tunnel_linux.go
```diff
@@ -158,47 +161,36 @@ func getCodeServerCredential(username, groupname string) (*syscall.SysProcAttr,
...
 		return nil, err
 	}
 
-	groups := parseGroupIDs(usr)
+	groupIds, err := usr.GroupIds()
+	if err != nil {
+		return nil, fmt.Errorf("failed to list groups of user %s: %w", username, err)
+	}
 
 	log.Debug().Msgf("Demoting code-server to user %s (group: %s).", username, groupname)
```
geunwoonoh: Suggestion: this log fires before assembly can fail, so it can claim a demotion that doesn't happen.

Moving it below the `DemotedSysProcAttr` error check keeps it truthful. Minor, but it costs nothing now that the call below it has a real failure path.

pkg/runner/tunnel_linux.go
```diff
@@ -158,47 +161,36 @@ func getCodeServerCredential(username, groupname string) (*syscall.SysProcAttr,
...
 	}
+	return sysProcAttr, nil
+}
 
-	groups := make([]uint32, 0, len(groupIds))
-	for _, gidStr := range groupIds {
-		gid, err := strconv.ParseUint(gidStr, 10, 32)
+// codeServerIDs cannot defer to utils.Demote: Demote treats an empty groupname
+// as "do not demote", which would leave the code-server session running as root.
+func codeServerIDs(usr *user.User, groupname string) (uint32, uint32, error) {
```
geunwoonoh: Suggestion: the doc comment says only what the function *isn't*.

Per [Go Doc Comments](https://go.dev/doc/comment) the first sentence should state what the function does — a godoc reader currently learns why it can't defer to `utils.Demote` but not what it returns. The rationale is worth keeping, just not as the whole comment.

Two same-typed unnamed `uint32` returns are also ambiguous at the call site; naming them fixes that without touching the body.

```suggestion
// codeServerIDs resolves the uid and gid to demote code-server to: groupname
// when set, otherwise the user's primary group.
//
// It cannot defer to utils.Demote, which treats an empty groupname as "do not
// demote" and would leave the session running as root.
func codeServerIDs(usr *user.User, groupname string) (uid, gid uint32, err error) {
```

pkg/runner/tunnel_linux.go
jisung-02: Fixed in 8fde79b. `getCodeServerCredential` now wraps with `failed to build credentials for user %s`, matching the `GroupIds()` path directly above it.

Checked the sibling demotion path for the same gap: `pty_unix.go:22` also returns the `DemotedSysProcAttr` error bare, and `pty.go:113` wraps it with only `failed to set process credentials`, so no identity reaches the operator there either. Left alone — `username` is not in scope in that function, and the `safeUint32Credentials` error above it is bare too, so it is at least internally consistent. Happy to carry the uid through in a follow-up if you think it is worth it.

pkg/utils/privilege_unix.go
jisung-02: Fixed in 8fde79b — `log.Warn()`, with the reason noted in place since the level choice is the whole point of the change.

Verified your `config.go` citation independently: `zerolog.SetGlobalLevel(zerolog.InfoLevel)` in the else branch at `pkg/config/config.go:111-115`, so Debug was indeed unreachable in production.

One adjacent spot left alone: `tunnel_linux.go:155` logs the not-running-as-root fallback at Debug, while `privilege_unix.go:89` logs the same situation at Warn. Pre-existing, and arguably a routine path for code-server rather than a degradation, so I did not fold it into this change.

pkg/utils/privilege_unix.go
jisung-02: Fixed in 6dd877c, though not by any of the three options — I read the requirement as "keep it off the exported signature" rather than "remove the value", which lets the numeric compare stand.

The 3-value form is now unexported `demotedSysProcAttr`, which `Demote` calls directly since both live in this package. `DemotedSysProcAttr` is `(*syscall.SysProcAttr, error)`, and the two external call sites no longer discard anything.

On your caveat, I did not want to trade the compare away. `slices.Contains(groupIds, grp.Gid)` matches raw strings from `os/user`, and in pure-Go mode both sides come from parsing `/etc/group`, so a non-normalised entry (`020` vs `20`) would slip past a check that the current `ParseUint` + `uint32` compare catches. Rare, but closing exactly that class of drift is what this PR is for. The result struct would have added a type to carry one bool that both call sites reach through for the only field they want. Cost of my version is the two-line wrapper.

pkg/utils/privilege_unix.go
jisung-02: Documented in e480620, then removed again in 6dd877c along with the return value itself — see my reply on the `groupInList` suggestion for why.

The exported signature is now `(*syscall.SysProcAttr, error)`, so there is no longer an undocumented return to explain; the sentence moved onto the unexported `demotedSysProcAttr`, where the bool now lives. Net effect is what you asked for by a different route: `go doc` no longer shows a naked bool at all.

pkg/utils/privilege_unix.go
jisung-02: Fixed in bd006e5, taken on your framing: three lines of bookkeeping out of a six-line loop, not the nanoseconds. I did not reproduce the benchmark — the reasoning that a ~250ns saving is invisible against the fork+exec that follows every demotion holds on its own, and `pkg/executor/executor.go:489` confirms the call frequency you cite.

`gid` is already `groups[0]` before the loop, so `slices.Contains(groups, group)` subsumes the map seed. `TestResolveGroups` passes unchanged, including `duplicate supplementary gids deduped before capping`, which is the case that would fail if the primary-gid path had regressed.

pkg/runner/tunnel_linux.go
jisung-02: Fixed in 8fde79b — moved below the error check. It landed in the same commit as the username wrapping, since both changes sit on those four lines.

Your framing is right about the cause: before this PR nothing below that log could fail, because the old `parseGroupIDs` used `continue` for an unparsable gid and `return nil` for a `GroupIds()` failure. The PR is what turned the log into a claim that could be false.

pkg/runner/tunnel_linux.go
jisung-02: Fixed in e480620. Took your wording, compressed to three lines: first sentence states what it resolves, the `utils.Demote` rationale follows it. Named returns applied.

Tried the same named-return fix on `parseUserCredentials` (`tunnel_linux.go:43`), which has the identical `(uint32, uint32, error)` ambiguity, and reverted it. Its body already binds `uid` and `gid` as the `uint64` results of `ParseUint`, so naming the returns gives `no new variables on left side of :=`. Fixing it properly means renaming those locals, which is more than a signature nit — left for a follow-up.

pkg/utils/privilege_unix_test.go
jisung-02: Applied in a9d8e6c, with one deviation: only the dedup case was deleted outright. Your `:24-50` reference holds for it — the `resolveGroups` table runs the same `80,80,12` input with `maxGroups: 3`, which is strictly stronger, since it also proves dedup precedes capping. The wrapper version ran without a cap on linux, so it could not have caught an ordering regression.

The `no supplementary groups` case turned out not to be covered there: `TestResolveGroups` has no nil/empty-list row (`empty gid rejected` is `[]string{""}`, a different input), so deleting it would have dropped that path. Moved it into `TestResolveGroups` instead, without the `uid` field, which is not `resolveGroups`'s concern.

The three remaining cases are the contract you listed, minus the cap, which has its own test.

Noting that you flagged this one as not worth a change by itself — done at the author's request, not read as a blocking ask.

geunwoonoh (approved):

PR #392, opened 2026-07-30
refactor(runner): drop dead localSudoRequests map and its read paths

jisung-02:
## Background

`AuthManager.localSudoRequests` has no write path. When it was introduced (5589152) it held sudo requests from non-Websh (local) users while the agent waited for a server decision. b5c7de9 then changed local requests to be answered synchronously inside `handleSudoApprovalRequest`, on the grounds that alpacon-server rejects them anyway, and deleted the registration while leaving the lookups behind. 47bbe5c later added one more read of the now-empty map in `isRequestPending`.

The map is therefore always empty, so the local branches in `HandleSudoApprovalResponse`, `isRequestPending`, and `cleanupTimeoutRequest` can never match. Reading the code suggests local requests are tracked as pending, which misrepresents what actually happens (an immediate synchronous answer).

Reviewing the completion channel on the way out surfaced a second, unrelated defect in the same function, which this PR also fixes. Review then added a regression test and two small cleanups. Each part is described separately below.

### Removing the dead map

Removed the `localSudoRequests` field, its two initialization sites, and all four places that read it.

- `HandleSudoApprovalResponse`: the "2. find in local user requests" block and the debug log that dumped the whole map
- `isRequestPending`: the local-request lookup branch
- `cleanupTimeoutRequest`: the "2. Local user" block
- The initialization in the `NewEmptyAuthManager` and `newTestAuthManager` helpers

The `// 1. Alpacon user` / `// 2. Local user` numbering comments went with them, since only one branch remains in each function.

`handleSudoApprovalRequest` also allocated `completionChan` at the top of the function even though the local path returns before ever using it. The allocation now happens where the channel is registered.

This part changes no behavior. Local sudo is still decided synchronously according to the `blockLocalSudo` policy, and Alpacon-user requests still register in `SessionInfo.Requests`.

### Fixing the dropped completion signal

`completionChan` was unbuffered while `signalCompletion` sends on it with a non-blocking `select`, so the signal is dropped whenever `HandleSudoApprovalResponse` runs before the waiter reaches its `select`. That window is not just the few instructions after the channel is registered: it spans the blocking REST call in `sendSudoRequestWithRetry`, so a response arriving on the websocket read loop while that call is in flight is lost.

The channel is now buffered with capacity 1, so the first signal always lands. This matches the other two non-blocking send sites in the tree, both already buffered-of-1 (`pkg/runner/client.go:76`, `pkg/pluginclient/client.go:215`); `completionChan` was the only outlier. `close(ch)` was rejected as an alternative because `removeCompletionChannel` and the timeout path can reach the same request, so a second close would panic.

Copilot raised this on the moved allocation line. The defect predates this PR: only the position of `make(...)` changed, not its buffering.

The `default` branch in `signalCompletion` was commented as "Channel already signaled or closed". Nothing in the tree closes a completion channel, and a send on a closed channel panics rather than taking `default`, so half of that comment described an unreachable case. It now states what capacity 1 actually makes the branch mean: a signal is already buffered.

### Responding to review

The buffering fix now has a regression test in `pkg/runner/auth_manager_completion_test.go`. Review was right that the property which regressed is testable even though the race is not. The test suggested in review allocated the channel itself, so it would have kept passing once the production capacity was dropped, pinning nothing. The allocation moved into a `newCompletionChannel` helper instead, so the test exercises the channel `handleSudoApprovalRequest` actually gets. Confirmed by temporarily restoring the unbuffered channel: the new test fails, and passes again once the capacity is back.

Two further tests cover the surrounding contract: that `signalCompletion` never blocks on a full buffer, since it runs on the websocket read loop, and that a signal for an already-removed request is a no-op, which the timeout path can cause.

`newTestAuthManager` now delegates to `NewEmptyAuthManager()`. The two helpers were byte-identical, which is why the dead-field deletion above had to be made twice.

## Safety

This touches the sudo approval path, so the authorization logic was checked separately from the cleanup. The branches that decide approval (the `blockLocalSudo` check and the session lookup in `pkg/runner/auth_manager.go:384-402`) are unchanged, and everything removed is a lookup against a map that is always empty.

The dropped-signal fix does not change any approval outcome either. When the signal is lost, `HandleSudoApprovalResponse` has already written the response to the PAM client (`auth_manager.go:566`) and closed the connection (`:577`); at the 30s timeout `isRequestPending` returns false because the response handler removed the entry, so the timeout branch takes its no-op path. The cost is a goroutine and a map entry that linger for 30 seconds, not a wrong decision. The fix removes that lingering.

One narrow window stays open, raised in review: the request becomes findable by `HandleSudoApprovalResponse` when it is registered into `SessionInfo.Requests`, a few lines before the completion channel is registered. A response landing in that gap is still dropped. It is only reachable from a stale or duplicated `request_id`, since the request has not been sent to the server yet, and closing it properly means restructuring the critical section. Tracked in #395.

## Tests

- `go build ./...` and `go vet ./...` pass
- `go test ./... -p 1` passes (45 packages); `go test ./pkg/runner/ -p 1` re-run after each commit
- The removal adds no tests: it deletes unreachable branches, so there is no new behavior to verify, and no test covered the three functions to begin with.
- Three tests added in `pkg/runner/auth_manager_completion_test.go`. An end-to-end reproduction of the dropped signal remains impractical, needing a live Unix socket, a server round trip, and a won race inside `sendSudoRequestWithRetry`, but the property that regressed is checkable on its own. An earlier revision of this section called the fix untestable, which conflated the race with the property.
- The regression test was verified to fail without the fix, not just to pass with it: reverting the capacity to `make(chan struct{})` makes `TestNewCompletionChannel_SurvivesSignalBeforeReceive` fail.
- CI was green on 24878c6 across all 24 checks, covering the distro build-and-test jobs, golangci-lint, govulncheck, CodeQL, and the go.mod tidy check. The run for the two review-response commits on top is in flight as this is written.
- Locally, `golangci-lint run ./pkg/runner/...` reports `commit_types.go:154 field expireDate is unused`. That finding is present on main as well and is unrelated to this change. The local version is v2.12.2 while CI pins v2.9.0.

## Follow-ups

Three items raised in review are tracked separately rather than folded in here, so this PR keeps its no-behavior-change, no-API-change scope.

- #394: remove the write-only `SudoRequest.RequestID` field, which is the same dead-field class this PR deletes but changes an exported type
- #395: register the request and its completion channel in one critical section, closing the narrow window described under Safety
- #396: the retry-failure path sends two responses and closes the connection twice; predates this PR and needs the PAM client's read behavior confirmed first

## Checklist

- [x] Traced the code paths for every change: the removal and the two cleanups are behavior-preserving, and the buffering fix affects only cleanup timing
- [x] No remaining references to the removed symbol (`grep -rn localSudoRequests`)
- [x] Checked the rest of the tree for the same unbuffered-plus-non-blocking-send pattern; no other occurrence
- [x] Build, vet, and the full test suite pass
- [x] The regression test was verified to fail without the fix

### DIFF
--- a/pkg/runner/auth_manager.go
+++ b/pkg/runner/auth_manager.go
@@ -745,7 +745,8 @@ func (am *AuthManager) signalCompletion(requestID string) {
 		select {
 		case ch <- struct{}{}:
 		default:
-			// Channel already signaled or closed
+			// Already signaled: the capacity-1 buffer is full and the waiter
+			// only needs one signal. Nothing closes these channels.
 		}
 	}
 }
--- a/pkg/runner/auth_manager.go
+++ b/pkg/runner/auth_manager.go
@@ -438,11 +438,7 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		Str("command_id", sudoApprovalReq.CommandID).
 		Msg("Alpacon user sudo request")
 
-	// Store completion channel so HandleSudoApprovalResponse can unblock the wait below.
-	// Buffered so signalCompletion's non-blocking send lands even if the response is
-	// processed before this goroutine reaches the select.
-	completionChan := make(chan struct{}, 1)
-	am.storeCompletionChannel(sudoApprovalReq.RequestID, completionChan)
+	completionChan := am.newCompletionChannel(sudoApprovalReq.RequestID)
 
 	// Send Sudo Approval request to the alpacon-server with retry
 	if err := am.sendSudoRequestWithRetry(sudoApprovalReq); err != nil {
@@ -724,10 +720,15 @@ func RegisterCommandPID(pid int, commandID, username string) func() {
 	return func() { am.RemovePIDCommandMapping(pid, commandID) }
 }
 
-func (am *AuthManager) storeCompletionChannel(requestID string, ch chan struct{}) {
+// Capacity 1 is load-bearing: signalCompletion sends without blocking, so an
+// unbuffered channel drops the signal when the response beats the waiter's select.
+func (am *AuthManager) newCompletionChannel(requestID string) chan struct{} {
+	ch := make(chan struct{}, 1)
 	am.mu.Lock()
 	am.completionChannels[requestID] = ch
 	am.mu.Unlock()
+
+	return ch
 }
 
 func (am *AuthManager) removeCompletionChannel(requestID string) {
@@ -745,8 +746,7 @@ func (am *AuthManager) signalCompletion(requestID string) {
 		select {
 		case ch <- struct{}{}:
 		default:
-			// Already signaled: the capacity-1 buffer is full and the waiter
-			// only needs one signal. Nothing closes these channels.
+			// Already signaled; the waiter only needs one.
 		}
 	}
 }
--- a/pkg/runner/auth_manager.go
+++ b/pkg/runner/auth_manager.go
@@ -438,8 +438,10 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		Str("command_id", sudoApprovalReq.CommandID).
 		Msg("Alpacon user sudo request")
 
-	// Store completion channel so HandleSudoApprovalResponse can unblock the wait below
-	completionChan := make(chan struct{})
+	// Store completion channel so HandleSudoApprovalResponse can unblock the wait below.
+	// Buffered so signalCompletion's non-blocking send lands even if the response is
+	// processed before this goroutine reaches the select.
+	completionChan := make(chan struct{}, 1)
 	am.storeCompletionChannel(sudoApprovalReq.RequestID, completionChan)
 
 	// Send Sudo Approval request to the alpacon-server with retry
--- a/pkg/runner/auth_manager.go
+++ b/pkg/runner/auth_manager.go
@@ -133,7 +133,6 @@ type AuthManager struct {
 	pidToSessionMap    map[int]*SessionInfo
 	controlClient      *ControlClient
 	listener           net.Listener
-	localSudoRequests  map[string]*SudoRequest
 	completionChannels map[string]chan struct{}
 	session            *scheduler.Session
 	blockLocalSudo     bool
@@ -154,7 +153,6 @@ func GetAuthManager(controlClient *ControlClient, session *scheduler.Session) *A
 	authManagerOnce.Do(func() {
 		authManager = &AuthManager{
 			pidToSessionMap:    make(map[int]*SessionInfo),
-			localSudoRequests:  make(map[string]*SudoRequest),
 			completionChannels: make(map[string]chan struct{}),
 			session:            session,
 		}
@@ -162,10 +160,6 @@ func GetAuthManager(controlClient *ControlClient, session *scheduler.Session) *A
 
 	if authManager.controlClient == nil {
 		authManager.controlClient = controlClient
-	}
-
-	if authManager.localSudoRequests == nil {
-		authManager.localSudoRequests = make(map[string]*SudoRequest)
 	}
 
 	if authManager.completionChannels == nil {
@@ -386,9 +380,6 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		return
 	}
 
-	// Create completion channel to signal when response is received
-	completionChan := make(chan struct{})
-
 	sid, sidOK := sessionID(sudoApprovalReq.PID)
 	am.mu.Lock()
 	session, exists := am.lookupSessionLocked(sid, sidOK, sudoApprovalReq.PPID)
@@ -447,7 +438,8 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		Str("command_id", sudoApprovalReq.CommandID).
 		Msg("Alpacon user sudo request")
 
-	// Store completion channel for this request
+	// Store completion channel so HandleSudoApprovalResponse can unblock the wait below
+	completionChan := make(chan struct{})
 	am.storeCompletionChannel(sudoApprovalReq.RequestID, completionChan)
 
 	// Send Sudo Approval request to the alpacon-server with retry
@@ -545,7 +537,6 @@ func (am *AuthManager) HandleSudoApprovalResponse(response SudoApprovalResponse)
 	am.mu.Lock()
 	var sudoRequest *SudoRequest
 
-	// 1. find in alpacon user requests
 	for _, session := range am.pidToSessionMap {
 		if req, exists := session.Requests[response.RequestID]; exists {
 			delete(session.Requests, response.RequestID)
@@ -554,22 +545,10 @@ func (am *AuthManager) HandleSudoApprovalResponse(response SudoApprovalResponse)
 			break
 		}
 	}
-
-	// 2. find in local user requests
-	if sudoRequest == nil {
-		if req, exists := am.localSudoRequests[response.RequestID]; exists {
-			delete(am.localSudoRequests, response.RequestID)
-			sudoRequest = req
-			log.Debug().Msgf("Found local user request for ID: %s", response.RequestID)
-		} else {
-			log.Debug().Msgf("Request ID %s not found in localSudoRequests", response.RequestID)
-		}
-	}
 	am.mu.Unlock()
 
 	if sudoRequest == nil {
 		am.mu.RLock()
-		log.Debug().Msgf("Current localSudoRequests: %+v", am.localSudoRequests)
 		for _, session := range am.pidToSessionMap {
 			log.Debug().Msgf("Session %s requests: %+v", session.SessionID, session.Requests)
 		}
@@ -775,16 +754,10 @@ func (am *AuthManager) isRequestPending(requestID string) bool {
 	am.mu.RLock()
 	defer am.mu.RUnlock()
 
-	// Check alpacon user requests
 	for _, session := range am.pidToSessionMap {
 		if _, exists := session.Requests[requestID]; exists {
 			return true
 		}
-	}
-
-	// Check local user requests
-	if _, exists := am.localSudoRequests[requestID]; exists {
-		return true
 	}
 
 	return false
@@ -802,7 +775,6 @@ func (am *AuthManager) Stop() {
 func (am *AuthManager) cleanupTimeoutRequest(requestID string, approved bool, reason string) {
 	am.mu.Lock()
 
-	// 1. Alpacon user
 	for _, session := range am.pidToSessionMap {
 		if req, exists := session.Requests[requestID]; exists {
 			delete(session.Requests, requestID)
@@ -813,17 +785,6 @@ func (am *AuthManager) cleanupTimeoutRequest(requestID string, approved bool, re
 			}
 			return
 		}
-	}
-
-	// 2. Local user
-	if req, exists := am.localSudoRequests[requestID]; exists {
-		delete(am.localSudoRequests, requestID)
-		am.mu.Unlock()
-		if req.Connection != nil {
-			am.sendSudoApprovalResponse(req.Connection, SudoApprovalRequest{RequestID: requestID}, approved, reason)
-			_ = req.Connection.Close()
-		}
-		return
 	}
 
 	am.mu.Unlock()
--- a/pkg/runner/auth_manager_tracker_test.go
+++ b/pkg/runner/auth_manager_tracker_test.go
@@ -13,7 +13,6 @@ import (
 func newTestAuthManager() *AuthManager {
 	return &AuthManager{
 		pidToSessionMap:    make(map[int]*SessionInfo),
-		localSudoRequests:  make(map[string]*SudoRequest),
 		completionChannels: make(map[string]chan struct{}),
 	}
 }
--- a/pkg/runner/auth_manager_tracker_test.go
+++ b/pkg/runner/auth_manager_tracker_test.go
@@ -7,14 +7,9 @@ import (
 	"time"
 )
 
-// newTestAuthManager returns an AuthManager with only the fields that the
-// tracker tests need populated. It intentionally does not start the
-// socket listener, so tests stay fast and isolated.
+// The shipped helper already builds exactly what these tests need.
 func newTestAuthManager() *AuthManager {
-	return &AuthManager{
-		pidToSessionMap:    make(map[int]*SessionInfo),
-		completionChannels: make(map[string]chan struct{}),
-	}
+	return NewEmptyAuthManager()
 }
 
 // TestAddPIDCommandMapping_RegistersCommandKind verifies that Commands
--- a/pkg/runner/auth_manager_completion_test.go
+++ b/pkg/runner/auth_manager_completion_test.go
@@ -1,0 +1,54 @@
+package runner
+
+import (
+	"testing"
+	"time"
+)
+
+// The channel must come from newCompletionChannel, not from the test: a channel
+// the test allocates itself would keep passing after the production capacity is
+// dropped, pinning nothing.
+func TestNewCompletionChannel_SurvivesSignalBeforeReceive(t *testing.T) {
+	am := newTestAuthManager()
+	ch := am.newCompletionChannel("req-1")
+
+	// The websocket read loop can process the response while the waiter is still
+	// inside sendSudoRequestWithRetry, before it reaches its select.
+	am.signalCompletion("req-1")
+
+	select {
+	case <-ch:
+	default:
+		t.Fatal("completion signal was dropped before the waiter received it")
+	}
+}
+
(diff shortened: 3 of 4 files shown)


pkg/runner/auth_manager.go
```diff
@@ -447,7 +438,8 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		Str("command_id", sudoApprovalReq.CommandID).
 		Msg("Alpacon user sudo request")
 
-	// Store completion channel for this request
+	// Store completion channel so HandleSudoApprovalResponse can unblock the wait below
+	completionChan := make(chan struct{})
 	am.storeCompletionChannel(sudoApprovalReq.RequestID, completionChan)
```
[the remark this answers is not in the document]
jisung-02: Fixed in bd38405. `completionChan` is now `make(chan struct{}, 1)`.

The window is wider than the few instructions after `storeCompletionChannel`: it spans the blocking REST call in `sendSudoRequestWithRetry` (auth_manager.go:446), so a response arriving on the websocket read loop while that call is still in flight would hit `default` and be dropped. The waiter then sits for 30s even though the PAM client already received its answer and its connection was closed at auth_manager.go:577 — so the cost is a lingering goroutine rather than a wrong authorization decision, but it is a real leak.

Buffering also matches the convention already in the tree: the other two non-blocking send sites are both buffered-of-1 (`pkg/runner/client.go:76`, `pkg/pluginclient/client.go:215`). `completionChan` was the only outlier.

Rejected `close(ch)` as an alternative — `removeCompletionChannel` and the timeout path can reach the same request, so a second close would panic.

pkg/runner/auth_manager.go
```diff
@@ -447,7 +438,10 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		Str("command_id", sudoApprovalReq.CommandID).
 		Msg("Alpacon user sudo request")
 
-	// Store completion channel for this request
+	// Store completion channel so HandleSudoApprovalResponse can unblock the wait below.
+	// Buffered so signalCompletion's non-blocking send lands even if the response is
+	// processed before this goroutine reaches the select.
+	completionChan := make(chan struct{}, 1)
```
junho226: Warning: this fix is testable—the PR body's claim that it isn't conflates the *race* with the *property*.

The body says the fix "is not covered either: reproducing the dropped signal means winning a race inside `sendSudoRequestWithRetry`, and the path needs a live Unix socket plus a server round trip." That's true of an end-to-end reproduction. But the property that regressed—a signal sent before the receiver reaches its `select` must not be dropped—is directly testable against `storeCompletionChannel`/`signalCompletion` using the existing `newTestAuthManager()` helper. No socket, no server, no race to win.

I confirmed this in a worktree: the test below passes on this branch, and against the pre-fix unbuffered channel the `default` branch fires (i.e. it fails).

```go
func TestSignalCompletion_NotDroppedBeforeSelect(t *testing.T) {
	am := newTestAuthManager()
	ch := make(chan struct{}, 1)
	am.storeCompletionChannel("req-1", ch)

	// Response processed before the waiter reaches its select.
	am.signalCompletion("req-1")

	select {
	case <-ch:
	default:
		t.Fatal("completion signal was dropped")
	}
}
```

Worth adding before merge. Without it, a future refactor can drop the `, 1` and every test stays green—which is exactly how the field this PR deletes became dead in the first place.

pkg/runner/auth_manager.go
```diff
@@ -447,7 +438,10 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
...
 	am.storeCompletionChannel(sudoApprovalReq.RequestID, completionChan)
```
junho226: Suggestion: one narrow window stays open after this fix.

The buffered channel closes the wide window (the blocking REST call), but the request becomes findable by `HandleSudoApprovalResponse` at line 428, while the channel is only stored here at 445—after `am.mu.Unlock()` at 432. A response landing in that gap still hits `signalCompletion`'s `exists == false` path and is dropped, reproducing the same 30-second lingering goroutine this fix targets.

Not a blocker: it's practically unreachable today, since the request hasn't reached the server yet and only a stale or duplicated `request_id` could get there. But registering the channel inside the same critical section makes "registered ⇒ signalable" hold unconditionally instead of resting on timing:

```go
	completionChan := make(chan struct{}, 1)
	session.Requests[sudoApprovalReq.RequestID] = &SudoRequest{
		RequestID:  sudoApprovalReq.RequestID,
		Connection: unixConn,
	}
	am.completionChannels[sudoApprovalReq.RequestID] = completionChan
	am.mu.Unlock()
```

The trade-off is honest: this moves `make(...)` back up and partially undoes the readability win you were after here. Your call whether the invariant is worth it.

pkg/runner/auth_manager_tracker_test.go
```diff
@@ -13,7 +13,6 @@ import (
 func newTestAuthManager() *AuthManager {
 	return &AuthManager{
 		pidToSessionMap:    make(map[int]*SessionInfo),
-		localSudoRequests:  make(map[string]*SudoRequest),
 		completionChannels: make(map[string]chan struct{}),
 	}
```
junho226: Suggestion: after this PR, `newTestAuthManager` is byte-identical to `NewEmptyAuthManager()`.

Both live in `package runner` with no build tag, so this test file can call the exported helper directly. This PR had to make the same one-line deletion in both places—that duplicated edit is the drift cost of keeping them separate, and the next field added to `AuthManager` pays it again.

```suggestion
	return NewEmptyAuthManager()
```

(The doc comment above still reads correctly with the delegation, since `NewEmptyAuthManager` documents the same "only what the tracker tests need, no socket listener" intent.)

junho226 (dismissed):
## Summary

Two changes in one PR, correctly described as separate in the body:

1. **Dead-code removal.** `AuthManager.localSudoRequests` has no write path, so its four read sites (`HandleSudoApprovalResponse`, `isRequestPending`, `cleanupTimeoutRequest`, plus a debug dump) could never match. Removed along with the field, its two init sites, and both test-helper initializations.
2. **A real defect fix.** `completionChan` becomes `make(chan struct{}, 1)`, so `signalCompletion`'s non-blocking send can no longer be dropped when the response is processed before the waiter reaches its `select`.

I verified the dead-map claim independently rather than taking it from the body: `git grep localSudoRequests main` returns only reads and `delete`s—no assignment anywhere in the tree—so all four removed branches were genuinely unreachable. `go vet ./pkg/runner/` is clean and `go test ./pkg/runner/` passes on `bd38405`. (`go build ./...` fails in a fresh worktree because the gitignored Ent output isn't generated yet—a setup artifact, not a PR defect; I did not verify the full-tree build.)

## Critical issues (0)

None. The removal is behavior-preserving and the buffering fix is correct—the channel is per-request and freshly allocated, so the first signal always lands in the empty buffer.

## Warnings (1)

**1. `pkg/runner/auth_manager.go:444`—the buffering fix is testable, contrary to the PR body.** See the inline comment. Short version: the body conflates the *race* with the *property*. The property that regressed is testable in ~15 lines against `storeCompletionChannel`/`signalCompletion` with the existing `newTestAuthManager()` helper—no socket, no server, no race. Worth landing before merge; without it a future refactor can drop the `, 1` and every test stays green, which is exactly how the field this PR deletes became dead in the first place.

## Suggestions (4)

**2. `pkg/runner/auth_manager_tracker_test.go:14`**—`newTestAuthManager` is now byte-identical to `NewEmptyAuthManager()`. Inline.

**3. `pkg/runner/auth_manager.go:445`**—one narrow window remains open between request registration and channel storage. Inline.

**4. `pkg/runner/auth_manager.go:748`—stale comment, now contradicted by the design** (outside the diff, so noted here):

```go
default:
	// Channel already signaled or closed
```

Nothing in the tree closes a completion channel, and the PR body explicitly rejects `close(ch)`. A send on a closed channel panics rather than taking `default`, so "or closed" describes an impossible branch. With capacity 1, `default` now means exactly one thing: already signaled.

**5. Answering your checklist question on `SudoRequest.RequestID`.** Verified: written at line 429, read nowhere. `SudoRequest` is constructed only in `pkg/runner`, and the map key already carries the id.

My recommendation: **remove it, but in a follow-up, not here.** It is the same write-only-dead-field class this PR exists to delete, so leaving it lets the codebase keep making the claim the PR is disproving. The counter-argument (exported field on an exported type in a `/v2` module) is real but thin in practice—nothing outside `pkg/runner` touches the type, and alpamon ships as an agent binary rather than a consumed library. Keeping it separate preserves this PR's "no behavior change, no API change" framing and gives the removal its own release-note line.

**Pre-existing, out of scope, noted only because this PR is the cleanup pass over this function:** the retry-failure path at 448-455 sends two responses on the same connection and closes it twice—`sendSudoApprovalResponse` at 450, then `cleanupTimeoutRequest` finds the request registered at 428 and sends + closes again at 785-786, then 453 closes once more. Harmless if the PAM client reads once and disconnects.

## Highlights

- **The buffered-of-1 convention claim checks out.** The only three non-blocking send sites in production code are `pkg/runner/client.go:271` (`CollectorRestartChan`, cap 1), `pkg/pluginclient/client.go:230` (`configUpdateCh`, cap 1), and this one. `completionChan` was the outlier; it now matches.
- **Rejecting `close(ch)` is correct**, and for the stated reason—`removeCompletionChannel` and the timeout path can both reach the same request, so a second close would panic.
- **The commit archaeology (5589152 → b5c7de9 → 47bbe5c) makes "why is this dead" answerable without re-deriving it.** That is what let me verify the claim in one grep instead of reconstructing the history.
- **Correctly scoped safety analysis.** The `blockLocalSudo` branch and the session lookup at 384-402 are untouched, and the write-error path's consequence (a lingering goroutine, not a wrong decision) is stated accurately rather than overstated.

---

**Verdict: approved.** No critical issues. The one warning—a ~15-line regression test pinning the buffering property—is worth landing before merge, but the code is correct as-is and I don't want to block a cleanup on it.

**Checklist**

- [x] Code follows stack conventions (sentence-case comments, no spaced em-dashes)
- [ ] Tests included—none added; see warning 1, the buffering property is cheaply testable
- [x] No security vulnerabilities—authorization branches unchanged, verified separately
- [x] Documentation updated if needed—n/a

pkg/runner/auth_manager.go
jisung-02: Fixed in 33a8b55, though not with the test as written.

The suggested test allocates the channel itself (`ch := make(chan struct{}, 1)`), so it exercises the test's own channel rather than the production one. Reverting `:444` to `make(chan struct{})` leaves it green: it asserts the property against a channel that already has the property. That is the same "every test stays green" failure mode you were guarding against.

So the allocation moved into a `newCompletionChannel(requestID)` helper, which is what the test calls. The capacity and the comment explaining why it is load-bearing now live in one place, and the call site drops from two lines to one.

Verified the way you'd want it verified. With the helper's capacity reverted to `make(chan struct{})`, `TestNewCompletionChannel_SurvivesSignalBeforeReceive` fails with "completion signal was dropped before the waiter received it"; restoring `, 1` makes it pass.

Two more tests cover the adjacent contract: that `signalCompletion` does not block on a full buffer, since it runs on the websocket read loop, and that signalling an already-removed request is a no-op, which the timeout path can cause.

pkg/runner/auth_manager.go
jisung-02: Confirmed, and filed as #395 rather than fixed here.

The window is real: the request is findable at `:428` while the channel is only registered after the unlock, so a response arriving in that gap hits `exists == false` and is dropped exactly as before the fix.

Not folding it in, for the reason you named plus one more. It is only reachable from a stale or duplicated `request_id`, and your snippet moves `make(...)` back to the call site, which now also un-pins the regression test added in 33a8b55, since that test calls `newCompletionChannel`. #395 records both the invariant argument and an alternative that avoids the trade-off: split the helper into a lock-free inner form plus a locking wrapper, then call the inner one inside the existing critical section. The capacity and the test entry point stay in the helper that way.

pkg/runner/auth_manager_tracker_test.go
jisung-02: Fixed in 4157fe9. `newTestAuthManager` now returns `NewEmptyAuthManager()`.

Confirmed the two bodies were identical before the change. The doc comment is down to one line, since the delegation itself says the rest.

jisung-02:
Thanks for the review. All four inline items are handled, two of them as follow-up issues. Pushed as 24878c6, 33a8b55, and 4157fe9; replies to the individual threads are inline.

Suggestion 4, the stale comment on `signalCompletion`'s `default` branch: fixed in 24878c6, then shortened in 33a8b55 to `// Already signaled; the waiter only needs one.` You were right that "or closed" described a branch the code cannot reach. I also swept the other non-blocking send sites for the same wrong claim, since a comment that describes an impossible case tends to be copied around. `pkg/runner/client.go` logs "already requested, skipping duplicate signal" and `pkg/pluginclient/client.go` says "Channel full: drop pending and enqueue newest", both accurate; `internal/pool/pool.go`'s "Job queue is closed" is a receive-with-ok on a queue that really does get closed. This was the only wrong one.

Suggestion 5, `SudoRequest.RequestID`: taking your recommendation. Filed as #394 instead of removing it here, so this PR keeps its no-API-change framing and the removal gets its own release-note line.

Your out-of-scope note on the retry-failure path is #396. I traced it before filing and it differs slightly from the description: `sendSudoApprovalResponse` only writes, it never closes, so the connection is closed twice (inside `cleanupTimeoutRequest`, then again at `:449` where the error is discarded) rather than three times. Two responses do go out, and the second is worse than merely duplicated: `cleanupTimeoutRequest` passes `SudoApprovalRequest{RequestID: requestID}`, so `username`, `pid`, `ppid`, `command`, `session_id`, and `command_id` are all empty in it. Whether that reaches the PAM client at all depends on its read behavior, which lives in another repo and I have not checked, so #396 states that as an open question rather than assuming it.

The PR body is updated to match. The Tests section no longer claims the buffering fix is untestable, it records that the regression test was verified to fail without the fix, and there is a new Follow-ups section pointing at #394, #395, and #396.

junho226 (approved):
## Re-review: delta since bd38405

Delta pass over `bd38405..4157fe9` (3 commits, +64 −14 in 3 files), not a re-read of the whole PR — my earlier approval was dismissed by the push.

### You were right about my test, and the fix is better than the one I proposed

My snippet allocated the channel itself, so it asserted the property against a channel that already had it. That is the exact failure mode I opened the warning with, and I walked into it. Moving the allocation into `newCompletionChannel` fixes it at the root: the capacity, the comment explaining why the capacity is load-bearing, and the entry point the test calls are now one thing, so there is no longer a place to drop `, 1` that the test cannot see.

Verified rather than taken from the reply: with the helper reverted to `make(chan struct{})`, `TestNewCompletionChannel_SurvivesSignalBeforeReceive` fails with `completion signal was dropped before the waiter received it`; restoring `, 1` makes it pass. Your claim holds exactly as stated.

`newTestAuthManager` → `NewEmptyAuthManager()` (4157fe9) is behavior-preserving — the two bodies were byte-identical, `NewEmptyAuthManager` is in the same package with no build tag, and `storeCompletionChannel` has no leftover callers anywhere in the tree. The `default`-branch comment (24878c6) now describes the only branch that is reachable.

Both deferrals are filed with accurate bodies (#394 for the write-only `RequestID`, #395 for the registration window), and #395's alternative — a lock-free inner form plus a locking wrapper, called inside the existing critical section — is the right shape. It keeps the capacity and the test entry point in the helper, which is the property this delta just bought.

### Suggestion (1, non-blocking)

**[pkg/runner/auth_manager_completion_test.go:48-54]** `TestSignalCompletion_UnknownRequestIsNoop` survives deleting the guard it reads as covering.

Its doc comment points at `signalCompletion` reaching "nothing registered", which is the `exists == false` path. I checked rather than reasoned it: replacing

```go
	if exists {
		select {
		case ch <- struct{}{}:
		default:
		}
	}
```

with an unguarded `select` leaves all three tests in the file passing. A send on a nil channel inside a `select` with a `default` is never ready, so `default` is taken — from here the nil case and the full-buffer case are indistinguishable.

So the `exists` check is redundant with the `select`, and what the test pins is one step removed: that a future refactor to a bare `ch <- struct{}{}` would hang forever on the unregistered path. That is worth having on the websocket read loop, and it is a different claim than the comment makes. Either say what it pins and drop the now-redundant guard, or keep the guard as stated intent and note that the `select`'s `default` is what makes the unregistered path safe. The second is smaller and keeps `signalCompletion` readable without knowing nil-channel `select` semantics — probably why the guard is there.

Given you explicitly checked the other two tests for being green for the wrong reason, this is the one spot in the delta where the answer is "partly".

### Verified

- `go vet ./pkg/runner/` clean; full `go test ./pkg/runner/` green (50 s).
- Capacity revert → the new regression test fails, as claimed.
- `exists`-guard removal → all three new tests still pass (the basis for the suggestion above).

---

**Verdict**: APPROVED

PR #399, opened 2026-07-31
refactor(runner): drop redundant SudoRequest.RequestID field

### LINKED ISSUE #394: Cleanup: remove the write-only SudoRequest.RequestID field
## Background

While removing the write-only dead field `AuthManager.localSudoRequests` in #392, one more field of the same class turned up. `SudoRequest.RequestID` (`pkg/runner/auth_manager.go:59`) is assigned at `pkg/runner/auth_manager.go:429` and read nowhere in the tree.

This was independently verified during the #392 review. `SudoRequest` is constructed only inside `pkg/runner`, and the request id is already carried by the key of the `SessionInfo.Requests` map, so the field just duplicates the key.

It was left out of #392 because it is an exported field on an exported type, which makes its removal an API surface change and blurs that PR's "no behavior change, no API change" scope. Splitting it out follows the reviewer's recommendation.

## Proposed change

- Remove the `RequestID` field from `SudoRequest`
- Drop the assignment from the struct literal at `pkg/runner/auth_manager.go:428-431`

## Notes

`SudoApprovalRequest.RequestID` and `SudoApprovalResponse.RequestID` are fields on different types that go over the socket protocol, so they are unaffected. Alpamon ships as an agent binary rather than a consumed library, so the practical fallout is nil, but the type is exported from the `/v2` module and the removal is worth a release-note line.

Line references are as of the #392 branch head.

Follow-up from #392.

jisung-02:
## Background

`SudoRequest` values are only ever reached through `SessionInfo.Requests`, a `map[string]*SudoRequest` that is already keyed by the request ID. The struct nevertheless carried a `RequestID` field holding the same value, and no code ever read it back off the struct. The same identifier existed in both the map key and the map value, leaving it ambiguous which one was authoritative.

## Changes

Removed the `RequestID` field from `SudoRequest` and dropped the corresponding assignment at its single construction site, `handleSudoApprovalRequest` (`pkg/runner/auth_manager.go:427`). `SudoRequest` now holds only `Connection net.Conn`.

To correlate a pending sudo request with its ID, use the `session.Requests` map key rather than a field on the value. `cleanupTimeoutRequest` (`pkg/runner/auth_manager.go:784`) already follows this: it passes the `requestID` parameter it used for the map lookup, not a field off the struct.

Collapsing the `SudoRequest` wrapper entirely into `map[string]net.Conn` was considered and rejected. It falls outside this branch's scope, and it would touch four files including `pkg/runner/pty.go` and the tests for little gain.

## Safety

`SudoRequest` has no JSON tags and is never serialized, so the wire format exchanged with alpacon-server is unaffected. The protocol-facing `SudoApprovalRequest.RequestID` and `SudoApprovalResponse.RequestID` fields are unchanged.

A repository-wide `git grep` across all file types confirms no remaining reader of the removed field: every surviving `SudoRequest` reference is the type definition, a map declaration or initialization, the construction site, or a local variable, and the only field access is `req.Connection`. Every surviving `RequestID` reference belongs to `SudoApprovalRequest` or `SudoApprovalResponse`. No Markdown, script, or config file mentions the field.

## Testing

- `go build ./pkg/runner/` passes
- `go vet ./pkg/runner/` passes
- `go test ./pkg/runner/ -p 1` passes (35.3s)

The sudo approval flow was not exercised manually on a real host.

Closes #394

### DIFF
--- a/pkg/runner/auth_manager.go
+++ b/pkg/runner/auth_manager.go
@@ -56,7 +56,6 @@ func (s *SessionInfo) effectiveKind() string {
 }
 
 type SudoRequest struct {
-	RequestID  string
 	Connection net.Conn
 }
 
@@ -426,7 +425,6 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 	}
 
 	session.Requests[sudoApprovalReq.RequestID] = &SudoRequest{
-		RequestID:  sudoApprovalReq.RequestID,
 		Connection: unixConn,
 	}
 	am.mu.Unlock()


junho226 (approved):
Dead field removal, and the claim that it is dead holds up.

## What I checked

| Item | Result |
| --- | --- |
| Remaining readers | None. Every surviving `RequestID` reference belongs to `SudoApprovalRequest` / `SudoApprovalResponse`, the type definitions (63, 76, 97), or `SudoApprovalRequest{RequestID: requestID}` at 784. Nothing reads it off a `SudoRequest` value |
| The two consumers, read directly | `HandleSudoApprovalResponse` (534-553) uses `sudoRequest` only for the nil check and `Connection`; `cleanupTimeoutRequest` (784) passes the `requestID` parameter it already used for the lookup. Matches the description |
| Unkeyed struct literals | The only `&SudoRequest{` is the keyed one at 427, so no positional literal breaks on a field removal |
| Whether the field was ever observable | `auth_manager.go:546` logs `session.Requests` with `%+v`, but the values are pointers, so that line prints addresses and the map keys — the request IDs were already visible there, and the struct field never was. Confirms it was dead rather than merely unread |
| Wire format | `SudoRequest` has no JSON tags and no serialization path; the two protocol structs are untouched |
| Local verification | `go build ./pkg/runner/`, `go vet ./pkg/runner/`, `go test ./pkg/runner/ -p 1` (ok, 58.6s) all pass |
| CI | Analyze (actions/go) and the full Build and Test distro matrix are green |

## Notes

Agreed on rejecting the collapse to `map[string]net.Conn`. It would spread across four files for no behavioral gain, and the named wrapper is what keeps a second per-request field cheap to add later.

Grounding the change in `cleanupTimeoutRequest`'s existing use of the map key is the right framing — this is conforming to a precedent already in the file, not introducing a rule.

**Approve.**

PR #404, opened 2026-08-11
fix(firewall): map dports in batch firewall rule conversion

jisung-02:
## Summary

Fixes a field dropped in the batch firewall rule conversion: `CommandData.ToArgs()` never mapped `dports` when converting the `rules` array (`[]map[string]any`) into `common.FirewallRule`, so multi-port rules reached the backend with no port match at all.

## Problem

`dports` was the only one of the 17 `FirewallRule` fields not mapped in the rules-array conversion. A rule created in the console as "allow tcp 80, 443" arrived at the agent with an empty `DPorts`, and the iptables backend simply omits the port match in that case:

```
# intended                                                 # actual (before this fix)
-A INPUT -p tcp -m multiport --dports 80,443 -j ACCEPT  →  -A INPUT -p tcp -j ACCEPT
```

The console shows two ports while the installed rule opens every port on the protocol. `ValidateRule` does not require any port to be present, so the broken rule passed validation silently.

This loop is not one conversion path among several — it is the sole conversion boundary for every firewall rule field. The handler reads nothing off `CommandArgs` except `ChainName`, `ChainNames`, `OldRuleID`, `Operation`, `RuleID` and `Rules`, so the top-level `args.DPorts` / `args.Protocol` / etc. copied at `command.go:164-198` have no reader anywhere in `pkg/executor/handlers/firewall/`. Anything dropped in this loop reaches no backend at all.

## Fix

Add the missing `dports` mapping to the rules conversion loop in `ToArgs()`. JSON numbers unmarshal as `float64`, so numeric elements are asserted as `float64` and converted to `int`, matching the existing `port_start`/`port_end` pattern in the same function. String elements fall back to `strconv.Atoi(strings.TrimSpace(...))`, so a sender that emits `["80", "443"]` — or `["80", " 443"]` — no longer lands in the exact failure this PR fixes through a different door.

Cross-repo contract check: every current server-side sender (batch, add/update, reorder) builds the payload via `FirewallRule.get_arguments()`, which emits `'dports': rule.get_dports_list()` as a list of ints, already stripped. The snapshot-restore path in `firewall/operations.py:261-263` instead builds `'dports': rule_data.get('dports', '').split(',')`, a list of strings split without stripping. Those elements can carry a leading space: `validate_dports` (`firewall/api/serializers.py:278-292`) strips each part for validation but stores `value` verbatim, so a rule entered as "80, 443" is snapshotted as that string and restored as `["80", " 443"]`. Hence the `TrimSpace` — without it the fallback would drop `" 443"` and install the rule one port short, a quieter version of the failure it was added to prevent.

That path is unreachable today because it carries `operation: 'restore'`, which is not among alpamon's firewall sub-operations (`common/types.go:58-63` has only batch, flush, delete, add, update) and is rejected as an unknown operation. The fallback means the day someone adds a `restore` case, that payload does not go live silently portless.

## Tests

- `TestParseCommandData_BatchRuleFields`: a payload populating all 17 rule fields, asserted against a fully populated `common.FirewallRule` in a single `assert.Equal`, so a future drop fails here whichever field it is. A second range rule pins that a `dports` rule does not contaminate its neighbour.
- `TestParseCommandData_BatchRuleDPortsAsStrings`: string `dports` are converted (`["80", "443"]` → `[80 443]`), a space-padded element survives (`["80", " 443"]` → `[80 443]`, the shape restore actually produces), and an unparseable element is skipped rather than poisoning the rule (`["80", "not-a-port"]` → `[80]`).
- Both fail when `internal/protocol/command.go` alone is reverted to the base commit. Dropping only the `TrimSpace` fails the padded case with `expected: []int{80, 443} / actual: []int{80}`.
- `go build ./...`, `go vet ./...`, `gofmt -l .` and `go test ./... -p 1` are clean.

## E2E verification

Verified the full path with a local stack (alpacon-server + celery worker/beat + backhaul-server + proxy-server, all run natively) and a Lima VM (ubuntu 26.04, arm64) running this branch's build as the agent: firewall API (alpacax workspace) → `apply_security_group_task` → Redis Stream → backhaul → alpamon (WebSocket) → iptables.

Batch apply with a dports rule ("8080,9090") plus a range rule (7000-7010):

```
-A INPUT -p tcp -m multiport --dports 8080,9090 -m comment --comment "rule_id:...,type:alpacon" -j ACCEPT
-A INPUT -p tcp -m tcp --dport 7000:7010 -m comment --comment "rule_id:...,type:alpacon" -j ACCEPT
```

Command result: `{"success":true,"applied_rules":2,"rolled_back":false}`

Update apply after PATCHing dports to "8080,9090,9443":

```
-A INPUT -p tcp -m multiport --dports 8080,9090,9443 -m comment --comment "rule_id:...,type:alpacon" -j ACCEPT
```

Since every operation flows through the same conversion loop (`args.Rules[i]`), this exercises the boundary that batch, single add/update and reorder all share. The range rule stayed untouched by dports in both runs.

## Follow-up (not in this PR)

`CommandData.Rules` is `[]map[string]any` and this loop is its only consumer in the tree. All 17 `json` tags on `common.FirewallRule` already match the wire keys the loop asserts by hand, so typing the field as `[]common.FirewallRule` would hand the mapping to `encoding/json`, delete the loop, and make this bug class impossible rather than merely fixed. Left out because strict unmarshaling turns a type mismatch into a hard `ParseCommandData` error instead of a silent skip — arguably the better behavior, but a behavior change well beyond a field-mapping fix.

The server's iptables and nftables adapters also emit `backend`, `table` and `family`, which have no field on `common.FirewallRule`. They do not bite: neither `iptables.go` nor `nftables.go` ever emits a `-t <table>` flag, so alpamon always works on the default table. That is an unimplemented feature rather than a drop this PR should fix.

### DIFF
--- a/internal/protocol/command.go
+++ b/internal/protocol/command.go
@@ -3,6 +3,7 @@ package protocol
 import (
 	"encoding/json"
 	"strconv"
+	"strings"
 
 	"github.com/alpacax/alpamon/v2/pkg/executor/handlers/common"
 )
@@ -249,10 +250,11 @@ func (c *CommandData) ToArgs() *common.CommandArgs {
 						rule.DPorts = append(rule.DPorts, int(n))
 					case string:
 						// Not every sender emits numbers: the snapshot-restore
-						// payload carries dports as strings. Dropping one leaves
-						// the rule with no port match, which opens every port on
-						// the protocol.
-						if port, err := strconv.Atoi(n); err == nil {
+						// payload carries dports as strings, split from a stored
+						// "80, 443" without trimming. Dropping an element leaves
+						// the rule short a port, or with no port match at all,
+						// which opens every port on the protocol.
+						if port, err := strconv.Atoi(strings.TrimSpace(n)); err == nil {
 							rule.DPorts = append(rule.DPorts, port)
 						}
 					}
--- a/internal/protocol/command.go
+++ b/internal/protocol/command.go
@@ -2,6 +2,7 @@ package protocol
 
 import (
 	"encoding/json"
+	"strconv"
 
 	"github.com/alpacax/alpamon/v2/pkg/executor/handlers/common"
 )
@@ -243,8 +244,17 @@ func (c *CommandData) ToArgs() *common.CommandArgs {
 			}
 			if v, ok := ruleMap["dports"].([]any); ok {
 				for _, p := range v {
-					if n, ok := p.(float64); ok {
+					switch n := p.(type) {
+					case float64:
 						rule.DPorts = append(rule.DPorts, int(n))
+					case string:
+						// Not every sender emits numbers: the snapshot-restore
+						// payload carries dports as strings. Dropping one leaves
+						// the rule with no port match, which opens every port on
+						// the protocol.
+						if port, err := strconv.Atoi(n); err == nil {
+							rule.DPorts = append(rule.DPorts, port)
+						}
 					}
 				}
 			}
--- a/internal/protocol/command.go
+++ b/internal/protocol/command.go
@@ -241,6 +241,13 @@ func (c *CommandData) ToArgs() *common.CommandArgs {
 			if v, ok := ruleMap["port_end"].(float64); ok {
 				rule.PortEnd = int(v)
 			}
+			if v, ok := ruleMap["dports"].([]any); ok {
+				for _, p := range v {
+					if n, ok := p.(float64); ok {
+						rule.DPorts = append(rule.DPorts, int(n))
+					}
+				}
+			}
 			if v, ok := ruleMap["icmp_type"].(string); ok {
 				rule.ICMPType = v
 			}
--- a/internal/protocol/command_test.go
+++ b/internal/protocol/command_test.go
@@ -3,6 +3,7 @@ package protocol
 import (
 	"testing"
 
+	"github.com/alpacax/alpamon/v2/pkg/executor/handlers/common"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -44,17 +45,28 @@ func TestParseCommandData_PackageProxyAbsent(t *testing.T) {
 	}
 }
 
-// TestParseCommandData_BatchRuleDPorts pins the batch rule conversion: a rule
-// carrying dports must reach CommandArgs with its ports intact. Dropping them
-// installs a rule with no port match at all, which opens every port on the
-// protocol instead of the two the console shows.
-func TestParseCommandData_BatchRuleDPorts(t *testing.T) {
+// TestParseCommandData_BatchRuleFields pins the batch rule conversion, which is
+// the only place a firewall rule field is converted: the handler reads nothing
+// off CommandArgs but ChainName, ChainNames, OldRuleID, Operation, RuleID and
+// Rules. Every field is asserted at once so a future drop fails here whichever
+// field it is. dports matters most: dropping it installs a rule with no port
+// match at all, which opens every port on the protocol instead of the two the
+// console shows.
+func TestParseCommandData_BatchRuleFields(t *testing.T) {
 	cmd := &Command{
 		ID:    "fw-batch-1",
 		Shell: "internal",
 		Line:  "batch",
 		Data: `{"chain_name": "ALPACON_INPUT", "operation": "batch", "rules": [
-			{"chain": "INPUT", "protocol": "tcp", "dports": [80, 443], "target": "ACCEPT"},
+			{
+				"chain_name": "ALPACON_INPUT", "method": "insert", "chain": "INPUT",
+				"protocol": "tcp", "port_start": 8000, "port_end": 8010,
+				"dports": [80, 443], "icmp_type": "echo-request",
+				"source": "10.0.0.0/8", "destination": "192.168.0.1",
+				"target": "ACCEPT", "description": "web", "priority": 10,
+				"rule_type": "alpacon", "rule_id": "rule-1", "old_rule_id": "rule-0",
+				"operation": "add"
+			},
 			{"chain": "INPUT", "protocol": "tcp", "port_start": 8000, "port_end": 8010, "target": "ACCEPT"}
 		]}`,
 	}
@@ -65,10 +77,51 @@ func TestParseCommandData_BatchRuleDPorts(t *testing.T) {
 	args := data.ToArgs()
 	require.Len(t, args.Rules, 2)
 
-	assert.Equal(t, []int{80, 443}, args.Rules[0].DPorts)
-	assert.Zero(t, args.Rules[0].PortStart)
+	assert.Equal(t, common.FirewallRule{
+		ChainName:   "ALPACON_INPUT",
+		Method:      "insert",
+		Chain:       "INPUT",
+		Protocol:    "tcp",
+		PortStart:   8000,
+		PortEnd:     8010,
+		DPorts:      []int{80, 443},
+		ICMPType:    "echo-request",
+		Source:      "10.0.0.0/8",
+		Destination: "192.168.0.1",
+		Target:      "ACCEPT",
+		Description: "web",
+		Priority:    10,
+		RuleType:    "alpacon",
+		RuleID:      "rule-1",
+		OldRuleID:   "rule-0",
+		Operation:   "add",
+	}, args.Rules[0])
 
+	// A dports rule must not contaminate a neighbouring range rule.
 	assert.Empty(t, args.Rules[1].DPorts)
 	assert.Equal(t, 8000, args.Rules[1].PortStart)
 	assert.Equal(t, 8010, args.Rules[1].PortEnd)
+}
+
+// TestParseCommandData_BatchRuleDPortsAsStrings covers senders that emit dports
+// as strings rather than numbers, which the snapshot-restore payload does.
+func TestParseCommandData_BatchRuleDPortsAsStrings(t *testing.T) {
+	cmd := &Command{
+		ID:    "fw-batch-2",
+		Shell: "internal",
+		Line:  "batch",
+		Data: `{"chain_name": "ALPACON_INPUT", "operation": "batch", "rules": [
+			{"chain": "INPUT", "protocol": "tcp", "dports": ["80", "443"], "target": "ACCEPT"},
+			{"chain": "INPUT", "protocol": "tcp", "dports": ["80", "not-a-port"], "target": "ACCEPT"}
+		]}`,
+	}
+
+	data, err := cmd.ParseCommandData()
+	require.NoError(t, err)
+
+	args := data.ToArgs()
+	require.Len(t, args.Rules, 2)
+
+	assert.Equal(t, []int{80, 443}, args.Rules[0].DPorts)
+	assert.Equal(t, []int{80}, args.Rules[1].DPorts)
 }
--- a/internal/protocol/command_test.go
+++ b/internal/protocol/command_test.go
@@ -104,7 +104,9 @@ func TestParseCommandData_BatchRuleFields(t *testing.T) {
 }
 
 // TestParseCommandData_BatchRuleDPortsAsStrings covers senders that emit dports
-// as strings rather than numbers, which the snapshot-restore payload does.
+// as strings rather than numbers, which the snapshot-restore payload does. Its
+// elements can carry a leading space: the server stores dports verbatim, so a
+// rule entered as "80, 443" is split into ["80", " 443"] on restore.
 func TestParseCommandData_BatchRuleDPortsAsStrings(t *testing.T) {
 	cmd := &Command{
 		ID:    "fw-batch-2",
@@ -112,7 +114,8 @@ func TestParseCommandData_BatchRuleDPortsAsStrings(t *testing.T) {
 		Line:  "batch",
 		Data: `{"chain_name": "ALPACON_INPUT", "operation": "batch", "rules": [
 			{"chain": "INPUT", "protocol": "tcp", "dports": ["80", "443"], "target": "ACCEPT"},
-			{"chain": "INPUT", "protocol": "tcp", "dports": ["80", "not-a-port"], "target": "ACCEPT"}
+			{"chain": "INPUT", "protocol": "tcp", "dports": ["80", "not-a-port"], "target": "ACCEPT"},
+			{"chain": "INPUT", "protocol": "tcp", "dports": ["80", " 443"], "target": "ACCEPT"}
 		]}`,
 	}
 
@@ -120,8 +123,9 @@ func TestParseCommandData_BatchRuleDPortsAsStrings(t *testing.T) {
 	require.NoError(t, err)
 
 	args := data.ToArgs()
-	require.Len(t, args.Rules, 2)
+	require.Len(t, args.Rules, 3)
 
 	assert.Equal(t, []int{80, 443}, args.Rules[0].DPorts)
 	assert.Equal(t, []int{80}, args.Rules[1].DPorts)
+	assert.Equal(t, []int{80, 443}, args.Rules[2].DPorts)
 }
--- a/internal/protocol/command_test.go
+++ b/internal/protocol/command_test.go
@@ -43,3 +43,32 @@ func TestParseCommandData_PackageProxyAbsent(t *testing.T) {
 		})
 	}
 }
+
+// TestParseCommandData_BatchRuleDPorts pins the batch rule conversion: a rule
+// carrying dports must reach CommandArgs with its ports intact. Dropping them
+// installs a rule with no port match at all, which opens every port on the
+// protocol instead of the two the console shows.
+func TestParseCommandData_BatchRuleDPorts(t *testing.T) {
+	cmd := &Command{
+		ID:    "fw-batch-1",
+		Shell: "internal",
+		Line:  "batch",
+		Data: `{"chain_name": "ALPACON_INPUT", "operation": "batch", "rules": [
+			{"chain": "INPUT", "protocol": "tcp", "dports": [80, 443], "target": "ACCEPT"},
+			{"chain": "INPUT", "protocol": "tcp", "port_start": 8000, "port_end": 8010, "target": "ACCEPT"}
+		]}`,
+	}
+
+	data, err := cmd.ParseCommandData()
+	require.NoError(t, err)
+
+	args := data.ToArgs()
+	require.Len(t, args.Rules, 2)
+
+	assert.Equal(t, []int{80, 443}, args.Rules[0].DPorts)
+	assert.Zero(t, args.Rules[0].PortStart)
+
+	assert.Empty(t, args.Rules[1].DPorts)
+	assert.Equal(t, 8000, args.Rules[1].PortStart)
+	assert.Equal(t, 8010, args.Rules[1].PortEnd)
+}


internal/protocol/command.go
```diff
@@ -241,6 +241,13 @@ func (c *CommandData) ToArgs() *common.CommandArgs {
 			if v, ok := ruleMap["port_end"].(float64); ok {
 				rule.PortEnd = int(v)
 			}
+			if v, ok := ruleMap["dports"].([]any); ok {
+				for _, p := range v {
+					if n, ok := p.(float64); ok {
+						rule.DPorts = append(rule.DPorts, int(n))
+					}
+				}
+			}
```
junho226, line 262: Parity check on this loop after the fix: I parsed the `json` tags off `common.FirewallRule` and matched them against every `ruleMap[...]` key read here — 17/17 fields now mapped, and no key read here lacks a struct field. `dports` was indeed the last gap.

Follow-up worth considering, not for this PR: all 17 of those tags already match the wire keys being asserted by hand, and `CommandData.Rules` (`[]map[string]any`) has no other consumer in the tree. Typing it as `[]common.FirewallRule` would let `encoding/json` do the mapping and let this whole ~57-line block go away — after which a newly added field is mapped automatically and this bug class cannot recur. The tradeoff is that strict unmarshaling turns a type mismatch into a hard `ParseCommandData` error rather than a silent skip, which is a behavior change beyond a one-field fix.

junho226 (commented):
## What this PR does

Adds the missing `dports` mapping to the `rules`-array conversion loop in `CommandData.ToArgs()`, so multi-port firewall rules reach the backend with their port list intact instead of installing a rule with no port match at all. Ships one regression test.

**Verdict: looks good to merge.** The fix is correct and minimal, follows the existing `port_start`/`port_end` pattern in the same loop, and the test is a real regression test — I confirmed it fails on the base commit. Below: one correction to the description and two follow-up suggestions. Nothing blocking.

---

### 1. Field-by-field parity audit — confirmed, `dports` was the last gap

I checked this mechanically rather than by eye: parsed the `json` tags off `common.FirewallRule` (`pkg/executor/handlers/common/args.go:116-134`) and matched them against every `ruleMap["..."]` key the loop reads.

| | |
|---|---|
| `FirewallRule` fields | 17 |
| mapped by the loop after this PR | 17 |
| loop keys with no matching struct field | 0 |

`chain_name`, `method`, `chain`, `protocol`, `port_start`, `port_end`, `dports`, `icmp_type`, `source`, `destination`, `target`, `description`, `priority`, `rule_type`, `rule_id`, `old_rule_id`, `operation` — all present. The description's "`dports` was the only one of the 17 not mapped" holds exactly.

**Parity against what alpacon-server actually sends** is the more interesting check. `IptablesAdapter.format_rule_for_agent` / `NftablesAdapter.format_rule_for_agent` emit 16 keys; the loop now consumes 13 of them. Three are dropped: `backend`, `table`, `family`. These have no field on `common.FirewallRule` at all — and they don't bite, because neither `iptables.go` nor `nftables.go` ever emits a `-t <table>` flag, so alpamon always works on the default table regardless of what the server sends. That is an unimplemented feature rather than a drop this PR should fix. Noting it only so the next person auditing this seam doesn't have to re-derive it.

(The UFW and firewalld adapters additionally emit `action` / `direction` / `interface` / `zone` / `service` / `firewalld_rule_type` / `rich_rule`. alpamon's `BackendType` is `none` / `iptables` / `nftables` only — there is no UFW or firewalld backend — so those keys have no consumer by design.)

### 2. Small correction to the description — the fix is more central than it claims

> The single-command path (`CommandData.DPorts`) was already mapped correctly; only the rules-array paths (batch, single add/update wrapped in `rules`, reorder) were affected.

The firewall handler reads exactly six fields off `CommandArgs`: `ChainName`, `ChainNames`, `OldRuleID`, `Operation`, `RuleID`, `Rules`. Every per-rule field — protocol, ports, source, target, priority — reaches a backend **only** through `args.Rules[i]`. The top-level `args.DPorts` / `args.Protocol` / etc. copied at `command.go:164-198` have no reader anywhere in `pkg/executor/handlers/firewall/`.

So this loop is not one of several conversion paths, it is the *sole* conversion boundary for every firewall rule field. That doesn't change the fix, but it raises the value of points 3 and 4.

### 3. Test coverage — real regression test, but narrow

I verified it is a genuine regression test by reverting only the fix and re-running it:

```
$ git checkout HEAD~1 -- internal/protocol/command.go
$ go test ./internal/protocol/ -run TestParseCommandData_BatchRuleDPorts
    Error:      	Not equal:
                	expected: []int{80, 443}
                	actual  : []int(nil)
--- FAIL: TestParseCommandData_BatchRuleDPorts (0.01s)
```

That is the right thing to have shipped. Pinning the second range rule to catch cross-index contamination is a good instinct too — a shared `append` target is exactly how this loop would break next.

The limitation: it asserts three fields (`DPorts`, `PortStart`, `PortEnd`). If `source` or `priority` stops being mapped tomorrow, this test still passes. Given point 2, a whole-struct assertion would be cheap insurance. Suggestion inline.

### 4. Structural — the duplication is the real defect class (follow-up, not this PR)

`CommandData.Rules` is `[]map[string]any`, and this loop is its only consumer in the whole tree. All 17 `json` tags on `common.FirewallRule` already match the wire keys the loop asserts by hand — so typing it as `[]common.FirewallRule` would hand the mapping to `encoding/json` and delete ~57 lines, after which a newly added field is mapped automatically and this bug class cannot recur.

Worth noting the batch path *cannot* simply call the single-rule converter in a loop — there is no shared converter. The top-level path is struct-literal copying from typed fields, while this one does untyped map assertions. The json-tag route is the equivalent that actually exists.

The catch that makes it a judgment call rather than a free win: strict unmarshaling turns a type mismatch into a hard `ParseCommandData` error instead of a silent skip. Arguably better — loud beats silent is this PR's own thesis — but a behavior change well beyond the scope of a one-field fix. Follow-up.

### 5. Firewall semantics round-trip — checks out

- **nil vs empty**: `append` on a nil slice means absent `dports` and `"dports": []` both yield `nil`. `iptables.go:79` gates on `len(rule.DPorts) > 0`, so both correctly fall through to the `PortStart`/`PortEnd` branch — no behavior difference. The `assert.Empty` on rule[1] covers it.
- **multiport syntax**: produces `-m multiport --dports 80,443`, matching what `IptablesAdapter` intends and the E2E output in the description.
- **port ranges**: unaffected; rule[1] pins `8000:8010` surviving alongside a dports rule.
- **truncation**: `int(n)` on a JSON number is fine here — the server sends ints via `get_dports_list()`, and `validator.go:57-62` bounds-checks every element afterward.

One asymmetry flagged inline: the `float64` assertion silently skips anything that is not a number.

## Checks run

Run in a clean worktree at `0706d24`. Note that `go generate ./pkg/db/ent` is required first — a fresh checkout has no generated ent code and `go build ./...` fails with `no required module provides package .../pkg/db/ent/cpu` until it runs.

| command | result |
|---|---|
| `go generate ./pkg/db/ent` | exit 0 |
| `go build ./...` | exit 0, clean |
| `go vet ./...` | exit 0, clean |
| `gofmt -l .` | no output, clean |
| `go test ./... -p 1` | **exit 0** — 40 packages ok, 0 FAIL |
| `go test ./internal/protocol/ -run TestParseCommandData -v` | 3/3 PASS including the new test |
| regression check (revert fix, rerun new test) | **fails as expected** — output quoted above |

**Could not run:**

- `golangci-lint run` — not installed in this environment (`which golangci-lint` returns not found). CI runs it via `golangci/golangci-lint-action@v9` (`.github/workflows/lint.yml:36`); I could not reproduce it locally, so lint findings beyond `go vet` and `gofmt` are unverified here.
- The E2E verification in the description (Lima VM, live iptables output) is **not independently verified** — no VM or live stack in this review environment. I checked the cross-repo contract by reading alpacon-server's backend adapters instead.

internal/protocol/command.go
```diff
@@ -241,6 +241,13 @@ func (c *CommandData) ToArgs() *common.CommandArgs {
 			if v, ok := ruleMap["port_end"].(float64); ok {
 				rule.PortEnd = int(v)
 			}
+			if v, ok := ruleMap["dports"].([]any); ok {
+				for _, p := range v {
+					if n, ok := p.(float64); ok {
```
junho226: Minor, and consistent with the surrounding style, so take it or leave it: this assertion silently drops any element that is not a JSON number. `dports: ["80", "443"]` yields `DPorts == nil`, iptables omits the port match, and the rule opens every port on the protocol — the exact failure this PR fixes, just reached through a different sender.

Your out-of-scope note is right that such a sender exists, but I think the reason it is unreachable is different from the one stated. `firewall/operations.py` builds `'dports': rule_data.get('dports', '').split(',')` for the snapshot-restore payload, and that payload carries `operation: 'restore'` — which is not in alpamon's dispatch switch at all (`common/types.go:59-63` has only batch/flush/delete/add/update), so it is rejected as an unknown operation before any rule is installed. `handleFirewallRollback` not consuming `args.Rules` is true, but it is not what stops it. Same conclusion, though the actual guard is the dispatch switch — so the day someone adds a `restore` case, that payload goes live and silently portless.

If you want cheap insurance now, a `strconv.Atoi` fallback on a string element, or a warn-level log on a skipped one, would keep it from being silent.

internal/protocol/command_test.go
```diff
@@ -43,3 +43,32 @@ func TestParseCommandData_PackageProxyAbsent(t *testing.T) {
...
+
+	args := data.ToArgs()
+	require.Len(t, args.Rules, 2)
+
+	assert.Equal(t, []int{80, 443}, args.Rules[0].DPorts)
+	assert.Zero(t, args.Rules[0].PortStart)
+
+	assert.Empty(t, args.Rules[1].DPorts)
+	assert.Equal(t, 8000, args.Rules[1].PortStart)
+	assert.Equal(t, 8010, args.Rules[1].PortEnd)
```
junho226: This is a genuine regression test — I reverted `command.go` to `HEAD~1` and it fails with `expected: []int{80, 443} / actual: []int(nil)`. Pinning rule[1]'s range against cross-index contamination is a good call too.

Suggestion: these assertions catch a `dports` drop but not the next one. If `source` or `priority` stops being mapped, this test still passes — and this loop is the only place any rule field is converted (the firewall handler reads nothing off `CommandArgs` except `ChainName`, `ChainNames`, `OldRuleID`, `Operation`, `RuleID`, `Rules`). Populating all 17 fields in the payload and doing a single `assert.Equal` against a fully populated `common.FirewallRule` would turn this into a test that fails on *any* future field drop, for about the same number of lines.

internal/protocol/command.go
jisung-02: Fixed in 3671e54a. The type assertion is now a switch with a `strconv.Atoi` fallback on string elements, so `dports: ["80", "443"]` yields `[80 443]` instead of `nil`. An element that parses as neither is still skipped, but it can no longer be a whole rule going portless just because the sender chose strings. `TestParseCommandData_BatchRuleDPortsAsStrings` covers both the string list and the unparseable element.

Thank you for the correction on why the restore payload is unreachable — you are right and my note was wrong. The guard is the dispatch switch: `restore` is not among the firewall sub-operations in `common/types.go:58-63` (batch, flush, delete, add, update only), so it is rejected as an unknown operation before `handleFirewallRollback` is ever relevant. I checked and there is indeed no `FirewallOpRestore` constant. The PR description now states it that way. That also makes the `Atoi` fallback worth more than cheap insurance: the day someone adds a `restore` case, the payload lands on a converter that already understands it.

internal/protocol/command_test.go
jisung-02: Taken in 3671e54a. `TestParseCommandData_BatchRuleDPorts` is now `TestParseCommandData_BatchRuleFields`: the payload populates all 17 fields and a single `assert.Equal` compares rule[0] against a fully populated `common.FirewallRule`, so a drop of `source` or `priority` fails here too. The second range rule stays as the cross-index contamination check.

Verified it still catches the original bug — reverting `internal/protocol/command.go` alone to the base commit fails it on the whole-struct comparison, `DPorts` being the differing field.

internal/protocol/command.go
jisung-02: Thank you for auditing this mechanically rather than by eye, and for stating the parity result as a count.

Agreed on the follow-up, and agreed it does not belong in this PR — the silent-skip to hard-error change is exactly the kind of thing that should land on its own so a regression in it is attributable. I have recorded both the proposal and the reason it was left out in the PR description, and as a `Rejected:` trailer on the commit so it survives in `git log` rather than only here.

Your point 2 was the more useful correction and I have folded it into the description: the top-level `args.DPorts` copied at `command.go:164-198` has no reader in `pkg/executor/handlers/firewall/` at all (`grep` for `args.DPorts|args.Protocol|args.PortStart|args.Source` there returns nothing), so this loop is the sole conversion boundary rather than one path among several. That is what makes the whole-struct assertion worth its lines.

On `backend` / `table` / `family`: noted, and also folded into the description so the next person auditing this seam does not re-derive it.

internal/protocol/command.go
```diff
@@ -241,6 +242,22 @@ func (c *CommandData) ToArgs() *common.CommandArgs {
...
+				for _, p := range v {
+					switch n := p.(type) {
+					case float64:
+						rule.DPorts = append(rule.DPorts, int(n))
+					case string:
+						// Not every sender emits numbers: the snapshot-restore
+						// payload carries dports as strings. Dropping one leaves
+						// the rule with no port match, which opens every port on
+						// the protocol.
+						if port, err := strconv.Atoi(n); err == nil {
```
junho226: Nit, non-blocking: `strconv.Atoi(strings.TrimSpace(n))` would make this robust against the exact payload it targets.

The restore payload's elements can carry a leading space. In alpacon-server, `validate_dports` (`firewall/api/serializers.py:278-292`) strips each part *for validation* but returns `value` unmodified, so `"80, 443"` is stored verbatim in the `dports` `CharField`. The snapshot stores that raw string (`firewall/operations.py:85`), and restore splits it without stripping (`:261-263`), yielding `["80", " 443"]`.

`strconv.Atoi(" 443")` returns an error, so that element is skipped and the rule installs one port short — a quieter version of the failure this PR exists to prevent, on the very payload the fallback was added for. The normal send path is unaffected: it goes through `get_dports_list()`, which strips and returns ints.

Unreachable today for the same reason the rest of the fallback is (`operation: 'restore'` is rejected as an unknown sub-operation), so this is only about the day someone wires that case up. If you take it, `["80", " 443"]` is worth adding to `TestParseCommandData_BatchRuleDPortsAsStrings` alongside the existing `not-a-port` case.

junho226 (commented):
## Re-review: `0706d24` → `3671e54`

One new commit since the last review, and it does two things: adds a `strconv.Atoi` fallback for string `dports` elements, and widens the batch test to assert the whole `FirewallRule` rather than three fields. Both are direct responses to the last round. Everything I checked before still holds, and the two suggestions that were actionable were taken. Still looks good to merge.

### Status of the previous findings

| # | Previous finding | Status |
|---|---|---|
| 1 | Field-by-field parity audit — 17/17 | **Still clean**, re-measured at `3671e54` |
| 2 | Description understates the fix's centrality | **Applied** — folded into the description |
| 3 | Regression test was narrow (3 fields) | **Fixed** — whole-struct assertion |
| 4 | Structural duplication (json-tag route) | **Declined with a reason** — correct call, not repeated below |
| 5 | Firewall semantics round-trip | **Re-checked** after the conversion changed; still holds |
| inline | `float64` assertion silently skips non-numbers | **Partially fixed** — strings now handled; other types still skipped, deliberately |

**1. Parity, re-measured at the new head.** I re-ran the same mechanical check rather than carrying the old number over — both sides of it were touched by this commit, so the count had to be re-established:

| | |
|---|---|
| `FirewallRule` json tags (`common/args.go`) | 17 |
| `ruleMap["..."]` keys read by the loop | 17 |
| tags with no loop key (dropped) | **0** |
| loop keys with no matching tag (orphan) | **0** |
| duplicate key reads | 0 |

Same 17/17 as before. The new commit did not disturb it.

**3. The test was broadened, and it is still a real regression test.** `TestParseCommandData_BatchRuleFields` populates all 17 keys and compares `args.Rules[0]` against a fully-populated `common.FirewallRule` in one `assert.Equal`, so a dropped `source` or `priority` now fails here too. I re-established the regression property by reverting `internal/protocol/command.go` alone to the base commit (`581302b`) — both new tests fail, and the whole-struct diff names the field:

```
--- FAIL: TestParseCommandData_BatchRuleFields (0.01s)
    @@ -7,6 +7,3 @@
      PortEnd: (int) 8010,
    - DPorts: ([]int) (len=2) { (int) 80, (int) 443 },
    + DPorts: ([]int) <nil>,
      ICMPType: (string) (len=12) "echo-request",
--- FAIL: TestParseCommandData_BatchRuleDPortsAsStrings (0.00s)
    expected: []int{80, 443}
    actual  : []int(nil)
```

One residual limit, stated for accuracy rather than as something to fix: a whole-struct `assert.Equal` catches a mapping being *removed*, because the expected literal names the value. It does not catch a *new* 18th field added to `FirewallRule` and to the wire but never added to the loop — that field is the zero value on both sides and the assertion passes. Narrow in practice: `DPorts` predates this loop (added in `10dcc7a`, loop in its current form from `7ad8e0c`), so the original bug was a day-one omission, which the new assertion would have caught. Closing the struct-growth case properly is the point-4 follow-up, not a test change.

**4. The json-tag refactor was declined, and the reason is right.** Strict unmarshaling converting a silent skip into a hard `ParseCommandData` error is a behavior change that deserves its own PR so a regression in it is attributable. Recording it as a `Rejected:` commit trailer rather than only in a review thread is the better place for it. No further comment from me.

**5. Semantics re-checked, since the conversion itself changed.** The `Atoi` path can produce values the `float64` path could not — `"0"`, `"-1"`, `"99999"` all parse. They do not reach a backend silently: `firewall.go:136,151` call `ValidateBatchRules` → `ValidateRule` → `ValidatePort`, which bounds every `DPorts` element to 1–65535 (`validator.go:57-62`, `:135-140`) and fails the rule loudly as `rule[i]: invalid dports[j]: port N out of range`. So the fallback widens what is *accepted* without widening what can be *installed*. nil-vs-empty, multiport syntax and range rules are untouched by this commit and still behave as described last time.

**The cross-repo claim behind the fallback checks out.** I verified it in alpacon-server rather than taking the description's word: `firewall/operations.py:261-263` builds `'dports': rule_data.get('dports', '').split(',')` — a list of strings — and `:293` ships it as `'operation': 'restore'`, which is not among `FirewallOpBatch/Flush/Delete/Add/Update` in `common/types.go:58-63`, so it is rejected as an unknown operation today. The correction about *why* it is unreachable (dispatch switch, not `handleFirewallRollback`) is accurate.

### New: one nit on the fallback itself

The fallback targets the restore payload specifically, so it is worth being exact about the shape that payload can take. `strconv.Atoi` rejects a leading space, and the stored value can contain one — details inline. One-line fix, non-blocking; the path is unreachable today for the same reason the rest of the fallback is.

### Checks run

Run at `3671e54a269c` in the existing review worktree. `go generate ./pkg/db/ent` is still required before building.

| command | result |
|---|---|
| `go generate ./pkg/db/ent` | exit 0 |
| `go build ./...` | exit 0, clean |
| `go vet ./...` | exit 0, clean |
| `gofmt -l .` | no output, clean |
| `go test ./... -p 1` | **exit 0** — 35 ok, 25 no-test-files, **0 FAIL** |
| `go test ./internal/protocol/ -run TestParseCommandData -v` | **4/4 PASS** (was 3/3; the string test is new) |
| regression check (revert `command.go` to `581302b`, rerun) | **both new tests fail** — output quoted above |
| parity re-measure at new head | **17/17, 0 dropped, 0 orphan** |

No movement against the `0706d24` baseline other than the added test.

**Could not run:**

- `golangci-lint run` — still not installed in this environment (`golangci-lint not found`). CI runs it via `golangci/golangci-lint-action@v9` (`.github/workflows/lint.yml`). Lint findings beyond `go vet` and `gofmt` remain unverified here; I am not implying lint passed.
- The Lima VM / live `iptables` E2E in the description is still **not independently verified** — no VM or live stack in this environment, as last time. I checked the cross-repo contract by reading alpacon-server's `firewall/operations.py` and `firewall/api/serializers.py` instead.

internal/protocol/command.go
jisung-02: Taken in bc0541c4 — `strconv.Atoi(strings.TrimSpace(n))`.

You are right that this is the payload the fallback was added for, which makes the leading space worth handling rather than a hypothetical. I followed your references in alpacon-server: `validate_dports` (`firewall/api/serializers.py:278-292`) strips each part inside the loop but returns `value`, so the `CharField` keeps `"80, 443"` verbatim; `firewall/operations.py:85` snapshots that raw string and `:261-263` splits it without stripping. So `["80", " 443"]` is the shape restore actually produces, and without the trim the rule installs one port short — quieter than the original bug and harder to notice.

Added `["80", " 443"]` to `TestParseCommandData_BatchRuleDPortsAsStrings` as you suggested, alongside the existing `not-a-port` case. Confirmed it is a real regression test for this specific line: dropping only the `TrimSpace` and leaving the rest of the fallback in place fails it with `expected: []int{80, 443} / actual: []int{80}`.

Still unreachable today for the reason you state, and the PR description now says so with the serializer and snapshot references included.

junho226 (approved):
Third pass, incremental over `3671e54a269c..bc0541c4aa58` (one commit, `fix(protocol): trim whitespace from string dports elements`, 2 files).

The nit is taken: `strconv.Atoi(strings.TrimSpace(n))` at `internal/protocol/command.go:257`. The comment above it now names where the whitespace comes from (a stored `"80, 443"` split without trimming) and distinguishes the two failure shapes — a rule short a port vs. a rule with no port match at all. The test gained a third rule with `dports: ["80", " 443"]` asserting `[80 443]`, and the doc comment on it records why the leading space is real rather than hypothetical. Nothing else changed in the range, and I found nothing new to raise.

Parity re-measured at this head, since `command.go` moved: parsing the `json` tags off `common.FirewallRule` (`pkg/executor/handlers/common/args.go`) against every `ruleMap["..."]` key the `rules` loop in `CommandData.ToArgs()` reads gives **17 tags / 17 keys / 17 mapped / 0 tags unread / 0 keys without a field**. Unchanged from the previous two heads.

The regression property still holds, in both directions:

- Reverting `command.go` to the branch base (`581302b`) fails the test: `expected: []int{80, 443} / actual: []int(nil)`.
- Reverting only the new `strings.TrimSpace` fails the new case specifically: `expected: []int{80, 443} / actual: []int{80}`. So the third rule is pinning the trim, not riding along on the earlier fix.

LGTM — good to merge.

### Checks run

At `bc0541c4aa58`, after `go generate ./pkg/db/ent`:

| Check | Result |
|---|---|
| `go build ./...` | exit 0, no output |
| `go vet ./...` | exit 0, no output |
| `gofmt -l .` | no output (clean) |
| `go test ./... -p 1` | `EXIT=0` — 80 packages: 45 `ok`, 35 `[no test files]`, 0 `FAIL` |
| `go test ./internal/protocol/ -run TestParseCommandData -v` | 4 tests + 2 subtests, all `PASS`, `ok ... 0.550s` |

Not run: `golangci-lint` is not installed in this environment, so nothing here speaks to it — CI's `golangci/golangci-lint-action@v9` is the only signal on lint. The Lima VM / live-iptables E2E in the description remains not independently verifiable here.

PR #413, opened 2026-08-13
fix(runner): send exactly one sudo approval response per request

### LINKED ISSUE #396: Duplicate sudo approval response on the retry-failure path
## Background

The retry-failure path in `handleSudoApprovalRequest` (`pkg/runner/auth_manager.go:444-450`) writes a response twice to the same connection and closes it twice. Traced through the code, the order is:

1. `:446` — `sendSudoApprovalResponse(unixConn, sudoApprovalReq, false, "Communication error")` writes the first response.
2. `:447` — `cleanupTimeoutRequest` looks the request up in `session.Requests`. It was registered at `:428` and has not been deleted yet, so the lookup always hits, and the function writes a second response and calls `req.Connection.Close()`.
3. `:449` — `_ = unixConn.Close()` closes a second time. The return value is discarded, so `os.ErrClosed` is swallowed silently.

The two responses also differ in content. `cleanupTimeoutRequest` passes only `SudoApprovalRequest{RequestID: requestID}`, so the second response carries empty `username`, `pid`, `ppid`, `command`, `session_id`, and `command_id`.

## Impact

Unverified. If the PAM client (alpamon-pam) reads one response and disconnects, the second write fails with EPIPE and only leaves an error log, which is harmless. If it keeps reading, it receives a second response with empty fields. The client's read behavior lives in a separate repository and was not checked here, so it should be confirmed before fixing.

## Proposed change

Make the path send one response and close once. Two directions:

- Add a variant of `cleanupTimeoutRequest` that only deregisters, and leave the response and the close to the caller.
- Drop the direct response at `:446` and pass the full `sudoApprovalReq` to `cleanupTimeoutRequest`.

The second also fixes the empty-field problem.

## Scope

This predates #392 and is unrelated to what that PR changes; it was noted in passing during the #392 review.

#384 touches the same function to consolidate the boilerplate shared by the three socket response writers. That is a separate concern, but both changes edit `sendSudoApprovalResponse` and its callers, so whichever lands second will need to rebase on the other.

Line references are as of the #392 branch head.

jisung-02:
Closes #396

## The bug

The retry-failure path in `handleSudoApprovalRequest` wrote a response and closed the socket, then `cleanupTimeoutRequest` looked the still-registered request up and wrote a second response and closed again.

The alpamon-pam client reads auth.sock until EOF and parses the whole buffer as one JSON document. Two concatenated responses fail parsing, so the deny reason never reaches PAM.

## The fix

**One ownership rule.** A pending request is answered and closed by whoever takes it out of `session.Requests`, and by nobody else.

- `cleanupTimeoutRequest` is renamed `finalizeRequest`—two of its three callers are not timeouts.
- The retry-failure path's direct write and duplicate `Close()` are gone.

**Whoever removes a session entry follows the same rule.** Taking a session entry out of the map puts its pending requests beyond `finalizeRequest`'s reach, so every function that does so answers and closes them:

- `RemovePIDSessionMapping`, `RemovePIDCommandMapping`: deny with `"Session ended"`
- `AddPIDSessionMapping`, `AddPIDCommandMapping`: deny a reused pid's leftovers with `"Session replaced"`

Without that, a PTY teardown during the up-to-25s retry left the socket open with nothing written, and the PAM client waited out its own timeout before falling back to its fail-open setting.

**Responses carry the full request.** Each pending request keeps the `SudoApprovalRequest` it was registered with, so every response carries the requester's fields instead of only the request id.

## Hardening

- Every write to the auth socket now carries `authSocketWriteTimeout` (5s). The deny paths run inline on PTY teardown and on Command registration, and `HandleSudoApprovalResponse` writes from the websocket read loop, so a peer that stopped reading must not stall any of them. The sibling response paths `sendIsAlpconResponse` and `sendSessionEventResponse` got the same deadline.
- A write that fails because the client is already gone—broken pipe, reset, closed socket, or the deadline—is a warning rather than an error, and `control_client.go` no longer repeats that error at error level. `authSocketWriteLevel` makes that decision for both `sendSudoApprovalResponse` and `sendIsAlpconResponse`; `check_user` is answered on every PAM sudo hook, so it produces the same noise from a client that already left.
- `HandleSudoApprovalResponse` closes the connection when its write fails: the request is already deregistered at that point, so nothing else could reclaim the descriptor.
- `AddPIDSessionMapping` fills in its defaults under `am.mu`—re-registering the entry already installed for a pid writes fields a concurrent lookup is reading. `putSession` becomes `putSessionLocked`, and both callers answer the displaced entry's requests after unlocking.

## Tests

New `pkg/runner/auth_manager_cleanup_test.go`, extended `auth_manager_completion_test.go`.

The central one is `TestHandleSudoApprovalRequest_RetryFailureAnswersOnce`: it runs issue #396's own path through `handleSudoApprovalRequest` and asserts exactly one response followed by EOF. A pre-cancelled `am.ctx` fails the retry immediately instead of spending its 25s budget.

Reintroducing the old double-write ordering (`sendSudoApprovalResponse` → `cleanupTimeoutRequest` → `Close()`) fails it, as it should:

```
--- FAIL: TestHandleSudoApprovalRequest_RetryFailureAnswersOnce (0.00s)
    auth_manager_cleanup_test.go:319: expected EOF after single response, got <nil>
```

The rest: `TestFinalizeRequest_*` (single response, no-op on an unregistered request, race against `HandleSudoApprovalResponse`), `TestRemovePID{Session,Command}Mapping_*`, `TestAddPID{Session,Command}Mapping_*`, `TestHandleSudoApprovalResponse_SignalsWaiterOnWriteFailure`, `TestHandleSudoApprovalRequest_LocalSudoRegistersNothing`.

```
gofmt -l pkg/runner/   # no output
go vet ./pkg/runner    # ok
go test ./pkg/runner -p 1 -count=1
ok  github.com/alpacax/alpamon/v2/pkg/runner  45.470s
```

All targeted tests also pass under `-race`.

- `HandleSudoApprovalResponse` no longer dumps every session's pending requests before the unmatched-response debug line; the line below it already said no request matched the id.

## Review notes

- The deny reason strings (`"Session ended"`, `"Session replaced"`, `"Communication error"`) never reach alpacon-server. `sendSudoApprovalResponse` writes only to auth.sock, and the sole outbound POST, `sendSudoRequestWithRetry`, sends a `SudoApprovalRequest`, which carries no reason field. Their only audience is alpamon-pam, so the compatibility question belongs in that repo. (Corrected from the original note, which asked about the server.)
- 5s for `authSocketWriteTimeout` sits between the 25s retry budget and the PAM client's own 30s, and every caller closes the connection right after the write, so no path accumulates two deadlines.
- `AddPIDSessionMapping` clears `session.CommandID` on every call. The only production caller (`pty.go:135`) passes a fresh pointer each time, but re-registering the same pointer is now pinned as supported behavior by a test.

### DIFF
--- a/pkg/runner/auth_manager.go
+++ b/pkg/runner/auth_manager.go
@@ -3,18 +3,22 @@ package runner
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"maps"
 	"net"
 	"os"
 	"path/filepath"
-	"strings"
+	"slices"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/alpacax/alpamon/v2/internal/retry"
 	"github.com/alpacax/alpamon/v2/pkg/scheduler"
 	"github.com/alpacax/alpamon/v2/pkg/utils"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 )
 
@@ -58,6 +62,9 @@ func (s *SessionInfo) effectiveKind() string {
 
 type SudoRequest struct {
 	Connection net.Conn
+	// Request is kept whole so whoever finalizes it can answer with the fields
+	// the PAM client sent.
+	Request SudoApprovalRequest
 }
 
 type SudoApprovalRequest struct {
@@ -159,6 +166,10 @@ const (
 	authRetryInitialInterval = 1 * time.Second
 	authRetryMaxInterval     = 10 * time.Second
 	authRetryTimeout         = 25 * time.Second // Less than PAM 30s timeout
+	// authSocketWriteTimeout bounds every write to auth.sock. Session teardown and
+	// Command registration deny leftover requests inline, so a peer that stopped
+	// reading must not stall them.
+	authSocketWriteTimeout = 5 * time.Second
 	// emitConcurrencyLimit caps in-flight access-event emit goroutines.
 	emitConcurrencyLimit = 16
 	// authSocketReadBufferSize bounds a single auth.sock request. The largest
@@ -484,8 +495,13 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 
 	session.Requests[sudoApprovalReq.RequestID] = &SudoRequest{
 		Connection: unixConn,
+		Request:    sudoApprovalReq,
 	}
+	// Created under the registration's lock: a teardown in between would signal a
+	// channel that does not exist yet, leaving this goroutine to its full timeout.
+	completionChan := am.newCompletionChannelLocked(sudoApprovalReq.RequestID)
 	am.mu.Unlock()
+	defer am.removeCompletionChannel(sudoApprovalReq.RequestID)
 
 	log.Debug().
 		Str("request_id", sudoApprovalReq.RequestID).
@@ -494,15 +510,9 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		Str("command_id", sudoApprovalReq.CommandID).
 		Msg("Alpacon user sudo request")
 
-	completionChan := am.newCompletionChannel(sudoApprovalReq.RequestID)
-
-	// Send Sudo Approval request to the alpacon-server with retry
 	if err := am.sendSudoRequestWithRetry(sudoApprovalReq); err != nil {
 		log.Error().Err(err).Msg("Failed to send sudo_approval request after retries")
-		am.sendSudoApprovalResponse(unixConn, sudoApprovalReq, false, "Communication error")
-		am.cleanupTimeoutRequest(sudoApprovalReq.RequestID, false, "Communication error")
-		am.removeCompletionChannel(sudoApprovalReq.RequestID)
-		_ = unixConn.Close()
+		am.finalizeRequest(sudoApprovalReq.RequestID, "Communication error")
 		return
 	}
 
@@ -514,21 +524,20 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 		// Response received and processed by HandleSudoApprovalResponse
 		log.Debug().Msgf("sudo_approval response received for request %s", sudoApprovalReq.RequestID)
 	case <-time.After(30 * time.Second):
-		// Prevent race condition: check if request still exists before cleanup
-		// (HandleSudoApprovalResponse may have already processed it)
+		// The take under am.mu is what settles the race; this check only keeps the
+		// timeout from warning about a request the response path already took.
 		if am.isRequestPending(sudoApprovalReq.RequestID) {
 			log.Warn().Msg("sudo_approval response timeout")
-			am.cleanupTimeoutRequest(sudoApprovalReq.RequestID, false, "Response timeout")
+			am.finalizeRequest(sudoApprovalReq.RequestID, "Response timeout")
 		} else {
 			log.Debug().Msgf("sudo_approval timeout triggered but request already handled: %s", sudoApprovalReq.RequestID)
 		}
 	case <-am.ctx.Done():
 		log.Debug().Msg("Context cancelled, cleaning up sudo_approval connection")
 		if am.isRequestPending(sudoApprovalReq.RequestID) {
-			am.cleanupTimeoutRequest(sudoApprovalReq.RequestID, false, "Service shutdown")
+			am.finalizeRequest(sudoApprovalReq.RequestID, "Service shutdown")
 		}
 	}
-	am.removeCompletionChannel(sudoApprovalReq.RequestID)
 }
 
 func (am *AuthManager) sendIsAlpconResponse(conn net.Conn, username, groupname string, pid, ppid int, isAlpconUser bool) {
@@ -547,6 +556,7 @@ func (am *AuthManager) sendIsAlpconResponse(conn net.Conn, username, groupname s
 		return
 	}
 
+	_ = conn.SetWriteDeadline(time.Now().Add(authSocketWriteTimeout))
 	_, err = conn.Write(responseJSON)
 	if err != nil {
 		log.Error().Err(err).Msg("Failed to send is_alpacon_response")
@@ -554,19 +564,21 @@ func (am *AuthManager) sendIsAlpconResponse(conn net.Conn, username, groupname s
 	}
 }
 
-// sendSudoApprovalResponse is used when there is something wrong sending the sudo approval request to the alpacon-server
-func (am *AuthManager) sendSudoApprovalResponse(conn net.Conn, sudo_approval_req SudoApprovalRequest, approved bool, reason string) {
+// sendSudoApprovalResponse writes the response alpamon itself decided on, for
+// every case the alpacon-server does not answer. Closing the connection is the
+// caller's job.
+func (am *AuthManager) sendSudoApprovalResponse(conn net.Conn, req SudoApprovalRequest, approved bool, reason string) {
 	response := SudoApprovalResponse{
 		Type:         "sudo_approval_response",
-		Username:     sudo_approval_req.Username,
-		Groupname:    sudo_approval_req.Groupname,
-		PID:          sudo_approval_req.PID,
-		PPID:         sudo_approval_req.PPID,
-		Command:      sudo_approval_req.Command,
-		IsAlpconUser: sudo_approval_req.IsAlpconUser,
-		SessionID:    sudo_approval_req.SessionID,
-		CommandID:    sudo_approval_req.CommandID,
-		RequestID:    sudo_approval_req.RequestID,
+		Username:     req.Username,
+		Groupname:    req.Groupname,
+		PID:          req.PID,
+		PPID:         req.PPID,
+		Command:      req.Command,
+		IsAlpconUser: req.IsAlpconUser,
+		SessionID:    req.SessionID,
+		CommandID:    req.CommandID,
+		RequestID:    req.RequestID,
 		Approved:     approved,
 		Reason:       reason,
 	}
@@ -577,28 +589,33 @@ func (am *AuthManager) sendSudoApprovalResponse(conn net.Conn, sudo_approval_req
 		return
 	}
 
+	_ = conn.SetWriteDeadline(time.Now().Add(authSocketWriteTimeout))
 	_, err = conn.Write(responseJSON)
 	if err != nil {
-		log.Error().Err(err).Msg("Failed to send sudo_approval_response")
-		return
+		logSudoResponseWriteError(err, req.RequestID)
 	}
 }
 
-// HandleSudoApprovalResponse is used to handle the sudo_approval response from the alpacon-server
+// logSudoResponseWriteError downgrades a client that is already gone to a
+// warning: by the time a denial reaches a torn-down session, sudo has usually
+// exited, and the write fails as a broken pipe, a reset, a closed socket, or
+// authSocketWriteTimeout expiring on a peer that stopped reading.
+func logSudoResponseWriteError(err error, requestID string) {
+	level := zerolog.ErrorLevel
+	if errors.Is(err, syscall.EPIPE) || errors.Is(err, syscall.ECONNRESET) ||
+		errors.Is(err, net.ErrClosed) || errors.Is(err, os.ErrDeadlineExceeded) {
+		level = zerolog.WarnLevel
+	}
+	log.WithLevel(level).Err(err).Str("request_id", requestID).Msg("Failed to send sudo_approval_response")
+}
+
+// HandleSudoApprovalResponse owns the response and close for the request it
+// takes, the way finalizeRequest does for the paths alpamon answers itself.
 func (am *AuthManager) HandleSudoApprovalResponse(response SudoApprovalResponse) error {
 	log.Info().Str("request_id", response.RequestID).Bool("approved", response.Approved).Msg("Processing sudo_approval response")
 
 	am.mu.Lock()
-	var sudoRequest *SudoRequest
-
-	for _, session := range am.pidToSessionMap {
-		if req, exists := session.Requests[response.RequestID]; exists {
-			delete(session.Requests, response.RequestID)
-			sudoRequest = req
-			log.Debug().Msgf("Found Alpacon user request for ID: %s", response.RequestID)
-			break
-		}
-	}
+	sudoRequest := am.takeRequestLocked(response.RequestID)
 	am.mu.Unlock()
 
 	if sudoRequest == nil {
@@ -607,6 +624,9 @@ func (am *AuthManager) HandleSudoApprovalResponse(response SudoApprovalResponse)
 			log.Debug().Msgf("Session %s requests: %+v", session.SessionID, session.Requests)
 		}
 		am.mu.RUnlock()
+
+		// Routine after a timeout: the waiter already answered and deregistered.
+		log.Warn().Str("request_id", response.RequestID).Msg("No pending sudo_approval request for response")
 
 		return fmt.Errorf("no pending sudo_approval request found for request_id: %s", response.RequestID)
 	}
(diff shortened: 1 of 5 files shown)


pkg/runner/auth_manager.go
```diff
@@ -831,21 +876,59 @@ func (am *AuthManager) Stop() {
...
+// by HandleSudoApprovalResponse.
+func (am *AuthManager) finalizeRequest(requestID, reason string) {
 	am.mu.Lock()
+	req := am.takeRequestLocked(requestID)
+	am.mu.Unlock()
 
+	if req == nil {
+		log.Warn().Msgf("No pending sudo request to finalize: %s", requestID)
+		return
+	}
```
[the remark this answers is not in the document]
jisung-02: Fixed in e25b2bea — downgraded to Debug with `request_id` and `reason` as structured fields.

You are right that `req == nil` is routine rather than exceptional at every call site: `HandleSudoApprovalResponse` or a session teardown can take the request between the `isRequestPending` check and the take under `am.mu`, and the retry-failure path races the same way while `sendSudoRequestWithRetry` is still running.

pkg/runner/auth_manager.go
```diff
@@ -547,26 +557,29 @@ func (am *AuthManager) sendIsAlpconResponse(conn net.Conn, username, groupname s
 		return
 	}
 
+	_ = conn.SetWriteDeadline(time.Now().Add(authSocketWriteTimeout))
 	_, err = conn.Write(responseJSON)
 	if err != nil {
 		log.Error().Err(err).Msg("Failed to send is_alpacon_response")
```
junho226: This one write path kept its error level while its two siblings moved off it: `sendSudoApprovalResponse` now routes through `logSudoResponseWriteError`, and `sendSessionEventResponse` logs at warn. `check_user` is answered on every PAM sudo hook, so a client that already disconnected produces exactly the broken-pipe noise this PR set out to demote, and it will be the loudest of the three. The deadline you added above is the part that matters, so this is not a blocker, but sharing the classifier here (or a small `logAuthSocketWriteError(err, msg, requestID)`) would make the rule uniform across all three writers.

pkg/runner/auth_manager.go
```diff
@@ -695,8 +734,11 @@ func (am *AuthManager) AddPIDCommandMapping(pid int, commandID, username string)
 		Requests:  make(map[string]*SudoRequest),
 	}
 	am.mu.Lock()
-	am.pidToSessionMap[pid] = info
+	replaced := am.putSessionLocked(pid, info)
 	am.mu.Unlock()
+
+	am.denyPendingRequests(replaced, "Session replaced")
```
junho226, line 737: Worth naming explicitly: this puts a socket write on the command-start path. `pidHook` in `pkg/executor/handlers/shell/shell.go:175` calls `RegisterCommandPID` right after the child process starts, so a displaced entry with N leftover requests can spend up to `5 x N` seconds here before the command proceeds (same shape in `AddPIDSessionMapping` for Websh setup). In practice N is 0, and when it is not, the peer is parked in `read()` so the write returns immediately, which is why `authSocketWriteTimeout` makes this acceptable rather than a defect. If you would rather keep it off the start path entirely, the requests are already detached from the map at this point, so nothing else can reclaim them and `go am.denyPendingRequests(replaced, "Session replaced")` would be safe. My preference is to leave it inline and keep the ordering obvious.

junho226 (dismissed):
## Summary

I traced the mechanism from the code rather than taking the description's word for it, and it holds. On `main` the retry-failure branch is a straight-line double send, not a race: it called `sendSudoApprovalResponse(unixConn, ...)` inline **without deregistering**, so `cleanupTimeoutRequest` on the very next line still found the request registered a few lines above, wrote a second response built from `SudoApprovalRequest{RequestID: requestID}` (every other field zero), and closed the connection; the trailing `_ = unixConn.Close()` then closed it again. Two producers on one connection, because the inline writer was not the one that took the request out of the map.

The fix is the right shape for that: make removal from `session.Requests` the token of ownership, and let only the taker write and close.

## Critical issues

None.

### At most once

`session.Requests` is mutated in exactly three places in non-test code: the insert at `auth_manager.go:496`, `takeRequestLocked` (delete one), and `takePendingRequestsLocked` (clear all). Both takers require `am.mu`, and every call site holds the write lock across the take. The write and the close then happen only in `denyPendingRequests` and `HandleSudoApprovalResponse`, each acting on a request it took. Two responses would need two successful takes of the same map entry, which the write lock excludes.

### At least once

The other direction matters as much—a request that is never answered hangs the PAM client until its own timeout—so I enumerated every exit from a registered request:

| Exit | Answered by |
| --- | --- |
| Retry failure | `finalizeRequest("Communication error")` |
| Completion channel | the taker already answered |
| 30s timeout | `finalizeRequest("Response timeout")` |
| `am.ctx.Done()` | `finalizeRequest("Service shutdown")` |
| Server response | `HandleSudoApprovalResponse` |
| Session teardown | `denyPendingRequests(..., "Session ended")` |
| Pid reuse | `denyPendingRequests(..., "Session replaced")` |

The two `isRequestPending` guards in front of `finalizeRequest` are only noise suppression, and that is what the new comment says: `finalizeRequest` re-takes under the lock and no-ops on nil, so the TOCTOU between check and take is harmless.

One residual zero-response path survives: a panic between registration and the select. `handleSudoRequest`'s deferred recover closes the socket, but the entry stays in `session.Requests` with a dead connection, so the client reads EOF with no JSON and the entry is reaped only at teardown. That predates this PR and needs a panic to reach, so it is not something to hold the merge for.

### Go concurrency

- No lock is held across a socket write. `denyPendingRequests` calls `signalCompletion`, which takes `am.mu.RLock()`, so it must run unlocked; all five callers unlock first, and the doc comment states the constraint.
- Completion channels are never closed—only sent to under a non-blocking `select`, then deleted—so `signalCompletion` is safe to call any number of times and cannot panic on a double send.
- `defer am.removeCompletionChannel` now covers the early return that previously needed a manual call, so no channel is left in the map.
- No unchecked type assertions, no `sync.Once`, no WaitGroup in the changed code.

### Idempotency at the boundary

The fix is agent-side only, and it has to be. alpamon-pam reads auth.sock to EOF and decodes the whole buffer as one JSON document, so the protocol carries no dedup and correctness rests entirely on alpamon writing once. The transport is a unix `SOCK_STREAM` with no retry layer above it, so there is no replay to defend against. There *is* dedup on the inbound side: a duplicate `sudo_approval_response` from the server finds no pending request in `takeRequestLocked` and returns without writing.

## Warnings

**`sendIsAlpconResponse` was left out of the log-level change** (`auth_manager.go:563`). `sendSudoApprovalResponse` now routes through `logSudoResponseWriteError` and `sendSessionEventResponse` logs at warn, but this one still logs every write failure at error level. `check_user` is answered on every PAM sudo hook, so a client that already left produces exactly the broken-pipe noise this PR set out to demote. This is Copilot's suppressed comment on f7b8afd and it is a fair one. Not a blocker—the deadline that actually matters was added—but the inconsistency will read as an oversight later.

## Suggestions

**Inline denials now sit on the session-start path.** `pidHook` in `pkg/executor/handlers/shell/shell.go:175` calls `RegisterCommandPID` right after the child starts, and `pty.go:135` runs during Websh setup; both can now spend up to `5 × N` seconds writing to N displaced clients before returning. In practice N is 0, and when it is not, the peer is parked in `read()` so the write returns immediately—`authSocketWriteTimeout` is what keeps this a note rather than a warning. If you want it off the start path entirely, the requests are already detached by then, so `go am.denyPendingRequests(replaced, ...)` would be safe. I would leave it inline.

**The unmatched-response branch lost its only above-debug trace.** `HandleSudoApprovalResponse` still loops over every session dumping `Requests` before the new debug line, and `control_client.go` now discards the returned error. A server answering an id alpamon never registered—a real protocol bug, as opposed to the late arrival this branch is written for—is now invisible above debug. Reasonable trade, but it is a trade, and the loop above it is redundant with the new line.

**On your own review questions**, two are answerable from this repo:

- The deny reason strings never reach alpacon-server. `sendSudoApprovalResponse` writes only to the unix socket; the sole outbound POST is `sendSudoRequestWithRetry`, which sends a `SudoApprovalRequest` carrying no reason. So the audience for `"Session ended"` / `"Session replaced"` / `"Communication error"` is alpamon-pam, and the compatibility question belongs in that repo, not the server's. Worth correcting in the PR body so the next reader does not go looking server-side.
- 5s for `authSocketWriteTimeout` sits sensibly between the 25s retry budget and the PAM client's 30s, and every caller closes the connection right after the write, so no path can accumulate two deadlines.

## Highlights

- **The completion-channel move is a second fix the description does not claim credit for.** Creating the channel inside the registration critical section (`newCompletionChannelLocked`, `:502`) closes a window where a teardown landing between the old `am.mu.Unlock()` and `newCompletionChannel` would signal a channel that did not exist yet, leaving the waiter parked the full 30s on a request that was already answered and whose connection was closed.
- `HandleSudoApprovalResponse` now closes the connection before it checks the write error. On `main` the error path returned without closing and leaked the descriptor.
- `AddPIDSessionMapping` filling a nil `Requests` map removes a latent nil-map-write panic: the insert at `:496` writes straight into `session.Requests`, and nothing on that path previously guaranteed a caller-supplied `SessionInfo` had one.
- The tests are the real thing, not happy-path decoration. `expectSingleResponse` asserts a decode **followed by EOF**, which pins at-most-once and at-least-once in a single assertion, and `registerPipeRequest` uses an unbuffered `net.Pipe` so a missing write blocks instead of passing silently. I verified they bite—see below.

## Verification

Everything below was run in a clean worktree at `f7b8afd`, on darwin/amd64 with go1.25.12 (the version `.github/workflows/build-and-test.yml` pins).

| Command | Result |
| --- | --- |
| `go generate ./pkg/db/ent` | ok (required before build; ent output is gitignored) |
| `go build -v ./...` | ok |
| `go vet ./...` | exit 0, no output |
| `gofmt -l .` | no output (ignoring generated `pkg/db/ent/`) |
| `go test -race -count=1 ./pkg/runner/` | `ok ... 54.760s` |
| `go test -race -count=1 ./...` | exit 0, 45 packages `ok` + 35 with no test files, no `FAIL`, no `DATA RACE` |
| `go test ./... -p 1 -count=1` (CI's own command) | exit 0, same 45 + 35, no `FAIL` |
| `CGO_ENABLED=0 GOOS=linux golangci-lint run --timeout=5m ./...` | `0 issues` |

`golangci-lint` was not installed on this machine. I installed v2.9.0 to match the version pinned in `.github/workflows/lint.yml`; CI may still resolve a different patch build. Run under darwin's default build tags it reports two `unused` findings, `cmd/alpamon/command/register/register.go:31` and `pkg/runner/commit_types.go:162`—both in files this PR does not touch, and both disappear under CI's `CGO_ENABLED=0 GOOS=linux`, so they are platform artifacts rather than anything this branch introduced.

**Regression check.** A true base-revert is not possible here: the new test files reference `finalizeRequest`, `SudoRequest.Request`, and `registerCompletionChannel`, none of which exist on `origin/main`, so they cannot compile against base. I reintroduced the two defects on top of the branch instead and confirmed the tests fail:

1. Restored the pre-fix ordering in the retry-failure path (inline `sendSudoApprovalResponse` → `finalizeRequest` → `Close`):

```
--- FAIL: TestHandleSudoApprovalRequest_RetryFailureAnswersOnce (0.01s)
    auth_manager_cleanup_test.go:319: expected EOF after single response, got <nil>
```

That is the exact failure quoted in the PR body, reproduced independently.

2. Dropped `denyPendingRequests` from `RemovePIDSessionMapping`:

```
--- FAIL: TestRemovePIDSessionMapping_DeniesPendingRequestsAndSignalsWaiter (2.00s)
    auth_manager_cleanup_test.go:175: decoding response: read pipe: i/o timeout
```

Both mutations were reverted and the worktree confirmed clean with `git status --short`.

**Not covered by any test:** the completion-channel-under-lock change. The suite pins the ownership rule, not the narrow window between registration and channel creation, and forcing that interleaving would need injected scheduling. Worth stating plainly rather than implying the coverage is total.

**Not run:** nothing else. Every command listed above executed to completion.

## Verdict

**APPROVE.** The named mechanism is the real one, exactly-once holds in both directions under the ownership rule, no lock is held across a socket write, and the regression tests genuinely fail without the fix. The findings above are all non-blocking; the `sendIsAlpconResponse` log level is the one I would take in a follow-up.

## Checklist

- [x] Bug mechanism confirmed in the code, not just the description
- [x] At-most-once verified (single take site per path, all under `am.mu`)
- [x] At-least-once verified (every exit path enumerated)
- [x] Go concurrency reviewed (no double close, no leaked goroutine or channel, no lock held across I/O)
- [x] Boundary idempotency considered (client-side fix, no transport retry, server-side duplicate handled)
- [x] Tests proven to fail without the fix (two mutation checks)
- [x] `go build` / `go vet` / `gofmt` / `go test -race` / `golangci-lint` all run and reported
- [x] Worktree restored clean

pkg/runner/auth_manager.go
jisung-02: Fixed in 7a41ef64. The level decision moved out of `logSudoResponseWriteError` into `authSocketWriteLevel(err) zerolog.Level`, and this write now shares it:

```go
_, err = conn.Write(responseJSON)
if err != nil {
	log.WithLevel(authSocketWriteLevel(err)).Err(err).Msg("Failed to send is_alpacon_response")
}
```

`sendSudoApprovalResponse` keeps its own wrapper only because it adds the `request_id` field. `sendSessionEventResponse` still logs unconditionally at warn, which stays correct for that path — it never wants error level.

The dead `return` at the end of the function went with it.

pkg/runner/auth_manager.go
jisung-02: Left inline, as you prefer. I walked the same path before deciding: `RegisterCommandPID` and `AddPIDSessionMapping` only reach `denyPendingRequests` for a pid that is being reused while requests are still registered against the old entry, so N is 0 on every ordinary start, and a peer that is parked in `read()` takes the write immediately. `authSocketWriteTimeout` is what bounds the pathological case, and moving the call to a goroutine would trade an obvious ordering for a bound we already have.

No code change for this one, so this thread is the record of the rejected alternative.

junho226 (approved):
## Round 2

Two commits since `f7b8afd`, both in `pkg/runner/auth_manager.go`, +13/−16. No production behavior on the exactly-once path changed, so the ownership argument from round 1 still stands as written.

| Prior finding | Status |
|---|---|
| Warning — `sendIsAlpconResponse` kept error level while its siblings moved off it | **Fixed** (`7a41ef6`) |
| Suggestion — inline denials on the session/command-start path | **Answered, and I agree** |
| Suggestion — the unmatched-response branch has no trace above debug | **Half taken** (`8806384`), with the remainder named honestly |
| Suggestion — deny-reason compatibility belongs to alpamon-pam | Answered in the PR thread |

### The classifier extraction is the right shape

`authSocketWriteLevel(err) zerolog.Level` now owns the decision and both writers consult it. That is better than what I asked for: I suggested sharing the classifier, and the split cleanly separates *which level* from *what to log*, so `logSudoResponseWriteError` keeps its `request_id` field without owning policy.

The doc comment moved to the classifier and was rewritten from "a denial reaches a torn-down session" to "a response reaches a torn-down session". That generalization is correct now that `check_user` answers go through it too — an easy thing to leave stale, and it wasn't.

One detail I checked rather than assumed: the `return` dropped from the error branch was the function's last statement, so removing it changes nothing. Confirmed by reading the whole function, not just the hunk.

### The unmatched-response branch

Half-taking this was the right call, and saying so plainly is better than a silent partial fix. The loop that went is a real improvement on its own terms: it reformatted every session's pending requests under a read lock to restate what the `log.Debug()` on the next line already said, so it cost a lock acquisition and a `%+v` of a map to add nothing.

What remains is what I named and you restated: a server answering a request id alpamon never registered is still indistinguishable, above debug level, from the ordinary late arrival this branch was written for. Separating those two is a different change with a different risk, and it does not belong in a PR about exactly-once. Leaving it explicit in the thread is the right resting place.

### Session/command-start path

Your walk matches mine. `denyPendingRequests` is only reached there for a pid being reused while requests are still registered against the old entry, so N is 0 on every ordinary start; when it is not, the peer is parked in `read()` and the write returns immediately. `authSocketWriteTimeout` bounds the pathological case. Moving it to a goroutine would trade a clear ordering for a bound that already exists. Keeping it inline is correct.

## New findings

None.

One observation, not a request: `authSocketWriteLevel` has no direct test. Neither did `logSudoResponseWriteError` before this PR, and the refactor is a pure extraction — same condition, same two branches — so this is not a regression and I would not add a test for a six-line log-level helper. Recording it so the coverage claim stays accurate rather than implied.

## Verification

Run in the worktree at `8806384`, go1.25.12 darwin/amd64.

| Command | Result |
|---|---|
| `go generate ./pkg/db/ent` | ok (required before build; ent output is gitignored) |
| `go build ./...` | pass, exit 0 |
| `go vet ./...` | pass, exit 0, no output |
| `gofmt -l .` | pass, no files listed (ignoring generated `pkg/db/ent/`) |
| `go test -race -count=1 ./...` | **pass** — exit 0, 45 packages `ok`, no `FAIL`, no `DATA RACE`. `pkg/runner` 87.1s |
| `CGO_ENABLED=0 GOOS=linux golangci-lint run --timeout=5m ./...` | **pass** — `0 issues.` |

`golangci-lint` is not on this machine's default `PATH`; I ran the v2.9.0 binary installed under `$(go env GOPATH)/bin` to match the version pinned in `.github/workflows/lint.yml`. CI may still resolve a different patch build.

**Not re-run:** the two mutation checks from round 1 (restoring the pre-fix ordering, and dropping `denyPendingRequests` from `RemovePIDSessionMapping`). The delta does not touch either code path — it is confined to log-level classification and a removed debug loop — so round 1's result carries over unchanged. Saying which evidence is carried forward rather than freshly measured.

**Could not run:** nothing.

## Verdict

**APPROVE.** The warning is closed by a cleaner change than the one I proposed, and both suggestions got a reasoned answer rather than a reflexive edit. Nothing new opened.

## Checklist

- [x] Prior findings closed or fairly answered
- [x] Tests pass under `-race`
- [x] Lint clean under CI settings
- [x] No new defects in the delta

Issue #393, opened 2026-07-30
Support Arch and Alpine: a package-manager axis the server understands


jisung-02:
## Background

#348 split the platform value into three axes while adding openSUSE/SLES support (`pkg/utils/platform.go`).

| Axis | Values | Purpose |
| --- | --- | --- |
| `PlatformLike` | debian / rhel / darwin / windows | what we report to the server |
| `PackageManager` | apt / yum / zypper / brew / none | what we run locally |
| `PlatformID` | raw os-release id | the one distro-level branch (Tumbleweed) |

That split is what made "report rhel, run zypper" possible for SUSE. It held because the three things the server gates on `platform` (rpm packaging, the `wheel` sudo group, shadow-utils account tooling) are identical on RHEL and SUSE.

Arch and Alpine do not satisfy that premise, so there is no value to report for them and `utils.InitPlatform()` dies at startup. `pkg/utils/platform_test.go:75` pins these distributions as unclassified.

## Why no existing value covers them

| | Packaging | Sudo group | Account tooling | init |
| --- | --- | --- | --- | --- |
| Arch | pacman | `wheel` | shadow-utils | systemd |
| Alpine | apk | `wheel` | BusyBox `adduser`/`addgroup` | OpenRC |

`Server.platform` accepts only `debian`/`rhel`/`darwin`/`windows`, and the field is write-once. Reporting Arch as `rhel` makes the server compose `yum` commands; reporting it as `debian` makes it compose `apt-get`. Either fails, and a wrong value can only be corrected by registering the host again. The behavior-bucket argument used for SUSE does not transfer here.

## Work required

Agent side:

- Add `pacman` and `apk` to the `PackageManager` axis. The branch points are the three switches in `pkg/executor/handlers/system/system.go` (`:204` upgrade, `:500` uninstall, `:593` system update).
- Add both distribution ids to `ResolvePlatform`, which cannot happen until the `PlatformLike` value they report is decided (see the server side below).
- Alpine's account tooling differs. `pkg/executor/handlers/user/user.go:198,412` and `group.go:131,196` sit on the `PlatformLike` axis and invoke `/usr/sbin/useradd` and `/usr/sbin/groupadd`. A default Alpine install has no shadow package, and BusyBox `adduser`/`addgroup` take a different option set. Decide whether to require shadow as a dependency or branch on BusyBox.
[...]

Issue #348, opened 2026-07-04
openSUSE/SLES: platform misclassification (suse family falls back to debian)


eunyoung14:
## Summary

alpamon does not recognize the SUSE family (openSUSE Leap/Tumbleweed, SLES). Both platform-detection paths fall back to `debian`, which is worse than "unsupported"—it is **silent misclassification** that breaks user provisioning and sudo handling on the host.

## Where it happens

1. `cmd/alpamon/command/register/register.go` — `detectPlatform()` has no case for `opensuse-leap` / `opensuse-tumbleweed` / `sles` / `opensuse`, so the default fallback returns `"debian"`.
2. `pkg/executor/handlers/user/user.go:150-215` — the adduser handler branches on `utils.PlatformLike`; with the wrong `debian` value it takes the Debian path.

## Consequences of the wrong `debian` value

- The Debian path calls `/usr/sbin/adduser` (the Debian perl wrapper), which **does not exist on openSUSE/SLES** → user provisioning fails.
- The reported platform drives server-side sudo-group selection (`wheel` vs `sudo`); openSUSE/SLES use `wheel` and have no `sudo` group, so grants silently break.
- The platform value is effectively fixed after registration, so a wrong value persists until the host re-registers.

## Fix (small)

Map the SUSE family → **`rhel`** at both detection sites (~20 lines + tests). openSUSE/SLES are rpm-based, use the `wheel` group, and ship shadow-utils (`useradd`/`groupadd`/`usermod`), so the existing `rhel` handler path works verbatim—no new platform value needed.

Also:
- Note openSUSE/SLES as best-effort in the README support matrix.
- Verify in a Leap smoke test: openSUSE's default sudoers uses `Defaults targetpw` with `%wheel` commented out, so `wheel` membership alone does not confer sudo—needs a documented prerequisite or a `sudoers.d` drop-in.
- Hosts already registered as openSUSE carry a wrong (fixed) platform value and must re-register after this fix.

## Follow-up (separate)

Full official support also needs an openSUSE/SLES rpm distribution channel, a `zypper` install path, a Leap CI image, and `alpamon-pam` verification on the SUSE PAM stack—tracked separately from this detection fix.

Issue #129, opened 2025-11-14
systemd service restarts immediately after quit command


geunwoonoh:
### Description
When the `quit` command is executed via the internal command system, the alpamon service terminates and immediately restarts due to the `Restart=always` setting in the systemd service file.

**Steps to reproduce:**
1. Install alpamon as a systemd service
2. Execute internal command: `quit`
3. Service terminates and immediately restarts

### Expected Behavior
When the `quit` command is issued, the service should:
- ✅ Terminate gracefully (exit code 0)
- ✅ Stay stopped until manually restarted
- ✅ Still auto-start on system boot/reboot
- ✅ Still auto-restart on crashes/failures (non-zero exit codes)

### Root Cause
The current systemd service configuration in [`configs/alpamon.service`](https://github.com/alpacax/alpamon/blob/main/configs/alpamon.service) has:
```ini
[Service]
Restart=always
```
This causes systemd to restart the service regardless of the exit code.

### Proposed Solution
Change the `Restart` directive to `on-failure`:
```ini
[Service]
Restart=on-failure
```
With this change:
- Normal shutdown (quit command → exit code 0) = no restart
- Abnormal termination (crashes → exit code ≠ 0) = auto restart
- System reboot = auto start (controlled by WantedBy=multi-user.target)

### Affected Code
- [configs/alpamon.service:11](https://github.com/alpacax/alpamon/blob/main/configs/alpamon.service#L11) - Service configuration

### Additional Context
The quit command properly returns exit code 0 for graceful shutdown, so this is purely a systemd configuration issue rather than a code issue.

PR #414, opened 2026-08-13
fix(utils): fail once on an unclassifiable platform instead of looping

### LINKED ISSUE #397: Consider failing once, instead of restart-looping, on an unclassifiable distribution
## Background

`utils.InitPlatform()` ends in `log.Fatal` when it cannot classify the host (`pkg/utils/platform.go:167`), which exits 1. `configs/alpamon.service` sets `Restart=on-failure` and carries no `RestartPreventExitStatus` or `StartLimit*`, so systemd brings the agent straight back up, it fails at the same line, and the cycle repeats.

This is the symptom #348 reported for openSUSE. That fix, and the Arch/Alpine work proposed in #393, both widen which distributions classify, which removes the loop for those hosts specifically. The failure mode itself is untouched, so anything still unclassified keeps looping: Gentoo, NixOS, a derivative with an unfamiliar os-release id.

It may be worth separating the two concerns. Which distributions we accept is one question; how we decline the rest is another, and the second one seems cheap to improve.

## What an operator sees today

The journal fills with one repeated line, `Unsupported platform: os=linux distribution="gentoo"`, and there is little to distinguish "stopped because of a configuration problem" from "crashing repeatedly". `InitPlatform()` runs at `cmd/alpamon/command/root.go:107`, before `pidfile.WritePID` on line 110, so no PID file is left behind either. Nothing is corrupted, but there is also no state pointing at the cause.

## A possible approach

Give a restart-proof failure its own exit code and tell systemd not to retry it:

- have the unclassified-platform path exit with a configuration-error code, `EX_CONFIG` (78) being the usual convention
- add `RestartPreventExitStatus=78` to `configs/alpamon.service`

The unit would then settle in `failed` after one attempt, the reason would stay readable in the journal, and `systemctl status alpamon` would show it. #129 already established the "do not restart a deliberate exit" shape for the `quit` command; this extends the same idea to a configuration failure, which `Restart=on-failure` currently treats as transient.

Two things worth deciding rather than assuming:

- how far to extend the distinction. Other startup paths also exit 1, for instance the PID file failure at `root.go:113`. Platform classification is the clearest case; the rest may or may not belong.
- `pkg/plugin/plugin.go:177` calls `InitPlatform()` too, in a context that is not the systemd unit, so it may want different handling.

Not urgent, and not a blocker for #393 or #348, but it does look like the piece that would stop this class of report from recurring.

## References

- #348: the openSUSE case where this surfaced
- #393: adds distributions to the classifier, and notes this symptom as out of scope
- #129: the earlier "stay stopped after a deliberate exit" fix for `quit`

jisung-02:
## Background

`utils.InitPlatform()` ended in `log.Fatal` when the classifier could not map the host, which logs and exits 1. The systemd unit sets `Restart=on-failure`, so exit 1 reads as transient and the unit relaunches straight into the same failure. It ends at the default start limit (`DefaultStartLimitBurst=5` over `DefaultStartLimitIntervalSec=10s`) reporting:

```
start request repeated too quickly
```

That is what `systemctl status alpamon` shows. The line that actually names the cause — `Unsupported platform: os=linux distribution="gentoo"` — sits buried in five repetitions of journal output. `InitPlatform` also runs before `pidfile.WritePID`, so there is no PID file left behind to point at either.

An unclassifiable distribution is not a transient condition. Restarting cannot clear it.

## Changes

The classification failure now travels back as a sentinel instead of killing the process:

- `pkg/utils/platform.go`: `InitPlatform` returns an error wrapping `ErrUnsupportedPlatform` rather than calling `log.Fatal`. The host lookup splits from the classification half (`applyPlatform`), the same way `DetectRegistrationPlatform` already splits from `ResolveRegistrationPlatform`, so the sentinel and the globals it writes are reachable without a host. The host half is shared too: `rawHostPlatform` replaces the `if runtime.GOOS == "linux" { host.Info() }` block that `InitPlatform` and `DetectRegistrationPlatform` each carried, returning the raw id and leaving the wrapping to callers whose error texts differ.
- `pkg/utils/exitcode.go`: new `ConfigErrorExitCode = 78` (`sysexits.h` `EX_CONFIG`), and `StartupExitCode(err)` which maps a startup failure to its exit status — the sentinel to 78, anything else to a restartable 1.
- `cmd/alpamon/command/root.go:107` and `pkg/plugin/plugin.go:183`: both callers route their failure through `StartupExitCode`. `pkg/plugin` returns rather than exits, so its deferred cleanup chain and its own exit path still run — both were skipped by `log.Fatal`.
- `pkg/plugin` package doc: plugin repos import this SDK but ship their own unit files, so the doc states the other half of the contract — a plugin unit needs `RestartPreventExitStatus=78` or it keeps restart-looping on the code the SDK now returns.
- `configs/alpamon.service`: `RestartPreventExitStatus=78` alongside the existing `Restart=on-failure`.
- `README.md`: what an operator sees after an EX_CONFIG exit, in the Service management section.

systemd no longer restarts the unit by itself: it exits 78 once and stays `failed` with `systemctl status` naming the reason. `RestartPreventExitStatus` only blocks the automatic restart, and this repo also ships `configs/alpamon-restart.timer`, which `scripts/postinstall.sh` enables. That timer sends an explicit `systemctl restart` five minutes after activation and once per boot, and an external restart is not blocked — so the unit takes that one extra attempt, exits 78 again, and settles back into `failed`. `OnActiveSec` without `OnUnitActiveSec` fires once per activation, so this is a single retry, not a loop. This is the same "do not restart a deliberate exit" shape #129 established for `quit`.

The exit code is duplicated in `configs/alpamon.service` because systemd parses `RestartPreventExitStatus` as a literal and cannot read a Go constant. A test pins the two together (see Testing).

Deliberately left out of scope:

- `host.Info()` failures still exit 1. A missing `os-release` is not a verdict on which distributions are supported, and unlike classification it may clear on a retry.
- The PID file failure at `root.go:113` still exits 1. A stale pidfile is the kind of thing a restart does clear.
- Which distributions classify is untouched. That stays with #348 and #393.

The same "no restart can clear this" failure class still exits 1 elsewhere — `pkg/config/config.go:97,102,108,120` and `root.go:178` are all `log.Fatal`. Those are follow-up work, not part of this change.

## Testing

New in `pkg/utils`:

- `TestApplyPlatform_UnsupportedWrapsSentinel`: the classification failure wraps `ErrUnsupportedPlatform` and names the distribution, and leaves `PlatformLike` / `PackageManager` / `PlatformID` untouched. Without the wrap, `StartupExitCode` silently falls back to a restartable exit 1.
- `TestApplyPlatform_SetsGlobals`: all three globals land from their own source across debian, rhel, SUSE, darwin and windows.
- `TestStartupExitCode`: a wrapped sentinel maps to 78 and a host lookup failure to 1. The callers themselves end in `os.Exit` or a return code, neither reachable from a test, so the mapping is what gets pinned.
- `TestConfigErrorExitCodeMatchesUnitFile`: `configs/alpamon.service` sets `RestartPreventExitStatus=78` as an active directive under `[Service]`. Matched line by line rather than as a substring, since a commented-out or misplaced directive is inert to systemd and would otherwise keep the test green while the restart loop returns. Confirmed by mutating the unit file each way and watching the test fail.

`TestResolvePlatform_CIHost` re-read the host itself and called `ResolvePlatform` directly, so the CI matrix never entered `InitPlatform`. It is now `TestInitPlatform_CIHost` and calls what the agent actually runs at startup.

`applyPlatform`, `SetPlatformLike`, `SetPackageManager` and `SetPlatformID` go from 0% to full statement coverage.

Local run on darwin/arm64: `gofmt -l .`, `go build ./...` and `go vet ./...` clean, `go test ./... -p 1 -count=1` all `ok` (`pkg/utils 0.414s`, `pkg/plugin 0.284s`).

Not verified: this has not been run under a real systemd unit to observe the `failed` state end to end. The assertion above is that the unit file and the constant agree, and that `StartupExitCode` maps the sentinel to 78 — not that systemd behaves as documented.

Closes #397

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,8 @@ sudo systemctl restart alpamon
 sudo journalctl -u alpamon -f
 ```
 
+The unit also sets `RestartPreventExitStatus=78`. Alpamon exits with 78 (`EX_CONFIG`) when startup hits a condition a restart cannot clear, such as a Linux distribution it cannot classify. systemd then leaves the service in `failed` instead of restarting it, so `systemctl status alpamon` still shows the reason. Fix the underlying condition and start the service again.
+
 The unit sets `KillMode=process`, so restarting or stopping alpamon signals only the agent itself, not the whole control-group. Sessions it launched—Websh terminals and detached jobs such as `tmux`, `screen`, or `nohup`'d commands—keep running across a restart, the same way `sshd` leaves active login sessions alone. Reattach to them from a new session after the agent reconnects.
 
 ### macOS (launchd)
--- a/README.md
+++ b/README.md
@@ -147,9 +147,9 @@ sudo systemctl restart alpamon
 sudo journalctl -u alpamon -f
 ```
 
-The unit also sets `RestartPreventExitStatus=78`. Alpamon exits with 78 (`EX_CONFIG`) when startup hits a condition a restart cannot clear, such as a Linux distribution it cannot classify. systemd then leaves the service in `failed` instead of restarting it, so `systemctl status alpamon` still shows the reason. Fix the underlying condition and start the service again.
-
 The unit sets `KillMode=process`, so restarting or stopping alpamon signals only the agent itself, not the whole control-group. Sessions it launched—Websh terminals and detached jobs such as `tmux`, `screen`, or `nohup`'d commands—keep running across a restart, the same way `sshd` leaves active login sessions alone. Reattach to them from a new session after the agent reconnects.
+
+The unit also sets `RestartPreventExitStatus=78`. Alpamon exits with 78 (`EX_CONFIG`) when startup hits a condition a restart cannot clear, such as a Linux distribution it cannot classify. systemd then leaves the service in `failed` instead of restarting it, so `systemctl status alpamon` still shows the reason. Fix the underlying condition and start the service again.
 
 ### macOS (launchd)
 
--- a/cmd/alpamon/command/root.go
+++ b/cmd/alpamon/command/root.go
@@ -2,7 +2,6 @@ package command
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -107,10 +106,7 @@ func runAgent(ready chan<- struct{}) {
 	// platform
 	if err := utils.InitPlatform(); err != nil {
 		log.Error().Err(err).Msg("Failed to initialize platform.")
-		if errors.Is(err, utils.ErrUnsupportedPlatform) {
-			os.Exit(utils.ConfigErrorExitCode)
-		}
-		os.Exit(1)
+		os.Exit(utils.StartupExitCode(err))
 	}
 
 	// Pid
--- a/cmd/alpamon/command/root.go
+++ b/cmd/alpamon/command/root.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -104,7 +105,13 @@ func runAgent(ready chan<- struct{}) {
 	updater.CleanupStaleOld()
 
 	// platform
-	utils.InitPlatform()
+	if err := utils.InitPlatform(); err != nil {
+		log.Error().Err(err).Msg("Failed to initialize platform.")
+		if errors.Is(err, utils.ErrUnsupportedPlatform) {
+			os.Exit(utils.ConfigErrorExitCode)
+		}
+		os.Exit(1)
+	}
 
 	// Pid
 	pidFilePath, err := pidfile.WritePID(pidfile.FilePath(name))
--- a/configs/alpamon.service
+++ b/configs/alpamon.service
@@ -12,6 +12,8 @@ KillMode=process
 SELinuxContext=-unconfined_u:unconfined_r:unconfined_t:s0
 AppArmorProfile=-unconfined
 Restart=on-failure
+# EX_CONFIG from utils.ConfigErrorExitCode: a startup failure no restart can clear.
+RestartPreventExitStatus=78
 StandardOutput=journal
 StandardError=journal
 SyslogIdentifier=alpamon
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -7,6 +7,12 @@
 //
 // A plugin defines a Plugin value and either calls Plugin.Run directly or
 // embeds it under a Cobra root command via NewRootCmd.
+//
+// Run exits with utils.ConfigErrorExitCode (78) when startup hits a condition
+// no restart can clear, such as a host the platform classifier cannot map. A
+// plugin's service unit should carry RestartPreventExitStatus=78 alongside its
+// Restart= setting, the way configs/alpamon.service does; without it systemd
+// relaunches straight back into the same failure.
 package plugin
 
 import (
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -11,6 +11,7 @@ package plugin
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -174,7 +175,13 @@ func (p *Plugin) run() int {
 	if p.InitLogger != nil {
 		p.InitLogger()
 	}
-	utils.InitPlatform()
+	if err := utils.InitPlatform(); err != nil {
+		log.Error().Err(err).Msg("Failed to initialize platform.")
+		if errors.Is(err, utils.ErrUnsupportedPlatform) {
+			return utils.ConfigErrorExitCode
+		}
+		return 1
+	}
 
 	pidFilePath, err := pidfile.WritePID(pidfile.FilePath(p.Name))
 	if err != nil {
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -17,7 +17,6 @@ package plugin
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -183,10 +182,7 @@ func (p *Plugin) run() int {
 	}
 	if err := utils.InitPlatform(); err != nil {
 		log.Error().Err(err).Msg("Failed to initialize platform.")
-		if errors.Is(err, utils.ErrUnsupportedPlatform) {
-			return utils.ConfigErrorExitCode
-		}
-		return 1
+		return utils.StartupExitCode(err)
 	}
 
 	pidFilePath, err := pidfile.WritePID(pidfile.FilePath(p.Name))
--- a/pkg/utils/exitcode_test.go
+++ b/pkg/utils/exitcode_test.go
@@ -1,11 +1,31 @@
 package utils
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
 	"testing"
 )
+
+// The callers turn this into os.Exit / a return code, neither reachable from a test, so the mapping itself is what gets pinned.
+func TestStartupExitCode(t *testing.T) {
+	tests := []struct {
+		name string
+		err  error
+		want int
+	}{
+		{"wrapped sentinel", fmt.Errorf("%w: os=linux distribution=%q", ErrUnsupportedPlatform, "gentoo"), ConfigErrorExitCode},
+		{"host lookup failure", fmt.Errorf("failed to retrieve platform information: %w", errors.New("no os-release")), 1},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := StartupExitCode(tt.err); got != tt.want {
+				t.Errorf("StartupExitCode(%v) = %d, want %d", tt.err, got, tt.want)
+			}
+		})
+	}
+}
 
 // systemd parses RestartPreventExitStatus as a literal and cannot read the Go constant, so drift between the two silently restores the restart loop this code exists to stop.
 func TestConfigErrorExitCodeMatchesUnitFile(t *testing.T) {
--- a/pkg/utils/exitcode_test.go
+++ b/pkg/utils/exitcode_test.go
@@ -1,0 +1,20 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+)
+
+// systemd parses RestartPreventExitStatus as a literal and cannot read the Go constant, so drift between the two silently restores the restart loop this code exists to stop.
+func TestConfigErrorExitCodeMatchesUnitFile(t *testing.T) {
+	unit, err := os.ReadFile("../../configs/alpamon.service")
+	if err != nil {
+		t.Fatalf("failed to read the unit file: %v", err)
+	}
+	want := fmt.Sprintf("RestartPreventExitStatus=%d", ConfigErrorExitCode)
+	if !strings.Contains(string(unit), want) {
+		t.Errorf("configs/alpamon.service must contain %q", want)
+	}
+}
--- a/pkg/utils/exitcode_test.go
+++ b/pkg/utils/exitcode_test.go
@@ -14,7 +14,17 @@ func TestConfigErrorExitCodeMatchesUnitFile(t *testing.T) {
 		t.Fatalf("failed to read the unit file: %v", err)
 	}
 	want := fmt.Sprintf("RestartPreventExitStatus=%d", ConfigErrorExitCode)
-	if !strings.Contains(string(unit), want) {
-		t.Errorf("configs/alpamon.service must contain %q", want)
+	// A commented-out or misplaced directive is inert to systemd, so a substring match would pass on a unit that still restart-loops.
+	var section string
+	for line := range strings.SplitSeq(string(unit), "\n") {
+		line = strings.TrimSpace(line)
+		if strings.HasPrefix(line, "[") && strings.HasSuffix(line, "]") {
+			section = line
+			continue
+		}
+		if section == "[Service]" && line == want {
+			return
+		}
 	}
+	t.Errorf("configs/alpamon.service must set %q under [Service]", want)
 }
--- a/pkg/utils/platform.go
+++ b/pkg/utils/platform.go
@@ -17,6 +17,10 @@ var (
 	PlatformID     string
 )
 
+// ErrUnsupportedPlatform marks the classification failure callers translate
+// into ConfigErrorExitCode.
+var ErrUnsupportedPlatform = errors.New("unsupported platform")
+
 // Mirrors alpacon-server servers/constants.py PLATFORM_ALIASES 1:1 — never add ids the server lacks: write-once Server.platform makes a divergence permanent.
 var platformAliases = map[string]string{
 	"debian":    "debian",
@@ -119,18 +123,26 @@ func ResolveServerPlatform(goos, explicit string, detect func() (string, error))
 	return explicit, "", nil
 }
 
+// The raw distribution id for this host; empty off Linux, where GOOS alone classifies. The callers' error texts differ, so wrapping stays with them.
+func rawHostPlatform() (string, error) {
+	if runtime.GOOS != "linux" {
+		return "", nil
+	}
+	info, err := host.Info()
+	if err != nil {
+		return "", err
+	}
+	return info.Platform, nil
+}
+
 // The value register and migrate send; errors instead of defaulting because the server persists it write-once, fixable only by re-registering.
 func DetectRegistrationPlatform() (string, error) {
-	var raw string
-	if runtime.GOOS == "linux" {
-		info, err := host.Info()
-		if err != nil {
-			return "", fmt.Errorf(
-				"failed to detect the host platform: %w.\n"+
-					"Pass --platform debian|rhel to skip detection if you know the host is compatible",
-				err)
-		}
-		raw = info.Platform
+	raw, err := rawHostPlatform()
+	if err != nil {
+		return "", fmt.Errorf(
+			"failed to detect the host platform: %w.\n"+
+				"Pass --platform debian|rhel to skip detection if you know the host is compatible",
+			err)
 	}
 	return ResolveRegistrationPlatform(runtime.GOOS, raw)
 }
@@ -149,19 +161,11 @@ func ResolveRegistrationPlatform(goos, raw string) (string, error) {
 	return like, nil
 }
 
-// ErrUnsupportedPlatform marks the classification failure callers translate
-// into ConfigErrorExitCode.
-var ErrUnsupportedPlatform = errors.New("unsupported platform")
-
 // Classifies the host once at startup; the error is terminal for the caller because every handler switches on these values and guessing would misreport to the server.
 func InitPlatform() error {
-	var raw string
-	if runtime.GOOS == "linux" {
-		info, err := host.Info()
-		if err != nil {
-			return fmt.Errorf("failed to retrieve platform information: %w", err)
-		}
-		raw = info.Platform
+	raw, err := rawHostPlatform()
+	if err != nil {
+		return fmt.Errorf("failed to retrieve platform information: %w", err)
 	}
 	return applyPlatform(runtime.GOOS, raw)
 }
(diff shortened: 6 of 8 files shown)


pkg/utils/exitcode_test.go
```diff
@@ -0,0 +1,20 @@
...
+// systemd parses RestartPreventExitStatus as a literal and cannot read the Go constant, so drift between the two silently restores the restart loop this code exists to stop.
+func TestConfigErrorExitCodeMatchesUnitFile(t *testing.T) {
+	unit, err := os.ReadFile("../../configs/alpamon.service")
+	if err != nil {
+		t.Fatalf("failed to read the unit file: %v", err)
+	}
+	want := fmt.Sprintf("RestartPreventExitStatus=%d", ConfigErrorExitCode)
+	if !strings.Contains(string(unit), want) {
+		t.Errorf("configs/alpamon.service must contain %q", want)
+	}
```
geunwoonoh: Warning: `strings.Contains` also passes on a commented-out directive.

This test exists to catch drift, and it misses one form of exactly that drift. Comment the directive out in the unit file:

```
# RestartPreventExitStatus=78
```

The substring is still present, so the test stays green while systemd ignores the directive and the restart loop quietly returns. Matching against the whole file also accepts the directive misplaced under `[Unit]`.

Parsing line by line and accepting only an active directive closes both:

```suggestion
	found := false
	for _, line := range strings.Split(string(unit), "\n") {
		line = strings.TrimSpace(line)
		if strings.HasPrefix(line, "#") || strings.HasPrefix(line, ";") {
			continue
		}
		if line == want {
			found = true
			break
		}
	}
	if !found {
		t.Errorf("configs/alpamon.service must contain an active %q directive", want)
	}
```

configs/alpamon.service
```diff
@@ -12,6 +12,8 @@ KillMode=process
 SELinuxContext=-unconfined_u:unconfined_r:unconfined_t:s0
 AppArmorProfile=-unconfined
 Restart=on-failure
+# EX_CONFIG from utils.ConfigErrorExitCode: a startup failure no restart can clear.
+RestartPreventExitStatus=78
```
geunwoonoh, line 16: Warning: an explicit `systemctl restart` bypasses this directive.

The repo already ships a restart timer, and `scripts/postinstall.sh:154-156` enables and starts it:

```
configs/alpamon-restart.service: ExecStart=/bin/systemctl restart alpamon.service
configs/alpamon-restart.timer:   OnActiveSec=5min, WantedBy=multi-user.target
```

`RestartPreventExitStatus` only blocks systemd's *automatic* restart. A `systemctl restart` arriving from outside is not blocked and clears the failed state. On an unclassifiable host the real sequence is:

1. postinstall starts the unit, it exits 78, systemd leaves it `failed` with no auto-restart
2. five minutes later the timer restarts it, it exits 78, `failed` again

`OnActiveSec` with `Persistent=false` and no `OnUnitActiveSec` fires once per timer activation, so this is **not a loop**, and the end state is still `failed` with the reason readable. The change achieves what it set out to. The timer is `WantedBy=multi-user.target` though, so one extra attempt also follows every boot.

No code change needed here. What is inaccurate is the description's "The unit now settles in `failed` after one attempt", on both the install and the boot path.

pkg/plugin/plugin.go
```diff
@@ -174,7 +175,13 @@ func (p *Plugin) run() int {
 	if p.InitLogger != nil {
 		p.InitLogger()
 	}
-	utils.InitPlatform()
+	if err := utils.InitPlatform(); err != nil {
```
geunwoonoh, line 183: Warning: the plugin side only gets half the fix.

`pkg/plugin` is the public SDK that plugin repos (alpamon-dhcp-plugin and friends) import. It now returns 78, but the half that actually stops the loop, `RestartPreventExitStatus=78`, lives only in this repo's `configs/alpamon.service`. Plugin unit files do not have it, so a plugin still restart-loops on exit 78.

Not a regression against the previous exit 1 loop, just unfinished. Stating the contract in the package doc comment keeps plugin authors from missing it:

```go
// Package plugin ...
//
// A plugin's service unit should set RestartPreventExitStatus to
// utils.ConfigErrorExitCode (78); Run returns it when startup hits a
// condition no restart can clear.
```

cmd/alpamon/command/root.go
```diff
@@ -104,7 +105,13 @@ func runAgent(ready chan<- struct{}) {
...
 
 	// platform
-	utils.InitPlatform()
+	if err := utils.InitPlatform(); err != nil {
+		log.Error().Err(err).Msg("Failed to initialize platform.")
+		if errors.Is(err, utils.ErrUnsupportedPlatform) {
+			os.Exit(utils.ConfigErrorExitCode)
+		}
+		os.Exit(1)
+	}
```
geunwoonoh, line 110: Suggestion: extracting this mapping closes the gap the description already names.

These six lines are identical to `pkg/plugin/plugin.go:178-184`. The description notes *"Not verified: the `errors.Is` to `os.Exit(78)` mapping in `root.go` and `plugin.go` has no test"*, diagnosed as untestable without injecting process exit. Pull the mapping out and it is a pure function that tests directly:

```go
// pkg/utils/exitcode.go
// StartupExitCode maps a startup failure to its exit status: a condition no
// restart can clear gets ConfigErrorExitCode, anything else stays restartable.
func StartupExitCode(err error) int {
	if errors.Is(err, ErrUnsupportedPlatform) {
		return ConfigErrorExitCode
	}
	return 1
}
```

```go
// root.go
if err := utils.InitPlatform(); err != nil {
	log.Error().Err(err).Msg("Failed to initialize platform.")
	os.Exit(utils.StartupExitCode(err))
}

// plugin.go
if err := utils.InitPlatform(); err != nil {
	log.Error().Err(err).Msg("Failed to initialize platform.")
	return utils.StartupExitCode(err)
}
```

Left as a plain block rather than a committable suggestion, since it needs the new `utils` helper first.

Separately, on line 109: switching to `log.Error()` is correct, since zerolog's `log.Fatal()` calls `os.Exit(1)`. Worth knowing that anything alerting on `"level":"fatal"` in the journal loses this event, and that `Msg("Failed to initialize platform.")` restates what the error text already carries.

pkg/utils/platform.go
```diff
@@ -150,30 +149,39 @@ func ResolveRegistrationPlatform(goos, raw string) (string, error) {
 	return like, nil
 }
 
-// Classifies the host once at startup; unclassifiable is fatal because every handler switches on these values and guessing would misreport to the server.
-func InitPlatform() {
+// ErrUnsupportedPlatform marks the classification failure callers translate
+// into ConfigErrorExitCode.
+var ErrUnsupportedPlatform = errors.New("unsupported platform")
```
geunwoonoh: Suggestion: Go convention groups sentinel errors near the top of the file with the other package-level vars. This one sits mid-file, away from the `PlatformLike` / `PackageManager` / `PlatformID` block at lines 13-17. Discoverability only, not behavior.

pkg/utils/platform.go
```diff
@@ -150,30 +149,39 @@ func ResolveRegistrationPlatform(goos, raw string) (string, error) {
...
 		info, err := host.Info()
 		if err != nil {
-			log.Error().Err(err).Msg("Failed to retrieve platform information.")
-			os.Exit(1)
+			return fmt.Errorf("failed to retrieve platform information: %w", err)
 		}
 		raw = info.Platform
 	}
+	return applyPlatform(runtime.GOOS, raw)
+}
```
geunwoonoh, line 171: Suggestion: the host lookup block sits in two places verbatim.

`DetectRegistrationPlatform` at lines 123-137 carries the same `if runtime.GOOS == "linux" { host.Info() }`. This PR split the pure half to restore the symmetry with `ResolveRegistrationPlatform`; folding the host half completes it. The two error messages differ, so the shared helper should return the raw id and leave wrapping to the callers:

```go
// The raw distribution id for this host; empty on non-Linux, where GOOS alone classifies.
func rawHostPlatform() (string, error) {
	if runtime.GOOS != "linux" {
		return "", nil
	}
	info, err := host.Info()
	if err != nil {
		return "", err
	}
	return info.Platform, nil
}
```

README.md
```diff
@@ -147,6 +147,8 @@ sudo systemctl restart alpamon
 sudo journalctl -u alpamon -f
 ```
 
+The unit also sets `RestartPreventExitStatus=78`. Alpamon exits with 78 (`EX_CONFIG`) when startup hits a condition a restart cannot clear, such as a Linux distribution it cannot classify. systemd then leaves the service in `failed` instead of restarting it, so `systemctl status alpamon` still shows the reason. Fix the underlying condition and start the service again.
```
geunwoonoh: Suggestion: "also" refers forward, not back.

This paragraph opens with "The unit **also** sets `RestartPreventExitStatus=78`", but the earlier unit setting it points back to, "The unit sets `KillMode=process`", is the paragraph *after* it. Dropping "also" or swapping the two paragraphs puts the reading order right.

geunwoonoh (dismissed):
## Review summary

`utils.InitPlatform()` now returns an `ErrUnsupportedPlatform` sentinel instead of ending in `log.Fatal`, and both callers translate it into exit 78 (`EX_CONFIG`). `RestartPreventExitStatus=78` on the unit breaks the restart loop, and a test pins the constant to the unit file.

The direction is right. Separating "no restart can clear this" from "transient" by exit code, deliberately leaving `host.Info()` failures on exit 1, and splitting `applyPlatform` so the sentinel and the globals it writes are reachable without a host: each choice is justified in the change itself.

**Local verification** (darwin/arm64, after regenerating ent): `go build ./...`, `go vet ./...` and `gofmt -l .` all clean, `go test ./pkg/utils/ -count=1` reports `ok 0.425s`. Coverage confirms `applyPlatform` at 100% and `SetPlatformLike` / `SetPackageManager` / `SetPlatformID` at 100% each, so the coverage claim in the description holds.

### Critical issues (0)

None.

### Warnings (3)

**1. [`pkg/utils/exitcode_test.go:17`] `strings.Contains` also passes on a commented-out directive**

This test exists to catch drift, and it misses one form of exactly that drift. Comment the directive out in the unit file:

```
# RestartPreventExitStatus=78
```

The substring is still present, so the test stays green while systemd ignores the directive and the restart loop quietly returns. Matching against the whole file also accepts the directive misplaced under `[Unit]`. Parsing line by line and accepting only an active directive closes both.

**2. [`configs/alpamon-restart.timer`, `scripts/postinstall.sh:154-156`] An explicit `systemctl restart` bypasses `RestartPreventExitStatus`**

The repo already ships a restart timer, and postinstall enables and starts it:

```
configs/alpamon-restart.service: ExecStart=/bin/systemctl restart alpamon.service
configs/alpamon-restart.timer:   OnActiveSec=5min, WantedBy=multi-user.target
```

`RestartPreventExitStatus` only blocks systemd's *automatic* restart. A `systemctl restart` arriving from outside is not blocked and clears the failed state. So on an unclassifiable host the real sequence is:

1. postinstall starts the unit, it exits 78, systemd leaves it `failed` with no auto-restart
2. five minutes later the timer restarts it, it exits 78, `failed` again

`OnActiveSec` with `Persistent=false` and no `OnUnitActiveSec` fires once per timer activation, so this is **not a loop**, and the end state is still `failed` with the reason readable. The change achieves what it set out to. The timer is `WantedBy=multi-user.target` though, so one extra attempt also follows every boot.

No code change needed. What is inaccurate is the description's **"The unit now settles in `failed` after one attempt"**, on both the install and the boot path. Something like "the automatic restart loop is gone, and apart from the single attempt the restart timer contributes, the unit stays failed" would match what happens.

**3. [`pkg/plugin/plugin.go:178`] The plugin side only gets half the fix**

`pkg/plugin` is the public SDK that plugin repos (alpamon-dhcp-plugin and friends) import. It now returns 78, but the half that actually stops the loop, `RestartPreventExitStatus=78`, lives only in this repo's `configs/alpamon.service`. Plugin unit files do not have it, so a plugin still restart-loops on exit 78.

This is not a regression against the previous exit 1 loop, just unfinished. Stating the contract in the SDK doc comment keeps plugin authors from missing it:

```go
// Package plugin ...
//
// A plugin's service unit should set RestartPreventExitStatus to
// utils.ConfigErrorExitCode (78); Run returns it when startup hits a
// condition no restart can clear.
```

### Suggestions (5)

**4. [`cmd/alpamon/command/root.go:108-114`, `pkg/plugin/plugin.go:178-184`] Extracting the duplicated mapping closes the gap the description already names**

The same six lines appear in both callers. The description notes *"Not verified: the `errors.Is` to `os.Exit(78)` mapping in `root.go` and `plugin.go` has no test"*, diagnosed as untestable without injecting process exit. Pull the mapping out and it is a pure function that tests directly:

```go
// pkg/utils/exitcode.go
// StartupExitCode maps a startup failure to its exit status: a condition no
// restart can clear gets ConfigErrorExitCode, anything else stays restartable.
func StartupExitCode(err error) int {
	if errors.Is(err, ErrUnsupportedPlatform) {
		return ConfigErrorExitCode
	}
	return 1
}
```

```go
// root.go
if err := utils.InitPlatform(); err != nil {
	log.Error().Err(err).Msg("Failed to initialize platform.")
	os.Exit(utils.StartupExitCode(err))
}

// plugin.go
if err := utils.InitPlatform(); err != nil {
	log.Error().Err(err).Msg("Failed to initialize platform.")
	return utils.StartupExitCode(err)
}
```

**5. [`pkg/utils/platform.go:157-167` and `123-137`] The host lookup block sits in two places verbatim**

`InitPlatform` and `DetectRegistrationPlatform` each carry the same `if runtime.GOOS == "linux" { host.Info() }`. This PR split the pure half to restore the symmetry with `ResolveRegistrationPlatform`; folding the host half completes it. The two error messages differ, so the shared helper should return the raw id and leave wrapping to the callers:

```go
// The raw distribution id for this host; empty on non-Linux, where GOOS alone classifies.
func rawHostPlatform() (string, error) {
	if runtime.GOOS != "linux" {
		return "", nil
	}
	info, err := host.Info()
	if err != nil {
		return "", err
	}
	return info.Platform, nil
}
```

**6. [`pkg/utils/platform.go:154`] Sentinel error placement**

Go convention groups sentinel errors near the top of the file with the other package-level vars. This one sits at line 154, mid-file, away from the `PlatformLike` block at lines 13-17. Discoverability only, not behavior.

**7. [`cmd/alpamon/command/root.go:109`, `pkg/plugin/plugin.go:179`] The log level dropped from fatal to error**

Switching to `log.Error()` is correct, since zerolog's `log.Fatal()` calls `os.Exit(1)`, which is the thing being avoided. Worth knowing that anything alerting on `"level":"fatal"` in the journal loses this event. Also `Msg("Failed to initialize platform.")` restates what the error text already carries (`failed to retrieve platform information: ...`). Ignore if no such alerting exists.

**8. [`README.md:150`] "also" refers forward, not back**

The new paragraph opens with `The unit **also** sets RestartPreventExitStatus=78`, but the earlier unit setting it points back to, `The unit sets KillMode=process`, is the paragraph *after* it. Dropping "also" or swapping the two paragraphs puts the reading order right.

### Duplication, dead code, performance

- **Dead code**: none. `ErrUnsupportedPlatform` and `ConfigErrorExitCode` are both used by both callers, and `SetPlatformLike` / `SetPackageManager` / `SetPlatformID` are used throughout the `pkg/executor/handlers/{user,group,system}` tests.
- **Duplication**: items 4 (six lines of caller mapping, twice) and 5 (the host lookup block, twice) are all of it. The supported-distribution sentence appears in both `applyPlatform` and `ResolveRegistrationPlatform`, but the audiences differ (operator vs. registration CLI user) and so does the advice, so those should stay apart.
- **Performance**: no impact. `InitPlatform` runs once at startup, and `log.Fatal` to returning an error is one allocation apart. The added `os.ReadFile` is test-only. `applyPlatform` writing globals is safe: startup is single-goroutine, so no race.

### Highlights

- The `applyPlatform` split follows the existing `DetectRegistrationPlatform` / `ResolveRegistrationPlatform` pattern, so anyone who knows this file has nothing new to learn.
- `savePlatformGlobals` ties global restoration to `t.Cleanup`, which is what lets `TestInitPlatform_CIHost` run the real startup path without leaking into other tests.
- `TestApplyPlatform_UnsupportedWrapsSentinel` asserts that a failed classification leaves the globals alone. Breaking that would report a partially classified host to the server, and the contract is now pinned.
- `TestApplyPlatform_SetsGlobals` includes the SUSE row (`like=rhel`, `pkgManager=zypper`, `id=opensuse-tumbleweed`), which is what actually proves the three globals come from three different sources.
- The description names what it left out of scope and what the follow-ups are (`pkg/config/config.go:97,102,108,120`, `root.go:182`), and states its own unverified areas.
- Commits are split fix / test / docs.

---

**Verdict**: APPROVED

Behavior is correct and build, vet and tests are all clean, so nothing blocks the merge. Warning 1 (the hole in the guard test) and Warning 3 (the plugin SDK contract) are worth folding into this PR; Warning 2 needs only a wording fix in the description.

**Checklist**:
- [x] Code follows stack conventions (`gofmt`, `%w` wrapping, sentinel plus `errors.Is`)
- [x] Tests included and passing (`pkg/utils` ok, `applyPlatform` at 100%)
- [x] No security vulnerabilities (no change to credential or input handling)
- [x] Documentation updated (README systemd section, wording fix suggested)
- [ ] Exit 78 contract documented for the plugin SDK (Warning 3)

pkg/utils/exitcode_test.go
jisung-02: Fixed in 1227799f. The test now scans line by line and accepts the directive only as an active line under `[Service]`, so both shapes you named are rejected. Verified by mutating the unit file each way and watching it fail:

```
# case A: directive commented out
exitcode_test.go:29: configs/alpamon.service must set "RestartPreventExitStatus=78" under [Service]
# case B: directive moved under [Unit]
exitcode_test.go:29: configs/alpamon.service must set "RestartPreventExitStatus=78" under [Service]
```

configs/alpamon.service
jisung-02: You are right, and the description was wrong about it. Rewritten: the automatic restart is gone, but `configs/alpamon-restart.timer` still sends one explicit `systemctl restart` five minutes after activation and once per boot, which `RestartPreventExitStatus` does not block — so the unit takes that one extra attempt, exits 78 again, and settles back into `failed`. Also noted that `OnActiveSec` without `OnUnitActiveSec` makes it a single retry rather than a loop. No code change.

pkg/plugin/plugin.go
jisung-02: Fixed in dd0e90ad. The package doc now states the other half of the contract:

```
Run exits with utils.ConfigErrorExitCode (78) when startup hits a condition
no restart can clear, such as a host the platform classifier cannot map. A
plugin's service unit should carry RestartPreventExitStatus=78 alongside its
Restart= setting, the way configs/alpamon.service does; without it systemd
relaunches straight back into the same failure.
```

cmd/alpamon/command/root.go
jisung-02: Fixed in c04d6692. `utils.StartupExitCode(err)` now carries the mapping and both callers are one line each. `TestStartupExitCode` pins both arms — a wrapped sentinel to 78, a host lookup failure to a restartable 1 — so the gap the description named is closed.

pkg/utils/platform.go
jisung-02: Fixed in 8bc5453d. `ErrUnsupportedPlatform` now sits beside the `PlatformLike` / `PackageManager` / `PlatformID` block at the top of the file.

pkg/utils/platform.go
jisung-02: Fixed in 8bc5453d, taking your `rawHostPlatform` shape as written. Both `InitPlatform` and `DetectRegistrationPlatform` call it and keep their own wrapping, since one addresses an operator reading the journal and the other a user running `register`.

README.md
jisung-02: Fixed in 6f4cb238. Swapped the two paragraphs, so the `KillMode=process` one now comes first and "also" points back at it.

geunwoonoh (approved):

PR #416, opened 2026-08-14
build(go): raise the go directive to 1.25.13

jisung-02:
## Summary

`govulncheck` found four standard-library vulnerabilities reachable from our
code on go1.25.12. All four are fixed in go1.25.13, so raising the `go`
directive is the entire remedy.

**No Go functionality changed.** Not a line of application code was touched:
the fixes live in the toolchain that compiles us, and the only Go source
change in this branch is the `go` directive in `go.mod`.

## Vulnerabilities cleared

| ID | Package | Issue |
| --- | --- | --- |
| [GO-2026-6218](https://pkg.go.dev/vuln/GO-2026-6218) | `net/url` | quadratic complexity in `resolvePath` |
| [GO-2026-6090](https://pkg.go.dev/vuln/GO-2026-6090) | `crypto/tls` | unbounded post-handshake messages |
| [GO-2026-5972](https://pkg.go.dev/vuln/GO-2026-5972) | `encoding/asn1` | no maximum recursion depth |
| [GO-2026-5026](https://pkg.go.dev/vuln/GO-2026-5026) | `net/http` (idna) | ASCII-only Punycode labels not rejected |

Reachable call sites included `pkg/runner/ftp.go`, `pkg/runner/client.go`,
`pkg/migrate/migrate.go`, `pkg/scheduler/session.go`, and
`pkg/executor/handlers/file/file_io_unix.go`.

## Release notes

[Go 1.25.13](https://go.dev/doc/devel/release#go1.25.13), released
2026-08-13, carries security fixes to the `go` command, `crypto/tls`,
`encoding/asn1`, `encoding/xml`, `html/template`, `net/http`, and `net/url`,
plus bug fixes to the compiler, the runtime, `crypto/tls`, and `os`. Full
change list: [Go1.25.13 milestone](https://github.com/golang/go/issues?q=milestone%3AGo1.25.13+label%3ACherryPickApproved).

## CI change: one source of truth for the toolchain

Every `setup-go` call resolved its version from a `GO_VERSION` workflow env,
so the toolchain had two sources of truth and a bump had to land in four
files at once. `setup-go` pins `GOTOOLCHAIN=local`, which makes that drift
fatal instead of self-healing: the moment the resolved version lags behind
the `go` directive, every command that loads packages dies before it runs.

`alpacon-cli` hit exactly this on its own 1.25.13 bump. Its `govulncheck` job
resolved `1.25.x` to a cached 1.25.12 while `go.mod` already required
1.25.13, and the scan failed before it started:

```
govulncheck: loading packages: err: exit status 1: stderr: go: go.mod requires go >= 1.25.13 (running go 1.25.12; GOTOOLCHAIN=local)
```

Alpamon was one drifted bump away from the same failure, so this branch
removes the second source of truth rather than waiting for it to fire: every
`setup-go` now reads `go-version-file: go.mod`, and the `GO_VERSION` env is
gone from all three workflows. Alpamon installs `govulncheck` with `go
install` rather than through `golang/govulncheck-action`, so it does not need
the extra unwrapping that `alpacon-cli` required.

Dockerfiles stay on `FROM golang:1.25`—that tag already tracks the newest
patch.

## Review follow-ups

Four commits added after review, all on the same theme of leaving exactly one
place where a version or an assumption is written down.

| Commit | Change |
| --- | --- |
| `71cffffa` | `govulncheck` is installed at `@v1.7.0` instead of `@latest`, closing the last `GOTOOLCHAIN=local` drift path in CI |
| `2469eae2` | `README.md:185` no longer names a patch version; it points at the `go` directive in `go.mod` |
| `a01225b8` | `setup-go-generate` records in its `description` that it assumes a workspace-root checkout |
| `f04d2578` | `.goreleaser.yaml` no longer runs `go mod tidy` at release time |

The pin costs no coverage: `govulncheck` fetches its vulnerability data from
`vuln.go.dev` on every run, so only the scanner engine is held back, never the
CVE feed. v1.7.0 declares `go 1.25.0`, below the 1.25.13 in `go.mod`. The
version has to be bumped by hand, since Dependabot does not track versions
inside a `run` step.

Dropping the release-time `go mod tidy` is safe because `release.yml` calls
`build-and-test.yml` on the tag and `goreleaser` runs with
`needs: [build-and-test]`, so `verify-tidy` has already failed on any `go.mod`
or `go.sum` diff for the exact commit being released. The `go generate` hook
stays: the ent output is gitignored, so the release clone has none of it.

## Verification

Run locally on go1.25.13:

| Check | Result |
| --- | --- |
| `go build ./...` | passes |
| `go vet ./...` | passes |
| `go test ./... -p 1` | every package `ok`, no `FAIL` or `panic` |
| `go mod tidy` | no `go.sum` change |
| `gofmt -l` (excluding generated ent code) | clean |
| `alpamon --help` | runs |
| `govulncheck ./...` | `No vulnerabilities found` |

### DIFF
--- a/.github/actions/setup-go-generate/action.yml
+++ b/.github/actions/setup-go-generate/action.yml
@@ -3,17 +3,17 @@ description: >-
   Install the Go toolchain and generate the gitignored ent package. The tree
   imports ent subpackages that aren't checked in, so codegen must run before
   any command that loads packages (build, test, tidy, govulncheck, lint).
-inputs:
-  go-version:
-    description: Go version to install
-    required: true
 runs:
   using: composite
   steps:
+    # go.mod is the only source of truth for the toolchain. setup-go pins
+    # GOTOOLCHAIN=local, so a version resolved from anywhere else (a workflow
+    # env, a `1.25.x` wildcard hitting a stale runner cache) fails every
+    # package-loading command the moment it lags behind the go directive.
     - name: Set up Go
       uses: actions/setup-go@v6
       with:
-        go-version: ${{ inputs.go-version }}
+        go-version-file: go.mod
 
     - name: Generate go code
       run: go generate ./pkg/db/ent
--- a/.github/actions/setup-go-generate/action.yml
+++ b/.github/actions/setup-go-generate/action.yml
@@ -3,6 +3,8 @@ description: >-
   Install the Go toolchain and generate the gitignored ent package. The tree
   imports ent subpackages that aren't checked in, so codegen must run before
   any command that loads packages (build, test, tidy, govulncheck, lint).
+  Assumes the repository is checked out at the workspace root: both the go.mod
+  the toolchain is read from and the generate path resolve from there.
 runs:
   using: composite
   steps:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,9 +13,6 @@ on:
 permissions:
   contents: read
 
-env:
-  GO_VERSION: '1.25.12'
-
 jobs:
   linter:
     name: golangci-lint
@@ -29,8 +26,6 @@ jobs:
 
       - name: Set up Go and generate ent code
         uses: ./.github/actions/setup-go-generate
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - name: GolangCI-Lint
         uses: golangci/golangci-lint-action@v9
@@ -48,8 +43,6 @@ jobs:
 
       - name: Set up Go and generate ent code
         uses: ./.github/actions/setup-go-generate
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - name: Install govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -45,7 +45,7 @@ jobs:
         uses: ./.github/actions/setup-go-generate
 
       - name: Install govulncheck
-        run: go install golang.org/x/vuln/cmd/govulncheck@latest
+        run: go install golang.org/x/vuln/cmd/govulncheck@v1.7.0
 
       # Report findings without gating merges: a newly published CVE in a
       # transitive dep or stdlib would otherwise turn unrelated PRs red.
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -13,9 +13,6 @@ on:
 permissions:
   contents: read
 
-env:
-  GO_VERSION: '1.25.12'
-
 jobs:
   verify-tidy:
     name: Verify go.mod and go.sum are tidy
@@ -26,8 +23,6 @@ jobs:
 
       - name: Set up Go and generate ent code
         uses: ./.github/actions/setup-go-generate
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - name: Verify go.mod and go.sum are tidy
         run: |
@@ -151,8 +146,6 @@ jobs:
 
       - name: Set up Go and generate ent code
         uses: ./.github/actions/setup-go-generate
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - name: Build
         run: go build -v .
@@ -170,8 +163,6 @@ jobs:
 
       - name: Set up Go and generate ent code
         uses: ./.github/actions/setup-go-generate
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - name: Build
         run: go build -v .
@@ -258,8 +249,6 @@ jobs:
 
       - name: Set up Go and generate ent code
         uses: ./.github/actions/setup-go-generate
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - name: Build
         run: go build -v .
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,9 +8,6 @@ on:
 permissions:
   contents: write
 
-env:
-  GO_VERSION: '1.25.12'
-
 jobs:
   build-and-test:
     uses: ./.github/workflows/build-and-test.yml
@@ -58,7 +55,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v6
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: go.mod
 
       # GoReleaser needs write access to two repositories with a single
       # token: this repo (to attach release assets) and homebrew-alpacon
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -8,7 +8,6 @@ release:
 before:
   hooks:
     - go generate ./pkg/db/ent
-    - go mod tidy
 
 builds:
   - id: alpamon-linux
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ go build -o alpamon ./cmd/alpamon         # native
 GOOS=windows GOARCH=amd64 go build ./cmd/alpamon   # Windows cross-compile
 ```
 
-Go 1.25.12+ is required. `GOPATH/bin` should be on `PATH`. The generated Ent code is gitignored, so run `go generate ./pkg/db/ent` (see below) before the first build.
+The Go toolchain the `go` directive in `go.mod` names is required. `GOPATH/bin` should be on `PATH`. The generated Ent code is gitignored, so run `go generate ./pkg/db/ent` (see below) before the first build.
 
 ### Generate Ent schema code
 
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/alpacax/alpamon/v2
 
-go 1.25.12
+go 1.25.13
 
 require (
 	entgo.io/ent v0.14.5


.github/workflows/lint.yml
```diff
@@ -48,8 +43,6 @@ jobs:
 
       - name: Set up Go and generate ent code
         uses: ./.github/actions/setup-go-generate
-        with:
-          go-version: ${{ env.GO_VERSION }}
 
       - name: Install govulncheck
         run: go install golang.org/x/vuln/cmd/govulncheck@latest
```
geunwoonoh: Suggestion: this step is the last remaining GOTOOLCHAIN drift path, the same failure class this PR removes everywhere else.

Under `GOTOOLCHAIN=local`, `go install ...@latest` fails the moment govulncheck's own `go.mod` requires a Go newer than the 1.25.13 installed on the runner. Only the `govulncheck ./...` step below is `continue-on-error: true` — this install step is not, so on that day every unrelated PR goes red for a reason that has nothing to do with it.

Pinning the version, or extending `continue-on-error` to the install step, closes it. Outside this diff, but it is the same theme, so worth noting here.

.github/actions/setup-go-generate/action.yml
```diff
@@ -3,17 +3,17 @@ description: >-
...
   steps:
+    # go.mod is the only source of truth for the toolchain. setup-go pins
+    # GOTOOLCHAIN=local, so a version resolved from anywhere else (a workflow
+    # env, a `1.25.x` wildcard hitting a stale runner cache) fails every
+    # package-loading command the moment it lags behind the go directive.
     - name: Set up Go
       uses: actions/setup-go@v6
       with:
-        go-version: ${{ inputs.go-version }}
+        go-version-file: go.mod
```
geunwoonoh, line 18: Nit: `go.mod` here resolves relative to the workspace, not to the action's own directory.

Every current caller checks out at the workspace root, so this works today. If a future job ever passes `path:` to `actions/checkout`, the action dies with a confusing "file not found". A line in the action `description` noting the root-checkout assumption would save the next person the debugging round.

geunwoonoh (commented):
## Review summary

This is mergeable. Nothing blocks it. The one thing that should have ridden along in this commit is `README.md:185`, which still says `Go 1.25.12+ is required`. The three previous toolchain bumps (1.25.7 -> 1.25.10 -> 1.25.12) each updated that line in the same commit; this one missed it.

**Scope**: raises the `go` directive to 1.25.13 to clear four stdlib vulnerabilities, and removes the `GO_VERSION` env from three workflows so the toolchain has exactly one source of truth. Not a line of Go application code changed.

---

### Critical issues (0)

None.

### Warnings (1)

**1. `README.md:185` — the version line becomes wrong the moment this merges**

```
Go 1.25.12+ is required. `GOPATH/bin` should be on `PATH`. ...
```

From `git log -L 185,185:README.md`:

| Commit | Contents |
| --- | --- |
| `1e90317` | go.mod 1.25.10 + README 1.25.10 |
| `f745ff8` | go.mod 1.25.12 + README 1.25.12 |
| `46fe86e` (this PR) | go.mod 1.25.13, **README still 1.25.12** |

The premise of this PR is removing the second source of truth for the version, and this line is the last version literal left in the repo. Either bump it to 1.25.13, or drop the number entirely and point at the `go` directive in `go.mod` — the latter fits the PR's own thesis better.

### Suggestions (3)

Inline on the two that land in the diff. The third:

**`.goreleaser.yaml:11` — the release-time `go mod tidy` is redundant**

`before.hooks` re-runs `go mod tidy` on every release build, but the `verify-tidy` job already guarantees tidiness on every PR and push. It is duplicated work, and it leaves room for a release build to re-resolve dependencies and mutate `go.mod`/`go.sum` mid-release. Release builds are better off hermetic. (Outside this diff.)

### On Go conventions, duplication, and performance

**Go conventions.** No Go code changed, so there is no room for a convention violation. Pinning a patch version in the `go` directive matches Go 1.21+ semantics (it is a minimum-toolchain requirement) and matches what this repo has done since 1.25.10. The alternative — `go 1.25` plus `toolchain go1.25.13` — would not enforce the floor for anyone building from source, so the hard pin is the right call for a security bump. `gh search code` across the alpacax org finds no repo importing `github.com/alpacax/alpamon/v2`, so raising the floor strands no downstream consumer.

**Duplicated or dead code.** The removal of the `go-version` input is cleaned up completely: no `GO_VERSION` and no caller passing that input survives anywhere in the repo (verified by grep). `release.yml` keeping its own `actions/setup-go` instead of the shared composite action is not duplication but a correct choice — `.goreleaser.yaml`'s `before.hooks` already runs `go generate ./pkg/db/ent`, so reusing the composite action would run codegen twice.

**Performance.** Zero runtime impact. If anything it improves: GO-2026-6218, one of the four cleared here, is quadratic complexity in `net/url` `resolvePath`, which is a performance fix as much as a security one. CI time is unchanged, since an exact patch version was pinned before and after, so setup-go downloads the toolchain either way.

### Highlights

- **The PR body's claim is backed by the CI log.** The govulncheck job prints `Setup go version spec 1.25.13`, and the step env shows `GOTOOLCHAIN: local`. The failure condition that hit alpacon-cli is real, and this change makes it structurally unreachable.
- **Leaving the Dockerfiles on `FROM golang:1.25` is verified correct.** Per the Docker Hub API, `golang:1.25` and `golang:1.25.13` currently share the same linux/amd64 digest, `sha256:6be3daa1…`.
- **The comment on the non-gating govulncheck job** records why the step is soft-fail and which failures still fail the job, so the reasoning survives in the file.

### Local verification (go1.25.13 darwin/arm64)

| Check | Result |
| --- | --- |
| `go generate ./pkg/db/ent` -> `go build ./...` | passes |
| `go vet ./...` | passes |
| `go mod tidy` | no `go.mod`/`go.sum` diff |
| `gofmt -l` (excluding generated code) | clean |
| `govulncheck ./...` | `Your code is affected by 0 vulnerabilities.` |

A green govulncheck check is not evidence on its own, because the step is `continue-on-error: true`, so I read the actual job log. The two vulnerabilities left are unreachable and there is nothing to do about either: GO-2026-5942 (`net`) is fixed only in `go1.26.6` with no patch on the 1.25 line yet, and GO-2026-5932 (`golang.org/x/crypto@v0.52.0`) reports `Fixed in: N/A`. Neither has a call path.

---

**Verdict**: COMMENTED — 0 critical, 1 warning, 3 suggestions. Fix the one README line and this is an approve.

**Checklist**
- [x] Follows stack conventions
- [x] Tests included and passing (no Go code changed; all 25 CI checks pass)
- [x] No security vulnerabilities (0 reachable)
- [ ] Documentation updated (`README.md:185` not updated)

.github/workflows/lint.yml
jisung-02: Fixed in `71cffffa` — pinned to `@v1.7.0`.

I took the pin over extending `continue-on-error` to the install step: with the install soft-failed, `govulncheck ./...` would then hit "command not found" and soft-fail too, so the job would report green while scanning nothing. The pin fails loudly at bump time instead, which is the behaviour I want from a security job.

Worth noting the pin costs no coverage, since `govulncheck` fetches its vulnerability data from `vuln.go.dev` on every run — only the scanner engine is held back, never the CVE feed. `alpacon-event-server` already pins the same tool this way. v1.7.0 declares `go 1.25.0`, well below the 1.25.13 in `go.mod`.

I also considered `GOTOOLCHAIN=auto` for this job, as `alpacon-cli` does, and dropped it: it puts a second toolchain source back in right after this branch removed one. The tradeoff is that Dependabot cannot see a version inside a `run` step, so this pin needs a manual bump — recorded as a `Directive` trailer on the commit.

.github/actions/setup-go-generate/action.yml
jisung-02: Fixed in `a01225b8` — the assumption is now in the action `description`.

I widened it slightly from your wording: `go generate ./pkg/db/ent` in the second step resolves against the workspace for the same reason, so the note covers both rather than the `go-version-file` line alone.

Confirmed the premise before writing it: all six callers across `lint.yml` and `build-and-test.yml` use `actions/checkout` with no `path:`, so nothing is broken today. The commit carries a `Directive` trailer saying that a caller genuinely needing a non-root checkout should get an explicit path input on the action rather than have the note quietly stop being true.

geunwoonoh (approved):
## Re-review: `8760429` -> `f04d257`

Approving. All four points are addressed, one commit each, and each commit's stated reasoning holds up against the code.

| Finding | Commit | Resolution | How I checked it |
| --- | --- | --- | --- |
| Warning: `README.md:185` version literal | `2469eae` | Number dropped, points at the `go` directive | No patch-level Go literal left anywhere in the repo outside `go.mod` (grep) |
| `govulncheck@latest` | `71cffff` | Pinned to `@v1.7.0` | Tag exists on the module proxy; its own `go.mod` declares `go 1.25.0`, well under 1.25.13, so it installs under `GOTOOLCHAIN=local`. CI log shows `@v1.7.0` actually running |
| `go-version-file` path resolution | `a01225b` | Assumption recorded in the action `description` | All six callers use `actions/checkout` with no `path:` |
| Release-time `go mod tidy` | `f04d257` | Hook removed | Detail below |

### On the goreleaser change

You noted you could not run `goreleaser check` locally, so I dug into this one separately. **The change is safe** — but the reasoning needs one correction, because the argument as written does not actually cover the failure mode.

`pkg/db/ent/generate.go:3` runs `go run -mod=mod entc.go`, and `-mod=mod` lets that command write. It does: `go generate` on its own adds 18 lines to `go.sum` — hashes for entc's codegen-only dependencies (`ariga.io/atlas`, `hashicorp/hcl/v2`, `zclconf/go-cty`, and friends). Reproduced on a clean checkout of `f04d257`:

```
RESULT: go generate mutates go.sum
 go.sum | 18 ++++++++++++++++++
```

`verify-tidy` hides this. That job runs generate -> tidy -> diff, so tidy strips the 18 lines back out before the diff runs and the job goes green. So "verify-tidy already proved tidiness on the exact commit" is true about the *committed* files, but it does not establish that the release tree stays clean once the hook is gone. The old `tidy` hook was not only duplicated work — it was also undoing the mutation the `go generate` hook right above it had just made. Without it, the release build now carries a dirty `go.sum` from the generate hook onward.

That turns out not to matter, for reasons independent of `verify-tidy`. From goreleaser v2's `BuildPipeline`:

```go
git.Pipe{},      // get and validate git repo state   <- dirty check
...
before.Pipe{},   // run global hooks before build     <- hooks run here
```

The dirty-state validation completes before the hooks run, so a hook that dirties the tree cannot trip it. Downstream, the only pipe that touches module state is `gomod.Pipe`, which just runs `go list -m` to read the module path; `gomod.CheckGoModPipe` and `gomod.ProxyPipe` skip because `gomod.proxy` is unset in `.goreleaser.yaml`. And a `go.sum` with surplus entries is never an error in Go — only missing or mismatched hashes fail verification. Build inputs and artifacts are unchanged.

Worth keeping in mind if the hook list is ever reordered or a pipe that validates the worktree is added.

### Verification

All 25 checks pass on `f04d257`. Since the govulncheck step is `continue-on-error: true`, a green check is not evidence on its own, so I read the job log: `go install golang.org/x/vuln/cmd/govulncheck@v1.7.0` ran, and the scan reports `Your code is affected by 0 vulnerabilities.` Locally on go1.25.13: `go build ./...`, `go vet ./...`, `gofmt -l` (excluding generated code), and `go mod tidy` (no diff) are all clean, and `govulncheck ./...` agrees. The two remaining vulnerabilities are still unreachable with no patch available on the 1.25 line: GO-2026-5942 (`net`, fixed only in go1.26.6) and GO-2026-5932 (`golang.org/x/crypto@v0.52.0`, `Fixed in: N/A`).

---

**Verdict**: APPROVED — 0 critical, 0 warnings.

**Checklist**
- [x] Follows stack conventions
- [x] Tests included and passing (no Go code changed; 25/25 CI checks pass)
- [x] No security vulnerabilities (0 reachable)
- [x] Documentation updated (`README.md` now points at the `go` directive instead of carrying a literal)

PR #417, opened 2026-08-14
ci: move the Rocky Linux containers to rockylinux/rockylinux

### LINKED ISSUE #376: CI: consider moving the rockylinux container images to rockylinux/rockylinux
## Background

The Docker Hub images our CI and release workflows use (`rockylinux:8` and `rockylinux:9`, from the `library/rockylinux` repository) appear not to have received updates for a while. The most recent tag update listed there is 2024-05-30, and the highest versions available are 9.3 and 8.9, with no tags for 9.4 through 9.8.

By comparison, `rockylinux/rockylinux` shows tag updates as recent as 2026-07-12 and carries Rocky 10 tags. Checking inside the containers shows the difference:

| Image | rocky-release |
|---|---|
| `rockylinux:9` | 9.3-1.1.el9 |
| `rockylinux/rockylinux:9` | 9.8-1.1.el9 |

Since `$releasever` still resolves to 9.8, packages are pulled at current versions and builds pass, so nothing is broken by this today. That said, the base layer itself sits roughly 26 months behind, so it may not include fixes that have landed since, and our distro compatibility checks run against an environment somewhat older than what users are likely to be on. Given that alpamon is installed on user servers, it seems worth keeping that environment reasonably close to current.

## Suggested changes

| File | Line | Current | Proposed |
|---|---|---|---|
| `.github/workflows/build-and-test.yml` | 68 | `rockylinux:8` | `rockylinux/rockylinux:8` |
| `.github/workflows/build-and-test.yml` | 77 | `rockylinux:9` | `rockylinux/rockylinux:9` |
| `.github/workflows/build-and-test.yml` | 109 | `rockylinux:9` | `rockylinux/rockylinux:9` |
| `.github/workflows/release.yml` | 190 | `rockylinux:9` | `rockylinux/rockylinux:9` |

These correspond to the `rocky-8`, `rocky-9`, and `rocky-9-arm64` jobs, and the rpm install check in the release workflow.

## Points to verify

Swapping the image moves the base from 8.9 to 8.10 and from 9.3 to 9.8, so a little validation would help before committing to it:

- whether the `rocky-8`, `rocky-9`, and `rocky-9-arm64` jobs still build and test cleanly on the newer base
- whether the rpm install check in the release workflow still passes
- whether `rockylinux/rockylinux` maintains stable `8` and `9` tags. The tag listing referenced above covered only the 15 most recently updated tags, so the presence of 9.x minor tags has not been confirmed

## Notes

The 9.8 release version for `rockylinux/rockylinux:9` was confirmed directly with `rpm -q rocky-release` inside the container. The maintenance comparison rests on the tag update timestamps shown on Docker Hub; no official statement from the Rocky Linux project about the status of either repository was checked, so it would be worth confirming that before treating this as settled.

Related URL: https://hub.docker.com/r/rockylinux/rockylinux

jisung-02:
## Background

The distro compatibility jobs and the release rpm install check pull `rockylinux:8` and `rockylinux:9` from Docker Hub's `library/rockylinux`. That repository's most recent tag update is 2024-05-30, and its highest versions are 8.9 and 9.3. `rockylinux/rockylinux` publishes 8.10 and 9.8, with its `9` tag updated 2026-06-12.

Nothing is broken by the old images today. `$releasever` still resolves to the current minor, so packages install at current versions and the jobs pass. The concern is that the base layer itself is roughly 26 months behind, so the environment the compatibility checks run against is older than what alpamon is actually installed on.

## Changes

Four `container:` values swap from `rockylinux:N` to `rockylinux/rockylinux:N`:

| File | Line | Job |
|---|---|---|
| `.github/workflows/build-and-test.yml` | 72 | `rocky-8` |
| `.github/workflows/build-and-test.yml` | 81 | `rocky-9` |
| `.github/workflows/build-and-test.yml` | 118 | `rocky-9-arm64` |
| `.github/workflows/release.yml` | 191 | rpm install check |

No other change. The base moves from 8.9 to 8.10 and from 9.3 to 9.8.

## Testing

Both tags were confirmed to exist and to publish the architectures the matrix needs, via the Docker Hub API:

```
rockylinux/rockylinux:8 -> last_updated 2024-06-05, [arm64, amd64]
rockylinux/rockylinux:9 -> last_updated 2026-06-12, [amd64, arm64, ppc64le, s390x]
```

The `rocky-8`, `rocky-9` and `rocky-9-arm64` jobs on this PR are what verify the new base builds and tests cleanly.

Not verified: the rpm install check in `release.yml` runs only on a release, so its new base stays unverified until the next tag is cut. No official Rocky Linux statement about the maintenance status of either repository was checked — the comparison rests on Docker Hub tag timestamps and on `rpm -q rocky-release` inside the containers.

Closes #376

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -188,7 +188,7 @@ jobs:
             arch: amd64
             install_cmd: apt-get install -y -o Acquire::Retries=3
           - pkg_type: rpm
-            container: rockylinux:9
+            container: rockylinux/rockylinux:9
             arch: amd64
             install_cmd: dnf install -y
     container: ${{ matrix.container }}
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -69,7 +69,7 @@ jobs:
             runner: ubuntu-latest
           # RHEL/CentOS yum-based (amd64)
           - distro: rocky-8
-            container: rockylinux:8
+            container: rockylinux/rockylinux:8
             pkg_manager: yum
             runner: ubuntu-latest
           - distro: rhel-8
@@ -78,7 +78,7 @@ jobs:
             runner: ubuntu-latest
           # RHEL/CentOS dnf-based (amd64)
           - distro: rocky-9
-            container: rockylinux:9
+            container: rockylinux/rockylinux:9
             pkg_manager: dnf
             runner: ubuntu-latest
           - distro: almalinux-9
@@ -115,7 +115,7 @@ jobs:
             pkg_manager: apt-get
             runner: ubuntu-24.04-arm
           - distro: rocky-9-arm64
-            container: rockylinux:9
+            container: rockylinux/rockylinux:9
             pkg_manager: dnf
             runner: ubuntu-24.04-arm
           - distro: amazonlinux-2023-arm64


.github/workflows/release.yml
```diff
@@ -188,7 +188,7 @@ jobs:
             arch: amd64
             install_cmd: apt-get install -y -o Acquire::Retries=3
           - pkg_type: rpm
-            container: rockylinux:9
+            container: rockylinux/rockylinux:9
```
geunwoonoh, line 191: Suggestion: `rockylinux/rockylinux:9` is a mutable major-version tag, so CI runs are not fully reproducible. Pinning to `:9.8` or a digest is an option if reproducibility matters. Since these jobs are meant to validate against what Rocky Linux currently ships, the mutable tag matches the intent and the previous setup worked the same way, so no change is needed. Also note this rpm install check only runs on a release, so the new base stays unverified here until the next tag is cut; risk is low since the same image already passed in `build-and-test.yml`.

geunwoonoh (approved):
## Review

**Stack**: common (the diff touches only two workflow YAML files, so no Go-specific rules apply)
**Changes**: +4 -4 in 2 files

### Summary

Moves the container images used by the three distro-compat CI jobs (`rocky-8`, `rocky-9`, `rocky-9-arm64`) and the release workflow's rpm install check from Docker Hub's `library/rockylinux` (no tag updates since 2024-05-30, highest versions 8.9/9.3) to `rockylinux/rockylinux` (8.10/9.8). Four image names change; nothing else does.

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (1)

1. **[.github/workflows/release.yml:191]** Major-version tags like `rockylinux/rockylinux:9` are mutable, so CI runs are not fully reproducible. Pinning to `:9.8` or a digest is an option if reproducibility matters. That said, these jobs exist to validate compatibility against what Rocky Linux currently ships, so a mutable tag tracking the latest minor matches the intent, and the previous setup worked the same way. No change needed; noting it for the record.

### Additional checks (Go conventions, duplication, performance)

- **Go conventions**: no Go code changes in this PR, so there is nothing to review on that front; `golangci-lint` and `govulncheck` both pass.
- **Duplicate/unused code**: `rockylinux/rockylinux:9` appears three times, but GitHub Actions has no YAML anchors and each matrix include lists its own `container`, matching the file's existing pattern. Verified by grep that no old `rockylinux:8`/`rockylinux:9` references remain anywhere in the repo; `Dockerfiles/` has no Rocky variant, and `"rocky"` in `pkg/utils/platform.go:35` is runtime distro detection, unrelated to image names. Also confirmed `matrix.container` is only interpolated into `container.image` (`build-and-test.yml:42`, `release.yml:194`), so no string comparison depends on the old image names.
- **Performance**: no runtime code changes, so the only surface is CI time. The three rocky jobs ran in 1m56s–2m19s on this PR, in line with the other distros. A newer base (9.3 → 9.8) means `dnf install` fetches a smaller update delta, so if anything this is marginally faster. No regression.

### Highlights

- The diff is exactly the four lines the issue asks for, with no unrelated cleanup.
- The PR body separates what was verified (tags and architectures via the Docker Hub API) from what was not (the `release.yml` path stays unverified until the next tag is cut).
- The three rocky jobs passed on this PR with the new base; all 25 checks are green.

### Verdict: APPROVED

**Checklist**:
- [x] Code follows stack conventions (keeps the existing matrix pattern in both workflows)
- [x] Tests included and passing (the three rocky jobs verify this change directly)
- [x] No security vulnerabilities (moves to the official Rocky Linux org image with current patches)
- [x] Documentation updated if needed (none required; the Rocky mentions in README describe supported platforms, unrelated to CI images)

One follow-up: the rpm install check at `release.yml:191` only runs on a release, so confirm that job passes when the next tag is cut. Risk is low since the same image already passed in `build-and-test.yml`.

Issue #370, opened 2026-07-22
test: gradually unify assertion style on testify across the test suite


jisung-02:
## Background

Our test suite currently mixes two assertion styles. Around 27 of the ~99 `_test.go` files already use testify (`assert`/`require`), while the rest hand-write checks like `if got != want { t.Errorf(...) }` and `if err != nil { t.Fatalf(...) }`. Both styles work fine, so this is not about correctness—it is just that the inconsistency makes the suite a little harder to read and review, and hand-written assertions occasionally lead to small slips in error messages or comparison direction.

As a follow-up to #369 (the Go 1.25 idiom refresh), it might be worth gradually aligning the value-comparison-heavy files on testify, so the style is more predictable across the codebase.

## Candidates that would benefit most

Files where the same assertion patterns repeat a lot, so the readability gain is clearest:

- `pkg/executor/handlers/user/user_test.go`
- `pkg/executor/handlers/shell/shell_test.go`
- `pkg/executor/handlers/system/system_test.go`
- `pkg/updater/updater_test.go`
- `pkg/runner/auth_manager_tracker_test.go`
- `pkg/signing/keymanager_test.go`
- `pkg/utils/archive_test.go`
- `pkg/executor/handlers/file/multipart_stream_test.go`
- `pkg/utils/fs_test.go`
- `pkg/migrate/migrate_test.go`

## Partial candidates

For these, only the plain value-checking tests would move to testify; the concurrency/timing tests in the same file are better left as they are:

- `pkg/runner/terminal_manager_test.go`
- `pkg/executor/integration_test.go`
- `pkg/executor/chunk_writer_test.go`
- `pkg/executor/process_tree_*_test.go`
- `pkg/executor/regression_test.go`

## Probably out of scope

Files whose core is concurrency, timing, or network setup, where swapping assertions buys little:

- `pkg/logsink/writer_test.go` (UDS listener + WaitGroup setup)
- `internal/pool/pool_leak_test.go` (`runtime.NumGoroutine()` polling)
- `pkg/runner/pty_recovery_test.go`
- benchmark files and very small files with only a handful of assertions

## Suggested approach

- Behavior stays the same—this is only an assertion-style swap (`require.NoError`, `assert.Equal`, `assert.Contains`, `assert.Nil`, etc.)
- `require` where the test must stop on failure, `assert` where it can keep checking
- Split into small PRs (per file or small groups) to keep reviews light; no rush to do it all at once

[...]

PR #422, opened 2026-08-14
test(signing): unify assertion style and consolidate test fixtures

jisung-02:
Second PR for #370. Scope is `pkg/signing`—`keymanager_test.go` and `verifier_test.go`, which are the only test files in the package, so it reads as testify-only after this.

The first PR (#421) settled how `require` and `assert` get chosen. Here the original code had already made that choice: every `t.Fatal*` stops the test and every `t.Error*` does not. The conversion maps `t.Fatal*` to `require` and `t.Error*` to `assert`, one for one, so what each check does on failure is preserved and the rule itself is something a reviewer can check in a line.

## What changed

Eight commits, one idea each.

`test(signing): unify assertion style on testify`—all 54 hand-written checks become testify. No test was added, removed, or renamed by this commit.

`test(signing): pin each error check to the error it expects`—twelve assertions only checked that *an* error came back. `TestVerifyCommand_NilPublicKey` was the weakest: any `VerifyCommand` failure satisfied it, including one that never reached the key size check. The two signature paths now assert `ErrSignatureMismatch`, the one error worth retrying after a key refresh and therefore the one that has to be told apart from the rest; the others assert the message the code returns.

`test(signing): build keys and key servers through checked helpers`—sixteen call sites dropped the `ed25519.GenerateKey` error and used the key on the next line, and ten tests built their own `httptest` handler, two of them byte for byte identical, every one dropping the error from `json.NewEncoder.Encode`. `newTestKey` checks the generation once. `keyServer` covers the seven handlers that answer every request the same way; the three whose response depends on the request keep a handler but write through `writeKeyResponse`.

`test(signing): table-drive the verifier cases and sign through one helper`—five `VerifyCommand` error tests differed only in the signature they fed in and the message they expected, and three `BuildCanonicalPayload` tests only in the command and the JSON they compared against. `testCommand` builds the repeated command literal and `signedCommand` absorbs the build-sign-encode block the three signature tests each repeated.

`test(signing): order tests the way the code is ordered`—declaration moves only, so each test file follows the order its source file declares.

`test(signing): publish the recorded query from the handler`—two variables were written on the server's goroutine and read on the test's with nothing between them. A single `atomic.Pointer[url.Values]` removes the question rather than answering it, and retires the `authEnvPresent = true // reset` line with it.

`test(signing): keep the shared fixtures in their own file`—`newTestKey` is used by both test files but sat above the keymanager-only HTTP fixtures, where it reads as keymanager's. With every fixture in `fixtures_test.go`, each test file matches one source file.

`test(signing): drop the comments the assertions already make`—21 comments deleted, five compressed. One asserted wider than its code: the comment claiming the payload must contain raw UTF-8 bytes sat above two checks that only assert the escaped forms are absent.

```
pkg/signing/fixtures_test.go   |  80 +++++++
pkg/signing/keymanager_test.go | 505 +++++++++++-----------------------
pkg/signing/verifier_test.go   | 324 +++++++++----------------
3 files changed, 342 insertions(+), 567 deletions(-)
```

## Four conversions that are not mechanical

These are the spots where the obvious testify call is not equivalent to the code it replaces.

**`fetchCount.Load()` returns `int32`.** `assert.Equal(t, 1, fetchCount.Load())` fails on the type mismatch rather than the value, so the expected side is written `int32(1)`.

**`assert.Equal` on `[]byte` separates nil from empty.** The canonical payload comparisons pass `string(payload)` to keep the meaning of the original string comparison, and the nil payload check uses `assert.Nil`—`assert.Equal(t, nil, payload)` compares an untyped nil against an interface holding a nil `[]byte` and fails even when the payload is nil. Same finding as #421.

**Conditions ORed into one `if` split into one assertion each.** Pass and fail are unchanged; a failure now names which half broke.

**The `ed25519` key comparisons use `assert.Equal`, not `assert.True(pub.Equal(key))`.** `assert.True` prints only `false`. The two differ solely when one side is nil and the other empty, since testify falls back to `reflect.DeepEqual` for a named byte-slice type, and every compared key here is a 32-byte `GenerateKey` result that already passed `require.NoError`.

## Eight tests are renamed

Folding into tables turns eight top-level functions into subtests. The number of cases that run is unchanged at 28.

```
TestBuildCanonicalPayload                  -> TestBuildCanonicalPayload/all_fields_set
TestBuildCanonicalPayload_EmptyAnalyzedAt  -> TestBuildCanonicalPayload/empty_analyzed_at
TestBuildCanonicalPayload_HTMLChars        -> TestBuildCanonicalPayload/html_characters
TestVerifyCommand_EmptySignature           -> TestVerifyCommand_Errors/empty_signature
TestVerifyCommand_InvalidBase64            -> TestVerifyCommand_Errors/invalid_base64
TestVerifyCommand_WrongSignatureSize       -> TestVerifyCommand_Errors/wrong_signature_size
TestVerifyCommand_NilPublicKey             -> TestVerifyCommand_Errors/nil_public_key
TestVerifyCommand_NilCommand               -> TestVerifyCommand_Errors/nil_command
```

Anything filtering with `-run` on those names needs updating.

## What behaves differently

Two assertion strengthenings, both deliberate, both in the direction of catching more: `ErrSignatureMismatch` on the two signature paths, and the pinned messages on the other ten.

One fixture change. `testKeyResponse` always sets `valid_from`, so `TestKeyManager_InvalidAlgorithm` and `TestKeyManager_InvalidKeySize` now serve a field they did not before. Nothing reads it—`publicKeyResponse.ValidFrom` is parsed and never used, and both tests turn on `algorithm` and `public_key`.

## Verification

No production code changed.

- `go test ./... -p 1` passes, 45 packages ok
- `go test ./pkg/signing/ -count=20 -race -shuffle=on` passes
- `go vet ./pkg/signing/...` and `gofmt -l pkg/signing/` clean
- Each of the eight commits was checked out on its own and passes `go test ./pkg/signing/`, so any point in the range bisects

Since the value of a style PR is that "behavior unchanged" can be checked by reading, the claim was also measured rather than asserted. Twenty-one bugs were planted in the production code one at a time, and the pre-conversion test files were compared against the post-conversion ones on which tests each bug fails, with the eight renames mapped:

```
18/21 mutations caught identically
```

The eighteen cover HTML escaping, skipped U+2028 unescaping, corrupted literal backslashes, an empty slice returned where nil is expected, a dropped payload field, signature verification disabled, the public key size check removed, a nil command accepted, caching disabled, expiry ignored, `expires_at` ignored, the `auth_env` parameter dropped, corrupted key bytes, a corrupted key id, a mismatched kid accepted, the algorithm check removed, the key size check removed, and a stale key served after a failed refresh.

The three that differ are all cases the new tests catch and the old ones did not: returning a generic error instead of `ErrSignatureMismatch`, removing the signature size check, and removing the empty signature check. Nothing got weaker.

Separately, each of the ten pinned messages was changed to `boom` in turn, and every time exactly the one test naming that message failed. Dropping the `auth_env` parameter and always sending it both fail `TestKeyManager_AuthEnvQueryParam` and nothing else.

## Left for follow-ups on #370

- The rest of the candidate list, with the hand-written checks each still holds: `user` (72), `shell` (61), `system` (122), `updater` (56), `auth_manager_tracker` (50), `multipart_stream` (47), `migrate` (40)
- What #421 left in `pkg/utils`: six `os.CreateTemp("")` sites where `t.TempDir()` would clean up on panic, the order-dependent `TestGetCopyPath` subtests, and the 14 test files still on raw `t.Fatal`/`t.Error`

## Checklist

- [x] Every commit builds and passes tests in isolation
- [x] No production code touched; the number of cases that run is unchanged at 28
- [x] Coverage compared before and after with 21 planted bugs
- [x] `go test ./... -p 1` passes

Refs #370

### DIFF
--- a/pkg/signing/fixtures_test.go
+++ b/pkg/signing/fixtures_test.go
@@ -14,7 +14,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func newTestKey(t *testing.T) (ed25519.PublicKey, ed25519.PrivateKey) {
+const testServerID = "server-456"
+
+func testKey(t *testing.T) (ed25519.PublicKey, ed25519.PrivateKey) {
 	t.Helper()
 	pub, priv, err := ed25519.GenerateKey(nil)
 	require.NoError(t, err)
@@ -37,27 +39,29 @@ func writeKeyResponse(t *testing.T, w http.ResponseWriter, resp publicKeyRespons
 	assert.NoError(t, json.NewEncoder(w).Encode(resp))
 }
 
-// keyServer serves resp for every request, counting fetches when count is non-nil.
-func keyServer(t *testing.T, count *atomic.Int32, resp publicKeyResponse) *httptest.Server {
-	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
-		if count != nil {
-			count.Add(1)
-		}
+// keyServer serves resp for every request, counting fetches.
+func keyServer(t *testing.T, resp publicKeyResponse) (*httptest.Server, *atomic.Int32) {
+	var count atomic.Int32
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		count.Add(1)
 		writeKeyResponse(t, w, resp)
 	}))
+	t.Cleanup(srv.Close)
+	return srv, &count
 }
 
-func newTestServer(t *testing.T, pub ed25519.PublicKey, kid string) *httptest.Server {
-	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+// pathCheckingKeyServer is like keyServer but 404s outside the public-key path.
+func pathCheckingKeyServer(t *testing.T, pub ed25519.PublicKey, kid string) *httptest.Server {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path != "/api/commands/public-key/" {
 			http.NotFound(w, r)
 			return
 		}
 		writeKeyResponse(t, w, testKeyResponse(pub, kid))
 	}))
+	t.Cleanup(srv.Close)
+	return srv
 }
-
-const testServerID = "server-456"
 
 func testCommand(sig string) *protocol.Command {
 	return &protocol.Command{
--- a/pkg/signing/fixtures_test.go
+++ b/pkg/signing/fixtures_test.go
@@ -22,7 +22,6 @@ func newTestKey(t *testing.T) (ed25519.PublicKey, ed25519.PrivateKey) {
 }
 
 // testKeyResponse is what the AI server returns for an active Ed25519 key.
-// Tests that need an invalid response override a field on the result.
 func testKeyResponse(pub ed25519.PublicKey, kid string) publicKeyResponse {
 	return publicKeyResponse{
 		Algorithm: "Ed25519",
--- a/pkg/signing/fixtures_test.go
+++ b/pkg/signing/fixtures_test.go
@@ -1,0 +1,81 @@
+package signing
+
+import (
+	"crypto/ed25519"
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+
+	"github.com/alpacax/alpamon/v2/internal/protocol"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func newTestKey(t *testing.T) (ed25519.PublicKey, ed25519.PrivateKey) {
+	t.Helper()
+	pub, priv, err := ed25519.GenerateKey(nil)
+	require.NoError(t, err)
+	return pub, priv
+}
+
+// testKeyResponse is what the AI server returns for an active Ed25519 key.
+// Tests that need an invalid response override a field on the result.
+func testKeyResponse(pub ed25519.PublicKey, kid string) publicKeyResponse {
+	return publicKeyResponse{
+		Algorithm: "Ed25519",
+		PublicKey: base64.StdEncoding.EncodeToString(pub),
+		KeyID:     kid,
+		ValidFrom: "2026-01-01T00:00:00Z",
+	}
+}
+
+func writeKeyResponse(t *testing.T, w http.ResponseWriter, resp publicKeyResponse) {
+	t.Helper()
+	w.Header().Set("Content-Type", "application/json")
+	assert.NoError(t, json.NewEncoder(w).Encode(resp))
+}
+
+// keyServer serves resp for every request, counting fetches when count is non-nil.
+func keyServer(t *testing.T, count *atomic.Int32, resp publicKeyResponse) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		if count != nil {
+			count.Add(1)
+		}
+		writeKeyResponse(t, w, resp)
+	}))
+}
+
+func newTestServer(t *testing.T, pub ed25519.PublicKey) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/api/commands/public-key/" {
+			http.NotFound(w, r)
+			return
+		}
+		writeKeyResponse(t, w, testKeyResponse(pub, "key-test-123"))
+	}))
+}
+
+const testServerID = "server-456"
+
+func testCommand(sig string) *protocol.Command {
+	return &protocol.Command{
+		ID:         "cmd-123",
+		Shell:      "system",
+		Line:       "whoami",
+		User:       "root",
+		Group:      "root",
+		AnalyzedAt: "2026-03-01T12:00:00+00:00",
+		Signature:  sig,
+	}
+}
+
+// signedCommand returns a command carrying a signature valid for serverID.
+func signedCommand(t *testing.T, priv ed25519.PrivateKey, serverID string) *protocol.Command {
+	t.Helper()
+	cmd := testCommand("")
+	cmd.Signature = base64.StdEncoding.EncodeToString(ed25519.Sign(priv, BuildCanonicalPayload(cmd, serverID)))
+	return cmd
+}
--- a/pkg/signing/fixtures_test.go
+++ b/pkg/signing/fixtures_test.go
@@ -47,13 +47,13 @@ func keyServer(t *testing.T, count *atomic.Int32, resp publicKeyResponse) *httpt
 	}))
 }
 
-func newTestServer(t *testing.T, pub ed25519.PublicKey) *httptest.Server {
+func newTestServer(t *testing.T, pub ed25519.PublicKey, kid string) *httptest.Server {
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path != "/api/commands/public-key/" {
 			http.NotFound(w, r)
 			return
 		}
-		writeKeyResponse(t, w, testKeyResponse(pub, "key-test-123"))
+		writeKeyResponse(t, w, testKeyResponse(pub, kid))
 	}))
 }
 
--- a/pkg/signing/keymanager_test.go
+++ b/pkg/signing/keymanager_test.go
@@ -50,7 +50,7 @@ func TestIsLocalEnv(t *testing.T) {
 
 func TestKeyManager_Refresh(t *testing.T) {
 	pub, _ := newTestKey(t)
-	server := newTestServer(t, pub)
+	server := newTestServer(t, pub, "key-test-123")
 	defer server.Close()
 
 	km := NewKeyManager(server.URL, 3600, "", server.Client())
--- a/pkg/signing/keymanager_test.go
+++ b/pkg/signing/keymanager_test.go
@@ -2,8 +2,6 @@ package signing
 
 import (
 	"crypto/ed25519"
-	"encoding/base64"
-	"encoding/json"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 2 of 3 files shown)


geunwoonoh (dismissed):
## Summary

This PR moves the two `pkg/signing` test files onto testify and pulls the shared fixtures into `fixtures_test.go`. It is the second piece of #370 and touches no production code. The `t.Fatal*` → `require` and `t.Error*` → `assert` mapping is one for one, so what each check does on failure is preserved.

## What I verified independently

Rather than take the PR body at its word, I reproduced the claims.

| Check | Result |
|---|---|
| `go test ./... -p 1` (45 packages) | all pass |
| `go test ./pkg/signing -count=5 -race -shuffle=on` | passes |
| `go vet`, `gofmt -l` | clean |
| Coverage, `main` vs branch | **84.5% → 84.5%, identical per-function** |
| Cases executed | 30 PASS lines (22 top-level + 8 subtests), 28 leaf cases, unchanged |

I also planted three mutations in the production code and ran both suites against each:

| Planted bug | `main` tests | This PR's tests |
|---|---|---|
| Remove the `empty signature` check | **missed** | caught |
| Remove the `invalid signature size` check | **missed** | caught |
| Disable `ed25519.Verify` | caught | caught |

The claim that nothing got weaker and two checks got stronger holds up. It is rare for a style PR to carry evidence this concrete.

## Critical issues (0)

None.

## Warnings (1)

**1. `fixtures_test.go:56` and `keymanager_test.go:64` — a magic string now split across files**

`newTestServer` hardcodes the kid `"key-test-123"`, while the assertion that checks it lives in a different file.

```go
// fixtures_test.go:56
writeKeyResponse(t, w, testKeyResponse(pub, "key-test-123"))

// keymanager_test.go:64
assert.Equal(t, "key-test-123", km.keyID)
```

On `main` both sat in `keymanager_test.go`, so the connection was visible on one screen. Moving the fixture introduced the coupling, so this one is on the PR. Either take the kid as a parameter or lift it to a shared constant:

```go
func newTestServer(t *testing.T, pub ed25519.PublicKey, kid string) *httptest.Server {
```

## Suggestions (7)

**1. `fixtures_test.go:41` — `nil` passed as `count *atomic.Int32` to mean "don't count"**

Three of the seven call sites pass `keyServer(t, nil, resp)`. A nil pointer used as an optional-behavior switch is not idiomatic Go, and the other four sites each carry a separate `var fetchCount atomic.Int32`. Letting the helper own the counter drops both the nil branch and five declarations:

```go
func keyServer(t *testing.T, resp publicKeyResponse) (*httptest.Server, *atomic.Int32) {
	var count atomic.Int32
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		count.Add(1)
		writeKeyResponse(t, w, resp)
	}))
	return srv, &count
}
```

**2. `keymanager_test.go` — `defer server.Close()` repeated 11 times**

`keyServer` and `newTestServer` already take `t`, so a `t.Cleanup(srv.Close)` inside the helpers removes 11 lines at the call sites. No test closes its server early, so this is safe.

**3. `keymanager_test.go:269` — an assertion the next line already covers**

```go
assert.Contains(t, *q, "auth_env")
assert.Equal(t, "dev", q.Get("auth_env"))
```

`q.Get` returns `""` for an absent key, so the second line fails either way. The symmetry with the `assert.NotContains` below reads well, but the first line adds nothing.

**4. `keymanager_test.go:204` — an assertion two identically-wrong keys would pass**

```go
assert.Equal(t, key1, key2)
```

This proves the cache hit but never that either key is `pub`. One more line, `assert.Equal(t, pub, key1)`, closes it. The weakness predates the PR, but it sits squarely under this PR's own theme of pinning each check to what it expects.

**5. `fixtures_test.go:60` — a const declared between two functions**

`testServerID` sits between `newTestServer` and `testCommand`. Go convention puts it at the top of the file or immediately beside its only user.

**6. `fixtures_test.go` — two naming shapes for sibling helpers**

`newTestKey` and `newTestServer` take the `new` prefix; `keyServer`, `testKeyResponse`, `testCommand` and `signedCommand` do not. `newTestServer` and `keyServer` both return `*httptest.Server`, so the names give no hint that the difference between them is the path check.

**7. `keymanager_test.go:16,34` — two tables still iterate without `t.Run`**

`TestResolveAuthEnv` and `TestIsLocalEnv` loop without subtests. The Go stack rules ask for the `t.Run` shape, and since this PR added `t.Run` to the other two tables, the file is now inconsistent with itself. The `"ResolveAuthEnv(%q)"` message argument does name the failing case, so this is not urgent.

## Duplicate and unused code

- All eight fixtures are used; no dead helpers.
- Of the 80 lines in `fixtures_test.go`, only `newTestKey` (6 lines) is shared by both test files. `testServerID`, `testCommand` and `signedCommand` are verifier-only; `testKeyResponse`, `writeKeyResponse`, `keyServer` and `newTestServer` are keymanager-only. The "one test file per source file" argument is coherent, but the cost is that a reader of `verifier_test.go` has to jump to a third file to see what `testCommand` contains. This is a judgment call rather than a defect, so it is worth considering whether keeping only the genuinely shared `newTestKey` there reads better.
- **Outside this PR's scope**: `publicKeyResponse.ValidFrom` (`keymanager.go:59`) is parsed and never read anywhere in production. Only the test fixture populates it. Not something to change here, but a candidate for separate cleanup.

## Performance

Production code is untouched, so there is no runtime impact. On the test side the only addition is one nil check per request in `keyServer`, which is negligible; the package still runs in roughly 0.4s. Neither a regression nor an improvement.

## Coverage gaps (follow-ups for #370)

These predate the PR.

- `RefreshAndGet` is at **0%**. It is the retry path in `pkg/runner/signing.go:91` that refetches the key after a signature mismatch, which is exactly the scenario this PR just pinned `ErrSignatureMismatch` to. The most natural next test.
- `GetPublicKeyForKID` 70.6%, `fetchKeyLocked` 82.6%
- Untested branches: empty `key_id`, `maxResponseSize` overflow, malformed JSON body, and the TTL fallback when `expires_at` fails to parse.

## Highlights

- **`assert`, not `require`, inside the HTTP handlers** (`fixtures_test.go:37`). `require` calls `FailNow`, which must not be called off the test goroutine. Easy to get wrong, and it is right here.
- **`assert.Nil` for the nil `[]byte` payload** (`verifier_test.go:68`). `assert.Equal(t, nil, payload)` compares an untyped nil against an interface holding a nil `[]byte` and would not pass.
- **`assert.Equal(t, int32(1), ...)`**. testify compares types too, so a bare `1` would fail on the type rather than the value.
- **`atomic.Pointer[url.Values]` to publish the query** (`keymanager_test.go:256`). For the record, running `main`'s original version under `-race -count=50` does not flag anything, because `net/http` internals happen to supply the happens-before edge. So this is hardening rather than a live bug fix, and the PR body's framing of removing the question instead of answering it is accurate.
- **The eight renamed tests are documented with a mapping table**, including the effect on `-run` filters.
- Eight commits, one idea each, every one passing on its own.

---

**Verdict: APPROVED**

Nothing here blocks the merge. The one warning (the split kid string) is worth folding in now; the rest can go to follow-ups.

**Checklist**

- [x] Follows stack conventions (table-driven, testify, gofmt, vet)
- [x] Tests included and passing (45 packages, including race and shuffle)
- [x] No security issues (no hardcoded secrets; every test key is generated at runtime)
- [x] No docs needed (production code unchanged)
- [x] No coverage regression (84.5% unchanged, two additional mutations caught)

geunwoonoh (approved):
## Re-review: `2d4465f` → `0c0a25b`

Three new commits, each answering one part of the previous review.

```
02ff6ec  parameterize the server key ID instead of hardcoding it
a63c412  let key-server helpers own their counter and cleanup
0c0a25b  subtest the auth-env and local-env tables
```

## All eight points addressed, verified against the code

| # | Point | Done | What I checked |
|---|---|---|---|
| **W1** | kid magic string split across files | ✅ | `pathCheckingKeyServer(t, pub, kid)` now takes it as a parameter. The call site at `keymanager_test.go:75` passes `"key-test-123"` and line 78 asserts it, so the pair is visible on one screen again |
| **S1** | `nil` passed as the counter | ✅ | `keyServer` owns the counter and returns `(*httptest.Server, *atomic.Int32)`. No `keyServer(t, nil` remains; the three sites that don't need it use `server, _ :=` |
| **S2** | `defer server.Close()` ×11 | ✅ | `t.Cleanup(srv.Close)` inside both helpers. The three that remain belong to hand-built inline servers, which is the right place to keep them |
| **S3** | Redundant `assert.Contains` | ✅ | Removed; only `assert.Equal(t, "dev", q.Get("auth_env"))` is left |
| **S4** | Only `key1 == key2` asserted | ✅ | `assert.Equal(t, pub, key1)` added at `keymanager_test.go:197` |
| **S5** | Const wedged between functions | ✅ | `testServerID` moved directly below the imports |
| **S6** | Two naming shapes for sibling helpers | ✅ | `newTestKey`→`testKey`, `newTestServer`→`pathCheckingKeyServer`. The `new` prefix is gone, and the new name explains by itself how it differs from `keyServer` |
| **S7** | Two tables still looping without `t.Run` | ✅ | Both converted to subtests |

## Re-verification

| Check | Result |
|---|---|
| `go build ./...` | ok |
| `go test ./... -p 1` (45 packages) | all pass |
| `go test ./pkg/signing -count=5 -race -shuffle=on` | passes |
| `go vet`, `gofmt -l` | clean |
| Coverage | **84.5%, unchanged** |
| PASS lines | 28 → 43, purely from splitting 13 table rows into named subtests; the assertions are the same |

I planted the mutations again, adding one that removes the kid match in `GetPublicKeyForKID` to check whether the refactor leaked any detection strength.

| Planted bug | Result |
|---|---|
| Remove the `empty signature` check | caught |
| Remove the `invalid signature size` check | caught |
| Disable `ed25519.Verify` | caught |
| Drop the kid equality condition in `GetPublicKeyForKID` (new) | caught |

Nothing got weaker.

## One new suggestion (not blocking)

**`keymanager_test.go:30,49` — subtest names contain `/`, which breaks `-run` isolation**

`t.Run(tt.serverURL, ...)` uses the URL verbatim, and `/` is the subtest separator in `-run` patterns:

```
$ go test -run 'TestResolveAuthEnv/https://dev.alpacon.io$'
    --- PASS: TestResolveAuthEnv/https://dev.alpacon.io
    --- PASS: TestResolveAuthEnv/https://dev.alpacon.io/     <- matched too
```

So the one pair this table exists to tell apart, with and without the trailing slash, cannot be run individually. A `name` field, which is the conventional shape anyway, fixes it:

```go
tests := []struct {
	name      string
	serverURL string
	want      string
}{
	{"dev host", "https://dev.alpacon.io", "dev"},
	{"dev host with trailing slash", "https://dev.alpacon.io/", "dev"},
	...
}
```

The previous form (`assert.Equal(..., "ResolveAuthEnv(%q)", tt.serverURL)`) did not have this problem, but it also did not split failures into named cases. This version is better; the `name` field would finish the job. Fine as a follow-up.

## Follow-ups carried over from the previous review

- `RefreshAndGet` at 0% coverage, the signature-mismatch retry path in `pkg/runner/signing.go:91`
- `publicKeyResponse.ValidFrom` is parsed and never read in production
- The locality trade-off in `fixtures_test.go`

---

**Verdict: APPROVED**

**Checklist**

- [x] Follows stack conventions (table-driven, testify, gofmt, vet)
- [x] Tests included and passing (45 packages, including race and shuffle)
- [x] No security issues (no hardcoded secrets; every test key is generated at runtime)
- [x] No docs needed (production code unchanged)
- [x] No coverage regression (84.5% unchanged; four mutations caught)

PR #427, opened 2026-08-18
test(system): unify assertion style on testify

jisung-02:
Third PR for #370. Scope is `pkg/executor/handlers/system`—`system_test.go`, the only test file in the package, so it reads as testify-only after this.

The rule the first two PRs settled applies unchanged here: the original code had already chosen where a failure must stop, since every `t.Fatal*` aborts and every `t.Error*` keeps checking. The conversion maps `t.Fatal*` to `require` and `t.Error*` to `assert`, one for one, so what each check does on failure is preserved and the rule is something a reviewer can check by reading. 122 hand-written checks in, 35 `require` and 89 `assert` out; no test added, removed, or renamed.

## What changed

Three commits, one idea each.

`test(system): unify assertion style on testify`—all 122 checks become testify.

`test(system): pin the version lookup failure to its message`—`TestSystemHandler_Upgrade_VersionLookupFailureSelfUpdate` asserted only that *an* error came back, so any darwin-path failure satisfied it, including one raised before the version lookup was reached. It now asserts the message `handleUpgrade` returns for that path (`pkg/executor/handlers/system/system.go:243`). This is the only assertion in the PR that got stronger rather than equal; every other error check was already pinned to a message by the code being converted.

`test(system): drop the comments the assertions already make`—six comments restated the line under them, three of them inside a test already named `TestSystemHandler_PoolShutdown`. Comments carrying a why stay, including the note that a bare `zypper update` against stale metadata exits 0 without upgrading.

```
pkg/executor/handlers/system/system_test.go | 504 ++++++++--------------------
1 file changed, 136 insertions(+), 368 deletions(-)
```

## Five conversions that are not mechanical

`TestSystemHandler_Commands` collapses to one assertion. It compared the length, returned early on mismatch, then compared every index in a loop. `assert.Equal(t, expected, commands)` covers length, order and contents together.

Conditions ORed into one `if` split into one assertion each, so a failure names which half broke. Two of those ORs were also guarding an index access (`len(shell.Args) != 2 || shell.Args[0] != "-c" || ...`). Those keep the guard as `if assert.Len(t, shell.Args, 2) { ... }`: `require.Len` would abort the test, and the originals were `t.Errorf`, so the assertions after the block have to keep running. In `TestSystemHandler_Upgrade_PackageProxy` that is the difference between reporting a wrong argv and reporting a wrong argv plus whatever the six proxy env assertions find.

`TestSystemHandler_UnknownCommand` drops its bare error check. `assert.ErrorContains` fails on a nil error too, and the original dereferenced `err.Error()` two lines after a *non-fatal* nil check, so a nil error panicked the test instead of reporting it.

`TestSystemHandler_Update` turns "output empty and exit 1" into a plain `if`. The original `if output == "" && exitCode == 1` only entered its body when both held, so the `t.Errorf` inside never described what the failure was. It is now `if exitCode == 1 { assert.NotEmpty(t, output, ...) }`.

`TestSystemHandler_Uninstall` keeps one raw `t.Fatal`. It sits in a `select` against `time.After`, waiting for the uninstall goroutine to finish. That is synchronization, not a value check, and testify has no equivalent for it.

## Verification

No production code changed. `git diff origin/main..HEAD --name-only` is one test file.

The same 87 subtests pass before and after. Names collected with `go test -v` on `origin/main`'s copy of the file and on this branch, sorted and diffed:

```
$ diff old-names.txt new-names.txt && echo NAMES_IDENTICAL
NAMES_IDENTICAL
87
```

`gofmt -l` is empty, `go vet ./pkg/executor/handlers/system/` is clean, and the package passes under `-race -shuffle=on`:

```
ok  github.com/alpacax/alpamon/v2/pkg/executor/handlers/system  8.604s
```

Each of the three commits was built and tested on its own before the next one was written, so none of them is a broken intermediate state.

Behavior preservation was also checked by mutation: ten bugs were planted in `system.go` one at a time—dropping the zypper informational exit codes, changing `zypperLockAttempts`, breaking the refresh ordering, shortening the argv, and six others—and the old and new test files were run against each. Nine were caught identically by both. One was caught by neither, and it is a pre-existing gap rather than something this PR introduces: removing `--collect` from the `systemd-run` schedule still passes, because `TestSystemHandler_Uninstall_RetriesWithoutCollect` keys its mock result on the `--collect` form, so a code path that never uses `--collect` succeeds on the first attempt and still sets `retried`. Worth a follow-up in the package's own issue, not here.

## Follow-up, deliberately not in this PR

Roughly 25 tests repeat the same six-to-eight line mock, pool and context setup. A `newTestHandler` helper would cut about 200 more lines, but it rewrites test bodies rather than assertions and belongs in its own PR, where the diff can be read as one idea.

Part of #370. The remaining candidates are `user_test.go` (72 checks), `shell_test.go` (61), `updater_test.go` (56), `auth_manager_tracker_test.go` (50), `multipart_stream_test.go` (47) and `migrate_test.go` (40).

### DIFF
--- a/pkg/executor/handlers/system/system_test.go
+++ b/pkg/executor/handlers/system/system_test.go
@@ -9,9 +9,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
 	"github.com/alpacax/alpamon/v2/internal/pool"
 	"github.com/alpacax/alpamon/v2/pkg/agent"
 	"github.com/alpacax/alpamon/v2/pkg/config"
@@ -19,6 +16,8 @@ import (
 	"github.com/alpacax/alpamon/v2/pkg/updater"
 	"github.com/alpacax/alpamon/v2/pkg/utils"
 	"github.com/alpacax/alpamon/v2/pkg/version"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // MockWSClient is a mock implementation of WSClient for testing
--- a/pkg/executor/handlers/system/system_test.go
+++ b/pkg/executor/handlers/system/system_test.go
@@ -385,7 +385,6 @@ func TestSystemHandler_PoolShutdown(t *testing.T) {
 	handler := NewSystemHandler(mockExec, mockWS, ctxManager, workerPool, newMockVersionResolver(), nil)
 	ctx := context.Background()
 
-	// Shutdown pool first
 	_ = workerPool.Shutdown(100 * time.Millisecond)
 	ctxManager.Shutdown()
 
@@ -393,11 +392,9 @@ func TestSystemHandler_PoolShutdown(t *testing.T) {
 		Target: "alpamon",
 	}
 
-	// Should handle pool submission failure gracefully
 	exitCode, output, err := handler.Execute(ctx, common.Restart.String(), args)
 
 	require.NoError(t, err)
-	// Should still return success message even if pool submission fails
 	assert.Equal(t, 0, exitCode)
 	assert.Contains(t, output, "restart")
 }
@@ -631,11 +628,8 @@ func TestSystemHandler_Update(t *testing.T) {
 	// This test depends on the actual platform
 	exitCode, output, err := handler.Execute(ctx, common.Update.String(), args)
 
-	// Should not return an error
 	require.NoError(t, err)
-	// exitCode could be 0 (success) or 1 (platform not supported)
 	assert.Contains(t, []int{0, 1}, exitCode)
-	// Output should be present
 	if exitCode == 1 {
 		assert.NotEmpty(t, output, "an unsupported platform must say so")
 	}
--- a/pkg/executor/handlers/system/system_test.go
+++ b/pkg/executor/handlers/system/system_test.go
@@ -9,6 +9,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
 	"github.com/alpacax/alpamon/v2/internal/pool"
 	"github.com/alpacax/alpamon/v2/pkg/agent"
 	"github.com/alpacax/alpamon/v2/pkg/config"
@@ -166,9 +169,7 @@ func TestSystemHandler_Name(t *testing.T) {
 	defer ctxManager.Shutdown()
 
 	handler := NewSystemHandler(mockExec, mockWS, ctxManager, workerPool, newMockVersionResolver(), nil)
-	if handler.Name() != common.System.String() {
-		t.Errorf("expected name %q, got %q", common.System.String(), handler.Name())
-	}
+	assert.Equal(t, common.System.String(), handler.Name())
 }
 
 func TestSystemHandler_Commands(t *testing.T) {
@@ -192,16 +193,7 @@ func TestSystemHandler_Commands(t *testing.T) {
 		common.ByeBye.String(),
 	}
 
-	if len(commands) != len(expected) {
-		t.Errorf("expected %d commands, got %d", len(expected), len(commands))
-		return
-	}
-
-	for i, cmd := range commands {
-		if cmd != expected[i] {
-			t.Errorf("command %d: expected %q, got %q", i, expected[i], cmd)
-		}
-	}
+	assert.Equal(t, expected, commands)
 }
 
 func TestSystemHandler_Restart_Collector(t *testing.T) {
@@ -221,18 +213,10 @@ func TestSystemHandler_Restart_Collector(t *testing.T) {
 
 	exitCode, output, err := handler.Execute(ctx, common.Restart.String(), args)
 
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	if exitCode != 0 {
-		t.Errorf("expected exit code 0, got %d", exitCode)
-	}
-	if !mockWS.RestartCollectorCalled {
-		t.Error("expected RestartCollector to be called")
-	}
-	if !strings.Contains(output, "restarted") {
-		t.Errorf("expected output to contain 'restarted', got %q", output)
-	}
+	require.NoError(t, err)
+	assert.Equal(t, 0, exitCode)
+	assert.True(t, mockWS.RestartCollectorCalled)
+	assert.Contains(t, output, "restarted")
 }
 
 func TestSystemHandler_Restart_Default(t *testing.T) {
@@ -252,15 +236,9 @@ func TestSystemHandler_Restart_Default(t *testing.T) {
 
 	exitCode, output, err := handler.Execute(ctx, common.Restart.String(), args)
 
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	if exitCode != 0 {
-		t.Errorf("expected exit code 0, got %d", exitCode)
-	}
-	if !strings.Contains(output, "restart") {
-		t.Errorf("expected output to mention restart, got %q", output)
-	}
+	require.NoError(t, err)
+	assert.Equal(t, 0, exitCode)
+	assert.Contains(t, output, "restart")
 	// Give time for the pool task to execute
 	time.Sleep(100 * time.Millisecond)
 }
@@ -282,15 +260,9 @@ func TestSystemHandler_Restart_Alpamon(t *testing.T) {
 
 	exitCode, output, err := handler.Execute(ctx, common.Restart.String(), args)
 
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	if exitCode != 0 {
-		t.Errorf("expected exit code 0, got %d", exitCode)
-	}
-	if !strings.Contains(output, "restart") {
-		t.Errorf("expected output to mention restart, got %q", output)
-	}
+	require.NoError(t, err)
+	assert.Equal(t, 0, exitCode)
+	assert.Contains(t, output, "restart")
 }
 
 func TestSystemHandler_Quit(t *testing.T) {
@@ -308,15 +280,9 @@ func TestSystemHandler_Quit(t *testing.T) {
 
 	exitCode, output, err := handler.Execute(ctx, common.Quit.String(), args)
 
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
-	}
-	if exitCode != 0 {
-		t.Errorf("expected exit code 0, got %d", exitCode)
-	}
-	if !strings.Contains(output, "shutdown") {
-		t.Errorf("expected output to mention shutdown, got %q", output)
-	}
+	require.NoError(t, err)
+	assert.Equal(t, 0, exitCode)
+	assert.Contains(t, output, "shutdown")
 }
 
 func TestSystemHandler_Reboot(t *testing.T) {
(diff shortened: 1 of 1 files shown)


pkg/executor/handlers/system/system_test.go
```diff
@@ -1053,18 +883,14 @@ func TestSystemHandler_Upgrade_UsesZypper(t *testing.T) {
...
+		assert.NotContains(t, joined, "apt-get ", "a zypper host must not run apt")
 	}
+	require.GreaterOrEqual(t, updatedAt, 0, "a zypper update command must run, got %+v", mockExec.GetExecutedCommands())
 	// Against stale metadata a bare `update` exits 0 without upgrading, so the
 	// refresh must survive refactors.
-	if refreshedAt < 0 || refreshedAt > updatedAt {
-		t.Errorf("update must be preceded by a refresh, got %+v", mockExec.GetExecutedCommands())
-	}
+	assert.True(t, refreshedAt >= 0 && refreshedAt <= updatedAt,
+		"the update must be preceded by a refresh, got %+v", mockExec.GetExecutedCommands())
```
geunwoonoh: Suggestion: this is the one compound condition that escapes the rule this PR sets out, and `assert.True` reports only "should be true" on failure. The `require.GreaterOrEqual` two lines above already establishes `updatedAt >= 0`, so the split costs nothing.

```suggestion
	require.GreaterOrEqual(t, refreshedAt, 0, "a refresh must run, got %+v", mockExec.GetExecutedCommands())
	assert.Less(t, refreshedAt, updatedAt, "the update must be preceded by a refresh, got %+v", mockExec.GetExecutedCommands())
```

Verified locally: `gofmt` clean and `TestSystemHandler_Upgrade_UsesZypper` passes with this applied.

pkg/executor/handlers/system/system_test.go
```diff
@@ -715,18 +606,10 @@ func TestSystemHandler_Upgrade_VersionLookupFailureSelfUpdate(t *testing.T) {
...
-	if called {
-		t.Error("self-update must not run without a target version")
-	}
-	if !strings.Contains(output, "GitHub") {
-		t.Errorf("expected lookup failure message, got %q", output)
-	}
+	require.ErrorContains(t, err, "failed to retrieve the latest Alpamon version from GitHub")
+	assert.Equal(t, 1, exitCode)
+	assert.False(t, called, "self-update must not run without a target version")
+	assert.Contains(t, output, "GitHub")
```
geunwoonoh: Suggestion: the second commit pinned the error to its message, but this output check still rests on the single word `GitHub`. `system.go:242` returns the full sentence, so it can be pinned the same way — or dropped, since it now largely restates the `require.ErrorContains` above.

```suggestion
	assert.Contains(t, output, "Failed to retrieve the latest Alpamon version from GitHub.")
```

Verified locally: passes with this applied.

pkg/executor/handlers/system/system_test.go
```diff
@@ -9,6 +9,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
geunwoonoh: Suggestion (low priority): testify sits in its own group ahead of the local group here (three groups). That matches the std → third-party → local rule in our Go stack guide, but #422 merged alpacax and testify into a single group in `pkg/signing/fixtures_test.go` four days ago, and most other testify files in the repo do the same.

Both are gofmt-clean, so nothing is broken — but the series should settle on one and stay there. No change needed in this PR if the answer is "the stack guide wins and #422 was the outlier".

geunwoonoh (dismissed):
## Review

A testify conversion of the only test file in `pkg/executor/handlers/system`. Third in the #370 series, following the rule #421 (`pkg/utils`) and #422 (`pkg/signing`) settled: `t.Fatal*` to `require`, `t.Error*` to `assert`. No production code touched.

**I counted the mapping rather than taking it on trust.**

| | main | this branch |
|---|---|---|
| `t.Fatal*` | 36 | 1 raw (the `select` synchronization) + 35 `require` |
| `t.Error*` | 86 | 89 `assert` |

The three extra `assert` calls reconcile exactly against what the description claims: +5 from the OR splits (`PackageProxy` +2, `UsesZypper`, `Uninstall_UsesZypper` and `VersionLookupFailureProceeds` +1 each), -2 from the two collapses (`Commands` folded into one line, `UnknownCommand`'s bare nil check dropped). Nothing disappeared unaccounted for.

### What I ran

```
gofmt -l pkg/executor/handlers/system/     empty
go vet ./pkg/executor/handlers/system/     clean
go test -count=1                           ok  8.4s
go test -count=1 -race -shuffle=on         ok  8.5s
subtests                                   87, matching the description
gh pr checks 427                           22/22 pass, 0 failures
```

Performance is unchanged, measured rather than assumed. Three runs on `main` and three on this branch: 7.58/7.61/7.60s against 7.60/7.61/7.77s. This package's wall time is dominated by fixed `time.Sleep` and timeout waits inside the tests. Neither the `assert.NotContains` calls that moved inside loops nor the `GetExecutedCommands()` arguments now evaluated eagerly for `%+v` messages register, because `common/testing.go:129` returns the slice header with no lock and no copy.

### Critical issues (0)

None.

### Warnings (0)

I checked the argument order on all ~20 `assert.Equal` calls; none are reversed. There is also no place where a `require` relaxed to `assert` opens a nil dereference or an index panic: `require.NotNil` precedes every `shell` dereference, and `if assert.Len(...)` guards `shell.Args[0]`.

### Suggestions (4)

**1. `system_test.go:892` — the one compound condition that escapes the rule this PR sets**

The description states the rule that ORed conditions get split so a failure names which half broke. This line is the sole exception, and `assert.True` reports only "should be true". The `require.GreaterOrEqual` right above already establishes `updatedAt >= 0`, so splitting is cheap:

```go
require.GreaterOrEqual(t, refreshedAt, 0, "a refresh must run, got %+v", mockExec.GetExecutedCommands())
assert.Less(t, refreshedAt, updatedAt, "the update must be preceded by a refresh, got %+v", mockExec.GetExecutedCommands())
```

**2. `system_test.go:612` — the commit's intent does not reach the output check**

The second commit pinned the error to its message, but the output check in the same test still rests on the single word `"GitHub"`. `system.go:242` returns `"Failed to retrieve the latest Alpamon version from GitHub."`, so either pin the output the same way or drop the line, which now largely restates the error assertion.

**3. `system_test.go:27,36` — `MockWSClient.ShutDownCalled` is write-only**

`m.ShutDownCalled = true` is set and never asserted anywhere in the package. Not introduced here, but it is a dead field, and `golangci-lint`'s `unused` will not catch it because the write counts as a use. Either remove it or assert on it in the `Quit`/`Shutdown` tests, which would add a little coverage.

**4. `system_test.go:12-13` — import grouping diverges from its sibling PR**

testify sits in its own group ahead of the local group (three groups). That matches the std to third-party to local rule in our Go stack guide, but #422 merged alpacax and testify into one group in `pkg/signing/fixtures_test.go` four days ago, and most other testify files in the repo do the same. Both are gofmt-clean, so this is not urgent, but the series should settle on one.

### On the duplication

39 tests repeat the same six-to-eight line mock, pool and context setup: `common.NewMockCommandExecutor(t)` 39 times, `NewSystemHandler(mockExec, mockWS, ctxManager, workerPool, ...)` 39 times, `defer ctxManager.Shutdown()` 38 times. It is the largest remaining cleanup in the file, and the description already names it, sizes it at roughly 200 lines behind a `newTestHandler` helper, and defers it.

**I agree with deferring it.** Substituting assertions and rewriting test bodies are different kinds of diff, and mixing them would cost this PR the property that makes it reviewable: that every change can be checked line against line.

No other dead code or unused helpers. `findExecutedShell`, `setPackageManagerAndID`, `lockedThenFreeExecutor` and all four `MockVersionResolver` fields are genuinely used.

### Highlights

- **`TestSystemHandler_UnknownCommand` (line 344) fixes a real latent panic.** The old code checked `err == nil` with a *non-fatal* `t.Error`, then dereferenced `err.Error()` two lines later, so a nil error panicked the test instead of reporting it. `assert.ErrorContains` fails cleanly on nil. A genuine fix buried in a refactor, and properly disclosed in the description.
- **The `if assert.Len(t, shell.Args, 2) { ... }` guard** (lines 473, 578) is the right call. `require.Len` would have aborted and skipped the six proxy env assertions in `PackageProxy`; keeping the guard preserves that the originals were `t.Errorf`.
- Three commits, one idea each, and the third keeps the comments that carry a *why*, including the note that a bare `zypper update` against stale metadata exits 0 without upgrading.
- Behavior preservation was checked by mutation, and the one mutant neither file caught is correctly identified as a pre-existing gap (`Uninstall_RetriesWithoutCollect` keys its mock result on the `--collect` form) rather than something this PR introduces.

---

**Verdict: APPROVED**

None of the four suggestions block the merge. 1 and 2 are one-line changes worth making here; 3 and 4 can ride along with the next PR in the series or the follow-up helper PR.

**Checklist**
- [x] Follows stack conventions (gofmt and go vet clean; testify use matches the Go stack testing rules)
- [x] Tests included and passing (`-race -shuffle=on` included, CI 22/22)
- [x] No security concerns (test-only change, no hardcoded secrets)
- [x] No documentation update needed (no public behavior change)

pkg/executor/handlers/system/system_test.go
jisung-02: Applied in fe0ce70b, with one deviation from the suggestion: both halves stay `assert` rather than the first becoming `require`.

```go
assert.GreaterOrEqual(t, refreshedAt, 0, "a zypper refresh command must run, got %+v", mockExec.GetExecutedCommands())
assert.LessOrEqual(t, refreshedAt, updatedAt, "the update must be preceded by a refresh, got %+v", mockExec.GetExecutedCommands())
```

The original was a `t.Errorf`, so raising the first half to `require` would make this the only line in the file whose fatality differs from the code it replaced, which is the property the PR is built on. Nothing is lost by keeping it non-fatal: a `refreshedAt` of -1 fails on the first assertion and names the reason, and the second one passing spuriously on -1 does not hide it.

`LessOrEqual` rather than `Less` for the same reason, matching the original `refreshedAt <= updatedAt`. The two indices can never be equal in practice, so the strictness is not observable either way.

pkg/executor/handlers/system/system_test.go
jisung-02: Applied in 4f627b1e, taking the pin rather than the drop.

```go
assert.Contains(t, output, "Failed to retrieve the latest Alpamon version from GitHub.")
```

Kept rather than dropped because `handleUpgrade` returns two different strings on that path, `system.go:242-243`: the capitalized sentence that reaches the operator as output, and the lowercase error. The `require.ErrorContains` above covers only the second one, so removing this line would leave the output unchecked.

pkg/executor/handlers/system/system_test.go
jisung-02: Merged into one group in b134d7b8, and the same change is on the `user` branch for the next PR in the series.

I counted the repo before deciding. Of the 23 test files that import both testify and a local package, 22 keep them in one group and this file was the only split. Nothing enforces either shape: there is no `.golangci.yml` in the repo, so golangci-lint runs its default linters and none of them check import order, and `gofmt` only sorts within a group. Both forms were clean, which is why this survived.

So the tiebreaker is that the series should not add a second style. Following the stack guide instead would mean converting 22 unrelated files, which is not this series' job.

geunwoonoh (approved):

PR #428, opened 2026-08-18
test(user): unify assertion style on testify

jisung-02:
Fourth PR for #370. Scope is `pkg/executor/handlers/user`—`user_test.go`, the only test file in the package, so it reads as testify-only after this.

The rule the earlier PRs settled applies unchanged: the original code had already chosen where a failure must stop, since every `t.Fatal*` aborts and every `t.Error*` keeps checking. The conversion maps `t.Fatal*` to `require` and `t.Error*` to `assert`, one for one. All 27 `t.Fatal*` became `require` and all 45 `t.Error*` became `assert`; no test was added or removed, and exactly one subtest was renamed.

## What changed

Eleven commits, one idea each. The first five are the conversion; the last six answer the review on it and are described under [The review](#the-review).

| Commit | What it does |
|---|---|
| `1683e54f` | the conversion: every hand-written check in the file becomes testify |
| `f93376dd` | testify joins the local import group, matching the 22 other files in the repo that import both |
| `9ca0ca31` | the group-add table asserted `err` only inside a branch that never ran |
| `1c22d398` | a subtest name carried `/`, which is the `-run` separator |
| `1d698b85` | both tables accepted any error; each row now pins the one it expects |
| `97c4718a` | a row comment said the opposite of the field list below it |
| `1deafdc1` | two tables ran the same scenario with identical args |
| `0c36ed88` | one failing row asserted an exit code and nothing else |
| `affbf18f` | the forbid column compared slice elements while its message printed the joined string |
| `b3490e79` | `wantCode` was a non-varying column, the same reason `calledGroups` had already left |
| `7b535409` | the `groupadd -f` scan existed in two copies |
| `b4964b7d` | an assertion message that fit only one of its two rows |

```
pkg/executor/handlers/user/user_test.go | 439 +++++++++++++-------------------
1 file changed, 170 insertions(+), 269 deletions(-)
```

Commits three through five answer suggestions carried over from the earlier reviews in this series rather than the conversion itself: #427 asked for error checks to be pinned to their message instead of accepting any error, #422's re-review flagged `/` in subtest names for breaking `-run` isolation, and #422 flagged an assertion that two equally wrong values would have passed.

## The counts close exactly

The conversion turned 72 checks into 79 assertions: +9 from splitting compound conditions, -2 from folding two `if` pairs. `72 + 9 - 2 = 79`.

The three follow-up commits take that to 82, which is `require` 30 plus `assert` 52 as counted in the file:

| Commit | Delta | Why |
|---|---|---|
| `9ca0ca31` | +1 `assert` | the output each row produces is now pinned; it was discarded before |
| `1d698b85` | +1 `assert`, +1 `require` | `Validate` splits into three cases instead of two, and the new helper uses `require.ErrorAs` |

Nine conditions ORed into one `if` became one assertion each in the conversion, so a failure names which half broke:

| Original | Became |
|---|---|
| `(err != nil) != tt.wantErr` (`Execute`, `Validate`) | `if tt.wantErr { assert.Error } else { assert.NoError }` |
| `err != nil \|\| exitCode != 0` (two call sites) | `require.NoError` + `require.Equal` |
| `!C("--gid 5001") \|\| !C("alpacon")` | two `assert.Contains` |
| `!C("uid 9999") \|\| !C("requested uid 1001")` | two `assert.Contains` |
| `Invoked(adduser) \|\| Invoked(useradd)` | two `assert.False` |
| `C("created") \|\| !C("already exists but failed…")` | `assert.NotContains` + `assert.Contains` |
| `!C("already exists") \|\| C("added successfully")` | `assert.Contains` + `assert.NotContains` |

The `err != nil || exitCode != 0` pair is the only split that produces two `require` calls rather than two `assert` calls. The original aborted on either half, so both halves stay fatal.

The first row of that table was superseded by `1d698b85`, described below. The `wantErr` boolean it introduced is gone.

## Each row pins the error it expects

`assert.Error(t, err)` passes on whatever came back, so `Execute`'s unknown-command and adduser-failure rows shared one expectation, and `Validate`'s four failing rows shared one across three different validation causes.

`Execute` carries `wantErrMsg` and asserts `EqualError`:

```go
wantErrMsg:   "unknown user command: unknownuser",
wantErrMsg:   "user add error",
```

`Validate` splits by error kind, because the two are not the same shape. The unknown command returns a plain `fmt.Errorf`, so it takes `wantErrMsg`. The rest return `validator.ValidationErrors`, one line per field, so they take `wantErrFields` and assert the exact field set:

```go
wantErrFields: []string{"UID", "GID", "Comment", "HomeDirectory", "Shell", "Groupname"},
wantErrFields: []string{"UID", "GID", "HomeDirectory"},
wantErrFields: []string{"UID"},
```

```go
switch {
case tt.wantErrMsg != "":
	assert.EqualError(t, err, tt.wantErrMsg)
case len(tt.wantErrFields) > 0:
	assert.ElementsMatch(t, tt.wantErrFields, validationFields(t, err))
default:
	assert.NoError(t, err)
}
```

Pinning the whole multi-line validator string would restate the library's own wording, and `assert.ErrorContains` on one representative field would not separate the rows: the missing-fields row and the IAM row both flag `UID`. The field set is what each row is actually about. Every expectation was read off an actual run rather than derived from the struct tags.

The row comment claiming the adduser failure was "returned as part of output, not actual go error" is gone with this commit. The mock returns a real error, which the pinned message now shows.

## The group-add table asserted an error it never reached

`TestUserHandler_AddUserWithGroups` guarded its only `err` assertion behind a branch that could not run:

```go
// before
if !tt.groupService.AddUserToGroupsCalled {
	assert.NoError(t, err)
}
assert.Equal(t, tt.calledGroups, tt.groupService.AddUserToGroupsCalled, "AddUserToGroups called")
```

Both rows set additional groups, so `AddUserToGroups` always runs and the branch never executed. Replacing its body with `t.Fatal("UNREACHABLE BRANCH")` keeps the test passing, which is how this was confirmed. The `err` from `Execute` was therefore unchecked, and since `err` was used nowhere else, deleting the branch alone does not compile.

`calledGroups` was `true` in both rows, which left the table with no varying expectation. It becomes `wantOutput`, so the soft-failure path the test exists for is finally covered—the output was discarded before:

```go
// after
exitCode, output, err := handler.Execute(ctx, "adduser", args)

// A group-add failure reaches the operator through the output, not err.
assert.NoError(t, err)
assert.Equal(t, 0, exitCode, "the create path exits 0 whether or not the group-add step fails")
assert.Equal(t, tt.wantOutput, output)
assert.True(t, tt.groupService.AddUserToGroupsCalled, "AddUserToGroups must run when the args carry additional groups")
```

`handleAddUser` returns a nil `err` on every path this test reaches, including the group-add soft failure at `pkg/executor/handlers/user/user.go:310`, so `NoError` is right for both rows.

## The renamed subtest

`"adduser IAM User path must still require uid/gid/home"` spanned three levels of the `-run` pattern, since `/` is the separator, so the case could not be selected on its own. It is now `"...require uid, gid and home"`, and it is the only name in the file that carried a slash.

## Three loops replaced by an equivalent the package already has

Each was checked against the helper's source, not assumed.

The `forbidFlags` nested loop becomes `assert.NotContains(t, joined, forbid)`. The conversion first kept the original `a == forbid` element comparison; the review pointed out that the failure message printed `joined` either way, so the two columns of the table read as one kind of check and were not. Matching the forbid column against the joined string as well makes both columns substring rules, and it is the stronger of the two: `--uid=1001` is not element-equal to `--uid` and used to pass.

The `useradd` scan in `TestUserHandler_AddUser_ServiceAccountGroupaddFails` becomes `assert.False(t, mock.Invoked("/usr/sbin/useradd"), ...)`. `Invoked` (`pkg/executor/handlers/common/testing.go:134`) walks the same `m.commands` slice with the same exact `c.Name == program` match, and the rest of this file already uses it.

The `groupadd -f` / `useradd` ordering check becomes `assert.Less(t, groupaddDashFIndex, useraddIndex, ...)`, which prints both indices that the original message spelled out by hand. The scan that produces the first index is now `invokedAt`, described below.

## The review

The review on the conversion raised one warning and five suggestions. All six are taken, one commit each.

**The warning** was a comment saying `Shell is defaulted by the helper` three lines above a `wantErrFields` list containing `Shell`. If it were defaulted it could not fail validation. `types.go:52` states the opposite and says why: defaulting `Shell` before `ValidateStruct` would mask an empty `shell` payload and hand a service account an interactive `/bin/bash`. The comment predates this branch, but turning `wantErr: true` into an explicit field list is what made it contradict the line below it, so it is fixed here.

**Two tables ran the same scenario.** `TestUserHandler_Execute/adduser success debian` and `TestUserHandler_AddUserWithGroups/add user to groups success` took character-identical args, down to `Groups: []uint64{1002, 1003}`. Both used to discard the output, so the overlap did not show; pinning the output string exposed it. Dropping `Groups` from the `Execute` row leaves it covering the plain create path, which no other row covers.

**One failing row stayed unpinned.** `"adduser missing username"` asserted `wantCode: 1` and nothing else. That path exits through `user.go:118` as `return 1, err.Error(), nil`, so the string an operator sees rides in `output` and `wantErrMsg` cannot reach it. A `wantOutputPart` column covers that shape, and every failing row in the table now states what it expects to be told.

**`wantCode` was a column both rows filled identically.** `calledGroups` had left the same table for exactly that reason, so the rule was applied once and followed once. The expectation is worth keeping, just not as a column: a group-add failure must not change the exit code, which reads better as a literal `0` with that sentence attached.

**The `groupadd -f alpacon` scan existed twice**, character for character, in `TestUserHandler_AddUser_UidLess` and `TestUserHandler_AddUser_Idempotent_ServiceAccountNameExists`. It is now one helper:

```go
func invokedAt(mock *common.MockCommandExecutor, program string, args ...string) int {
	for i, c := range mock.GetExecutedCommands() {
		if c.Name == program && slices.Equal(c.Args, args) {
			return i
		}
	}
	return -1
}
```

The review suggested an `InvokedWithArgs(program string, args ...string) bool` on the shared mock next to `Invoked`, or a small helper inside this file. This takes the second, and returns an index rather than a bool. The index is what the uid-less test needs, since it asserts that `groupadd -f` ran before adduser/useradd, and a bool cannot express that. It stays in this file because both call sites are here: putting it on `MockCommandExecutor` would add an exported API to a mock nine packages share, for two callers.

The match is the whole argument list rather than the leading two the loops compared. Both call sites run `groupadd -f alpacon` with exactly those arguments, so nothing changes today, and an argument added later will say so instead of passing quietly.

## Verification

No production code changed. `git diff origin/main..HEAD --name-only` is one test file.

The 48 test and subtest PASS lines are the same before and after, apart from the deliberate rename. Names were collected with `go test -v` on `origin/main`'s copy of the file and on this branch, sorted and diffed:

```
$ diff names-old.txt names-new.txt
26c26
<     --- PASS: TestUserHandler_Validate/adduser_IAM_User_path_must_still_require_uid/gid/home
---
>     --- PASS: TestUserHandler_Validate/adduser_IAM_User_path_must_still_require_uid,_gid_and_home
```

`gofmt -l` is empty, `go vet ./pkg/executor/handlers/user/` is clean, and the package passes under `-race -shuffle=on`:

```
ok  github.com/alpacax/alpamon/v2/pkg/executor/handlers/user  1.512s
```

`go test ./... -p 1` passes across all 45 packages.

Behavior preservation for the conversion was checked by mutation. Six bugs were planted in `user.go` one at a time and the old and new test files were run against each:

| Planted bug | Old tests | New tests |
|---|---|---|
| `groupadd -f` failure returned as exit 0 | CAUGHT | CAUGHT |
| lookup error ignored, blind-creates | CAUGHT | CAUGHT |
| uid conflict masked | CAUGHT | CAUGHT |
| `groupadd -f` skipped for service accounts | CAUGHT | CAUGHT |
| `groupadd` runs regardless of the lookup gate | CAUGHT | CAUGHT |
| `--uid` leaks onto the uid-less path | CAUGHT | CAUGHT |

The last two target the folded assertions specifically, since those are the two places where one assertion now covers what two `if` blocks used to.

The three follow-up commits were probed the same way, by breaking each new assertion and confirming it reports:

| Probe | Result |
|---|---|
| an extra field added to a `wantErrFields` set | `FAIL … extra elements in list A` |
| `"user add error"` changed to `"user add errorX"` | `FAIL TestUserHandler_Execute/adduser_failure` |
| `assert.NoError(t, err)` flipped to `require.Error` | both rows FAIL, so the assertion runs on both |

The six review commits were probed the same way:

| Probe | Result |
|---|---|
| `wantOutputPart` changed to a string the output cannot contain | `FAIL … does not contain "Nonexistent"` |
| `--shell` added to the debian row's `forbidFlags` | `FAIL … should not contain "--shell"` |
| `invokedAt` asked for `-f wrong` | both uid-less rows FAIL |
| `invokedAt` asked for `-f` alone, one argument short | debian row FAILs |

`user.go` was restored from a saved copy after each run; `git status` shows only the test file modified.

## Left for a follow-up

The `/usr/sbin/*` program paths repeat as literals: `useradd` 13 times, `adduser` 11, `groupadd` 8. Constants would make a typo in a `mock.Invoked` argument a compile error rather than a silently false assertion, since `Invoked` returns `false` for any name it does not recognise. This was tried and dropped on purpose: production hardcodes the same paths (`user.go:161,218,271,414,466`), so a test-side constant duplicates them, and sharing one constant with production would let a wrong path pass its own test. Doing it properly touches both sides, which is a different diff from an assertion-style pass.

## Remaining candidates

Part of #370, which stays open. Next: `shell_test.go` (61 checks, plus three smaller files in the same package), `updater_test.go` (56), `auth_session_event_test.go` (54), `auth_manager_tracker_test.go` (50), `multipart_stream_test.go` (47).

### DIFF
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -616,7 +616,7 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 
 			// A group-add failure reaches the operator through the output, not err.
 			assert.NoError(t, err)
-			assert.Equal(t, 0, exitCode, "a group-add failure must not change the exit code")
+			assert.Equal(t, 0, exitCode, "the create path exits 0 whether or not the group-add step fails")
 			assert.Equal(t, tt.wantOutput, output)
 			assert.True(t, tt.groupService.AddUserToGroupsCalled, "AddUserToGroups must run when the args carry additional groups")
 		})
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -79,7 +79,8 @@ func TestUserHandler_Execute(t *testing.T) {
 				HomeDirectory: "/home/testuser",
 				Shell:         "/bin/bash",
 				Groupname:     "testgroup",
-				Groups:        []uint64{1002, 1003},
+				// No additional Groups: the group-add path is covered by
+				// TestUserHandler_AddUserWithGroups, which pins its output.
 			},
 			setupMock: func(mock *common.MockCommandExecutor) {
 				mock.SetResult(fmt.Sprintf("/usr/sbin/adduser --home /home/testuser --shell /bin/bash --uid %d --gid %d --gecos Test User --disabled-password testuser", 1001, 1001), 0, "User created", nil)
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -462,7 +462,8 @@ func TestUserHandler_Validate(t *testing.T) {
 				Username: "testuser",
 				// IsServiceAccount=false (default), so UID/GID/HomeDirectory
 				// are required_unless and also missing. Comment/Groupname are
-				// required unconditionally. Shell is defaulted by the helper.
+				// required unconditionally. Shell carries `validate:"required"`
+				// and is deliberately not defaulted (types.go:52), so it fails too.
 			},
 			wantErrFields: []string{"UID", "GID", "Comment", "HomeDirectory", "Shell", "Groupname"},
 		},
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -467,7 +467,7 @@ func TestUserHandler_Validate(t *testing.T) {
 			wantErr: false,
 		},
 		{
-			name: "adduser IAM User path must still require uid/gid/home",
+			name: "adduser IAM User path must still require uid, gid and home",
 			cmd:  "adduser",
 			args: &common.CommandArgs{
 				Username:  "john",
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -280,7 +280,7 @@ func TestUserHandler_AddUser_UidLess(t *testing.T) {
 				assert.Contains(t, joined, want, "flag %q must be present", want)
 			}
 			for _, forbid := range tt.forbidFlags {
-				assert.NotContains(t, target.Args, forbid, "args: %s", joined)
+				assert.NotContains(t, joined, forbid, "args: %s", joined)
 			}
 
 			// Service account must NOT use the gid-based groupadd path
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -18,9 +18,23 @@ import (
 
 	"github.com/alpacax/alpamon/v2/pkg/executor/handlers/common"
 	"github.com/alpacax/alpamon/v2/pkg/utils"
+	"github.com/go-playground/validator/v10"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+// validationFields names the struct fields a ValidateStruct error flags, so a
+// table row can pin exactly which fields failed instead of accepting any error.
+func validationFields(t *testing.T, err error) []string {
+	t.Helper()
+	var verrs validator.ValidationErrors
+	require.ErrorAs(t, err, &verrs)
+	fields := make([]string, 0, len(verrs))
+	for _, verr := range verrs {
+		fields = append(fields, verr.Field())
+	}
+	return fields
+}
 
 // MockGroupService implements services.GroupService for testing
 type MockGroupService struct {
@@ -52,7 +66,7 @@ func TestUserHandler_Execute(t *testing.T) {
 		setupMock    func(*common.MockCommandExecutor)
 		groupService *MockGroupService
 		wantCode     int
-		wantErr      bool
+		wantErrMsg   string
 	}{
 		{
 			name: "adduser success debian",
@@ -72,7 +86,6 @@ func TestUserHandler_Execute(t *testing.T) {
 			},
 			groupService: &MockGroupService{},
 			wantCode:     0,
-			wantErr:      false,
 		},
 		{
 			name: "deluser success",
@@ -86,7 +99,6 @@ func TestUserHandler_Execute(t *testing.T) {
 			},
 			groupService: &MockGroupService{},
 			wantCode:     0,
-			wantErr:      false,
 		},
 		{
 			name: "moduser success",
@@ -102,7 +114,6 @@ func TestUserHandler_Execute(t *testing.T) {
 			},
 			groupService: &MockGroupService{},
 			wantCode:     0,
-			wantErr:      false,
 		},
 		{
 			name:         "unknown command",
@@ -110,7 +121,7 @@ func TestUserHandler_Execute(t *testing.T) {
 			args:         &common.CommandArgs{},
 			groupService: &MockGroupService{},
 			wantCode:     1,
-			wantErr:      true,
+			wantErrMsg:   "unknown user command: unknownuser",
 		},
 		{
 			name: "adduser missing username",
@@ -121,7 +132,6 @@ func TestUserHandler_Execute(t *testing.T) {
 			},
 			groupService: &MockGroupService{},
 			wantCode:     1,
-			wantErr:      false,
 		},
 		{
 			name: "adduser failure",
@@ -140,7 +150,7 @@ func TestUserHandler_Execute(t *testing.T) {
 			},
 			groupService: &MockGroupService{},
 			wantCode:     1,
-			wantErr:      true, // error is returned as part of output, not actual go error
+			wantErrMsg:   "user add error",
 		},
 	}
 
@@ -163,13 +173,13 @@ func TestUserHandler_Execute(t *testing.T) {
 
 			exitCode, output, err := handler.Execute(ctx, tt.cmd, tt.args)
 
-			if tt.wantErr {
-				assert.Error(t, err)
+			if tt.wantErrMsg != "" {
+				assert.EqualError(t, err, tt.wantErrMsg)
 			} else {
 				assert.NoError(t, err)
 			}
 			assert.Equal(t, tt.wantCode, exitCode)
-			if exitCode == 0 && !tt.wantErr {
+			if exitCode == 0 && tt.wantErrMsg == "" {
 				assert.NotEmpty(t, output, "a successful command must carry output")
 			}
 		})
@@ -423,10 +433,14 @@ func TestUserHandler_Validate(t *testing.T) {
 	handler := newTestUserHandler(common.NewMockCommandExecutor(t), &MockGroupService{})
 
 	tests := []struct {
-		name    string
-		cmd     string
-		args    *common.CommandArgs
-		wantErr bool
+		name string
+		cmd  string
+		args *common.CommandArgs
+		// wantErrMsg pins a non-validator error to its exact message;
+		// wantErrFields pins a validator error to the exact set of fields it
+		// must flag. Both empty means the args must validate.
+		wantErrMsg    string
+		wantErrFields []string
 	}{
 		{
 			name: "adduser valid",
@@ -440,7 +454,6 @@ func TestUserHandler_Validate(t *testing.T) {
 				Shell:         "/bin/bash",
 				Groupname:     "testgroup",
 			},
-			wantErr: false,
 		},
 		{
 			name: "adduser missing required fields",
@@ -451,7 +464,7 @@ func TestUserHandler_Validate(t *testing.T) {
 				// are required_unless and also missing. Comment/Groupname are
 				// required unconditionally. Shell is defaulted by the helper.
 			},
-			wantErr: true,
+			wantErrFields: []string{"UID", "GID", "Comment", "HomeDirectory", "Shell", "Groupname"},
 		},
 		{
 			name: "adduser uid-less service account valid",
@@ -464,7 +477,6 @@ func TestUserHandler_Validate(t *testing.T) {
 				IsServiceAccount: true,
 				// UID/GID/HomeDirectory intentionally omitted (OS auto-assign)
 			},
-			wantErr: false,
 		},
 		{
 			name: "adduser IAM User path must still require uid, gid and home",
@@ -477,7 +489,7 @@ func TestUserHandler_Validate(t *testing.T) {
 				// IsServiceAccount=false (default)
 				// UID/GID/HomeDirectory missing: must fail
 			},
-			wantErr: true,
+			wantErrFields: []string{"UID", "GID", "HomeDirectory"},
 		},
 		{
 			name: "adduser IAM User with uid=0 must fail (cannot silently auto-assign)",
@@ -492,7 +504,7 @@ func TestUserHandler_Validate(t *testing.T) {
 				Groupname:     "alpacon",
 				// IsServiceAccount=false: uid=0 must be rejected
 			},
-			wantErr: true,
+			wantErrFields: []string{"UID"},
 		},
 		{
 			name: "deluser valid",
@@ -500,7 +512,6 @@ func TestUserHandler_Validate(t *testing.T) {
 			args: &common.CommandArgs{
 				Username: "testuser",
 			},
-			wantErr: false,
 		},
 		{
 			name: "moduser valid",
@@ -510,22 +521,24 @@ func TestUserHandler_Validate(t *testing.T) {
 				Groupnames: []string{"sudo"},
 				Comment:    "Updated",
 			},
-			wantErr: false,
 		},
 		{
-			name:    "unknown command",
-			cmd:     "unknown",
-			args:    &common.CommandArgs{},
-			wantErr: true,
+			name:       "unknown command",
+			cmd:        "unknown",
+			args:       &common.CommandArgs{},
+			wantErrMsg: "unknown user command: unknown",
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			err := handler.Validate(tt.cmd, tt.args)
-			if tt.wantErr {
-				assert.Error(t, err)
-			} else {
+			switch {
+			case tt.wantErrMsg != "":
+				assert.EqualError(t, err, tt.wantErrMsg)
+			case len(tt.wantErrFields) > 0:
+				assert.ElementsMatch(t, tt.wantErrFields, validationFields(t, err))
+			default:
 				assert.NoError(t, err)
 			}
 		})
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -538,7 +538,7 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 		setupMock    func(*common.MockCommandExecutor)
 		groupService *MockGroupService
 		wantCode     int
-		calledGroups bool
+		wantOutput   string
 	}{
 		{
 			name: "add user to groups success",
@@ -547,7 +547,7 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 			},
 			groupService: &MockGroupService{},
 			wantCode:     0,
-			calledGroups: true,
+			wantOutput:   "User 'testuser' added successfully",
 		},
 		{
 			name: "add user to groups failure",
@@ -556,7 +556,7 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 			},
 			groupService: &MockGroupService{AddUserToGroupsError: errors.New("failed to add to groups")},
 			wantCode:     0, // Still want 0 for user creation, group error is logged
-			calledGroups: true,
+			wantOutput:   "User 'testuser' created but failed to add to groups: failed to add to groups",
 		},
 	}
 
@@ -588,14 +588,13 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 				Groups:        []uint64{1002, 1003},
 			}
 
-			exitCode, _, err := handler.Execute(ctx, "adduser", args)
+			exitCode, output, err := handler.Execute(ctx, "adduser", args)
 
-			// A group-add failure is reported through the group service, not err.
-			if !tt.groupService.AddUserToGroupsCalled {
-				assert.NoError(t, err)
-			}
+			// A group-add failure reaches the operator through the output, not err.
+			assert.NoError(t, err)
 			assert.Equal(t, tt.wantCode, exitCode)
-			assert.Equal(t, tt.calledGroups, tt.groupService.AddUserToGroupsCalled, "AddUserToGroups called")
+			assert.Equal(t, tt.wantOutput, output)
+			assert.True(t, tt.groupService.AddUserToGroupsCalled, "AddUserToGroups must run when the args carry additional groups")
 		})
 	}
 }
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -67,6 +67,9 @@ func TestUserHandler_Execute(t *testing.T) {
 		groupService *MockGroupService
 		wantCode     int
 		wantErrMsg   string
+		// wantOutputPart pins the diagnostic a failing row hands the
+		// operator, for the paths that report through output, not err.
+		wantOutputPart string
 	}{
 		{
 			name: "adduser success debian",
@@ -131,8 +134,9 @@ func TestUserHandler_Execute(t *testing.T) {
 				UID: 1001,
 				GID: 1001,
 			},
-			groupService: &MockGroupService{},
-			wantCode:     1,
+			groupService:   &MockGroupService{},
+			wantCode:       1,
+			wantOutputPart: "Username",
 		},
 		{
 			name: "adduser failure",
@@ -180,6 +184,9 @@ func TestUserHandler_Execute(t *testing.T) {
 				assert.NoError(t, err)
 			}
 			assert.Equal(t, tt.wantCode, exitCode)
+			if tt.wantOutputPart != "" {
+				assert.Contains(t, output, tt.wantOutputPart)
+			}
 			if exitCode == 0 && tt.wantErrMsg == "" {
 				assert.NotEmpty(t, output, "a successful command must carry output")
 			}
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -16,11 +16,10 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-
 	"github.com/alpacax/alpamon/v2/pkg/executor/handlers/common"
 	"github.com/alpacax/alpamon/v2/pkg/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 // MockGroupService implements services.GroupService for testing
--- a/pkg/executor/handlers/user/user_test.go
+++ b/pkg/executor/handlers/user/user_test.go
@@ -559,7 +559,6 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 		name         string
 		setupMock    func(*common.MockCommandExecutor)
 		groupService *MockGroupService
-		wantCode     int
 		wantOutput   string
 	}{
 		{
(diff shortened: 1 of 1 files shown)


pkg/executor/handlers/user/user_test.go
```diff
@@ -498,7 +464,7 @@ func TestUserHandler_Validate(t *testing.T) {
 				// are required_unless and also missing. Comment/Groupname are
 				// required unconditionally. Shell is defaulted by the helper.
 			},
-			wantErr: true,
+			wantErrFields: []string{"UID", "GID", "Comment", "HomeDirectory", "Shell", "Groupname"},
```
geunwoonoh, line 481: Warning: the comment three lines above says `Shell is defaulted by the helper`, but if it were defaulted it could not appear in `wantErrFields`. The actual contract is the reverse, and `types.go:52-58` spells out why:

```go
// Shell is intentionally NOT defaulted: defaulting it before
// ValidateStruct would mask an empty `shell` payload ...
```

The comment is a context line this PR did not touch, but turning `wantErr: true` into an explicit field list means a line that used to be merely vague is now contradicted one line away. Someone reading this table later could take the `Shell` entry for a typo and delete it.

Suggested reword of the comment block above:

```go
				// IsServiceAccount=false (default), so UID/GID/HomeDirectory
				// are required_unless and also missing. Comment/Groupname are
				// required unconditionally. Shell carries `validate:"required"`
				// and is deliberately not defaulted (types.go:52), so it fails too.
```

pkg/executor/handlers/user/user_test.go
```diff
@@ -592,7 +560,7 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 			},
 			groupService: &MockGroupService{},
 			wantCode:     0,
-			calledGroups: true,
+			wantOutput:   "User 'testuser' added successfully",
```
geunwoonoh, line 575: Suggestion: this row and `TestUserHandler_Execute/adduser success debian` (line 72) take character-identical args, down to `Groups: []uint64{1002, 1003}`, with the same mock setup, the same platform and the same lookup fakes.

Before this PR both discarded the output, so the overlap did not show. Now that this row pins the output string exactly, the `Execute` row's `assert.NotEmpty(t, output)` is strictly contained in it. Dropping `Groups` from the `Execute` row would make the two tables cover different paths and remove the duplication.

pkg/executor/handlers/user/user_test.go
```diff
@@ -583,7 +551,7 @@ func TestUserHandler_AddUserWithGroups(t *testing.T) {
 		setupMock    func(*common.MockCommandExecutor)
 		groupService *MockGroupService
 		wantCode     int
-		calledGroups bool
+		wantOutput   string
```
geunwoonoh, line 567: Suggestion: `calledGroups` was replaced here on the grounds that both rows shared one value, so the column carried no expectation. `wantCode` is `0` in both rows of this same table and stayed. The same rule points at both, so either it goes as well or `calledGroups` had a reason to stay.

pkg/executor/handlers/user/user_test.go
```diff
@@ -161,14 +173,14 @@ func TestUserHandler_Execute(t *testing.T) {
...
+			} else {
+				assert.NoError(t, err)
 			}
-			if exitCode != tt.wantCode {
-				t.Errorf("Execute() exitCode = %v, want %v", exitCode, tt.wantCode)
-			}
-			if exitCode == 0 && output == "" && !tt.wantErr {
-				t.Error("Execute() returned success but no output")
+			assert.Equal(t, tt.wantCode, exitCode)
+			if exitCode == 0 && tt.wantErrMsg == "" {
```
geunwoonoh, line 204: Suggestion: the `"adduser missing username"` row (line 127) is the only failing row whose diagnostic stays unpinned — it asserts `wantCode: 1` and nothing else. That path exits through `user.go:118-120` as `return 1, err.Error(), nil`, so the string an operator actually sees rides in `output`. Every other failing row now pins its message.

A `wantOutputPart: "Username"` on that row, checked here, would put the whole table under one rule.

pkg/executor/handlers/user/user_test.go
```diff
@@ -258,21 +266,13 @@ func TestUserHandler_AddUser_UidLess(t *testing.T) {
...
-				}
+				assert.Contains(t, joined, want, "flag %q must be present", want)
 			}
 			for _, forbid := range tt.forbidFlags {
-				for _, a := range target.Args {
-					if a == forbid {
-						t.Errorf("flag %q must not appear in args, got: %s", forbid, joined)
-					}
-				}
+				assert.NotContains(t, target.Args, forbid, "args: %s", joined)
```
geunwoonoh: Suggestion: `wantFlags` above checks the joined string while `forbidFlags` checks slice elements, yet this failure message prints `joined`. That invites a reader to treat the two columns as the same kind of check.

This carries over the original nested loop's `a == forbid` semantics exactly, so behavior is unchanged — but a single-argument form such as `--uid=1001` would sail straight through the forbid check. Against the actual args of both rows today the stronger form still passes:

```suggestion
				assert.NotContains(t, joined, forbid, "args: %s", joined)
```

pkg/executor/handlers/user/user_test.go
```diff
@@ -804,9 +744,8 @@ func TestUserHandler_AddUser_Idempotent_ServiceAccountNameExists(t *testing.T) {
 			sawGroupaddDashF = true
 		}
 	}
-	if !sawGroupaddDashF {
-		t.Errorf("`groupadd -f alpacon` must still run for an existing service account; got %+v", mock.GetExecutedCommands())
-	}
+	assert.True(t, sawGroupaddDashF,
```
geunwoonoh: Suggestion: this `groupadd -f` scan is a character-for-character copy of the one in `TestUserHandler_AddUser_UidLess` (lines 254-262):

```go
if c.Name == "/usr/sbin/groupadd" && len(c.Args) >= 2 && c.Args[0] == "-f" && c.Args[1] == "alpacon" {
```

An `InvokedWithArgs(program string, args ...string) bool` next to `Invoked` (`common/testing.go:134`), or even a small helper inside this file, folds both. It also fits the way the rest of the file already leans on `Invoked`.

geunwoonoh (commented):
## PR review: #428 — test(user): unify assertion style on testify

**Stack**: golang · **Changes**: +139 −243 in 1 file

---

### Summary

The fourth PR in the #370 series. It moves the hand-written `t.Fatal*`/`t.Error*` checks in `user_test.go` onto testify while keeping the stop-vs-continue points the original code had already chosen, mapping `t.Fatal*` → `require` and `t.Error*` → `assert` one for one. On top of that it fixes three things: it revives an `err` check that was trapped behind an unreachable branch, it pins the rows that used to accept any error to a message or a field set, and it renames a subtest that collided with the `-run` separator.

No production code changed.

**What I verified directly**

| Check | Result |
|---|---|
| `gofmt -l pkg/executor/handlers/user/` | no output |
| `go vet ./pkg/executor/handlers/user/` | clean |
| `go test ./pkg/executor/handlers/... -race` | `ok` across all 9 packages |
| `wantOutput` strings vs production | match `user.go:309` and `user.go:321` |
| `wantErrMsg` vs production | match `user.go:71` and `user.go:108` |
| the import-grouping claim (22 files) | 21 files measured using the same style; consistent with the repo |

---

### Critical issues (0)

None.

---

### Warnings (1)

**1. `pkg/executor/handlers/user/user_test.go:465` — the comment says the opposite of the assertion right below it**

```go
// IsServiceAccount=false (default), so UID/GID/HomeDirectory
// are required_unless and also missing. Comment/Groupname are
// required unconditionally. Shell is defaulted by the helper.   // <- here
},
wantErrFields: []string{"UID", "GID", "Comment", "HomeDirectory", "Shell", "Groupname"},
```

If `Shell` were defaulted it could not appear in `wantErrFields`. The actual contract is the reverse, and `types.go:52-58` spells out why:

```go
// Shell is intentionally NOT defaulted: defaulting it before
// ValidateStruct would mask an empty `shell` payload ...
```

The comment itself is a context line this PR did not touch. But turning `wantErr: true` into an explicit field list means a line that used to be merely vague is now contradicted one line away. Someone reading this table later could take the `Shell` entry for a typo and delete it. A one-line fix:

```go
// IsServiceAccount=false (default), so UID/GID/HomeDirectory
// are required_unless and also missing. Comment/Groupname are
// required unconditionally. Shell carries `validate:"required"`
// and is deliberately not defaulted (types.go:52), so it fails too.
```

---

### Suggestions (5)

**1. `user_test.go:71-89` ↔ `user_test.go:556-563` — two tables run the same scenario twice**

`TestUserHandler_Execute/adduser success debian` and `TestUserHandler_AddUserWithGroups/add user to groups success` take character-identical args, down to `Groups: []uint64{1002, 1003}`, with the same mock setup, the same platform and the same lookup fakes.

Before this PR both discarded the output, so the overlap did not show. Now that `AddUserWithGroups` pins the output string exactly, the `Execute` row's `assert.NotEmpty(t, output)` is strictly contained in it. Dropping `Groups` from the `Execute` row would make the two tables cover different paths and remove the duplication.

**2. `user_test.go:127-135`, `176-184` — one failing row is the only one whose diagnostic stays unpinned**

The `"adduser missing username"` row asserts `wantCode: 1` and nothing else. That path exits through `user.go:118-120` as `return 1, err.Error(), nil`, so the string an operator actually sees rides in `output`. Every other failing row now pins its message; this one does not. A `wantOutputPart: "Username"` would put the whole table under one rule.

**3. `user_test.go:271-276` — two columns of the same table check different things**

```go
for _, want := range tt.wantFlags {
	assert.Contains(t, joined, want, ...)        // the joined string
}
for _, forbid := range tt.forbidFlags {
	assert.NotContains(t, target.Args, forbid, "args: %s", joined)   // slice elements
}
```

This carries over the original nested loop's `a == forbid` semantics exactly, so behavior is unchanged. But the failure message prints `joined` while the check is element equality, which invites a reader to treat the two columns as the same kind of check. A single-argument form such as `--uid=1001` would sail straight through the forbid check. `assert.NotContains(t, joined, forbid)` is stronger, and against the actual args of both rows today it still passes.

**4. `user_test.go:254-262` ↔ `user_test.go:741-746` — the `groupadd -f` scan exists in two copies**

```go
if c.Name == "/usr/sbin/groupadd" && len(c.Args) >= 2 && c.Args[0] == "-f" && c.Args[1] == "alpacon" {
```

An `InvokedWithArgs(program string, args ...string) bool` next to `Invoked` (`common/testing.go:134`), or even a small helper inside the file, folds both. It also fits the way the rest of the file already leans on `Invoked`.

**5. `user_test.go:554`, `562`, `571` — `wantCode` is a non-varying column too**

This PR replaced `calledGroups` with `wantOutput` on the grounds that both rows shared one value, so the column carried no expectation. `wantCode` is `0` in both rows of that same table and stayed. The same rule points at both, so either it goes as well or `calledGroups` had a reason to stay.

---

### Additional review requested (Go conventions, dead/duplicated code, performance)

**Go conventions** — clean. `gofmt` and `go vet` pass, `validationFields` carries `t.Helper()`, and the table-driven structure holds. This matches the golang stack rules ("Use testify for assertions", "Table-driven tests"). The imports put testify in the same group as the local packages; 21 test files in this repo do the same, so consistency wins here.

**Duplicated / unused code** — suggestions 1 and 4 above are the ones this PR's scope surfaces. One more that this PR did not introduce but is visible nearby: the platform save/restore boilerplate repeats 17 times in this file, 6 times in `group_test.go` and 6 times in `system_test.go`.

```go
originalPlatformLike := utils.PlatformLike
utils.SetPlatformLike(tt.platform)
t.Cleanup(func() { utils.SetPlatformLike(originalPlatformLike) })
```

A single `SetPlatform(t *testing.T, like string)` in `common/testing.go` would drop roughly 90 lines across the three files. That is a different kind of diff from an assertion-style pass, so it belongs in a follow-up.

**Performance** — neither a regression nor an improvement. This is a test-only change. `mock.GetExecutedCommands()`, passed as an assertion message argument, is evaluated on every call, but `common/testing.go:129-131` returns the slice header itself rather than a copy, and testify only runs `fmt.Sprintf` on failure. `assert.ElementsMatch` is O(n²) with n at most 6. The file is 243 lines shorter, which buys a marginally faster compile and nothing more.

**Security** — nothing to flag. No hardcoded credentials, no new input paths.

---

### Highlights

- **Removing the dead branch added a check.** `if !tt.groupService.AddUserToGroupsCalled { assert.NoError }` could never run, since both rows pass additional groups, which left `Execute`'s `err` effectively unchecked. It is now an unconditional `assert.NoError`. Confirming that with `t.Fatal("UNREACHABLE BRANCH")` is a sound way to establish it.
- **The `require`/`assert` mapping preserves the original abort points.** Reading through the diff I found no place where it flipped. Splitting `err != nil || exitCode != 0` into two `require` calls also matches the original meaning.
- **`validationFields` + `ElementsMatch` is the right call.** Pinning the validator's whole multi-line string would restate the library's own wording, and `ErrorContains` on a single representative field would not separate the two rows that both flag `UID`. The field set is what each row is actually about.
- **Dropping the `/` from the subtest name** makes the case selectable on its own with `-run`.
- **The verification section of the PR body is concrete** — six mutations, three reverse-checks on the new assertions, and a diff of the test-name list, all in a reproducible form.

---

**Verdict**: COMMENTED

Nothing here blocks a merge. The one warning is a single comment line and would be good to take in this PR; suggestions 1 and 2 fall out naturally from pushing this PR's own thesis — pin what each row expects — all the way. Suggestions 3, 4 and 5 are fine to defer.

**Checklist**
- [x] Code follows stack conventions (gofmt, vet, table-driven, testify)
- [x] Tests included and passing (`-race`, `ok` across 9 packages)
- [x] Documentation updated if needed (test-only, none required)
- [ ] Fix the comment at `user_test.go:465`

pkg/executor/handlers/user/user_test.go
jisung-02: Fixed in `97c4718a`, with your wording. `types.go:52` is the contract and the comment said the reverse; turning `wantErr: true` into an explicit field list is what put the two one line apart, so it belongs in this PR rather than a follow-up.

pkg/executor/handlers/user/user_test.go
jisung-02: Fixed in `1deafdc1`. `Groups` is off the `Execute` row, which now covers the plain create path — no other row does — and the group-add path belongs to the table that pins its output.

pkg/executor/handlers/user/user_test.go
jisung-02: Fixed in `b3490e79`. `wantCode` is gone for the reason you name.

The expectation it carried is still worth stating, just not as a column both rows fill identically:

```go
assert.Equal(t, 0, exitCode, "the create path exits 0 whether or not the group-add step fails")
```

`b4964b7d` fixes that message. My first draft read "a group-add failure must not change the exit code", which fits only the failure row — the success row breaking would have reported a cause it does not have.

pkg/executor/handlers/user/user_test.go
jisung-02: Fixed in `0c36ed88`. The row carries `wantOutputPart: "Username"`, checked with `assert.Contains` next to the exit-code assertion, so every failing row in the table now states what it expects to be told.

Probed by pointing it at a string the output cannot contain: the row fails with the validator's real message quoted, which confirms the diagnostic reaches `output` as you described.

pkg/executor/handlers/user/user_test.go
jisung-02: Fixed in `affbf18f`, taking your suggestion as written. Both rows pass unchanged, and adding `--shell` to the debian row's `forbidFlags` fails with `should not contain "--shell"`, so the check is live.

pkg/executor/handlers/user/user_test.go
jisung-02: Fixed in `7b535409`, taking the second of your two options — a helper inside this file.

It returns an index rather than a bool. `TestUserHandler_AddUser_UidLess` asserts that `groupadd -f` ran before adduser/useradd, so it needs the position, and a bool cannot express that. Both call sites are in this file, so the helper stays here instead of joining `Invoked` on the shared mock: that mock is used by nine packages, and this would be an exported API with two callers.

```go
func invokedAt(mock *common.MockCommandExecutor, program string, args ...string) int {
	for i, c := range mock.GetExecutedCommands() {
		if c.Name == program && slices.Equal(c.Args, args) {
			return i
		}
	}
	return -1
}
```

One difference from the loops it replaces: it compares the whole argument list, not the leading two. Both call sites run `groupadd -f alpacon` with exactly those arguments, so nothing changes today, and an argument added later will say so instead of passing quietly. Asking it for `-f` alone fails the debian row, which is the behaviour I wanted.

geunwoonoh (approved):
## Re-review: #428 — test(user): unify assertion style on testify

**Stack**: golang · **Changes**: +170 −269 in 1 file
**HEAD**: `b4964b7` (7 commits on top of `1d698b8`, the revision I reviewed)

---

### Summary

Every item from the previous review — one warning and five suggestions — is applied. The commits are split one per finding, and one more (`b4964b7`) is something you found yourself. Production code is still untouched.

I did not take the replies at face value; I broke each new assertion to confirm it actually reports.

### Verification of the applied findings

| Previous finding | Commit | Applied | How I verified it |
|---|---|---|---|
| W1 `Shell` comment contradiction | `97c4718` | verbatim | read the diff; matches `types.go:52-58` |
| S1 duplicated scenario | `1deafdc` | `Groups` removed | a probe confirms the `Execute` table now never reaches the group-add path at all |
| S2 unpinned failing row | `0c36ed8` | `wantOutputPart` added | `"Username"` → `"UsernameX"` fails |
| S3 `forbidFlags` asymmetry | `affbf18` | as suggested, checks `joined` | adding `--shell` to the debian row fails |
| S4 duplicated `groupadd -f` scan | `7b53540` | local `invokedAt` helper | `-f` → `-F` fails; shortening the args fails |
| S5 non-varying `wantCode` | `b3490e7` | column dropped, literal + message | `0` → `1` fails |
| (self-found) message fit only one row | `b4964b7` | reworded | read the diff |

**Build and test, measured**

| Check | Result |
|---|---|
| `gofmt -l pkg/executor/handlers/user/` | no output |
| `go vet ./pkg/executor/handlers/user/` | clean |
| `go test ./pkg/executor/handlers/... -race -shuffle=on` | `ok` across all 9 packages |
| PASS-line list vs `1d698b8` | 48 → 48, **identical** |
| `slices` requires Go 1.21+ | `go.mod` says `1.25.13` |

---

### Critical issues (0) · Warnings (0) · Suggestions (0)

Nothing new to raise. I dug into the two places I was unsure about and both hold up.

**`invokedAt`'s `slices.Equal` tightens the check, and that reads as deliberate.** The original scan was `len(c.Args) >= 2 && c.Args[0] == "-f" && c.Args[1] == "alpacon"` — a prefix match. The helper requires the full argument list to be equal. That is safe because `user.go:161` calls `Run(ctx, "/usr/sbin/groupadd", "-f", data.Groupname)` with exactly two arguments, so the assertion is now pinned to the real call shape rather than loosely matching it. The helper's doc comment says `exactly args`, so the contract is stated where a caller sees it. The commit subject says "fold", which suggests an equivalence, and this is slightly stronger than one — worth a mention, not a change.

**The two indices share a coordinate space.** `useraddIndex` comes from the `executed` loop and `groupaddDashFIndex` from inside `invokedAt`. `common/testing.go:129-131` returns `m.commands` itself rather than a copy, so both walk the same slice in the same order and the `assert.Less` comparison is meaningful.

---

### Highlights

- **The commit split is exact.** Six findings, six commits, each subject saying what it does. Easy to read, easy to revert individually.
- **`b4964b7` was not asked for.** Removing the `wantCode` column left the exit-code message describing only one of the two rows, so a break in the success row would have reported a cause it does not have. Catching that is the kind of thing that only comes from re-reading your own diff.
- **The replies carry their reasoning.** On S4 you explained why you picked the index-returning helper over the `common/testing.go` option — the ordering assertion needs the index. On S3 you reported checking that `--shell` actually fails. Both match what I measured independently.
- **Coverage went up, not down.** Dropping `Groups` from the `Execute` success row left a comment in its place explaining the split, which also stops someone from putting it back. The plain create path with no additional groups is newly covered; no row had it before.

---

**Verdict**: APPROVED

**Checklist**
- [x] Code follows stack conventions (gofmt, vet, table-driven, testify)
- [x] Tests included and passing (`-race -shuffle=on`, 9 packages)
- [x] No security vulnerabilities (test-only; no credentials, no new input paths)
- [x] Documentation updated if needed (none required)
- [x] All 6 previous findings applied, and each verified to actually report

One housekeeping note: the six review threads are still open. Nothing in them is outstanding, so feel free to resolve them.

PR #429, opened 2026-08-21
test: unify assertion style on testify in the remaining primary candidates

jisung-02:
Fifth and last PR for the primary candidate list in #370. Scope is the five files the issue named that were still unconverted: `multipart_stream_test.go`, `shell_test.go`, `migrate_test.go`, `auth_manager_tracker_test.go`, `updater_test.go`. With #421, #422, #427 and #428 merged, every file under the issue's "Candidates that would benefit most" now asserts through testify.

The rule the earlier PRs settled applies unchanged: `require` where a failed check makes the rest of the test meaningless, `assert` where the test can keep checking. No production code changed, and no test was added or removed.

This PR also writes that rule down, which nothing in the repo did before. See [The convention](#the-convention).

## What changed

Twenty-two commits, one idea each. The first five are the conversion, one file per commit. The next six are defects and leftovers found while reading the converted files. Then the convention, and last ten review follow-ups.

| Commit | What it does |
|---|---|
| `11a5fd69` | `multipart_stream_test.go`: every hand-rolled check becomes testify, and five copies of the parse-and-open-a-reader preamble become `requireMultipartReader` |
| `a6ecbd39` | `migrate_test.go`: the same conversion; `os.Stat` existence checks become `DirExists` / `FileExists`, and `Confirm(nil)` gets `assert.NotPanics` instead of a comment |
| `0394e09e` | `auth_manager_tracker_test.go`: the same conversion; lookups whose result the next line dereferences move to `require` |
| `a366f92e` | `updater_test.go`: the same conversion; `!os.IsNotExist(err)` becomes `assert.NoFileExists` |
| `cf5bdcb6` | `shell_test.go`: the same conversion; two hand-rolled found-flag loops become `assert.Contains` over `executedUsers` |
| `f36827a4` | `TestShellHandler_OrStopsOnSuccess` said in a comment that cmd2 must not run, and asserted only that cmd1's output was there. The mock returns `output2` for cmd2, so the run the comment forbids used to pass |
| `7b154f1e` | a no-op write in the pid-reuse test: `LookupPID` returns `TrackerEntry` by value, so the `StartedAt` assignment landed on a copy and the time gap its comment claimed never existed. No assertion read the field |
| `ae52d6a9` | the two recursive multipart tests compared field names against `"name"` literals while their siblings used `multipartFieldName`, and broke the loop on `err == io.EOF` instead of `errors.Is`. The discarded `io.ReadAll` error is now asserted |
| `b9f47e60` | three `os.ReadFile` calls dropped their error into `_`, so a failed read reported as a content mismatch against `""` instead of the real cause |
| `38a55c33` | `createTestArchiveNamed` builds the tar.gz in a `bytes.Buffer` instead of writing a temp file and reading it back, and checks both `Close` calls since they flush. `TestExtractBinary_NotFound` reuses it instead of hand-rolling the same archive. The two `TestDownloadFile_*` cases move next to their siblings |
| `fb5b8cf0` | single-use locals inlined in the `Commands` test |
| `f9c59ee5` | the convention, in `README.md`, `CLAUDE.md` and `.github/copilot-instructions.md` |
| `8a044a3c` | `requireMultipartReader` handed `params["boundary"]` to the reader without checking it, so a content type that parses but carries no boundary failed later as an opaque `boundary is empty`. `require.NotEmpty` moves the failure to the parsing step and names the content type |
| `bcdd02f7` | six prose spellings of "websh" become "Websh", and one "alpamon" becomes "Alpamon", per the proper-noun rule in `CLAUDE.md`. Values, identifiers and the `alpacon-server` repository name stay lowercase |
| `f4a36d14` | three comments used a spaced em-dash, which `CLAUDE.md` forbids. Two become colons, one a comma |
| `aa068009` | the six `assert.NoFileExists` calls become one `assertGone` helper per package that asserts `os.ErrNotExist` from `os.Stat`. The weakening turned out to be demonstrable, not theoretical: see [Testing](#testing) |
| `b2d3d28b` | the three convention copies disagreed, and all three offered `assert.NoFileExists` as the helper to prefer, which `aa068009` had just stopped using |
| `2d17a076` | the two `assertGone` copies become one `internal/testutil.AssertGone` taking `msgAndArgs ...any`. `aa068009` called the two copies the package boundary; the convention it wrote down makes a third copy a matter of time, and `internal/runnertest` is the precedent for a test-only internal package |
| `41fcacd0` | the three extract-binary tests opened with the same `t.TempDir()`, path join and `os.WriteFile`; `writeTestArchive` returns the archive path and the extraction destination beside it |
| `6fa0c5d8` | the pid-reuse test was the only lookup in the file that dropped `ok`, so an entry that stopped being found would have reported `"new-cmd" != ""` and hidden the cause |
| `df6ccd64` | the single-part roundtrip block, `NextPart` through the two names to the sha256 comparison, repeated in three tests and becomes `assertSinglePart` |
| `27714ca0` | nothing in the three convention copies said the other two existed. Each now carries an HTML comment naming all three, which renders as nothing and reaches whoever opens the file |

```
 .github/copilot-instructions.md                    |   2 +
 CLAUDE.md                                          |   3 +
 README.md                                          |  12 +
 internal/testutil/fs.go                            |  23 ++
 .../handlers/file/multipart_stream_test.go         | 256 ++++++--------
 pkg/executor/handlers/shell/shell_test.go          | 284 +++++-----------
 pkg/migrate/migrate_test.go                        | 184 ++++------
 pkg/runner/auth_manager_tracker_test.go            | 231 ++++---------
 pkg/updater/updater_test.go                        | 369 +++++++--------------
 9 files changed, 454 insertions(+), 910 deletions(-)
```

## The counts, and why two files do not map one for one

| File | `main` `t.Fatal*` | `main` `t.Error*` | `require` | `assert` | `t.Fatal` kept |
|---|---|---|---|---|---|
| `shell_test.go` | 20 | 41 | 20 | 41 | 0 |
| `auth_manager_tracker_test.go` | 15 | 35 | 16 | 37 | 0 |
| `multipart_stream_test.go` | 47 | 0 | 21 | 19 | 1 |
| `migrate_test.go` | 40 | 0 | 21 | 18 | 2 |
| `updater_test.go` | 37 | 19 | 26 | 21 | 0 |

`shell_test.go` maps exactly, because its original had already chosen where a failure must stop: every `t.Fatal*` aborts and every `t.Error*` keeps going, so the mapping is mechanical. `auth_manager_tracker_test.go` maps the same way apart from three additions described below.

The other three cannot map that way. `multipart_stream_test.go` and `migrate_test.go` used `t.Fatalf` for every check and `updater_test.go` for most of them, so a literal mapping would make every assertion fatal and lose the distinction the issue asks for. Those files were split by whether the following lines depend on the result: a failed `require` there would have left the next line reading a nil reader or an unwritten file anyway.

Three effects move the totals:

- Helper extraction folds repeats. `requireMultipartReader` covers 5 call sites, `assertSinglePart` 3, `createTestArchiveNamed` 3, `writeTestArchive` 3, `executedUsers` 2, and `testutil.AssertGone` 4 in `migrate_test.go` and 2 in `updater_test.go`. `requireMultipartReader` holds one `require` more than the assertions it replaced, because of the boundary check described below. `AssertGone` turns six `assert` calls into one, `assertSinglePart` turns nine `assert` and six `require` calls into three and two, and `writeTestArchive` turns three `require` calls into one. The counts above are per file, so a helper that moved to `internal/testutil` no longer shows in either.
- `assert.ErrorContains` folds pairs. Several files tested `err == nil` with `t.Error` and then called `err.Error()` on the next line, which would have panicked on the very failure being reported. One call covers both halves, without the hole.
- `||` conditions split into one assertion per half, so a failure names which half broke: 8 in `multipart_stream_test.go`, 3 in `shell_test.go`, 1 each in `migrate_test.go` and `auth_manager_tracker_test.go`.

The three `t.Fatal` calls that stay are all the timeout branch of a `select`, where there is no error value to assert on.

## Where the strength changed on purpose

Seven checks got stronger:

- `os.Stat` existence checks in `migrate_test.go` become `assert.DirExists` and `assert.FileExists`, which also pin the kind.
- The six path-absence checks keep the original `os.ErrNotExist` semantics through `internal/testutil.AssertGone`, now one helper for the whole module. The first pass used `assert.NoFileExists`, which is shorter and weaker; review caught it and [Testing](#testing) shows the case it let through.
- `requireMultipartReader` requires a non-empty `boundary`, which the original inline code never checked either. Without it a parsed content type with no boundary reached `multipart.NewReader` and failed there as `boundary is empty`, naming neither the content type nor the call site.
- `RegisterCommandPID`'s nil-cleanup check was `t.Error`, which reports and then calls that nil closure. `require.NotNil` stops there, and the call itself, whose comment said it must not panic, is now `assert.NotPanics`.
- `Confirm(nil)` was a bare call under a comment saying it must not panic; `assert.NotPanics` states it in code.
- The pid-reuse test read the `LookupPID` result without its `ok`, so a missing entry surfaced as `"new-cmd" != ""`. `require.True` stops the test at the lookup and says what was expected to survive.
- `assert.ErrorContains` closes the `err.Error()`-on-nil hole described above.

One got weaker, deliberately:

- `assert.ErrorIs(t, err, io.EOF)` replaces `err != io.EOF` at two places in `multipart_stream_test.go`. `errors.Is` unwraps, so a wrapped EOF would now pass. `multipart.Reader.NextPart` returns a bare `io.EOF`, and the looser form is the one that survives the stdlib wrapping it later.

## The convention

The issue says which style to use, but the repo did not. On `main` the suite splits 46 files that import testify against 71 that assert by hand; after this PR it is 51 against 66. Closing the primary list does not stop the next test from landing in the legacy style, so the rule is now written in three places, one per reader: `README.md` for people, `CLAUDE.md` for Claude Code, `.github/copilot-instructions.md` for Copilot.

`README.md` had no test instructions at all beyond the Docker distro matrix, so its new Tests section also carries the `go test` invocations and the reason `-p 1` is mandatory.

Three copies are only worth keeping if they agree, and the first pass failed that on both counts: the Copilot copy listed fewer legacy calls than the other two and blurred which one the exception covers, and all three named `assert.NoFileExists` as a helper to prefer. `b2d3d28b` fixes both, and the wording now carries the lesson the review produced: read what a helper accepts, because the shorter spelling is sometimes the weaker check.

The wording says `require` for a failed check that makes the rest meaningless, `assert` for the checks themselves, the specific helper over a hand-rolled comparison, and a bare `t.Fatal` only where there is no error value to assert on. Legacy tests convert while someone is already editing them, not as a separate sweep.

## Testing

No production code changed. `git diff origin/main..HEAD --name-only` is five test files, three documents and `internal/testutil/fs.go`. That last one is a `.go` file rather than a `_test.go` file, but it holds one assertion helper, lives under `internal/` so nothing outside the module can import it, and is imported only from `_test.go` files, which is the same arrangement `internal/runnertest` already uses. No shipped code path reaches it.

The PASS line set of each of the five packages is identical before and after. Names were collected with `go test -v` on `origin/main`'s copy of each file and on this branch, then sorted and diffed:

| Package | PASS | SKIP | `diff` |
|---|---|---|---|
| `pkg/executor/handlers/file` | 75 | 1 | empty |
| `pkg/executor/handlers/shell` | 52 | 0 | empty |
| `pkg/migrate` | 15 | 0 | empty |
| `pkg/runner` | 235 | 0 | empty |
| `pkg/updater` | 34 | 0 | empty |

No test was renamed, so unlike #428 there is no line to show here. No FAIL line appears in any run. The one SKIP is `TestFileHandler_Execute_Rm/refused_on_Windows`, a platform gate that lives in `file_test.go`, which this PR does not touch and which skips the same way on `main`.

`gofmt -l .` is empty and `go vet ./...` is clean. All 46 packages with tests pass under `go test ./... -p 1`, and the five touched packages also pass under `-race`:

```
ok  github.com/alpacax/alpamon/v2/pkg/executor/handlers/file   1.588s
ok  github.com/alpacax/alpamon/v2/pkg/executor/handlers/shell  1.478s
ok  github.com/alpacax/alpamon/v2/pkg/migrate                  14.988s
ok  github.com/alpacax/alpamon/v2/pkg/runner                    38.163s
ok  github.com/alpacax/alpamon/v2/pkg/updater                   1.757s
```

`assert.NoFileExists` was shown to be the weaker check before `aa068009` replaced it, rather than argued about. Creating a directory at `MarkerPath()` after `Confirm()` ran leaves `assert.NoFileExists` silent, while the path-absence helper on the same path in the same run reports:

```
--- FAIL: TestConfirm_RemovesMarkerAndBackup (0.01s)
    Error:      Expected error with "file does not exist" in chain but got nil.
    Messages:   Confirm did not remove the marker
```

The boundary assertion added in `8a044a3c` was seen failing before it was trusted. Forcing `params["boundary"] = ""` inside the helper fails all five call sites with the content type in the message:

```
--- FAIL: TestBuildMultipartStream_Roundtrip (0.00s)
    Error:      Should NOT be empty, but was
    Messages:   ct="multipart/form-data; boundary=d6593740a987c93cac042ceb027d27e7f9c60264f9e2c43287e4fd122d11"
```

The four helpers the review follow-ups introduced or tightened were each seen failing before being trusted, one mutation at a time, reverted after.

`testutil.AssertGone` pointed at a path that is still there:

```
--- FAIL: TestReplaceBinary (0.00s)
    Error:      Expected error with "file does not exist" in chain but got nil.
    Messages:   MUTANT: existing path must fail
```

`assertSinglePart` given the wrong file name, which is the argument the fold made a parameter:

```
--- FAIL: TestBuildMultipartStream_Roundtrip (0.00s)
    Error:      Not equal:
                expected: "f.WRONG"
                actual  : "f.bin"
```

The new `require.True` in the pid-reuse test, with the lookup pointed at a pid that was never registered:

```
--- FAIL: TestAddPIDCommandMapping_OverwritesStaleEntry (0.00s)
    Error:      Should be true
    Messages:   entry should survive the overwrite
```

`writeTestArchive` writing something other than the archive it was handed, which breaks all three of its callers at once and shows the helper is what feeds them:

```
--- FAIL: TestExtractBinary (0.00s)
--- FAIL: TestExtractBinary_WindowsExe (0.00s)
--- FAIL: TestExtractBinary_NotFound (0.00s)
    Error:      Error "failed to open gzip: unexpected EOF" does not contain "not found in archive"
```

The no-op `StartedAt` write in `7b154f1e` was confirmed dead before removal: `LookupPID` returns a value (`pkg/runner/auth_manager.go:787`), the assignment target is that copy, and no assertion in the test reads `StartedAt`. Deleting the block leaves `time` unused in the file, which is the compiler agreeing.

## Remaining

Part of #370, which stays open. The primary candidate list is complete; what remains are the partial candidates the issue lists, where only the plain value-checking tests should move and the concurrency and timing tests stay as they are: `terminal_manager_test.go`, `integration_test.go`, `chunk_writer_test.go`, `process_tree_*_test.go`, `regression_test.go`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,6 +60,7 @@ go test -v ./... -p 1
 go test -v ./pkg/collector/check/realtime/cpu
 ```
 
+<!-- This testing-conventions paragraph is mirrored in CLAUDE.md, README.md and .github/copilot-instructions.md. Edit all three together. -->
 Assert with [testify](https://github.com/stretchr/testify): `require` when a failed check makes the rest of the test meaningless, `assert` for the checks themselves. Prefer the specific helper over a hand-rolled comparison: `require.NoError`, `assert.ErrorContains`, `assert.DirExists`. Read what a helper accepts before reaching for it, because the shorter spelling is sometimes the weaker check. `assert.NoFileExists` is the example: it treats a directory at the path, and any other `Lstat` error, as absence, so a test that must prove a path was removed asserts `os.ErrNotExist` from `os.Stat` instead. Write every new test this way. Many older tests still call `t.Errorf` and `t.Fatalf` directly; convert one while you are already editing it rather than as a separate sweep. A bare `t.Fatal` stays correct where there is no error value to assert on, such as the timeout branch of a `select`.
 
 ### Running locally
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,6 +60,8 @@ go test -v ./... -p 1
 go test -v ./pkg/collector/check/realtime/cpu
 ```
 
+Assert with [testify](https://github.com/stretchr/testify): `require` when a failed check makes the rest of the test meaningless, `assert` for the checks themselves. Prefer the specific helper over a hand-rolled comparison—`require.NoError`, `assert.ErrorContains`, `assert.NoFileExists`. Write every new test this way. Many older tests still call `t.Errorf` and `t.Fatalf` directly; convert one while you are already editing it rather than as a separate sweep. A bare `t.Fatal` stays correct where there is no error value to assert on, such as the timeout branch of a `select`.
+
 ### Running locally
 ```bash
 # Run from source
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,7 +60,7 @@ go test -v ./... -p 1
 go test -v ./pkg/collector/check/realtime/cpu
 ```
 
-Assert with [testify](https://github.com/stretchr/testify): `require` when a failed check makes the rest of the test meaningless, `assert` for the checks themselves. Prefer the specific helper over a hand-rolled comparison—`require.NoError`, `assert.ErrorContains`, `assert.NoFileExists`. Write every new test this way. Many older tests still call `t.Errorf` and `t.Fatalf` directly; convert one while you are already editing it rather than as a separate sweep. A bare `t.Fatal` stays correct where there is no error value to assert on, such as the timeout branch of a `select`.
+Assert with [testify](https://github.com/stretchr/testify): `require` when a failed check makes the rest of the test meaningless, `assert` for the checks themselves. Prefer the specific helper over a hand-rolled comparison: `require.NoError`, `assert.ErrorContains`, `assert.DirExists`. Read what a helper accepts before reaching for it, because the shorter spelling is sometimes the weaker check. `assert.NoFileExists` is the example: it treats a directory at the path, and any other `Lstat` error, as absence, so a test that must prove a path was removed asserts `os.ErrNotExist` from `os.Stat` instead. Write every new test this way. Many older tests still call `t.Errorf` and `t.Fatalf` directly; convert one while you are already editing it rather than as a separate sweep. A bare `t.Fatal` stays correct where there is no error value to assert on, such as the timeout branch of a `select`.
 
 ### Running locally
 ```bash
--- a/pkg/executor/handlers/file/multipart_stream_test.go
+++ b/pkg/executor/handlers/file/multipart_stream_test.go
@@ -61,6 +61,23 @@ func requireMultipartReader(t *testing.T, body io.Reader, ct string) *multipart.
 	return multipart.NewReader(body, params["boundary"])
 }
 
+// assertSinglePart consumes the next part of mr and checks it is the content
+// field for fileName carrying payload. Comparing sha256 digests keeps a
+// megabyte-sized mismatch out of the failure output.
+func assertSinglePart(t *testing.T, mr *multipart.Reader, fileName string, payload []byte) {
+	t.Helper()
+	part, err := mr.NextPart()
+	require.NoError(t, err)
+	assert.Equal(t, multipartFieldContent, part.FormName())
+	assert.Equal(t, fileName, part.FileName())
+
+	got := sha256.New()
+	_, err = io.Copy(got, part)
+	require.NoError(t, err)
+	want := sha256.Sum256(payload)
+	assert.Equal(t, want[:], got.Sum(nil), "payload digest mismatch")
+}
+
 func TestBuildMultipartStream_Roundtrip(t *testing.T) {
 	payload := bytes.Repeat([]byte{0xAB}, 1<<20)
 	src := io.NopCloser(bytes.NewReader(payload))
@@ -69,16 +86,7 @@ func TestBuildMultipartStream_Roundtrip(t *testing.T) {
 	defer func() { _ = body.Close() }()
 
 	mr := requireMultipartReader(t, body, ct)
-	part, err := mr.NextPart()
-	require.NoError(t, err)
-	assert.Equal(t, multipartFieldContent, part.FormName())
-	assert.Equal(t, "f.bin", part.FileName())
-
-	got := sha256.New()
-	_, err = io.Copy(got, part)
-	require.NoError(t, err)
-	want := sha256.Sum256(payload)
-	assert.Equal(t, want[:], got.Sum(nil), "payload digest mismatch")
+	assertSinglePart(t, mr, "f.bin", payload)
 
 	_, err = mr.NextPart()
 	assert.ErrorIs(t, err, io.EOF)
@@ -182,16 +190,7 @@ func TestBuildMultipartStream_SmallPath_Roundtrip(t *testing.T) {
 	defer func() { _ = body.Close() }()
 
 	mr := requireMultipartReader(t, body, ct)
-	part, err := mr.NextPart()
-	require.NoError(t, err)
-	assert.Equal(t, multipartFieldContent, part.FormName())
-	assert.Equal(t, "small.bin", part.FileName())
-
-	got := sha256.New()
-	_, err = io.Copy(got, part)
-	require.NoError(t, err)
-	want := sha256.Sum256(payload)
-	assert.Equal(t, want[:], got.Sum(nil), "payload digest mismatch")
+	assertSinglePart(t, mr, "small.bin", payload)
 
 	_, err = mr.NextPart()
 	assert.ErrorIs(t, err, io.EOF)
@@ -228,16 +227,7 @@ func TestBuildMultipartStream_BufferedPath_Roundtrip(t *testing.T) {
 			require.NoError(t, err)
 			assert.Equal(t, contentLength, n, "wire size must match contentLength")
 
-			part, err := requireMultipartReader(t, &captured, ct).NextPart()
-			require.NoError(t, err)
-			assert.Equal(t, multipartFieldContent, part.FormName())
-			assert.Equal(t, "f.bin", part.FileName())
-
-			got := sha256.New()
-			_, err = io.Copy(got, part)
-			require.NoError(t, err)
-			want := sha256.Sum256(payload)
-			assert.Equal(t, want[:], got.Sum(nil), "payload digest mismatch")
+			assertSinglePart(t, requireMultipartReader(t, &captured, ct), "f.bin", payload)
 		})
 	}
 }
--- a/pkg/executor/handlers/file/multipart_stream_test.go
+++ b/pkg/executor/handlers/file/multipart_stream_test.go
@@ -175,7 +175,7 @@ func TestBuildMultipartStream_EarlyCloseNoLeak(t *testing.T) {
 // path (hint < multipartPipeBufSize, currently 4 MiB) produces a well-formed
 // multipart body with correct payload.
 func TestBuildMultipartStream_SmallPath_Roundtrip(t *testing.T) {
-	payload := bytes.Repeat([]byte{0xCD}, 512<<10) // 512 KiB — well below multipartPipeBufSize (4 MiB)
+	payload := bytes.Repeat([]byte{0xCD}, 512<<10) // 512 KiB, well below multipartPipeBufSize (4 MiB)
 	src := io.NopCloser(bytes.NewReader(payload))
 	body, ct, _, err := buildMultipartStream(src, "small.bin", false, int64(len(payload)))
 	require.NoError(t, err)
--- a/pkg/executor/handlers/file/multipart_stream_test.go
+++ b/pkg/executor/handlers/file/multipart_stream_test.go
@@ -10,6 +10,9 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type errReader struct {
@@ -44,84 +47,72 @@ func (e *errReader) Close() error {
 	return e.closeErr
 }
 
+// requireMultipartReader parses ct as a multipart/form-data content type and
+// returns a reader over body, failing the test when either step does not hold.
+func requireMultipartReader(t *testing.T, body io.Reader, ct string) *multipart.Reader {
+	t.Helper()
+	mt, params, err := mime.ParseMediaType(ct)
+	require.NoError(t, err, "ct=%q", ct)
+	require.Equal(t, "multipart/form-data", mt)
+	return multipart.NewReader(body, params["boundary"])
+}
+
 func TestBuildMultipartStream_Roundtrip(t *testing.T) {
 	payload := bytes.Repeat([]byte{0xAB}, 1<<20)
 	src := io.NopCloser(bytes.NewReader(payload))
 	body, ct, _, err := buildMultipartStream(src, "f.bin", false, -1)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	defer func() { _ = body.Close() }()
-	mt, params, err := mime.ParseMediaType(ct)
-	if err != nil || mt != "multipart/form-data" {
-		t.Fatalf("ct=%q err=%v", ct, err)
-	}
-	mr := multipart.NewReader(body, params["boundary"])
+
+	mr := requireMultipartReader(t, body, ct)
 	part, err := mr.NextPart()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if part.FormName() != "content" || part.FileName() != "f.bin" {
-		t.Fatalf("name=%q file=%q", part.FormName(), part.FileName())
-	}
+	require.NoError(t, err)
+	assert.Equal(t, "content", part.FormName())
+	assert.Equal(t, "f.bin", part.FileName())
+
 	got := sha256.New()
-	if _, err := io.Copy(got, part); err != nil {
-		t.Fatal(err)
-	}
+	_, err = io.Copy(got, part)
+	require.NoError(t, err)
 	want := sha256.Sum256(payload)
-	if !bytes.Equal(got.Sum(nil), want[:]) {
-		t.Fatal("payload digest mismatch")
-	}
-	if _, err := mr.NextPart(); err != io.EOF {
-		t.Fatalf("expected EOF, got %v", err)
-	}
+	assert.Equal(t, want[:], got.Sum(nil), "payload digest mismatch")
+
+	_, err = mr.NextPart()
+	assert.ErrorIs(t, err, io.EOF)
 }
 
 func TestBuildMultipartStream_Recursive(t *testing.T) {
 	src := io.NopCloser(strings.NewReader("zip-data"))
 	body, ct, _, err := buildMultipartStream(src, "tree.zip", true, -1)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 	defer func() { _ = body.Close() }()
-	_, params, _ := mime.ParseMediaType(ct)
-	mr := multipart.NewReader(body, params["boundary"])
+
+	mr := requireMultipartReader(t, body, ct)
 	sawName := false
 	for {
 		part, err := mr.NextPart()
 		if err == io.EOF {
 			break
 		}
-		if err != nil {
-			t.Fatal(err)
-		}
+		require.NoError(t, err)
 		if part.FormName() == "name" {
 			data, _ := io.ReadAll(part)
-			if string(data) != "tree.zip" {
-				t.Fatalf("name=%q", data)
-			}
+			assert.Equal(t, "tree.zip", string(data))
 			sawName = true
 		}
 	}
-	if !sawName {
-		t.Fatal("expected name field for recursive upload")
-	}
+	assert.True(t, sawName, "expected name field for recursive upload")
 }
 
 func TestBuildMultipartStream_SrcErrorPropagates(t *testing.T) {
 	er := &errReader{r: bytes.NewReader(bytes.Repeat([]byte{1}, 1024)), failAt: 256}
 	body, _, _, err := buildMultipartStream(er, "f", false, -1)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
+
 	_, err = io.Copy(io.Discard, body)
-	if err == nil || !strings.Contains(err.Error(), "synthetic") {
-		t.Fatalf("expected synthetic error, got %v", err)
-	}
+	assert.ErrorContains(t, err, "synthetic")
+
 	_ = body.Close()
-	if er.closeCnt == 0 {
-		t.Fatal("src.Close() was not called")
-	}
+	assert.NotZero(t, er.closeCnt, "src.Close() was not called")
 }
 
 // TestBuildMultipartStream_SrcCloseErrorPropagates exercises the demoted-cat
@@ -148,13 +139,11 @@ func TestBuildMultipartStream_SrcCloseErrorPropagates(t *testing.T) {
 				closeErr: errors.New("synthetic-close-fail"),
 			}
 			body, _, _, err := buildMultipartStream(er, "f.bin", false, tc.hint)
-			if err != nil {
-				t.Fatal(err)
-			}
+			require.NoError(t, err)
+
 			_, err = io.Copy(io.Discard, body)
-			if err == nil || !strings.Contains(err.Error(), "synthetic-close-fail") {
-				t.Fatalf("expected src.Close error to propagate, got %v", err)
-			}
+			assert.ErrorContains(t, err, "synthetic-close-fail", "expected src.Close error to propagate")
+
 			_ = body.Close()
 		})
 	}
@@ -166,9 +155,7 @@ func TestBuildMultipartStream_EarlyCloseNoLeak(t *testing.T) {
 	// pool puts and a pipe close remain), so the signal proves it exits.
 	er := &errReader{r: bytes.NewReader(bytes.Repeat([]byte{1}, 4<<20)), failAt: 1 << 30, closed: make(chan struct{})}
 	body, _, _, err := buildMultipartStream(er, "f", false, -1)
-	if err != nil {
-		t.Fatalf("buildMultipartStream: %v", err)
-	}
+	require.NoError(t, err, "buildMultipartStream")
 	buf := make([]byte, 64)
 	_, _ = body.Read(buf)
 	_ = body.Close()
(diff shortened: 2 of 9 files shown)


pkg/executor/handlers/file/multipart_stream_test.go
```diff
@@ -44,84 +47,73 @@ func (e *errReader) Close() error {
...
 
+// requireMultipartReader parses ct as a multipart/form-data content type and
+// returns a reader over body, failing the test when either step does not hold.
+func requireMultipartReader(t *testing.T, body io.Reader, ct string) *multipart.Reader {
+	t.Helper()
+	mt, params, err := mime.ParseMediaType(ct)
+	require.NoError(t, err, "ct=%q", ct)
+	require.Equal(t, "multipart/form-data", mt)
+	return multipart.NewReader(body, params["boundary"])
+}
```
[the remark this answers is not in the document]
jisung-02: Fixed in 8a044a3c. `require.NotEmpty(t, params["boundary"], "ct=%q", ct)` now sits next to the other two checks, so the helper fails at the parsing step and prints the content type that lacked the boundary instead of letting the reader report "boundary is empty" with no context.

Verified the assertion actually reports: forcing `params["boundary"] = ""` inside the helper fails all five call sites with `Should NOT be empty` and `ct="multipart/form-data; boundary=..."` in the message. Reverting it leaves the package green, and the whole suite still passes under `go test ./... -p 1` (46 packages, no failures).

The doc comment said "either step", which stopped counting the checks; it now names all three and records why the boundary one is there.

pkg/migrate/migrate_test.go
```diff
@@ -22,11 +25,19 @@ func setupTempDataDir(t *testing.T) string {
...
-		t.Fatalf("write %s: %v", path, err)
-	}
+	require.NoError(t, os.WriteFile(path, []byte(content), 0600), "write %s", path)
+}
+
+// assertGone fails unless path is absent, which os.Stat reports as
+// os.ErrNotExist. assert.NoFileExists is the shorter spelling but it treats a
+// directory at path, and any other Lstat error, as absence; every caller here
+// is proving that a cleanup step ran, so nothing but a missing path passes.
+func assertGone(t *testing.T, path, msg string) {
```
geunwoonoh: Suggestion: `assertGone` has an identical twin in `pkg/updater/updater_test.go:27`. The package boundary explains the two copies, but the convention this PR writes down asks every package to assert `os.ErrNotExist` from `os.Stat` when a path must be gone, so a third copy is a matter of time: `pkg/utils/fs_test.go`, `pkg/executor/handlers/file/file_io_test.go` and `cmd/alpamon/command/register/recovery_test.go` still use `os.IsNotExist` and are the next candidates. `internal/runnertest` is already a precedent for a test-only internal package, so one copy in `internal/testutil` is worth considering.

Two things to fold in while there: a `msgAndArgs ...any` signature like testify's lets callers format the message, and `os.Lstat` would also catch a dangling symlink (not relevant to the current callers).

pkg/runner/auth_manager_tracker_test.go
```diff
@@ -382,17 +292,9 @@ func TestLookupPID_Missing(t *testing.T) {
 func TestAddPIDCommandMapping_OverwritesStaleEntry(t *testing.T) {
 	am := newTestAuthManager()
 	am.AddPIDCommandMapping(4001, "old-cmd", "alice")
-	// Force a distinguishable time gap without sleeping.
-	if entry, ok := am.LookupPID(4001); ok {
-		entry.StartedAt = time.Now().Add(-time.Hour)
-	}
 	am.AddPIDCommandMapping(4001, "new-cmd", "bob")
 
 	entry, _ := am.LookupPID(4001)
```
geunwoonoh: Suggestion: this is the only lookup in the file that drops `ok`; every other one is guarded by `require.True(t, ok, ...)`. If a regression made the entry disappear, the failure would read `"new-cmd" != ""` and hide the cause. Pre-existing, but 7b154f1 edited this test, so "convert one while you are already editing it" applies.

```suggestion
	entry, ok := am.LookupPID(4001)
	require.True(t, ok, "entry should survive the overwrite")
```

pkg/updater/updater_test.go
```diff
@@ -184,29 +183,19 @@ func TestExtractBinary(t *testing.T) {
 
 	tempDir := t.TempDir()
 	archivePath := filepath.Join(tempDir, "test.tar.gz")
-	if err := os.WriteFile(archivePath, archive, 0644); err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, os.WriteFile(archivePath, archive, 0644))
```
geunwoonoh: Suggestion: this three-line preamble (`tempDir := t.TempDir()` → `archivePath := filepath.Join(tempDir, "test.tar.gz")` → `require.NoError(t, os.WriteFile(...))`) repeats at lines 184-186, 209-211 and 225-227. In the spirit of 38a55c3 reusing `createTestArchiveNamed`, a `writeTestArchive(t, archive []byte) string` helper would fold the three.

pkg/executor/handlers/file/multipart_stream_test.go
```diff
@@ -44,84 +47,77 @@ func (e *errReader) Close() error {
...
-		t.Fatal(err)
-	}
-	if part.FormName() != "content" || part.FileName() != "f.bin" {
-		t.Fatalf("name=%q file=%q", part.FormName(), part.FileName())
-	}
+	require.NoError(t, err)
+	assert.Equal(t, multipartFieldContent, part.FormName())
+	assert.Equal(t, "f.bin", part.FileName())
+
 	got := sha256.New()
```
geunwoonoh, line 74: Suggestion: the single-part roundtrip block (`NextPart` → `FormName`/`FileName` → sha256 → digest comparison) has the same shape here, at line 190 and at line 236. It could fold into `assertSinglePart(t, mr, fileName, payload)` the way `requireMultipartReader` folded the preamble. That widens this PR, so a follow-up is fine.

CLAUDE.md
```diff
@@ -60,6 +60,8 @@ go test -v ./... -p 1
 go test -v ./pkg/collector/check/realtime/cpu
 ```
 
+Assert with [testify](https://github.com/stretchr/testify): `require` when a failed check makes the rest of the test meaningless, `assert` for the checks themselves. Prefer the specific helper over a hand-rolled comparison: `require.NoError`, `assert.ErrorContains`, `assert.DirExists`. Read what a helper accepts before reaching for it, because the shorter spelling is sometimes the weaker check. `assert.NoFileExists` is the example: it treats a directory at the path, and any other `Lstat` error, as absence, so a test that must prove a path was removed asserts `os.ErrNotExist` from `os.Stat` instead. Write every new test this way. Many older tests still call `t.Errorf` and `t.Fatalf` directly; convert one while you are already editing it rather than as a separate sweep. A bare `t.Fatal` stays correct where there is no error value to assert on, such as the timeout branch of a `select`.
```
geunwoonoh, line 64: Suggestion: this paragraph now lives in three places in nearly the same words (here, `README.md:216`, `.github/copilot-instructions.md:32`). That matches the existing pattern, since copilot-instructions already mirrors the writing conventions from CLAUDE.md, so it is acceptable. The drift risk is real though: the three copies disagreed once inside this PR before b2d3d28 aligned them. Whoever edits this paragraph next needs to know to touch all three, and that fact is worth recording somewhere.

geunwoonoh (dismissed):
## Review

**Stack**: golang · **Changes**: +438 -893 in 8 files (five test files, three documents, no production code)

### Summary

This PR converts the last five primary candidates from #370 to testify. It does not stop at the mechanical `t.Fatal*` → `require`, `t.Error*` → `assert` mapping: the three files that used `t.Fatalf` for everything were re-split by whether the next line depends on the result. Six defects found during the conversion are fixed in their own commits, and the assertion convention, which nothing in the repo recorded before, is now written down in README, CLAUDE.md and copilot-instructions.

### Critical issues (0)

None.

### Warnings (0)

None. The one place where the semantics changed on purpose (`err != io.EOF` → `assert.ErrorIs`) is explained in the PR body and the reasoning holds.

### Suggestions (5)

All optional. The inline comments carry the same five points.

1. **`pkg/migrate/migrate_test.go:37`, `pkg/updater/updater_test.go:27`**: `assertGone` exists in two packages with the same body. The package boundary explains it, but the convention this PR writes down asks every package to assert `os.ErrNotExist` from `os.Stat` when a path must be gone, so a third copy is a matter of time (`pkg/utils/fs_test.go`, `pkg/executor/handlers/file/file_io_test.go` and `cmd/alpamon/command/register/recovery_test.go` still use `os.IsNotExist` and are the next candidates). There is already a precedent for a test-only internal package (`internal/runnertest`), so one copy in `internal/testutil` is worth considering. Two things to fold in while there: a `msgAndArgs ...any` signature like testify's lets callers format the message, and `os.Lstat` would also catch a dangling symlink (not relevant to the current callers).

2. **`pkg/runner/auth_manager_tracker_test.go:297`**: `entry, _ := am.LookupPID(4001)` is the only lookup in the file that drops `ok`; every other one is guarded by `require.True(t, ok, ...)`. If a regression made the entry disappear, the failure would read `"new-cmd" != ""` and hide the cause. It is pre-existing, but 7b154f1 edited this test, so the convention ("convert one while you are already editing it") applies.

3. **`pkg/updater/updater_test.go:184-186, 209-211, 225-227`**: the three-line preamble (`tempDir := t.TempDir()` → `archivePath := filepath.Join(tempDir, "test.tar.gz")` → `require.NoError(t, os.WriteFile(...))`) repeats three times. In the spirit of 38a55c3 reusing `createTestArchiveNamed`, a `writeTestArchive(t, archive []byte) string` helper would fold them.

4. **`pkg/executor/handlers/file/multipart_stream_test.go:77, 190, 236`**: the single-part roundtrip block (`NextPart` → `FormName`/`FileName` → sha256 → digest comparison) has the same shape in three tests. It could fold into `assertSinglePart(t, mr, fileName, payload)` the way `requireMultipartReader` folded the preamble, but that widens this PR, so a follow-up is fine.

5. **`CLAUDE.md:63`, `README.md:216`, `.github/copilot-instructions.md:32`**: the convention paragraph lives in three places in nearly the same words. That matches the existing pattern (copilot-instructions already mirrors the writing conventions from CLAUDE.md), so it is acceptable, but the drift risk is real: the three copies disagreed once inside this PR before b2d3d28 aligned them. Whoever edits this paragraph next needs to know to touch all three, and that fact is worth recording somewhere.

### Go conventions

- Helper names follow the testify convention of encoding fatal vs non-fatal in the prefix (`requireMultipartReader` stops the test, `assertGone` continues), and every helper calls `t.Helper()`.
- Import grouping: testify in the same group as the local imports matches what #427 and #428 settled.
- `errors.Is` for EOF, `os.FileMode(0755)` for a typed comparison, and the specific helpers (`assert.Positive`/`NotZero`/`Empty`/`DirExists`/`FileExists`) are used consistently.
- The small tidy-ups (`writeFile(t *testing.T, path, content string)`, `-1 * time.Minute` → `-time.Minute`) are idiomatic.
- testifylint (not part of CI) flags `assert.ErrorContains` at `multipart_stream_test.go:117` and `:150` under its `require-error` rule. That is a deliberate divergence (an error check that is the test's own assertion uses `assert` under the project convention), so it is not a problem here, but if testifylint is ever enabled, `require-error` will need to be disabled or excepted.

### Duplicate and unused code

- Every new helper is used: `requireMultipartReader` 5×, `executedUsers` 2×, `assertGone` 4×+2×, `createTestArchiveNamed` 3×. No unused imports (compile and vet are clean).
- Dead code is properly gone: the no-op assignment to a `TrackerEntry` copy (7b154f1) together with the `time` import it left behind, and the `strings`/`errors` imports the conversion made unused.
- What remains duplicated is suggestions 1, 3, 4 and 5 above.

### Performance

- Improvement: `createTestArchiveNamed` builds into a `bytes.Buffer` instead of a temp file. Each of the three calls loses a `CreateTemp` + write + `ReadFile` + `Remove` round trip, and a failure no longer leaves anything in `/tmp`.
- No regressions. The values handed to `assert.Equal` are 32-byte digests and short strings, so the diff output on failure stays small.
- Local runtimes: file 0.53s, shell 0.39s, migrate 13.9s, updater 0.47s, runner 35.1s. The slow two are dominated by pre-existing `time.Sleep(4 * time.Second)` calls in the watchdog tests, which this PR does not touch. Shortening them would need a clock injection in production code, so that is a separate issue.

### Highlights

- The defects found during the conversion were real: `TestShellHandler_OrStopsOnSuccess` did not forbid the cmd2 run its comment described (f36827a; the mock does return `output2` for cmd2), the `err == nil` reported via `t.Error` followed by `err.Error()` was a nil-panic hole that `ErrorContains` closes in one line, and three `os.ReadFile` errors discarded into `_` are asserted now.
- The weakness of `assert.NoFileExists` was demonstrated rather than argued, restored through `assertGone`, and the lesson went into the convention text. Checked against the testify v1.11.1 source: the description is accurate (any `Lstat` error and a directory both count as absence).
- The require/assert/`t.Fatal` count table in the PR body matches a recount of all five files.
- Seventeen commits, one idea each, made this easy to review.

### Verification

- `gofmt -l pkg/` is empty; `go vet` passes for darwin, linux and windows.
- `go test -count=1 -p 1` is `ok` for all five packages.
- golangci-lint v2.9.0 with the default set CI runs: nothing in the files this PR touches. The extra local findings (`commit_types.go:162 expireDate unused`, `updater_test.go:347 appendAssign`, …) are in code this PR does not change.
- CI: all 25 checks pass and the PR is mergeable; review approval was the only thing outstanding.

**Verdict**: APPROVED

- [x] Documentation updated

pkg/migrate/migrate_test.go
jisung-02: Done in `2d17a076`. The helper moved to `internal/testutil` as `AssertGone`, with the `msgAndArgs ...any` signature you asked for, and both packages now call it. You are right that the package-boundary argument only explains why the second copy appeared, not why it should stay.

Two notes on the rest.

`pkg/utils/fs_test.go` turned out not to need converting: it already asserts `assert.ErrorIs(t, err, os.ErrNotExist, ...)` inline at line 146. So the remaining candidates are `pkg/executor/handlers/file/file_io_test.go:69` and the three `os.IsNotExist` calls in `cmd/alpamon/command/register/recovery_test.go`. Neither file is touched by this branch, and the convention this PR writes down says to convert a legacy test while already editing it rather than as a separate sweep, so they wait for the next PR that opens them.

`os.Lstat` I left alone deliberately. All three copies of the convention say `os.Stat` in so many words, no current caller deals in symlinks, and changing the helper without the documents would put them out of step again, which is exactly what `b2d3d28b` had to repair. Worth doing together, in a change that says why a dangling symlink counts as present.

pkg/runner/auth_manager_tracker_test.go
jisung-02: Applied as suggested in `6fa0c5d8`.

pkg/updater/updater_test.go
jisung-02: Done in `41fcacd0`, as `writeTestArchive`. It returns the extraction destination as well as the archive path, because two of the three callers built that from the same `t.TempDir()` and no caller needs the directory for anything else. That keeps the temporary directory inside the helper instead of handing it back for one more join.

pkg/executor/handlers/file/multipart_stream_test.go
jisung-02: Done here rather than in a follow-up, in `df6ccd64`. All three blocks sit in the one file this branch already rewrote, so folding them does not widen the PR beyond the file it already owns.

The trailing `NextPart` returns `io.EOF` check stays outside `assertSinglePart`: the buffered-path test has no such check, so only two of the three callers would share it.

CLAUDE.md
jisung-02: Recorded in `27714ca0`. Each of the three copies now carries the same HTML comment above the paragraph, naming all three files and saying to edit them together. A comment renders as nothing, so `README.md` keeps reading as documentation for people, while anyone who opens the file to edit that paragraph sees the instruction first.

I put it in all three rather than only in `CLAUDE.md`, because the drift can start from any of them.

geunwoonoh (approved):
## Re-review of `27714ca`

Re-approving after the five follow-up commits. My earlier approval was dismissed by the push, and this replaces it. All five suggestions from that review are addressed, one commit each, and the delta is +66 / −67, so answering the review did not grow the code.

### Suggestions, one by one

| # | Suggestion | Commit | Outcome |
|---|---|---|---|
| 1 | `assertGone` duplicated across two packages | `2d17a07` | Moved to `internal/testutil` as `AssertGone`, with the `msgAndArgs ...any` signature; both packages call it |
| 2 | `entry, _ := am.LookupPID(4001)` drops `ok` | `6fa0c5d` | Applied as suggested |
| 3 | Archive-writing preamble repeated three times | `41fcacd` | `writeTestArchive` returns the archive path and the extraction destination; all three callers use both |
| 4 | Single-part roundtrip block repeated three times | `df6ccd6` | Folded into `assertSinglePart` |
| 5 | Convention paragraph mirrored in three files | `27714ca` | Each copy now carries the same HTML comment naming all three |

### The two declines hold up

Both were checked rather than taken on trust.

- **`os.Lstat` deferred.** All three convention copies say `os.Stat` in so many words, so changing the helper alone would reopen exactly the disagreement `b2d3d28` had to repair. No current caller deals in symlinks. Doing it together with the documents, in a change that says why a dangling symlink counts as present, is the right sequencing.
- **The trailing EOF check stays outside `assertSinglePart`.** Confirmed: it exists in two of the three callers (lines 92 and 196) and not in the buffered-path test, so folding it in would have forced a check on a caller that never had one.

The supporting claims check out too. `pkg/utils/fs_test.go` already asserts `assert.ErrorIs(t, err, os.ErrNotExist, ...)` at line 146, so it is not a conversion candidate, and the remaining ones really are `file_io_test.go:69` plus the three `os.IsNotExist` calls in `recovery_test.go`. Leaving those for the PR that next opens them is what the convention this branch writes down asks for.

### Verification on `27714ca`

- `gofmt -l pkg/ internal/` is empty.
- `go vet ./...` is clean on darwin; the touched packages are also clean under `GOOS=linux` and `GOOS=windows`.
- `go test -count=1 -p 1` is `ok` for all six packages, `internal/testutil` included.
- `go build ./cmd/alpamon` succeeds, and `go mod tidy` leaves `go.mod` and `go.sum` unchanged.
- golangci-lint v2.9.0 with the default set CI runs reports 6 issues across the repo, every one of them in a file this PR does not touch (`plugin_test.go`, `register.go`, `commit_types.go`).
- CI is green: 26 of 26 checks pass.
- The PR body's claim that no shipped code path reaches `internal/testutil` holds: neither `internal/testutil` nor `testify` appears in `go list -deps ./cmd/alpamon`. (`testing` does appear, but it is already there on `origin/main`, so it is unrelated to this change.)

### One optional nit, deliberately not a thread

In `.github/copilot-instructions.md:32` the HTML comment sits at column 0 between two list items, which splits the bullet list into two `<ul>` elements. Rendered through the GitHub markdown API, the first bullet ends up in its own list with a gap before the rest. Indenting the comment by two spaces keeps it one list. It is cosmetic, and that file is read mostly by machines, so it is not worth blocking on. `README.md` and `CLAUDE.md` are unaffected: the comment sits between paragraphs there and is hidden as intended.

Raising it in the body rather than as an inline comment on purpose, since a new unresolved thread would re-block the merge over a cosmetic point.

### Verdict

**APPROVED.** The follow-ups are well-scoped, the two declines are reasoned rather than convenient, and every claim in the replies was verifiable. Resolving my five threads alongside this review.

- [x] Documentation updated

PR #433, opened 2026-08-24
fix: stop one bad entry from failing a WebFTP folder download

### LINKED ISSUE #432: [BUG] WebFTP folder download fails when the folder contains a symlink to a directory
## Description

Downloading a folder via WebFTP fails with `failed to write zip entry: read <path>: is a directory` when the folder contains a symbolic link that points to a directory.

Root cause: in `pkg/utils/archive.go`, `CreateZip` walks the tree with `filepath.Walk`, whose `FileInfo` is Lstat-based. A symlink-to-directory entry therefore fails the `fi.IsDir()` check and is treated as a regular file. `addFileToZip` then calls `os.Open`, which follows the symlink and opens the directory, and `io.Copy` fails with EISDIR. The whole archive operation aborts, so downloading any tree that contains a directory symlink always fails. This layout is common in capistrano-style release directories (e.g. `releases/<ts>/shared -> ../../shared`), where this was observed in production.

Related: symlinks to regular files are currently dereferenced silently and archived as full copies; only directory symlinks abort the archive. A fix should decide symlink policy explicitly — skip symlink entries, store them as symlink entries, or resolve them via `os.Stat` (with cycle protection if resolved).

## Steps to Reproduce

```bash
mkdir -p /tmp/demo/real
echo hi > /tmp/demo/real/file.txt
ln -s real /tmp/demo/link
```

Then download `/tmp/demo` as a folder via WebFTP (triggers the internal `upload` command with a directory path, i.e. `CreateZip` with `recursive=true`).

## Expected Behavior

The folder download succeeds; directory symlinks are handled per a defined policy (skipped, stored as links, or resolved) instead of failing the whole archive.

## Actual Behavior

The internal command fails immediately and no archive is produced. The user-facing folder download fails every time for such folders.

## Error Message

```
upload "['/tmp/demo']" (result: failed to write zip entry: read /tmp/demo/link: is a directory, elapsed: 0.008)
```

jisung-02:
## Background

Downloading a folder through WebFTP fails with exit 1 whenever the folder contains a symlink to a directory (#432). `CreateZip` in `pkg/utils/archive.go` treated every non-directory walk entry as a regular file: `filepath.Walk` reports entries through `Lstat`, so `IsDir()` is false for a link to a directory, but `os.Open` follows the link, and `io.Copy` on the opened directory failed with `EISDIR`, aborting the whole archive. One link anywhere in the tree made the entire download fail.

The same all-or-nothing shape turned out to have three more ways in, all of them found in review: a socket named explicitly, a file the user cannot read, and a `Close` error thrown away. They are fixed here too, since fixing only the symlink path would leave the folder download failing whole for reasons the user can neither see nor avoid.

## Changes

Symlinks met while walking a directory are now stored as zip link entries: the target path as the entry body with `ModeSymlink` set in the external attributes, the layout `zip --symlinks` produces. Standard tools (`unzip`, `bsdtar`, macOS Archive Utility) restore them as real links. Storing a walked link rather than following it also means a self-referencing link cannot recurse and that link's target is not pulled into the archive.

An explicitly listed path is still followed when it is a symlink. `filepath.Walk` lstats its root and would otherwise not descend into a symlinked directory, so the root is resolved with `filepath.EvalSymlinks` first. A directory named this way is archived in full, including whatever part of its target tree lives outside the listed path.

Sockets, FIFOs and device nodes are never archived, wherever they are found. They used to reach `os.Open` like any regular file, which is wrong for all of them: a socket fails the open and aborts the whole archive, the same all-or-nothing failure this PR set out to fix, and opening a FIFO read-only waits for a writer that a download will never provide. None of the three has content an archive can carry. Sockets in particular are ordinary things to find in a home directory: gpg-agent, ssh-agent, `docker.sock`. One met while walking is dropped silently, since the user never picked it out; one named explicitly is reported in the skipped list described below, because a path chosen by hand should not disappear without a word.

Entry headers now carry the source file's mode. `zip.Writer.Create` leaves the header's mode bits unset, so every regular file was written as 0666 while symlinks kept their real mode through `CreateHeader` plus `SetMode`. `Unzip` passes `f.Mode()` straight to `os.OpenFile`, so an executable script inside a downloaded folder came back without its execute bit. Both entry kinds now go through one helper that sets the name, the compression method, the mode and the modification time; regular files stay on Deflate and link entries on Store.

The mode written to the entry is the permission bits plus the link flag, not the whole mode. Handing `SetMode` the source's full mode turned a 04755 file into an entry carrying setuid, and `UnzipReader` would pass that to `os.OpenFile`, which recreates a setuid file on extraction. Two smaller header fixes ride along: `Modified` is set from the source, where every entry used to carry the 1980 DOS epoch, and archiving `/` no longer produces absolute entry names, since `filepath.Base("/")` is `/`.

`zip.Writer.Close` writes the central directory, and both deferred closes used to throw their error away. A write that fails there — a full disk, an NFS mount that went away — leaves an archive with no readable index while `CreateZip` still returned nil, so the user was told the download succeeded and then handed a zip nothing will open, with nothing in the log either. Both closes now report, and an error from `Close` never overwrites one that was already on its way out.

One file the user cannot read no longer ends the download. A home directory holding a single 0600 file belonging to someone else produced no archive at all, and the console showed that file's error in place of the folder that was asked for. `CreateZip` now returns the paths it left out alongside the error, splitting a source that cannot be opened, which costs one entry, from a failure once an entry is already being written, which costs the archive. `archive/zip` cannot take back an entry it has begun writing, so a read that breaks mid-copy has to end the archive rather than leave a silently truncated file inside it. An unreadable subdirectory is reported once rather than once per file it hides, since `filepath.Walk` hands over the directory it could not read and carries on with its siblings.

This follows what the standard tools do. `cp`, `rsync`, `tar` and `zip` were each run against a chmod 000 file under a non-root user: all four copy or archive everything else, name the path they could not read, and exit non-zero — 1, 23, 2 and 18 respectively. None of them splits its behaviour by errno, so neither does this.

`handleUpload` names up to three skipped paths with their causes and collapses the rest into a count, so one unreadable directory cannot fill the console with a wall of paths. When something was skipped the summary drops the file count entirely: that count is every path the user asked for, skipped ones included, and printing it beside the skip list would contradict it. The exit code stays 0 for a partial archive, because `handleUpload` has only 0 and 1 and returning 1 would put back the all-or-nothing appearance this PR removes.

An archive that ended up holding nothing while something was skipped is the other boundary, and it is a failure. `CreateZip` returns an error there, `makeArchive` deletes the empty file, and the command exits 1. Without that floor the reachable case was an ordinary folder download: `parsePaths` only stats the directory, so it succeeds and `recursive` is true, `filepath.Walk` then fails on the root with EACCES, the callback notes it and returns nil, and the user received a 22-byte zip holding nothing but its central directory, with exit 0. The summary named the skipped path, but the web client can only key off the exit code, so it read a completed download. Skipping everything is not a partial success.

## Behavior change

A symlink to a file met while walking used to be dereferenced into a full copy: the download succeeded and the user received a real file carrying the link's name. It is now stored as a link entry instead. Nobody reported this as broken, so it is a policy change rather than a bug fix, and it is what #432 refers to as "silently dereferenced as full copies". Extractors on macOS and Linux recreate the link; Windows Explorer writes a small text file holding the target path. A symlink named explicitly on the command is still followed, so that path is unaffected.

A folder download that used to fail whole on one unreadable or irregular entry now succeeds with a partial archive, and the console message names what is missing and why. The command still exits 0 in that case, so a caller reading only the exit code sees success where it used to see failure. A download where every path was skipped still exits 1, since an archive holding nothing is a failed request rather than a partial one.

`CreateZip` returns two values now. The only caller in this repo is `makeArchive`; a caller that ignores the skipped list is telling the user an archive is complete when it may not be.

Known limitation, not addressed here: `UnzipReader` has no link branch, so re-uploading a downloaded archive with `allow_unzip` turns each link entry back into a plain text file holding the target path. Adding link creation there needs the zip-slip check extended to link targets, which the current entry-name check does not cover.

## Testing

`TestCreateZip_SymlinksStoredAsEntries` builds a directory link, a file link, a self-referencing loop (`loop -> .`), and a broken link (`broken -> nowhere`), then asserts the archive holds exactly one real file plus the four links, each with `ModeSymlink` set and the target path as its body. The broken link pins the change from the old failure, where `os.Open` killed the whole archive with "failed to open". `TestCreateZip_SymlinkedRootIsFollowed` and `TestCreateZip_SymlinkedFileIsFollowed` cover the explicitly listed directory and file, the latter asserting the entry is a plain file carrying the target's content. `TestCreateZip_FileModeIsPreserved` archives a 04755 script and a 0644 file, checks both modes on the entries, checks setuid is gone, and checks the owner-execute bit survives extraction. `TestCreateZip_ModTimeIsPreserved` pins the entry timestamp to the source's.

`TestCreateZip_IrregularEntriesAreSkipped` puts a FIFO and a bound unix socket next to a real file and requires the archive to hold the file alone. `TestCreateZip_ListedIrregularPathIsSkipped` covers a FIFO named alongside a regular file, the only shape a multi-select can produce, and asserts the FIFO comes back in the skipped list rather than failing the archive or vanishing. Both live in a `!windows` file, since `syscall.Mkfifo` does not exist there.

`TestCreateZip_UnreadableEntriesAreSkipped` builds a directory holding a readable file, a 0000 file and a 0000 subdirectory, then asserts the archive holds the readable file alone while both unreadable paths come back carrying `os.ErrPermission` — the subdirectory once, not once per file inside it. It skips under root, which reads a 0000 file whatever its mode says. `TestCreateZip_CloseFailureIsReported` writes to `/dev/full`, which accepts the open and fails every write with ENOSPC; `zip.Writer` buffers, so a small entry survives `io.Copy` and only fails once `Close` flushes it with the central directory. No such device exists on darwin, so that test is build-tagged linux and was run in a container. `TestCreateZip_NothingArchivedIsAnError` builds the folder-download shape the guard exists for, a 0000 directory named on its own, and requires an error carrying `os.ErrPermission` alongside the one skipped path; it sits beside the partial case so both sides of the boundary are pinned. `TestUploadSummary` covers the console message itself: nothing skipped, one path named with its cause, a requested path that was skipped, and a tail that collapses into a count.

Every test that does not care about skipping now goes through a `requireZip` helper that also requires the skipped list to be empty, so an entry cannot silently go missing from a test that was not looking for it.

Every one of these was seen failing first: against the unfixed code, or against a deliberate mutation of the fixed code (`os.Stat` swapped for `os.Lstat`, a branch switched off, `CreateHeader` reverted to `Create`, the skip bookkeeping reverted to `return err`, `uploadSummary` made to ignore its skipped argument, the empty-archive guard inverted to `archived && len(skipped) > 0`). Without the irregular-file branch the socket fails with "operation not supported on socket" and the FIFO test never returns. Full suite: `go test ./... -p 1` exit 0 on darwin, and `pkg/utils` plus `pkg/executor/handlers/file` re-run in a `golang:1.25` container for the linux-only test.

Verified end to end on the dev workspace: an agent built from this branch, registered in a container holding the link layout under `/tmp/demo` plus a FIFO, a bound unix socket, a 04755 script and a file stamped 2021-03-04. The WebFTP folder download completed (`upload` command exit 0 in 487ms, previously exit 1 with EISDIR), the server recorded `demo.zip` with `success: true`, and `zipinfo` on the S3 object shows six entries: the four `lrwxrwxrwx` links, the file at `100644` dated 2021 Mar 4, and the script at `100755` with setuid dropped. Neither the FIFO nor the socket appears. After extraction `cat demo/link/file.txt` prints the file's content through the restored link.

Closes #432

### DIFF
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -85,7 +85,9 @@ func newZipEntry(w *zip.Writer, archiveName string, fi os.FileInfo, method uint1
 		Name:   filepath.ToSlash(archiveName),
 		Method: method,
 	}
-	hdr.SetMode(fi.Mode())
+	// Permissions and the link flag only: setuid, setgid and sticky mean
+	// nothing in a download and Unzip would hand them to os.OpenFile.
+	hdr.SetMode(fi.Mode().Perm() | fi.Mode()&os.ModeSymlink)
 
 	zw, err := w.CreateHeader(hdr)
 	if err != nil {
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -11,9 +11,10 @@ import (
 
 // CreateZip creates a zip archive at destPath containing the specified paths.
 // If recursive is true and a path is a directory, its contents are included recursively.
-// Explicitly listed paths are followed if they are symlinks; symlinks met while
-// walking are stored as link entries (target path, not content), so cycles cannot
-// recurse and out-of-tree targets are not pulled in.
+// A symlink listed explicitly is followed, so its whole target tree is archived
+// even when that tree lives outside the listed path. A symlink met while walking
+// is stored as a link entry (target path, not content), so it cannot recurse into
+// a cycle and its target is not pulled in.
 func CreateZip(destPath string, paths []string, recursive bool) error {
 	f, err := os.Create(destPath)
 	if err != nil {
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -16,15 +16,25 @@ import (
 // is stored as a link entry (target path, not content), so it cannot recurse into
 // a cycle and its target is not pulled in. Sockets, FIFOs and device nodes met
 // while walking are skipped; listed explicitly, they are an error.
-func CreateZip(destPath string, paths []string, recursive bool) error {
+func CreateZip(destPath string, paths []string, recursive bool) (err error) {
 	f, err := os.Create(destPath)
 	if err != nil {
 		return fmt.Errorf("failed to create zip file: %w", err)
 	}
-	defer func() { _ = f.Close() }()
+	defer func() {
+		if cerr := f.Close(); cerr != nil && err == nil {
+			err = fmt.Errorf("failed to close zip file: %w", cerr)
+		}
+	}()
 
 	w := zip.NewWriter(f)
-	defer func() { _ = w.Close() }()
+	// Close writes the central directory, so discarding its error would
+	// report a truncated archive as a successful one.
+	defer func() {
+		if cerr := w.Close(); cerr != nil && err == nil {
+			err = fmt.Errorf("failed to finalize zip file: %w", cerr)
+		}
+	}()
 
 	for _, path := range paths {
 		info, err := os.Stat(path)
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -2,12 +2,29 @@ package utils
 
 import (
 	"archive/zip"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
 )
+
+var errNotRegular = errors.New("not a regular file")
+
+// SkippedEntry is a path left out of the archive, with the reason it was left
+// out. Everything else was archived.
+type SkippedEntry struct {
+	Path   string
+	Reason error
+}
+
+// unreadable separates a source that cannot be opened, which costs one entry,
+// from a failure once the entry is already being written, which costs the whole
+// archive. It carries the cause unchanged so the reported reason stays readable.
+type unreadable struct{ error }
+
+func (u unreadable) Unwrap() error { return u.error }
 
 // CreateZip creates a zip archive at destPath containing the specified paths.
 // If recursive is true and a path is a directory, its contents are included recursively.
@@ -16,10 +33,35 @@ import (
 // is stored as a link entry (target path, not content), so it cannot recurse into
 // a cycle and its target is not pulled in. Sockets, FIFOs and device nodes are
 // never archived, wherever they are found.
-func CreateZip(destPath string, paths []string, recursive bool) (err error) {
+//
+// A listed path that cannot be opened, or that is not a regular file, is left
+// out and reported in skipped, so one bad path does not cost the user the rest
+// of the archive. A failure to write the archive itself returns an error, and so
+// does a read that fails once an entry is already being written.
+func CreateZip(destPath string, paths []string, recursive bool) (skipped []SkippedEntry, err error) {
+	note := func(path string, reason error) {
+		// A PathError repeats the path the entry already carries, so keep
+		// the cause on its own and let the caller pair the two.
+		var perr *os.PathError
+		if errors.As(reason, &perr) {
+			reason = perr.Err
+		}
+		skipped = append(skipped, SkippedEntry{Path: path, Reason: reason})
+	}
+	// skip records a source that could not be opened and reports whether err
+	// was one, so both branches agree on what costs a single entry.
+	skip := func(path string, err error) bool {
+		var u unreadable
+		if !errors.As(err, &u) {
+			return false
+		}
+		note(path, u.error)
+		return true
+	}
+
 	f, err := os.Create(destPath)
 	if err != nil {
-		return fmt.Errorf("failed to create zip file: %w", err)
+		return nil, fmt.Errorf("failed to create zip file: %w", err)
 	}
 	defer func() {
 		if cerr := f.Close(); cerr != nil && err == nil {
@@ -39,7 +81,8 @@ func CreateZip(destPath string, paths []string, recursive bool) (err error) {
 	for _, path := range paths {
 		info, err := os.Stat(path)
 		if err != nil {
-			return fmt.Errorf("failed to stat %s: %w", path, err)
+			note(path, err)
+			continue
 		}
 
 		if info.IsDir() && recursive {
@@ -47,7 +90,8 @@ func CreateZip(destPath string, paths []string, recursive bool) (err error) {
 			// symlinked directory, so resolve the explicitly requested path.
 			root, err := filepath.EvalSymlinks(path)
 			if err != nil {
-				return fmt.Errorf("failed to resolve %s: %w", path, err)
+				note(path, err)
+				continue
 			}
 			base := filepath.Base(path)
 			if base == string(filepath.Separator) {
@@ -56,7 +100,10 @@ func CreateZip(destPath string, paths []string, recursive bool) (err error) {
 			}
 			err = filepath.Walk(root, func(fpath string, fi os.FileInfo, err error) error {
 				if err != nil {
-					return err
+					// Walk hands over the directory it could not read and
+					// carries on with its siblings once this returns nil.
+					note(fpath, err)
+					return nil
 				}
 				if fi.IsDir() {
 					return nil
@@ -66,31 +113,41 @@ func CreateZip(destPath string, paths []string, recursive bool) (err error) {
 					return err
 				}
 				name := filepath.Join(base, relPath)
-				if fi.Mode()&os.ModeSymlink != 0 {
-					return addSymlinkToZip(w, fpath, name, fi)
-				}
-				if !fi.Mode().IsRegular() {
+				switch {
+				case fi.Mode()&os.ModeSymlink != 0:
+					err = addSymlinkToZip(w, fpath, name, fi)
+				case !fi.Mode().IsRegular():
 					// Sockets, FIFOs and device nodes carry nothing to archive,
 					// and opening a FIFO blocks until a writer shows up.
 					return nil
+				default:
+					err = addFileToZip(w, fpath, name, fi)
 				}
-				return addFileToZip(w, fpath, name, fi)
+				if skip(fpath, err) {
+					return nil
+				}
+				return err
 			})
 			if err != nil {
-				return err
+				return skipped, err
 			}
 		} else {
 			if !info.Mode().IsRegular() {
 				// Failing here would cost the user every path listed alongside it.
+				note(path, errNotRegular)
 				continue
 			}
-			if err := addFileToZip(w, path, filepath.Base(path), info); err != nil {
-				return err
+			err := addFileToZip(w, path, filepath.Base(path), info)
+			if skip(path, err) {
+				continue
+			}
+			if err != nil {
+				return skipped, err
 			}
 		}
 	}
 
-	return nil
+	return skipped, nil
 }
 
 // newZipEntry opens an entry carrying the source mode bits, which w.Create
@@ -117,7 +174,7 @@ func newZipEntry(w *zip.Writer, archiveName string, fi os.FileInfo, method uint1
 func addSymlinkToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo) error {
 	target, err := os.Readlink(filePath)
 	if err != nil {
-		return fmt.Errorf("failed to read symlink %s: %w", filePath, err)
+		return unreadable{err}
 	}
 
 	zw, err := newZipEntry(w, archiveName, fi, zip.Store)
@@ -135,7 +192,7 @@ func addSymlinkToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo
 func addFileToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo) error {
 	f, err := os.Open(filePath)
 	if err != nil {
-		return fmt.Errorf("failed to open %s: %w", filePath, err)
+		return unreadable{err}
 	}
 	defer func() { _ = f.Close() }()
 
@@ -145,6 +202,9 @@ func addFileToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo) e
 	}
 
 	if _, err := io.Copy(zw, f); err != nil {
+		// A read failing here has already written a partial entry that
+		// archive/zip cannot take back, so it ends the archive rather than
+		// leaving a silently truncated file in it.
 		return fmt.Errorf("failed to write zip entry: %w", err)
 	}
 
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -14,8 +14,8 @@ import (
 // A symlink listed explicitly is followed, so its whole target tree is archived
 // even when that tree lives outside the listed path. A symlink met while walking
 // is stored as a link entry (target path, not content), so it cannot recurse into
-// a cycle and its target is not pulled in. Sockets, FIFOs and device nodes met
-// while walking are skipped; listed explicitly, they are an error.
+// a cycle and its target is not pulled in. Sockets, FIFOs and device nodes are
+// never archived, wherever they are found.
 func CreateZip(destPath string, paths []string, recursive bool) (err error) {
 	f, err := os.Create(destPath)
 	if err != nil {
@@ -81,7 +81,8 @@ func CreateZip(destPath string, paths []string, recursive bool) (err error) {
 			}
 		} else {
 			if !info.Mode().IsRegular() {
-				return fmt.Errorf("cannot archive %s: unsupported file type %s", path, info.Mode().Type())
+				// Failing here would cost the user every path listed alongside it.
+				continue
 			}
 			if err := addFileToZip(w, path, filepath.Base(path), info); err != nil {
 				return err
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -40,6 +40,10 @@ func CreateZip(destPath string, paths []string, recursive bool) error {
 				return fmt.Errorf("failed to resolve %s: %w", path, err)
 			}
 			base := filepath.Base(path)
+			if base == string(filepath.Separator) {
+				// Base of "/" is "/", which would make every entry name absolute.
+				base = ""
+			}
 			err = filepath.Walk(root, func(fpath string, fi os.FileInfo, err error) error {
 				if err != nil {
 					return err
@@ -82,8 +86,9 @@ func CreateZip(destPath string, paths []string, recursive bool) error {
 // would replace with 0666. Zip archive names use forward slashes (ZIP spec).
 func newZipEntry(w *zip.Writer, archiveName string, fi os.FileInfo, method uint16) (io.Writer, error) {
 	hdr := &zip.FileHeader{
-		Name:   filepath.ToSlash(archiveName),
-		Method: method,
+		Name:     filepath.ToSlash(archiveName),
+		Method:   method,
+		Modified: fi.ModTime(),
 	}
 	// Permissions and the link flag only: setuid, setgid and sticky mean
 	// nothing in a download and Unzip would hand them to os.OpenFile.
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -11,6 +11,9 @@ import (
 
 // CreateZip creates a zip archive at destPath containing the specified paths.
 // If recursive is true and a path is a directory, its contents are included recursively.
+// Explicitly listed paths are followed if they are symlinks; symlinks met while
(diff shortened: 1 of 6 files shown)


pkg/utils/archive_test.go
```diff
@@ -60,6 +62,79 @@ func TestCreateZip_RecursiveDirectory(t *testing.T) {
...
+
+	entries := make(map[string]*zip.File, len(r.File))
+	for _, f := range r.File {
+		entries[f.Name] = f
+	}
+
+	// One real file plus the three links: following any link would add
+	// duplicates or recurse into loop.
+	require.Len(t, entries, 4)
+	require.Contains(t, entries, "demo/real/file.txt")
```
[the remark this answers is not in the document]
jisung-02: Fixed in 247a7352. The test now asserts `len(r.File)` first, then the map length, so an archive that writes the same name twice fails instead of collapsing into a single key.

Verified the new assertion RED: making the walk emit each symlink entry twice fails with "should have 4 item(s), but has 7", while the map-only assertion still passed.

choongbum (approved):

pkg/utils/archive.go
```diff
@@ -28,25 +33,47 @@ func CreateZip(destPath string, paths []string, recursive bool) error {
...
+				}
+				return addFileToZip(w, fpath, name, fi)
 			})
 			if err != nil {
 				return err
 			}
 		} else {
-			if err := addFileToZip(w, path, filepath.Base(path)); err != nil {
+			if !info.Mode().IsRegular() {
+				return fmt.Errorf("cannot archive %s: unsupported file type %s", path, info.Mode().Type())
```
choongbum: Suggestion: consider skipping here as well, the way the walk branch does, rather than failing the whole archive.

This branch is only reachable in the bulk case. `makeArchive` returns the path untouched when `!bulk && !recursive`, so a single non-directory path never reaches `CreateZip` — a lone socket or FIFO is streamed directly and never archived. And a directory always takes the walk branch, since the caller passes `recursive || bulk`. That leaves exactly one way in: the user multi-selected several entries in the WebFTP UI and one of them was irregular.

So "listed explicitly" in practice means "the user selected it along with other things", which makes the two paths disagree about the same items:

```
[file.txt, sock]  (multi-select)  -> cannot archive .../sock: unsupported file type S---------
[file.txt, fifo]  (multi-select)  -> cannot archive .../fifo: unsupported file type p---------
parent directory  (walk)          -> ok, archive holds file.txt; sock and fifo skipped
```

Selecting a folder succeeds and quietly drops the socket; selecting that folder's contents fails the download outright. Same intent, opposite outcome — and per the PR description these are ordinary things to find in a home directory, which is also the place users are most likely to multi-select in.

That is the all-or-nothing shape this PR removes for the walk path, still present on the other one. Skipping in both places would also make the rule stated in the doc comment simpler: irregular entries are never archived, wherever they are found.

Related: `TestCreateZip_IrregularPathIsRejected` calls `CreateZip(dest, []string{fifo}, false)`, which pins a combination the caller cannot produce. If you keep the error, it would be worth exercising the bulk shape (`[]string{file, fifo}, true`) instead, since that is the one users actually hit.

pkg/utils/archive.go
```diff
@@ -28,25 +33,47 @@ func CreateZip(destPath string, paths []string, recursive bool) error {
...
+				return fmt.Errorf("failed to resolve %s: %w", path, err)
+			}
+			base := filepath.Base(path)
+			if base == string(filepath.Separator) {
+				// Base of "/" is "/", which would make every entry name absolute.
+				base = ""
+			}
+			err = filepath.Walk(root, func(fpath string, fi os.FileInfo, err error) error {
 				if err != nil {
 					return err
```
choongbum: Suggestion (follow-up, not for this PR): an unreadable entry still fails the whole archive.

This propagation, plus the `os.Open` in `addFileToZip`, means a single `EACCES` — one file the walker cannot open, or one directory it cannot read — aborts the archive and the download exits 1. That is the same failure mode as the symlink bug in the issue: one bad entry anywhere in the tree costs the user everything.

It is arguably more reachable than the symlink case, too. A root-owned file left in a home directory by an earlier `sudo` is enough, and unlike a symlink there is nothing about it that looks unusual to the person requesting the download.

Clearly outside the scope you drew for this change, and the fix is a real design question rather than a branch — collecting per-entry failures and reporting them alongside a partial archive is a different contract than the current all-or-nothing one, and it needs the server side to have somewhere to put that report. Worth an issue if you agree, so the symlink fix does not read as having closed the general case.

choongbum (dismissed):
Approving. The core fix is right and the reasoning behind each branch holds up under checking. Two suggestions inline, neither a regression — both describe failure modes that predate this PR and survive it, so they are follow-up material rather than blockers.

## What I verified

- **`os.Stat` for listed paths vs `Lstat` inside the walk.** The split is exactly what makes "listed links are followed, walked links are stored" work, and it is easy to get backwards. Resolving only the walk root with `EvalSymlinks` is the right scope — resolving each walked entry would reintroduce the cycle risk the PR set out to remove.
- **Entry names are unchanged for the ordinary case.** For `/a/b` containing `c`, both the old `Rel(Dir(path), fpath)` and the new `Join(Base(path), Rel(root, fpath))` produce `b/c`. The trailing-slash case actually improves: `/a/b/` used to yield `c` (the parent segment was dropped, because `Dir("/a/b/")` is `/a/b`) and now yields `b/c`. The `Base("/") == separator` guard is correct and worth the two lines.
- **Narrowing the entry mode to permissions plus the link flag is the right call.** Bits beyond that have no meaning for a download, and keeping them out is the conservative default. Doing it in the same change that starts preserving mode at all is what stops the two from drifting apart later.
- **Zip-slip is not widened.** `UnzipReader`'s `Rel`-based check rejects `../`-shaped entry names, and the caller sanitizes and absolutizes every path before `CreateZip` sees it, so no relative root reaches the name construction in the first place. The known limitation you documented about link entries on the extract side is called out clearly enough that whoever picks it up will know what has to come with it.
- Tests pass locally, and CI is green across all build targets.

## Nit (pre-existing)

`w.Close()` and `f.Close()` are both deferred with their errors discarded. `zip.Writer.Close` is what writes the central directory, so a failure there produces a truncated archive that `CreateZip` still reports as success. Untouched by this PR, but it is the same "the archive silently is not what the user asked for" family the PR is cleaning up, so it may be worth folding into whichever follow-up you pick up.

pkg/utils/archive.go
jisung-02: Agreed, and fixed in bc4e805d. You are right that "listed explicitly" only ever means "multi-selected alongside other things", so the two branches were disagreeing about the same socket depending on how the user picked it.

The `else` branch now `continue`s instead of returning, and the doc comment states the simpler rule you suggested: sockets, FIFOs and device nodes are never archived, wherever they are found.

`TestCreateZip_IrregularPathIsRejected` is gone, replaced by `TestCreateZip_ListedIrregularPathIsSkipped` on the shape you named, `[]string{file, fifo}` with `recursive: true`. Seen RED first by restoring the `return fmt.Errorf`, which fails with `cannot archive /var/folders/.../fifo: unsupported file type p---------`.

One thing I went further on afterwards. Skipping silently in both branches made a hand-picked socket disappear without a word, which is a smaller version of the same problem: the user selects three things, gets an archive holding two, and nothing says why. So a listed irregular path is now reported in the skipped list that the third suggestion introduced, while one met while walking stays silent since the user never picked it out. `errNotRegular` is the reason it carries.

pkg/utils/archive.go
jisung-02: Taken in this PR rather than deferred, in 5cadb67c. You are right that it is more reachable than the symlink case, and closing #432 while leaving it open would have made the fix read as more general than it is.

On the contract question, I checked what the standard tools do before choosing. `cp`, `rsync`, `tar` and `zip` were each run against a chmod 000 file under a non-root user in a `debian:bookworm` container:

```
cp -r     exit 1     copies a.txt and sub/,  "cannot open 'src/secret.txt' for reading: Permission denied"
rsync -a  exit 23    sends a.txt and sub/,   "send_files failed to open ...: Permission denied (13)"
tar -cf   exit 2     archives src/a.txt and src/sub/b.txt
zip -r    exit 18    archives src/a.txt and src/sub/b.txt
```

All four keep everything they can read, name the path they could not, and exit non-zero. None of them splits its behaviour by errno. So `CreateZip` builds the archive from what it can read and returns the rest as `[]SkippedEntry{Path, Reason}` alongside the error.

The line I ended up drawing is not read-vs-write but before-vs-during an entry. A source that cannot be opened costs one entry; a failure once `CreateHeader` has run costs the archive. `archive/zip` has no way to take back an entry it has begun writing, so treating a mid-`io.Copy` read failure as a skip would leave a silently truncated file inside an archive that reports success — worse than failing. `os.Open` and `os.Readlink` failures are marked with an unwrappable `unreadable` type on the way up so the two cases stay distinguishable; there is a comment at the `io.Copy` site saying why it is deliberately not one of them.

An unreadable directory is reported once rather than once per file it hides: `filepath.Walk` hands the directory itself to the callback with the error and carries on with the siblings once it returns nil.

On the server side having somewhere to put the report — I kept it inside the existing message rather than adding a channel for it. `handleUpload` names up to three skipped paths with their causes and collapses the rest into a count, so one unreadable directory cannot fill the console. When anything was skipped the summary drops the file count entirely, since that count is every path the user requested, skipped ones included, and printing it next to the skip list would contradict it:

```
Successfully uploaded 3 file(s).
Uploaded the archive, skipping 1 path(s): /home/u/secret.txt: permission denied
```

Exit code stays 0. `handleUpload` only has 0 and 1, and 1 would put back exactly the all-or-nothing appearance this removes.

Tests: `TestCreateZip_UnreadableEntriesAreSkipped` covers a 0000 file and a 0000 subdirectory next to a readable file, and `TestUploadSummary` covers the message. Both seen RED first.

pkg/utils/archive.go
```diff
@@ -2,75 +2,209 @@ package utils
...
+			if skip(path, err) {
+				continue
+			}
+			if err != nil {
+				return skipped, err
 			}
 		}
 	}
 
+	return skipped, nil
```
choongbum, line 168: 🟡 When every listed path is skipped, this returns a nil error and an empty archive, and the download reports success.

Measured on this branch:

```
CreateZip(dest, []string{gone1, gone2}, true)
  err     = <nil>
  skipped = 2
  entries = 0
  zip     = 22 bytes   (central directory only)
```

`handleUpload` then takes the `statusCode == http.StatusOK` branch and returns exit 0 with the summary. On `main` the same input exits 1.

The most reachable version is not a vanished path but an ordinary folder download. `parsePaths` stats the directory and succeeds, so `recursive` is true; `filepath.Walk` then fails on the root with `EACCES`, the new callback notes it and returns nil, and the user receives an empty zip with exit 0. The summary line does say `skipping 1 path(s): …: permission denied`, so nothing is hidden — but a caller that keys off the exit code, which is what the web client has to do, reads it as a completed download.

That is the direction of this change overshooting its own boundary: "one bad entry should not cost the user the rest" is right, and "every entry was bad, so the download succeeded" is the same all-or-nothing error in the other direction. It is also the one case where the old behavior was correct.

A guard where the two facts are both in hand closes it — nothing archived plus something skipped is a failure, not a partial success. Worth a test alongside `TestCreateZip_UnreadableEntriesAreSkipped`, since that one proves the partial case works and this is its boundary.

Everything else in the rework looks right to me; this is the only thing keeping the approval.

choongbum (commented):
All three points are addressed, and the reporting side went further than I asked for. Holding approval on one boundary case inline — everything else below is verification, not a request.

## What the rework got right

- **`skipped` reaches the user.** My comment said a partial archive needs somewhere to put the report or it is just a quieter failure, and `uploadSummary` is exactly that. Capping the named paths at three so one unreadable directory cannot flood the console is a nice touch, as is declining to print a success count that would contradict the list beside it.
- **The split between "costs one entry" and "costs the archive" is drawn in the right place.** `unreadable` marks a source that never opened; a read that fails after `newZipEntry` has already emitted a local header ends the archive, because `archive/zip` cannot take that entry back. Getting this backwards would have produced silently truncated downloads, which is worse than the bug being fixed.
- **Reporting the unreadable directory once, not once per file inside it.** Returning nil from the walk callback so `Walk` carries on with siblings is the right shape, and the test pins the count.
- **`Close` errors are now recovered through the named return** without clobbering an earlier failure. The `zip.Writer.Close` case is the one that matters, since that is where the central directory is written.
- Tests cover each branch I raised, including the multi-select shape that the old `TestCreateZip_IrregularPathIsRejected` could not reach. CI is 25/25.

One note on the earlier exchange: my suggestion to exercise `[]string{file, fifo}, true` was about reachability, and `TestCreateZip_ListedIrregularPathIsSkipped` says so in its own comment. Good.

pkg/utils/archive.go
jisung-02: Confirmed, and fixed in c534d3ed. Your reading is exactly right: the change had no floor, and "every entry was bad, so the download succeeded" is the same all-or-nothing error pointed the other way.

Reproduced on the branch before the fix, both shapes:

```
folder download: err=<nil> skipped=1 entries=0 bytes=22
vanished paths:  err=<nil> skipped=2 bytes=22
```

Twenty-two bytes is the central directory alone, matching your measurement.

The guard sits where both facts are in hand, at the end of `CreateZip`:

```go
if !archived && len(skipped) > 0 {
	return skipped, fmt.Errorf("nothing could be archived, skipped %d path(s): %w",
		len(skipped), skipped[0].Reason)
}
```

`archived` is set only after an entry is actually written, so it stays false when the walk root fails with EACCES. `makeArchive` already removes the archive on error, so the empty file does not survive, and `handleUpload` exits 1.

The condition is "nothing archived AND something skipped" rather than "nothing archived", because an empty folder is a legitimate download: nothing was skipped there, so it still archives cleanly and exits 0. A directory holding only sockets and FIFOs also stays a success, since a walked irregular entry is dropped silently by design and never reaches the skipped list.

`TestCreateZip_NothingArchivedIsAnError` pins it, built on the folder-download shape you called out rather than a vanished path: a 0000 directory named on its own, requiring an error that carries `os.ErrPermission` plus the one skipped path. It sits next to `TestCreateZip_UnreadableEntriesAreSkipped`, so the partial case and its boundary are asserted side by side.

Seen RED first, with the condition inverted to `archived && len(skipped) > 0`:

```
--- FAIL: TestCreateZip_NothingArchivedIsAnError
    Error: An error is expected but got nil.
```

That same inversion also turns the partial case into a failure, so `TestCreateZip_UnreadableEntriesAreSkipped` fails alongside it and neither side of the boundary can drift unnoticed.

Full suite green on darwin, and the PR body's Changes and Behavior change sections now describe the exit-code contract on both sides.

choongbum (approved):
Approving. The guard lands where it needed to, and the condition is better than what I suggested.

I asked for "nothing archived is a failure", which taken literally would have made an empty folder download start erroring. `!archived && len(skipped) > 0` avoids that by splitting the two ways an archive ends up empty — nothing was archivable, versus everything that was there failed. Verified on this branch:

```
empty directory       entries=0  skipped=0  err=<nil>
socket-only directory entries=0  skipped=0  err=<nil>
unreadable directory  entries=0  skipped=1  err=permission denied
```

The middle row is the one that shows the line is drawn correctly: a walked socket is dropped without a note, so it never trips the guard, while the same emptiness caused by an `EACCES` does. Wiring `archived` at the two points where an entry is actually written, rather than reading the entry count back off the writer, is also what keeps that distinction available.

The test wraps it up properly too — `assert.ErrorIs(err, os.ErrPermission)` rather than a string match, so it survives the message being reworded.

Local tests pass, CI is 25/25.

One optional polish, nothing blocking: on the error path `makeArchive` returns `nil` for `skipped`, so the detailed list is dropped and the user sees only the count plus the first reason from the wrapped error. That is already more than the old behavior gave, and the full list has nowhere better to go today — worth remembering if the failure ever needs to name paths.

Thanks for the three rounds on this. The end state handles more cases than the issue asked for, and each one is pinned by a test that fails without it.

PR #435, opened 2026-08-25
fix: restore symlinks when unzipping an uploaded archive

### LINKED ISSUE #434: [BUG] Unzip flattens symlink entries into text files
## Background

`UnzipReader` in `pkg/utils/archive.go` has no branch for symlink entries. It reads the entry name, creates a file and copies the body into it, whatever the entry's mode says. A zip link entry carries the link target as its body with `ModeSymlink` set in the external attributes, so extracting one produces a plain file whose content is the target path: a 4-byte text file named `link` holding `real`.

This shows up on the upload path. `pkg/executor/handlers/file/file.go:286` calls `UnzipReader` when `allow_unzip` is set, so downloading a folder and uploading the archive back turns every symlink in it into a text file, with no error and nothing in the logs.

Not a regression from #433. Before that fix, downloading a folder that contained a symlink to a directory failed outright, so there were no link entries to bring back. #433 made the agent produce them; reading them back was never implemented.

## Work

Creating real symlinks and hardening the path checks have to land together. The current zip-slip check validates entry names only, which is sufficient while extraction never creates a link, and stops being sufficient the moment it does.

1. Create the link. When an entry's mode has `ModeSymlink`, read the body as the target and call `os.Symlink` instead of writing a file. Decide what to do when the path already exists, since `os.Symlink` fails rather than replacing.

2. Validate the link target, not just the entry name. `archive.go:161` computes the entry path relative to `destDir` and rejects anything that escapes. A link target is never checked. A link named `a` that points at `../../etc` passes the name check, and a later entry named `a/passwd` also passes it while landing outside `destDir` through the link. Reject absolute targets, resolve relative ones against the directory the link will sit in, and require the result to stay inside `destDir`. Nothing to exploit today, because no link is ever created; this is the constraint that makes step 1 safe to add.

3. Decide how entry order is handled. Nothing constrains the order of entries in an archive, so a link can be created before entries whose path goes through it. Once links exist on disk, extraction has to keep a later entry from being written through one.

4. Decide the Windows policy. Creating a symlink there needs the `SeCreateSymbolicLink` privilege or developer mode. Pick one of: fail the extraction, skip the entry, or keep today's behavior of writing the target as a file, and say so in the doc comment.

5. Tests. A round trip through `CreateZip` and `Unzip` keeps a link as a link; an archive whose link target escapes `destDir` is rejected; the same archive with its entries reordered is still rejected; the Windows path behaves as decided.

## Reproducing

```
mkdir -p /tmp/demo/real && echo hi > /tmp/demo/real/file.txt
ln -s real /tmp/demo/link
```

Download `/tmp/demo` through WebFTP, then upload the resulting `demo.zip` back with `allow_unzip` set. `demo/link` comes back as a regular file containing `real`.

jisung-02:
## Background

A folder downloaded through WebFTP and uploaded back with `allow_unzip` lost every symlink in it. #433 taught `CreateZip` to store a symlink as a zip link entry, the target path as the entry body with `ModeSymlink` in the external attributes, but left the other half undone and said so: "`UnzipReader` has no link branch, so re-uploading a downloaded archive with `allow_unzip` turns each link entry back into a plain text file holding the target path."

`UnzipReader` in `pkg/utils/archive.go` branched only on `IsDir()`. Everything else went to `os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())`, and `os.OpenFile` ignores the type bits in its permission argument, so `ModeSymlink` was discarded and `io.Copy` wrote the target path as file content. A link named `demo/link -> real` came back as a regular 0755 file holding the four bytes `real`.

Opening a link branch opens routes out of the destination directory that the entry-name check does not close. Those are fixed here as separate commits, because shipping the branch without them would be shipping the escape.

## Changes

29 commits, 1484 insertions and 48 deletions across 7 files. Each commit is one idea and each has a test that was seen failing first.

Extraction now mirrors compression. `extractSymlink` and `extractFile` sit next to each other the way `addSymlinkToZip` and `addFileToZip` already do, and a link entry is recreated with `os.Symlink`.

### Containment

A parent component that is itself a link is refused: `os.MkdirAll` follows one, so `mkdirAllInside` walks the path a component at a time and stops at the first link. A component that is a regular file is refused for the same reason, since walking past it only surfaces later as a raw `ENOTDIR`. Both refusals name the entry rather than a link target, because the entry they refuse is usually a plain file and the link is a component of the path under it. The link in the way can come from an earlier entry in the same archive or from the destination directory itself, which on the upload path is the user's own, so neither the entry order nor a clean archive makes that check unnecessary. The destination is resolved once with `filepath.EvalSymlinks`, since comparing against it unresolved reads every extraction under `/tmp` on darwin as an escape.

The link target itself is refused when it is absolute, and when a relative one lands outside the destination. Deciding that by text does not work: `filepath.Join` folds `subdir/parent/..` down to `subdir`, while the kernel follows `subdir/parent` first and ends up wherever that link points. `isValidLinkTarget` walks the target component by component, follows every link already on disk, takes `..` one real level up, and checks at each step that the path is still inside the destination. CodeQL reports the same shape as `go/unsafe-unzip-symlink`, and its help calls out the `filepath.Rel` check by name as the ineffective one.

Two bounds sit on that walk, and they measure different things. `maxLinkResolution` counts the links followed, not the components visited, so a deep tree with no link in it is not mistaken for a chain. `maxLinkWalk` counts components across the whole walk, because every link followed splices a whole target in front of the parts still to walk and the hop bound alone leaves a small archive able to hold a worker for tens of seconds.

Resolution alone stays order-dependent, because a link the archive lists later can turn an earlier one into an escape. Two passes close that. Link entries are held back until every regular entry is out, and `recheckLinks` then walks each one again with the archive fully on disk. It sweeps every link rather than stopping at the first offender, and the caller runs it even when an entry ended the run, so no link is left holding an escape the sweep never looked at. Holding links back also gives Windows what it needs at creation time: `os.Symlink` there picks a directory link or a file link by what the target is when the link is made, and entry order does not promise the target has been extracted yet.

Deferring links opens one more window of its own. Link entries are written in archive order, so an earlier one can replace a directory on the path of a later one between the first pass and the write. `extractSymlink` therefore walks its parent chain again immediately before writing, which refuses a component that has become a link.

An entry whose name folds onto the destination itself is refused, or skipped where it is a directory entry. `filepath.Join` maps `.`, `./`, the empty name and `sub/../` onto the destination, and `isInside` passes them because the relative path is `.`. The destination is the caller's directory and not the archive's, so its mode must not be set from an entry and it must not be replaced by one.

### Modes

Directory entries keep the mode they carry. This branch is what taught `newZipEntry` to store `ModeDir` and `addDirToZip` to write the entry, so the archive carries the real mode; extraction was making every directory 0755, and alpamon runs as root on a normal install, so a private directory came back world-readable after a round trip. The mode lands after every entry rather than at mkdir time, because `zip -r` writes a directory entry before the entries below it and a read-only mode landing first would block them; deepest first, because a parent with no search permission blocks the chmod under it. `chmodDir` opens with `O_NOFOLLOW|O_DIRECTORY` and calls `fchmod` on the handle, so a link that took the directory's name cannot be chmod'd through.

A mode is read out of a header only when its creator says a Unix host wrote it, which is the condition `archive/zip` itself uses. For any other creator `archive/zip` derives the mode from FAT attributes, and a directory entry with no attribute bits set comes out 0666—a directory with no execute bit and so no way in. `AllowUnzip` extracts whatever archive the user uploaded, so that path is reachable.

Setuid, setgid and sticky are dropped on extraction. `newZipEntry` already strips them when writing an archive, and the comment there names this exact hazard: "`Unzip` would hand them to `os.OpenFile`". Only half of it was done. The write is the half that matters, because `writeFileAs` demotes by piping through a `tee` child process while `UnzipReader` runs in the alpamon process itself, which is root on a normal install, so an uploaded archive carrying a 04755 entry landed a root-owned setuid file in the user's directory.

### Concurrency

Every websocket command is submitted to the worker pool at `pkg/runner/client.go:380`, so two unzips into the same tree can run at once. Losing a race to another extraction is not a failure, and four places now say so: `os.Mkdir` losing the create, `createSymlink` losing the final retry to a winner that wrote the same target, `recheckLinks` finding a link another sweep already removed, and `isValidLinkTarget` finding a component swept between its `Lstat` and its `Readlink`. The last one mattered most: the sweep deletes what it refuses, so a link that was always valid could be removed for a race it had no part in.

One race is closed rather than tolerated. `extractFile` clears a link at its path and then opens it, and a second extraction can put one back in between; `createFile` opens with `O_NOFOLLOW` so the race fails the entry instead of writing one entry's content through another entry's link. Windows has no equivalent and that window stays open there, stated at the `noFollow` constant.

A mutex was rejected for all of this: the pool is in-process, but `Unzip` is exported and reachable from more than one process, so a lock would pass an in-process test while leaving the real case broken.

### Console output

Rejection messages quote the values they name, and values that cannot be quoted are refused at the gate. The entry name and the link target are arbitrary bytes out of the archive, and `pkg/executor/handlers/file/file.go:290` returns the message to the console as the command result. `%q` covers the messages this code writes, but an OS error raised past those checks embeds the raw path where `%q` cannot reach it, so an entry name or link target carrying control bytes is rejected before it can raise one.

Every OS error that does escape now names the entry it came from. `mkdirAllInside`, `extractSymlink` and `extractFile` used to hand back the raw error, which read as `remove /private/var/folders/.../out/d: directory not empty`—no archive, no entry name, and an absolute path off the host.

### Other

`UnzipReader` creates the destination directory when it is missing. Resolving it up front had quietly made it a precondition, where every entry used to call `os.MkdirAll` on its own parent. The upload path always hands over a directory that already holds the archive, but the function is exported and the next caller need not.

The link target body is read through an `io.LimitReader`, and a body longer than the limit is rejected instead of silently truncated. Without the length check the read was bounded but the result was used anyway, so an over-long target produced a link to a path the archive never named.

A link target is normalized with `filepath.ToSlash` on the way into the archive and on the way out of it, the way entry names already are. A backslash target archived on Windows arrived on Unix as a single opaque component, which came back a broken link and skipped the per-component checks on the way.

`CreateZip` gives a directory holding nothing an entry of its own. `filepath.Walk` hands over every directory it descends into and `CreateZip` returned from all of them, so a directory reached the archive only through the entries below it. An empty one had none, and every empty directory vanished from a downloaded tree. `isEmptyDir` answers the question the archive actually asks—whether anything in the directory will be archived—so a directory holding only sockets, FIFOs or device nodes keeps its entry too, rather than disappearing along with the special files the walk drops. Directories that carry archivable content still get no entry, so the archive an existing request produces is byte for byte what it was, and a directory entry does not count as content either, so an archive whose every file was skipped still fails rather than reporting a finished download.

`isEmptyDir` also stops at the first entry instead of reading and sorting the whole directory. `filepath.Walk` calls it once per directory, so it sits on the whole download path: 1888560 ns/op against 28096 ns/op on a 5000-entry directory, and about 8 percent off `CreateZip` over 40k files in 20 large directories.

## Behavior change

`UnzipReader` takes `*zip.Reader` instead of `*zip.ReadCloser`. The doc says the caller owns the reader and the function never calls `Close`, so the old signature asked for more than the work needs and shut out a caller holding an archive that was never a file. A caller holding a `*zip.ReadCloser` passes `&rc.Reader` and keeps closing it itself. Both call sites in this repo are updated; this is a breaking change for anyone outside it.

A link entry now produces a real symlink where it used to produce a regular file holding the target path. That is the fix, but it is visible to anyone who was reading those files as text.

An archive whose link target escapes the destination, or whose entry is written through a symlink or a regular file, now fails the extraction with `illegal link target in zip` or `illegal entry in zip`. Such an archive previously extracted without complaint, because the link was never created and the entry that would have gone through it was written as an ordinary file inside the destination.

A directory entry's mode is now applied. A tree downloaded and re-uploaded keeps its directory permissions instead of coming back 0755; an archive from a non-Unix creator still gets the default, since its header carries no mode anyone chose.

A link cycle is kept rather than refused. `x -> x` and mutually referring links resolve for nobody—past 64 hops the kernel has already given up, at 40 on Linux and 32 on darwin—so they cannot carry anything outside, and `CreateZip` archives them off a real disk. Anyone who follows one gets `ELOOP`, exactly as they would from a cycle the host made itself. The component bound still rejects, so an archive built to spin the walk is unaffected.

An extraction into a directory that does not exist creates it again, as it did before the destination was resolved up front.

A downloaded directory tree keeps its empty directories, and its directories that hold only special files. An archive of a tree with neither in it is unchanged.

On a Windows host that will not make links at all, the link entry is dropped and the rest of the archive still extracts. Failing the download there would be the all-or-nothing shape #432 set out to remove, and writing the file with the target as its content is the bug this PR fixes. `syscall.Errno.Is` does not map `ERROR_PRIVILEGE_NOT_HELD` to `os.ErrPermission`, so the check names the Windows error codes directly, in a build-tagged file beside a `!windows` counterpart that reports every failure: Linux returns `EPERM` for a filesystem without link support and for a plain permission denial alike, so there is nothing to key off there.

## Testing

`go test ./... -p 1 -count=1` on the pushed head exits 0 with 46 packages `ok` and no failures. `go test -race -count=1 ./pkg/utils/ ./pkg/executor/handlers/file/` prints `ok ... pkg/utils 3.092s` and `ok ... pkg/executor/handlers/file 2.025s`, and `go test ./pkg/utils/ -race -count=20` over the link and concurrency tests is clean. `gofmt -l pkg/` is empty, `go vet ./...` is clean on darwin, `GOOS=windows go vet ./pkg/utils/ ./pkg/executor/handlers/file/` and `GOOS=linux go build ./...` both pass, and `golangci-lint run ./pkg/utils/` reports `0 issues.`

Every new test was seen failing first, against the unfixed code or a deliberate mutation of the fixed code. The escapes were reproduced before they were closed:

| Reproduction | Before |
|---|---|
| `a -> "."` then `a/c -> "../secret"` | `Unzip` returned nil and left `out/c` resolving outside the destination |
| `escape1`, `escape2`, `subdir/parent -> ".."` | `escape2` survived, resolving to the destination's parent |
| `escape`, `subdir/parent -> ".."`, `bad -> /etc` | `bad` ended the run before the sweep; `escape` survived |
| directory entry named `./` with mode 0777 | destination went from 0750 to 0777, extraction reported success |
| link entry named `.` | the destination directory was replaced by a link |
| `x -> "x"` | `illegal link target in zip: "x" -> "x"`, and the link deleted |
| `sub/` from a FAT-creator header | extracted `drw-rw-rw-`, nothing could be written under it |
| `d` (file) then `d/child.txt` | `open .../out/d/child.txt: not a directory` |
| `src/sub/` holding only a FIFO | 0 zip entries and 0 skips: the directory vanished |
| target of 71 components with no link in it | `illegal link target in zip` on a target that was never malformed |

Containment and restoration have tests on both sides, so a guard cannot be tightened into refusing what it should allow: `TestUnzip_LinkTargetMayGoThroughALinkInsideRoot` restores an `indirect -> alias/file.txt` chain that stays inside, `TestUnzip_DeepLinkTargetWithoutLinksIsAccepted` accepts a target past the hop bound with no link in it, and `TestUnzip_LinkCycleIsKeptButUnresolvable` requires `ELOOP` from the path through a kept cycle rather than an extraction failure. The two escaping-link tables run both entry orderings, since nothing constrains the order entries appear in.

`TestUnzip_ConcurrentExtractionsIntoOneDestination` runs eight extractions of one archive into one directory at once and requires every one to succeed. `TestRecheckLinks_OffenderLeftOnDiskOutranksTheOnesRemoved` drives the sweep directly, because no archive can leave a link's parent read-only before it runs. The two remaining races—`createSymlink`'s final retry and the `Lstat`/`Readlink` window in `isValidLinkTarget`—have no test, because neither can be reproduced deterministically from the outside.

The work was also exercised outside the test binary, against archives written by the real `zip --symlinks` CLI rather than by `CreateZip`, with a build of the merge base alongside for comparison. Fifteen attack archives were contained: neither build wrote anything outside the destination, and the file planted outside as bait was never touched. The concurrency failure was confirmed as one this branch introduced, since the merge-base build reported no error over three rounds of twelve; after the fix, five rounds of twelve all succeeded with the extracted tree identical to the source.

The CodeQL alerts on this branch were read rather than assumed. `go/unsafe-unzip-symlink`, alert 169, was the real one, and CodeQL reports it as fixed: the walk uses `os.Lstat` and `os.Readlink` rather than the `filepath.EvalSymlinks` its help suggests, which is not usable here because `filepath.Join` has already folded the `..` away by the time it would run and a link target need not exist, but the query accepts it.

The CodeQL check on this head is red, with 12 alerts numbered 207 to 218: one `go/zipslip` and eleven `go/path-injection`. Nine of them are sinks already reviewed and dismissed as false positives on this PR, renumbered because this round moved the code, which is the same renumbering that turned alert 165 into 168. The `go/zipslip` one is that same alert again, on the entry loop, and it is also open against `main`.

Three are sinks this round added: `os.Lstat(dir.path)` in `applyDirModes`, and the `os.OpenFile` in `createFile` and in `chmodDir`. The taint is real, since every path in the extraction comes out of the archive, but each call sits behind `isInside`, `mkdirAllInside`, or the link-target walk, and the query does not read a `filepath.Rel` comparison inside a helper as a barrier. The two `OpenFile` sinks open with `O_NOFOLLOW`, so they refuse the final component the query is worried about. None of the twelve is dismissed yet; they are left for review rather than cleared ahead of it.

The Windows path is still unverified. `GOOS=windows go build ./pkg/utils/` and `GOOS=windows go vet` pass, and the windows CI job runs `go test ./... -p 1`, so the platform-independent tests do run there; what remains unverified is behavior no test can reach from CI—whether the entry is actually dropped on a host without `SeCreateSymbolicLink`, and the `createFile` window that `noFollow` cannot close on Windows. Both will need a host with those properties.

Closes #434

### DIFF
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -264,7 +264,7 @@ func UnzipReader(r *zip.ReadCloser, destDir string) error {
 	for _, f := range r.File {
 		fpath := filepath.Join(root, f.Name)
 		if !isInside(root, fpath) {
-			return fmt.Errorf("illegal file path in zip: %s", f.Name)
+			return fmt.Errorf("illegal file path in zip: %q", f.Name)
 		}
 
 		if f.FileInfo().IsDir() {
@@ -308,7 +308,7 @@ func isInside(root, path string) bool {
 func mkdirAllInside(root, dir, name string) error {
 	rel, err := filepath.Rel(root, dir)
 	if err != nil {
-		return fmt.Errorf("illegal file path in zip: %s", name)
+		return fmt.Errorf("illegal file path in zip: %q", name)
 	}
 	if rel == "." {
 		return nil
@@ -331,7 +331,7 @@ func mkdirAllInside(root, dir, name string) error {
 			return err
 		}
 		if fi.Mode()&os.ModeSymlink != 0 {
-			return fmt.Errorf("illegal link target in zip: %s is written through a symlink", name)
+			return fmt.Errorf("illegal link target in zip: %q is written through a symlink", name)
 		}
 	}
 
@@ -355,12 +355,12 @@ func extractSymlink(f *zip.File, root, fpath string) error {
 		return err
 	}
 	if len(body) > maxSymlinkTarget {
-		return fmt.Errorf("illegal link target in zip: %s exceeds %d bytes", f.Name, maxSymlinkTarget)
+		return fmt.Errorf("illegal link target in zip: %q exceeds %d bytes", f.Name, maxSymlinkTarget)
 	}
 
 	target := string(body)
 	if !isValidLinkTarget(root, fpath, target) {
-		return fmt.Errorf("illegal link target in zip: %s -> %s", f.Name, target)
+		return fmt.Errorf("illegal link target in zip: %q -> %q", f.Name, target)
 	}
 
 	// os.Symlink refuses an existing path where a regular entry would have
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -10,6 +10,20 @@ import (
 	"strings"
 )
 
+const (
+	// maxSymlinkTarget is past any platform's path limit, so a longer body is a
+	// malformed entry, not a target worth reading into memory.
+	maxSymlinkTarget = 4096
+
+	// maxLinkResolution bounds the chain isValidLinkTarget will walk, the way
+	// ELOOP bounds the kernel's own resolution.
+	maxLinkResolution = 64
+
+	// symlinkAttempts bounds createSymlink's retry, so a path that another
+	// extraction keeps refilling ends the entry instead of spinning on it.
+	symlinkAttempts = 3
+)
+
 var errNotRegular = errors.New("not a regular file")
 
 // SkippedEntry is a path left out of the archive, with the reason it was left
@@ -25,6 +39,12 @@ type SkippedEntry struct {
 type unreadable struct{ error }
 
 func (u unreadable) Unwrap() error { return u.error }
+
+// extractedLink is a link entry held back until the regular ones are out.
+type extractedLink struct {
+	file *zip.File
+	path string
+}
 
 // CreateZip creates a zip archive at destPath containing the specified paths.
 // If recursive is true and a path is a directory, its contents are included
@@ -225,14 +245,6 @@ func addSymlinkToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo
 	return nil
 }
 
-func isEmptyDir(path string) (bool, error) {
-	entries, err := os.ReadDir(path)
-	if err != nil {
-		return false, err
-	}
-	return len(entries) == 0, nil
-}
-
 // addDirToZip writes a directory entry. The trailing slash is what every other
 // zip reader keys off; the mode bit alone is not enough.
 func addDirToZip(w *zip.Writer, archiveName string, fi os.FileInfo) error {
@@ -260,6 +272,14 @@ func addFileToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo) e
 	}
 
 	return nil
+}
+
+func isEmptyDir(path string) (bool, error) {
+	entries, err := os.ReadDir(path)
+	if err != nil {
+		return false, err
+	}
+	return len(entries) == 0, nil
 }
 
 // Unzip extracts a zip archive to the specified destination directory.
@@ -337,12 +357,6 @@ func UnzipReader(r *zip.ReadCloser, destDir string) error {
 	return recheckLinks(root, links)
 }
 
-// extractedLink is a link entry held back until the regular ones are out.
-type extractedLink struct {
-	file *zip.File
-	path string
-}
-
 // recheckLinks walks every link the extraction wrote once the archive is fully
 // on disk. A target is checked against what exists when the link is written, so
 // a link the archive lists later can still turn an earlier one into an escape.
@@ -408,10 +422,6 @@ func mkdirAllInside(root, dir, name string) error {
 	return nil
 }
 
-// maxSymlinkTarget is past any platform's path limit, so a longer body is a
-// malformed entry, not a target worth reading into memory.
-const maxSymlinkTarget = 4096
-
 // extractSymlink is the mirror of addSymlinkToZip.
 func extractSymlink(f *zip.File, root, fpath string) error {
 	rc, err := f.Open()
@@ -434,8 +444,6 @@ func extractSymlink(f *zip.File, root, fpath string) error {
 
 	return createSymlink(target, fpath)
 }
-
-const symlinkAttempts = 3
 
 // createSymlink puts a link to target at fpath, over whatever is already there.
 // A regular entry gets that from O_TRUNC in one call, and this is the same
@@ -462,10 +470,6 @@ func createSymlink(target, fpath string) error {
 		}
 	}
 }
-
-// maxLinkResolution bounds the chain isValidLinkTarget will walk, the way ELOOP
-// bounds the kernel's own resolution.
-const maxLinkResolution = 64
 
 // isValidLinkTarget reports whether a link at fpath may carry target. It walks
 // the target component by component, following every link already on disk,
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -2,11 +2,13 @@ package utils
 
 import (
 	"archive/zip"
+	"cmp"
 	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 )
 
@@ -46,8 +48,10 @@ type unreadable struct{ error }
 
 func (u unreadable) Unwrap() error { return u.error }
 
-// extractedLink is a link entry held back until the regular ones are out.
-type extractedLink struct {
+// deferredEntry is an entry whose work waits until the regular entries are out:
+// a link, so what it points at exists first, or a directory's mode, so a
+// read-only one cannot block the entries below it.
+type deferredEntry struct {
 	file *zip.File
 	path string
 }
@@ -318,6 +322,9 @@ func Unzip(src, destDir string) error {
 // Nothing may leave destDir: an escaping entry name, a link target that lands
 // outside once every link on the way to it is followed, and an entry written
 // through a link are each refused.
+//
+// A directory entry's mode lands last, deepest first, so a read-only directory
+// cannot block the entries below it.
 func UnzipReader(r *zip.Reader, destDir string) error {
 	// destDir used to appear on its own with the first entry, back when every
 	// entry called os.MkdirAll on its own parent.
@@ -332,7 +339,8 @@ func UnzipReader(r *zip.Reader, destDir string) error {
 		return fmt.Errorf("failed to resolve destination directory: %w", err)
 	}
 
-	var links []extractedLink
+	var links []deferredEntry
+	var dirs []deferredEntry
 	for _, f := range r.File {
 		fpath := filepath.Join(root, f.Name)
 		if !isInside(root, fpath) {
@@ -343,6 +351,7 @@ func UnzipReader(r *zip.Reader, destDir string) error {
 			if err := mkdirAllInside(root, fpath, f.Name); err != nil {
 				return err
 			}
+			dirs = append(dirs, deferredEntry{file: f, path: fpath})
 			continue
 		}
 
@@ -354,7 +363,7 @@ func UnzipReader(r *zip.Reader, destDir string) error {
 		// by what the target is at creation time, and entry order does not
 		// promise the target has been extracted yet.
 		if f.Mode()&os.ModeSymlink != 0 {
-			links = append(links, extractedLink{file: f, path: fpath})
+			links = append(links, deferredEntry{file: f, path: fpath})
 			continue
 		}
 
@@ -382,6 +391,41 @@ func UnzipReader(r *zip.Reader, destDir string) error {
 		return rejected
 	}
 
+	return applyDirModes(dirs)
+}
+
+// applyDirModes lands each directory entry's mode once everything below it is
+// out, because a directory entry from zip -r comes before its children and a
+// read-only mode landing first would block them. Deepest first for the same
+// reason: a parent without search permission blocks the chmod below it. A
+// child's path is longer than its parent's, so longest first is deepest first.
+func applyDirModes(dirs []deferredEntry) error {
+	slices.SortStableFunc(dirs, func(a, b deferredEntry) int {
+		return cmp.Compare(len(b.path), len(a.path))
+	})
+
+	for _, dir := range dirs {
+		// A link entry that took the directory's name has nothing for this
+		// mode to land on, and os.Chmod would follow it to whatever it points
+		// at.
+		fi, err := os.Lstat(dir.path)
+		if err != nil {
+			return fmt.Errorf("failed to set mode of %q: %w", dir.file.Name, err)
+		}
+		if !fi.IsDir() {
+			return fmt.Errorf("illegal entry in zip: %q is no longer a directory", dir.file.Name)
+		}
+		// Perm only, for the reason extractFile gives. Zero is an archive that
+		// carried no mode at all, not a directory nobody may enter.
+		mode := dir.file.Mode().Perm()
+		if mode == 0 {
+			continue
+		}
+		if err := chmodDir(dir.path, mode); err != nil {
+			return fmt.Errorf("failed to set mode of %q: %w", dir.file.Name, err)
+		}
+	}
+
 	return nil
 }
 
@@ -392,7 +436,7 @@ func UnzipReader(r *zip.Reader, destDir string) error {
 // as it was. rejected is the first offender removed and left the first one
 // still on disk. The caller reports left ahead of everything else, since it
 // means destDir still holds an escape.
-func recheckLinks(root string, links []extractedLink) (rejected, left error) {
+func recheckLinks(root string, links []deferredEntry) (rejected, left error) {
 	for _, link := range links {
 		target, err := os.Readlink(link.path)
 		if err != nil {
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -15,8 +15,9 @@ const (
 	// malformed entry, not a target worth reading into memory.
 	maxSymlinkTarget = 4096
 
-	// maxLinkResolution bounds the chain isValidLinkTarget will walk, the way
-	// ELOOP bounds the kernel's own resolution.
+	// maxLinkResolution bounds the links isValidLinkTarget will follow, the
+	// way ELOOP bounds the kernel's own resolution. Components are not
+	// counted: a deep tree with no link in it is not a chain.
 	maxLinkResolution = 64
 
 	// symlinkAttempts bounds createSymlink's retry, so a path that another
@@ -495,11 +496,8 @@ func isValidLinkTarget(root, fpath, target string) bool {
 
 	current := filepath.Dir(fpath)
 	parts := strings.Split(target, "/")
-	for steps := 0; len(parts) > 0; steps++ {
-		if steps > maxLinkResolution {
-			return false
-		}
-
+	hops := 0
+	for len(parts) > 0 {
 		part := parts[0]
 		parts = parts[1:]
 
@@ -511,6 +509,10 @@ func isValidLinkTarget(root, fpath, target string) bool {
 		default:
 			next := filepath.Join(current, part)
 			if fi, err := os.Lstat(next); err == nil && fi.Mode()&os.ModeSymlink != 0 {
+				hops++
+				if hops > maxLinkResolution {
+					return false
+				}
 				link, err := os.Readlink(next)
 				// An absolute link is refused rather than followed: only a link
 				// the extraction did not write can be one, and reasoning about
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -363,35 +363,59 @@ func UnzipReader(r *zip.Reader, destDir string) error {
 		}
 	}
 
+	var failed error
 	for _, link := range links {
 		if err := extractSymlink(link.file, root, link.path); err != nil {
-			return err
+			failed = err
+			break
 		}
 	}
+	// Links already on disk were validated against an archive that was not
+	// fully written yet, so sweep them even when a later entry ended the run.
+	rejected, left := recheckLinks(root, links)
+	switch {
+	case left != nil:
+		return left
+	case failed != nil:
+		return failed
+	case rejected != nil:
+		return rejected
+	}
 
-	return recheckLinks(root, links)
+	return nil
 }
 
 // recheckLinks walks every link the extraction wrote once the archive is fully
 // on disk. A target is checked against what exists when the link is written, so
 // a link the archive lists later can still turn an earlier one into an escape.
-func recheckLinks(root string, links []extractedLink) error {
+// Every link is swept, since the ones after an offender are just as unchecked
+// as it was. rejected is the first offender removed and left the first one
+// still on disk. The caller reports left ahead of everything else, since it
+// means destDir still holds an escape.
+func recheckLinks(root string, links []extractedLink) (rejected, left error) {
 	for _, link := range links {
 		target, err := os.Readlink(link.path)
 		if err != nil {
-			// Dropped on a host that makes no links, so there is none to check.
+			// Never written: dropped on a host that makes no links, or listed
+			// after the entry that ended the extraction. Nothing to check.
 			continue
 		}
 		if isValidLinkTarget(root, link.path, filepath.ToSlash(target)) {
 			continue
 		}
-		if err := os.Remove(link.path); err != nil {
-			return err
+		// Another extraction into the same destination may have swept it first.
+		if err := os.Remove(link.path); err != nil && !os.IsNotExist(err) {
+			if left == nil {
+				left = fmt.Errorf("failed to remove link %q: %w", link.file.Name, err)
+			}
+			continue
 		}
-		return fmt.Errorf("illegal link target in zip: %q -> %q", link.file.Name, target)
+		if rejected == nil {
+			rejected = fmt.Errorf("illegal link target in zip: %q -> %q", link.file.Name, target)
+		}
 	}
 
-	return nil
+	return rejected, left
 }
 
 func isInside(root, path string) bool {
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -399,20 +399,35 @@ func extractSymlink(f *zip.File, root, fpath string) error {
 		return fmt.Errorf("illegal link target in zip: %q -> %q", f.Name, target)
 	}
 
-	// os.Symlink will not overwrite where a regular entry truncates, so clear
-	// the way and keep the two kinds symmetric.
-	if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
-		return err
-	}
+	return createSymlink(target, fpath)
+}
 
-	if err := os.Symlink(target, fpath); err != nil {
-		if symlinkUnsupported(err) {
+const symlinkAttempts = 3
+
+// createSymlink puts a link to target at fpath, over whatever is already there.
+// A regular entry gets that from O_TRUNC in one call, and this is the same
+// thing in two: commands run on a worker pool, so a second extraction can fill
+// the path back in between them, and losing that race is not a failure.
+func createSymlink(target, fpath string) error {
+	for attempt := 0; ; attempt++ {
+		if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
+			return err
+		}
+
+		err := os.Symlink(target, fpath)
+		if err == nil || symlinkUnsupported(err) {
 			return nil
 		}
-		return err
-	}
+		if !os.IsExist(err) || attempt == symlinkAttempts-1 {
+			return err
+		}
 
-	return nil
+		// Whoever won carries the same entry when the target matches, so
+		// there is nothing left to write.
+		if got, rerr := os.Readlink(fpath); rerr == nil && got == target {
+			return nil
+		}
+	}
 }
 
 // maxLinkResolution bounds the chain isValidLinkTarget will walk, the way ELOOP
@@ -475,7 +490,7 @@ func extractFile(f *zip.File, fpath string) error {
 	// os.OpenFile would write through a link sitting at the path, the one
 	// component mkdirAllInside does not cover.
 	if fi, err := os.Lstat(fpath); err == nil && fi.Mode()&os.ModeSymlink != 0 {
-		if err := os.Remove(fpath); err != nil {
+		if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
 			return err
 		}
 	}
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -244,18 +244,14 @@ func Unzip(src, destDir string) error {
 // Used when the caller has pre-validated the same handle to close TOCTOU windows.
 //
 // A link entry—the target path as the body with ModeSymlink set, the layout
-// CreateZip writes—comes back as a real symlink, except on a host that makes
-// no links, where the entry is dropped rather than failing the whole
-// extraction or landing as a file holding the target path.
+// CreateZip writes—comes back as a real symlink, dropped on a Windows host that
+// makes none rather than failing the whole extraction.
 //
-// Nothing may leave destDir: an escaping entry name is rejected, so is a link
-// target that resolves outside, and no entry is written through a link, since
-// one an earlier entry created—or one already sitting in destDir—turns a name
-// that passes into a route out.
+// Nothing may leave destDir: an escaping entry name, a link target resolving
+// outside, and an entry written through a link are each refused.
 func UnzipReader(r *zip.ReadCloser, destDir string) error {
-	// Comparing against destDir unresolved reads an extraction that merely
-	// goes through a link as an escape, which is every extraction under /tmp
-	// on darwin.
+	// Unresolved, every extraction under /tmp on darwin reads as an escape,
+	// since /tmp is itself a link.
 	root, err := filepath.EvalSymlinks(destDir)
 	if err != nil {
 		return fmt.Errorf("failed to resolve destination directory: %w", err)
@@ -293,7 +289,6 @@ func UnzipReader(r *zip.ReadCloser, destDir string) error {
 	return nil
 }
 
-// isInside reports whether path stays within root.
 func isInside(root, path string) bool {
 	rel, err := filepath.Rel(root, filepath.Clean(path))
 	if err != nil || filepath.IsAbs(rel) {
@@ -302,9 +297,8 @@ func isInside(root, path string) bool {
 	return rel != ".." && !strings.HasPrefix(rel, ".."+string(os.PathSeparator))
 }
 
-// mkdirAllInside walks dir one component at a time because os.MkdirAll follows
-// a symlink, which would let a link extracted earlier, or one already sitting
-// in destDir, put the entries below it outside root.
+// mkdirAllInside walks dir one component at a time because os.MkdirAll follows a
+// symlink, which would put the entries below one outside root.
 func mkdirAllInside(root, dir, name string) error {
 	rel, err := filepath.Rel(root, dir)
 	if err != nil {
@@ -319,9 +313,9 @@ func mkdirAllInside(root, dir, name string) error {
 		current = filepath.Join(current, part)
 		fi, err := os.Lstat(current)
 		if os.IsNotExist(err) {
-			// Commands run on a worker pool, so a second extraction may
-			// create the directory first. Losing that race is not a failure,
-			// but the check below still has to see what landed.
+			// Commands run on a worker pool, so another extraction may win
+			// the create. Not a failure, but the check below still needs
+			// to see what landed.
 			if err := os.Mkdir(current, 0755); err != nil && !os.IsExist(err) {
 				return err
 			}
@@ -338,12 +332,11 @@ func mkdirAllInside(root, dir, name string) error {
 	return nil
 }
 
-// maxSymlinkTarget is longer than any platform's path limit, so a body past it
-// is a malformed entry rather than a target worth reading into memory.
+// maxSymlinkTarget is past any platform's path limit, so a longer body is a
+// malformed entry, not a target worth reading into memory.
 const maxSymlinkTarget = 4096
 
-// extractSymlink restores a link entry, refusing a target that leaves root.
-// The mirror of addSymlinkToZip.
+// extractSymlink is the mirror of addSymlinkToZip.
 func extractSymlink(f *zip.File, root, fpath string) error {
 	rc, err := f.Open()
 	if err != nil {
@@ -363,8 +356,8 @@ func extractSymlink(f *zip.File, root, fpath string) error {
 		return fmt.Errorf("illegal link target in zip: %q -> %q", f.Name, target)
 	}
 
-	// os.Symlink refuses an existing path where a regular entry would have
-	// truncated it, so clear the way first and keep the two kinds symmetric.
+	// os.Symlink will not overwrite where a regular entry truncates, so clear
+	// the way and keep the two kinds symmetric.
 	if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
 		return err
 	}
@@ -379,9 +372,9 @@ func extractSymlink(f *zip.File, root, fpath string) error {
 	return nil
 }
 
-// isValidLinkTarget rejects an empty or absolute target, and a relative one
-// that resolves outside root. Zip entries carry slash-separated paths whatever
-// wrote them, so a leading slash is absolute on every platform.
+// isValidLinkTarget counts a leading slash as absolute itself: zip entries carry
+// slash-separated paths whatever wrote them, so filepath.IsAbs alone would let a
+// Unix-absolute target through on Windows.
 func isValidLinkTarget(root, fpath, target string) bool {
 	if target == "" || strings.HasPrefix(target, "/") || filepath.IsAbs(target) {
 		return false
@@ -389,19 +382,19 @@ func isValidLinkTarget(root, fpath, target string) bool {
 	return isInside(root, filepath.Join(filepath.Dir(fpath), filepath.FromSlash(target)))
 }
 
-// extractFile writes a regular entry. The mirror of addFileToZip.
+// extractFile is the mirror of addFileToZip.
 func extractFile(f *zip.File, fpath string) error {
-	// os.OpenFile would write through a link sitting at the path: the last
-	// component, the one mkdirAllInside does not cover.
+	// os.OpenFile would write through a link sitting at the path, the one
+	// component mkdirAllInside does not cover.
 	if fi, err := os.Lstat(fpath); err == nil && fi.Mode()&os.ModeSymlink != 0 {
 		if err := os.Remove(fpath); err != nil {
 			return err
 		}
 	}
 
-	// Perm only: extraction runs in the alpamon process, which is root on a
-	// normal install, so setuid, setgid and sticky off an entry would land on
-	// a root-owned file. newZipEntry drops the same bits on the way in.
+	// Perm only: extraction runs as root on a normal install, so setuid, setgid
+	// or sticky off an entry would land on a root-owned file. newZipEntry drops
+	// the same bits on the way in.
 	outFile, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode().Perm())
 	if err != nil {
 		return err
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -251,6 +251,12 @@ func Unzip(src, destDir string) error {
 // outside once every link on the way to it is followed, and an entry written
 // through a link are each refused.
 func UnzipReader(r *zip.ReadCloser, destDir string) error {
+	// destDir used to appear on its own with the first entry, back when every
+	// entry called os.MkdirAll on its own parent.
+	if err := os.MkdirAll(destDir, 0755); err != nil {
+		return fmt.Errorf("failed to create destination directory: %w", err)
+	}
+
 	// Unresolved, every extraction under /tmp on darwin reads as an escape,
 	// since /tmp is itself a link.
 	root, err := filepath.EvalSymlinks(destDir)
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -484,6 +484,13 @@ func hasControlBytes(s string) bool {
 	return strings.ContainsFunc(s, func(r rune) bool { return r < 0x20 || r == 0x7f })
 }
 
+// createFile opens path for writing, made or truncated, and where the host
+// has O_NOFOLLOW it fails instead of following a link that another extraction
+// into the same destination may have put there since extractFile looked.
+func createFile(path string, mode os.FileMode) (*os.File, error) {
+	return os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC|noFollow, mode)
+}
+
 func isInside(root, path string) bool {
 	rel, err := filepath.Rel(root, filepath.Clean(path))
 	if err != nil || filepath.IsAbs(rel) {
@@ -660,8 +667,9 @@ func isAbsTarget(target string) bool {
 
 // extractFile is the mirror of addFileToZip.
 func extractFile(f *zip.File, fpath string) error {
-	// os.OpenFile would write through a link sitting at the path, the one
-	// component mkdirAllInside does not cover.
+	// A link sitting at the path is the one component mkdirAllInside does not
+	// cover. createFile refuses to write through one, so it is cleared first
+	// rather than followed.
 	if fi, err := os.Lstat(fpath); err == nil && fi.Mode()&os.ModeSymlink != 0 {
 		if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
 			return err
@@ -671,7 +679,7 @@ func extractFile(f *zip.File, fpath string) error {
 	// Perm only: extraction runs as root on a normal install, so setuid, setgid
 	// or sticky off an entry would land on a root-owned file. newZipEntry drops
 	// the same bits on the way in.
-	outFile, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode().Perm())
+	outFile, err := createFile(fpath, f.Mode().Perm())
 	if err != nil {
 		return err
 	}
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -344,6 +344,12 @@ func UnzipReader(r *zip.Reader, destDir string) error {
 	var links []deferredEntry
 	var dirs []deferredEntry
 	for _, f := range r.File {
+		// The rejection below quotes the name, but an OS error raised further
+		// in embeds the raw path where %q cannot reach it, and it travels to
+		// the console as the command result.
+		if hasControlBytes(f.Name) {
+			return fmt.Errorf("illegal file path in zip: %q", f.Name)
+		}
 		fpath := filepath.Join(root, f.Name)
 		if !isInside(root, fpath) {
 			return fmt.Errorf("illegal file path in zip: %q", f.Name)
@@ -473,6 +479,11 @@ func recheckLinks(root string, links []deferredEntry) (rejected, left error) {
 	return rejected, left
 }
 
+// hasControlBytes reports whether s carries a byte a terminal would act on.
+func hasControlBytes(s string) bool {
+	return strings.ContainsFunc(s, func(r rune) bool { return r < 0x20 || r == 0x7f })
+}
+
 func isInside(root, path string) bool {
 	rel, err := filepath.Rel(root, filepath.Clean(path))
 	if err != nil || filepath.IsAbs(rel) {
@@ -539,6 +550,11 @@ func extractSymlink(f *zip.File, root, fpath string) error {
 	}
 
 	target := string(body)
+	// The same route the entry-name check closes: a failure past this point
+	// wraps the OS error, which embeds the raw target where %q cannot reach.
+	if hasControlBytes(target) {
+		return fmt.Errorf("illegal link target in zip: %q -> %q", f.Name, target)
+	}
 	if !isValidLinkTarget(root, fpath, target) {
 		return fmt.Errorf("illegal link target in zip: %q -> %q", f.Name, target)
 	}
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -256,7 +256,10 @@ func addSymlinkToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo
 		return err
 	}
 
-	if _, err := zw.Write([]byte(target)); err != nil {
+	// The separator the reading host expects, the same normalization entry
+	// names get: a backslash target written on Windows must not come back on
+	// Unix as a single opaque component.
+	if _, err := zw.Write([]byte(filepath.ToSlash(target))); err != nil {
 		return fmt.Errorf("failed to write zip entry: %w", err)
 	}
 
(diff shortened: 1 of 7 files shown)


pkg/utils/archive.go
```diff
@@ -242,45 +242,180 @@ func Unzip(src, destDir string) error {
...
 func UnzipReader(r *zip.ReadCloser, destDir string) error {
-	for _, f := range r.File {
-		fpath := filepath.Join(destDir, f.Name)
+	// Comparing against destDir unresolved reads an extraction that merely
+	// goes through a link as an escape, which is every extraction under /tmp
+	// on darwin.
+	root, err := filepath.EvalSymlinks(destDir)
+	if err != nil {
+		return fmt.Errorf("failed to resolve destination directory: %w", err)
+	}
```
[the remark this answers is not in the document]
jisung-02: Confirmed and fixed in 5ef5c66. `UnzipReader` now calls `os.MkdirAll(destDir, 0755)` before resolving it, which puts back what every entry used to do to its own parent.

The shipped caller does not hit this: `pkg/executor/handlers/file/file.go:286` passes `filepath.Dir(args.Path)`, and `writeFileAs` has just written the archive there. The function is exported, though, so the precondition was real for anyone else. `TestUnzip_MissingDestinationDirectoryIsCreated` covers it, and failed before the fix with `failed to resolve destination directory: ... no such file or directory`.

pkg/utils/archive.go
```diff
@@ -242,45 +242,180 @@ func Unzip(src, destDir string) error {
...
+	if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	if err := os.Symlink(target, fpath); err != nil {
+		if symlinkUnsupported(err) {
+			return nil
+		}
+		return err
 	}
```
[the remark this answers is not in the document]
jisung-02: Good catch, and taken: 682d2f5 defers link entries to a second pass, so every regular entry is out before the first `os.Symlink` runs. A link to a directory the same archive carries now finds that directory in place, which is what Windows needs to pick a directory link over a file link.

Not adopting the other option you offered, always dropping link entries on Windows: an archive round-tripped through WebFTP would silently lose its links on a host that can make them, and #432 set out to remove exactly that kind of all-or-nothing behaviour.

Deferring turned out to be worth having on its own. Validating a link target against the filesystem is order-dependent, because a link listed later can turn an earlier one into an escape, so `recheckLinks` walks every extracted link once the archive is fully on disk. A chain that still resolves through a link the same archive writes later is covered by `TestUnzip_LinkTargetThroughAnotherLinkIsRejected`, which runs both orderings.

One limit stated plainly: the Windows behaviour is reasoned, not observed. `GOOS=windows go build ./pkg/utils/` passes, but the windows CI job does not run `go test`.

pkg/utils/archive.go
```diff
@@ -242,45 +242,180 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to resolve destination directory: %w", err)
+	}
 
-		// Zip-slip protection using filepath.Rel (handles root dirs and all platforms)
-		rel, err := filepath.Rel(destDir, filepath.Clean(fpath))
-		if err != nil || filepath.IsAbs(rel) || rel == ".." || strings.HasPrefix(rel, ".."+string(os.PathSeparator)) {
-			return fmt.Errorf("illegal file path in zip: %s", f.Name)
+	for _, f := range r.File {
+		fpath := filepath.Join(root, f.Name)
```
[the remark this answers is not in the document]
jisung-02: True positive, and fixed in 682d2f5.

The query help is right about the shape that was here: `isValidLinkTarget` joined the target to the link's directory and compared with `filepath.Rel`, which is exactly the syntactic check the help calls ineffective. Reproduced it before fixing. An archive carrying `subdir/parent -> ..` and then `escape -> subdir/parent/..` passed, because `filepath.Join` folds `subdir/parent/..` down to `subdir`, and left `escape` pointing one level above the destination. Writing through it was still refused by `mkdirAllInside`, so nothing landed outside, but the link itself should never have been created.

The target is now walked component by component, following every link already on disk and taking `..` one real level up, with an inside-the-destination check at each step. An absolute link is refused rather than followed. Because a link listed later in the archive can turn an earlier one into an escape, link entries are extracted after the regular ones and `recheckLinks` walks each one again with the archive fully on disk. `TestUnzip_LinkTargetThroughAnotherLinkIsRejected` covers both orderings.

The recommendation names `filepath.EvalSymlinks`, which does not fit here: `filepath.Join` has already folded the `..` away by the time it would run, and a link target need not exist. The walk uses `os.Lstat` and `os.Readlink` instead, so this alert may well stay open even though the escape is closed.

pkg/utils/archive.go
```diff
@@ -242,45 +242,180 @@
...
+			continue
 		}
 
-		rc, err := f.Open()
-		if err != nil {
-			_ = outFile.Close()
+		if err := extractFile(f, fpath); err != nil {
 			return err
 		}
+	}
```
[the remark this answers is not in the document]
jisung-02: This alert is not new on this branch. The same rule is already open against `main` as alert 165, on the same loop at `pkg/utils/archive.go:246`; the code moved, so it was re-reported here under a new number.

This branch is the one that adds the validation the rule asks for: the joined path is checked with `isInside`, and `mkdirAllInside` refuses to walk through a symlink component so that a passing name cannot become a route out at extraction time. Leaving it open for now rather than dismissing it, since the query's own example sanitizer is a `strings.Contains(name, "..")` check that this code deliberately does not use.

pkg/utils/archive.go
```diff
@@ -242,45 +242,180 @@
...
+		return fmt.Errorf("illegal file path in zip: %q", name)
+	}
+	if rel == "." {
+		return nil
+	}
+
+	current := root
+	for _, part := range strings.Split(rel, string(os.PathSeparator)) {
+		current = filepath.Join(current, part)
+		fi, err := os.Lstat(current)
```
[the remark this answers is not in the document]
jisung-02: This answer covers all nine `go/path-injection` alerts on this file (170 through 178).

Every one of them lands on an operation the branch already guards. Alerts 170 to 172 are the `os.Lstat`, `os.Mkdir` and `os.Lstat` inside `mkdirAllInside`, which walks the path one component at a time precisely so that no component is a symlink and every prefix stays inside the destination. Alerts 173 to 175 are in `extractSymlink`, after `isInside` on the entry name and after the link-target walk. Alerts 176 to 178 are in `extractFile`, after the same entry-name check, with the final component removed first if a link is sitting there.

The taint reaching them is real, the escape is not: the guard is a `filepath.Rel` comparison inside a helper, which the query does not appear to recognise as a barrier. Not suppressing them in this PR. This repo does use `// lgtm[go/path-injection]` with a written reason elsewhere, for example `pkg/executor/handlers/file/file.go:292`, so that route is open if the alerts survive the fix in 682d2f5.

pkg/utils/archive.go
```diff
@@ -242,45 +295,260 @@ func Unzip(src, destDir string) error {
...
+		}
+
+		if err := extractFile(f, fpath); err != nil {
+			return err
+		}
+	}
+
+	for _, link := range links {
+		if err := extractSymlink(link.file, root, link.path); err != nil {
 			return err
```
geunwoonoh: **Critical**: returning here skips `recheckLinks` entirely, so every link already written stays unchecked against the entries that came after it.

Repro, extracted into a fresh destination:

```
entries: subdir/keep, escape -> subdir/parent/.., subdir/parent -> .., bad -> /etc

err = illegal link target in zip: "bad" -> "/etc"
  escape SYMLINK -> "subdir/parent/.."
  resolves to .../001          <- the parent of destDir
  destDir    is .../001/out
  ESCAPED    = true
```

`bad` is rejected at creation, so the function returns before the recheck runs, and `escape` is left resolving outside destDir. Nothing is written outside, but the link the PR set out to never leave behind is still there, and a later extraction into that directory now fails at `mkdirAllInside` forever.

Keep the creation error and let the recheck run regardless:

```go
	var failed error
	for _, link := range links {
		if err := extractSymlink(link.file, root, link.path); err != nil {
			failed = err
			break
		}
	}
	// Links already on disk were validated against an archive that was not
	// fully written yet, so sweep them even when a later entry ended the run.
	if err := recheckLinks(root, links); err != nil && failed == nil {
		failed = err
	}
	return failed
```

`recheckLinks` already skips links that were never created, since `os.Readlink` fails on them.

pkg/utils/archive.go
```diff
@@ -242,45 +295,260 @@ func Unzip(src, destDir string) error {
...
+			// Dropped on a host that makes no links, so there is none to check.
+			continue
+		}
+		if isValidLinkTarget(root, link.path, filepath.ToSlash(target)) {
+			continue
+		}
+		if err := os.Remove(link.path); err != nil {
 			return err
 		}
+		return fmt.Errorf("illegal link target in zip: %q -> %q", link.file.Name, target)
```
geunwoonoh: **Critical**: this returns on the first offender, so any link after it in `links` is never re-checked. Those links passed validation at creation time only because the later entries did not exist yet, which is the exact case this function was added for.

Repro with two links of the same shape:

```
entries: subdir/keep, escape1 -> subdir/parent/.., escape2 -> subdir/parent/.., subdir/parent -> ..

err = illegal link target in zip: "escape1" -> "subdir/parent/.."
  escape1        absent
  escape2        SYMLINK -> "subdir/parent/.."  (resolves to .../001, out=.../001/out)
```

`escape2` survives, resolving to the parent of destDir.

Sweep every link, then report:

```go
func recheckLinks(root string, links []extractedLink) error {
	var rejected error
	for _, link := range links {
		target, err := os.Readlink(link.path)
		if err != nil {
			// Dropped on a host that makes no links, so there is none to check.
			continue
		}
		if isValidLinkTarget(root, link.path, filepath.ToSlash(target)) {
			continue
		}
		if err := os.Remove(link.path); err != nil {
			return err
		}
		if rejected == nil {
			rejected = fmt.Errorf("illegal link target in zip: %q -> %q", link.file.Name, target)
		}
	}

	return rejected
}
```

pkg/utils/archive.go
```diff
@@ -242,45 +295,260 @@ func Unzip(src, destDir string) error {
...
+	// Zip entries carry slash-separated paths whatever wrote them, so a leading
+	// slash is absolute even where filepath.IsAbs alone would miss it.
+	if target == "" || strings.HasPrefix(target, "/") || filepath.IsAbs(target) {
+		return false
+	}
+
+	current := filepath.Dir(fpath)
+	parts := strings.Split(target, "/")
+	for steps := 0; len(parts) > 0; steps++ {
+		if steps > maxLinkResolution {
```
geunwoonoh: **Warning**: `steps` increments once per path component, not once per symlink followed, so this refuses targets that contain no links at all.

```
target = "d/d/d/.../file.txt"    (71 components, 0 symlinks)  -> illegal link target in zip
target = "././././.../file.txt"  (71 components, all ".")     -> illegal link target in zip
```

The comment on `maxLinkResolution` says it bounds the chain "the way ELOOP bounds the kernel's own resolution", but the kernel bounds link traversals (40), not path depth. A deep tree downloaded through WebFTP and uploaded back fails the whole extraction.

Count hops instead:

```go
	current := filepath.Dir(fpath)
	parts := strings.Split(target, "/")
	hops := 0
	for len(parts) > 0 {
		part := parts[0]
		parts = parts[1:]

		switch part {
		case "", ".":
			continue
		case "..":
			current = filepath.Dir(current)
		default:
			next := filepath.Join(current, part)
			if fi, err := os.Lstat(next); err == nil && fi.Mode()&os.ModeSymlink != 0 {
				hops++
				if hops > maxLinkResolution {
					return false
				}
				// ... existing expansion
				continue
			}
			current = next
		}

		if !isInside(root, current) {
			return false
		}
	}
```

Termination still holds: expansions are capped by `maxLinkResolution` and each one adds at most `maxSymlinkTarget` components.

pkg/utils/archive.go
```diff
@@ -242,45 +295,260 @@ func Unzip(src, destDir string) error {
...
+	var links []extractedLink
+	for _, f := range r.File {
+		fpath := filepath.Join(root, f.Name)
+		if !isInside(root, fpath) {
+			return fmt.Errorf("illegal file path in zip: %q", f.Name)
 		}
 
 		if f.FileInfo().IsDir() {
-			if err := os.MkdirAll(fpath, 0755); err != nil {
+			if err := mkdirAllInside(root, fpath, f.Name); err != nil {
```
geunwoonoh, line 390: **Warning**: the directory entry carries its mode now, but `mkdirAllInside` hardcodes 0755, so it is dropped.

```
archive entry "src/private/"  mode=drwx------
extracted                     mode=drwxr-xr-x   <- source was drwx------
```

This PR is what made empty directories survive at all, so it is also what makes the widening visible: on a root install a private directory comes back world-readable after a download and re-upload. Files already keep `f.Mode().Perm()`.

Directory entries are only ever written for empty directories, so a narrow mode here cannot block a child write:

```go
		if f.FileInfo().IsDir() {
			if err := mkdirAllInside(root, fpath, f.Name); err != nil {
				return err
			}
			// Only an empty directory gets an entry, so a narrow mode here
			// has no child left to block.
			if err := os.Chmod(fpath, f.Mode().Perm()); err != nil {
				return err
			}
			continue
		}
```

pkg/utils/archive.go
```diff
@@ -229,6 +274,14 @@ func addFileToZip(w *zip.Writer, filePath, archiveName string, fi os.FileInfo) e
 	return nil
 }
 
+func isEmptyDir(path string) (bool, error) {
+	entries, err := os.ReadDir(path)
+	if err != nil {
+		return false, err
+	}
+	return len(entries) == 0, nil
```
geunwoonoh: **Suggestion**: `os.ReadDir` reads every entry and sorts them by name to answer a question that needs one entry. `filepath.Walk` calls this per directory, so it lands on the whole download path.

```
BenchmarkIsEmptyDir_ReadDir-14           1888560 ns/op   (5000-entry directory)
BenchmarkIsEmptyDir_Readdirnames1-14       28096 ns/op   -> 67x

BenchmarkCreateZip_TreeFat  (40k files across 20 large directories)
  as written    2598418396 ns/op
  Readdirnames  2378277000 ns/op   -> ~8% end to end
```

`errors` and `io` are already imported.

```suggestion
	f, err := os.Open(path)
	if err != nil {
		return false, err
	}
	defer func() { _ = f.Close() }()
	if _, err := f.Readdirnames(1); err != nil {
		if errors.Is(err, io.EOF) {
			return true, nil
		}
		return false, err
	}
	return false, nil
```

pkg/utils/archive.go
```diff
@@ -242,45 +295,260 @@ func Unzip(src, destDir string) error {
...
+	return createSymlink(target, fpath)
+}
+
+// createSymlink puts a link to target at fpath, over whatever is already there.
+// A regular entry gets that from O_TRUNC in one call, and this is the same
+// thing in two: commands run on a worker pool, so a second extraction can fill
+// the path back in between them, and losing that race is not a failure.
+func createSymlink(target, fpath string) error {
+	for attempt := 0; ; attempt++ {
+		if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
```
geunwoonoh, line 616: **Suggestion**: this and the other bare `return err` sites (`mkdirAllInside`, `extractSymlink`, `extractFile`, `recheckLinks`) hand raw OS errors to `pkg/executor/handlers/file/file.go:290`, which returns them to the console as the command result.

A link entry whose name collides with a non-empty directory in the same archive produces:

```
remove /private/var/folders/.../out/d: directory not empty
```

No archive, no entry name, and an internal absolute path. The golang stack rules ask for wrapping; the entry name is the part that makes this actionable.

```go
return fmt.Errorf("failed to restore link %q: %w", name, err)
```

pkg/utils/archive.go
```diff
@@ -242,45 +295,260 @@ func Unzip(src, destDir string) error {
...
+		case "..":
+			current = filepath.Dir(current)
+		default:
+			next := filepath.Join(current, part)
+			if fi, err := os.Lstat(next); err == nil && fi.Mode()&os.ModeSymlink != 0 {
+				link, err := os.Readlink(next)
+				// An absolute link is refused rather than followed: only a link
+				// the extraction did not write can be one, and reasoning about
+				// where it lands is not worth the archive.
+				if err != nil || strings.HasPrefix(filepath.ToSlash(link), "/") || filepath.IsAbs(link) {
```
geunwoonoh: **Suggestion**: the same absolute-target test appears at line 482 and here. Worth one helper so a future tightening cannot land on only one of them.

```go
// isAbsTarget reports whether a link target is absolute. Zip entries carry
// slash-separated paths whatever wrote them, so a leading slash counts even
// where filepath.IsAbs alone would miss it.
func isAbsTarget(s string) bool {
	s = filepath.ToSlash(s)
	return strings.HasPrefix(s, "/") || filepath.IsAbs(s)
}
```

pkg/utils/archive.go
```diff
@@ -242,45 +295,260 @@ func Unzip(src, destDir string) error {
...
 // Used when the caller has pre-validated the same handle to close TOCTOU windows.
+//
+// A link entry—the target path as the body with ModeSymlink set, the layout
+// CreateZip writes—comes back as a real symlink, dropped on a Windows host that
+// makes none rather than failing the whole extraction.
+//
+// Nothing may leave destDir: an escaping entry name, a link target that lands
+// outside once every link on the way to it is followed, and an entry written
+// through a link are each refused.
 func UnzipReader(r *zip.ReadCloser, destDir string) error {
```
geunwoonoh: **Suggestion** (optional, fine as a follow-up): the doc says "Caller owns r" and the function never calls `Close`, so `*zip.Reader` is the type that matches what it actually needs. The PR description makes this argument itself for the destination directory: "the function is exported and the next caller need not" be the upload path. The current signature shuts out a caller holding an in-memory archive. The one call site becomes `&rc.Reader`.

pkg/utils/archive_test.go
```diff
@@ -314,3 +316,316 @@ func TestCreateZipAndUnzip_RoundTrip(t *testing.T) {
...
+func TestUnzip_EscapingLinkTargetIsRejected(t *testing.T) {
+	// Both names pass the entry-name check; only the link target puts the
+	// second write outside.
+	escaping := []zipEntry{
+		{name: "a", body: "../../etc", isLink: true},
+		{name: "a/passwd", body: "malicious"},
+	}
+	// Nothing constrains entry order, so a link arriving after the path that
+	// goes through it must be rejected just the same.
+	for name, entries := range map[string][]zipEntry{
```
geunwoonoh: **Suggestion**: Go randomizes map iteration, so subtest order changes per run and CI logs stop diffing cleanly. Both this table and the one at line 508 are affected. The golang stack rules call for a slice of structs:

```go
	for _, tt := range []struct {
		name    string
		entries []zipEntry
	}{
		{"link first", escaping},
		{"link last", []zipEntry{escaping[1], escaping[0]}},
	} {
		t.Run(tt.name, func(t *testing.T) {
```

Separately, two gaps worth a test each: nothing asserts what is left in destDir after a rejection, which is how the Critical finding on `recheckLinks` passed CI, and nothing covers directory permissions across a round trip.

geunwoonoh (changes requested):
## Review summary

**Stack**: golang · **Changes**: +630 -21 across 4 files · **CI**: 20/20 green, no unresolved threads

This restores zip link entries as real symlinks and closes the escape routes that opening that branch creates. Nine commits, each doing one thing, each with a test that was seen failing first. The structural calls are right: deferring link entries to a second pass is the correct answer to entry-order dependence, and resolving the target component by component instead of comparing with `filepath.Rel` is correct where CodeQL's own suggested fix would not have been.

One invariant the PR states for itself does not hold. I reproduced it two ways.

---

### Critical (1)

**1. A rejected extraction leaves an escaping symlink on disk** — `pkg/utils/archive.go:353`, `pkg/utils/archive.go:376`

The description says "removing the offending link before failing the extraction". Only the first offender is removed. Two paths leak:

**(a) `recheckLinks` returns on the first offender**, so every link after it in `links` is never re-checked. Creation-time validation passed only because the later links did not exist yet, so the recheck is the only barrier those links have.

```
entries: subdir/keep, escape1 -> subdir/parent/.., escape2 -> subdir/parent/.., subdir/parent -> ..

err = illegal link target in zip: "escape1" -> "subdir/parent/.."
  escape1        absent
  escape2        SYMLINK -> "subdir/parent/.."  (resolves to .../001, out=.../001/out)  <- survives
```

**(b) A failure in `extractSymlink` skips `recheckLinks` entirely.** The second loop returns raw, so every link already created stays unchecked.

```
entries: subdir/keep, escape -> subdir/parent/.., subdir/parent -> .., bad -> /etc

err = illegal link target in zip: "bad" -> "/etc"
  escape SYMLINK -> "subdir/parent/.."
  resolves to .../001          <- the parent of destDir
  destDir    is .../001/out
  ESCAPED    = true
```

**On impact, stated plainly**: nothing is written outside destDir. What survives is a symlink in the user's own directory, which that user could have created themselves, so this is not a privilege escalation. Two things still argue for fixing it before merge. The contract the PR sets for itself is not met, and a directory holding that link fails every later extraction at `mkdirAllInside`: poison it once and uploads to that path stay broken.

The fix is about ten lines across two functions; see the inline comments.

---

### Warnings (2)

**2. `maxLinkResolution` counts path components, not symlink hops** — `pkg/utils/archive.go:489`

The comment claims ELOOP semantics, but the kernel bounds link traversals (40), not path depth. `steps` increments per component, so a legitimate target with no symlinks in it at all is refused:

```
target = "d/d/d/.../file.txt"    (71 components, 0 symlinks)  -> illegal link target in zip
target = "././././.../file.txt"  (71 components, all ".")     -> illegal link target in zip
```

Download a deep tree through WebFTP, upload it back, and the whole extraction fails. Not a hole, a false rejection, but it costs the user their data either way.

**3. Empty-directory permissions are flattened to 0755** — `pkg/utils/archive.go:328`

This PR is what taught `newZipEntry` to store `ModeDir` and `addDirToZip` to write the entry, so the archive now carries the real mode. Extraction drops it: `mkdirAllInside` hardcodes 0755.

```
archive entry "src/private/"  mode=drwx------
extracted                     mode=drwxr-xr-x   <- source was drwx------
```

Alpamon runs as root on a normal install, so a private directory comes back world-readable after a round trip. Files keep `f.Mode().Perm()` already; directories are the half that was left out.

---

### Suggestions (5)

**4. `isEmptyDir` reads and sorts the whole directory to answer one question** — `pkg/utils/archive.go:278`

`os.ReadDir` reads every entry and sorts by name. Answering "is it empty" needs one entry. `filepath.Walk` calls this per directory, so it sits on the whole download path. Measured:

```
BenchmarkIsEmptyDir_ReadDir-14           1888560 ns/op   (5000-entry directory)
BenchmarkIsEmptyDir_Readdirnames1-14       28096 ns/op   -> 67x

BenchmarkCreateZip_TreeFat  (40k files across 20 large directories)
  as written    2598418396 ns/op
  Readdirnames  2378277000 ns/op   -> ~8% end to end
```

I measured the other direction too. Memoizing the per-component `Lstat` walk in `mkdirAllInside` moved `BenchmarkUnzip_Tree` from 796ms to 781ms, about 2%. Not worth the cache: leave that code as it is.

**5. Bare OS errors reach the console** — `pkg/utils/archive.go:454`

`mkdirAllInside`, `extractSymlink`, `extractFile`, `createSymlink`, and `recheckLinks` return OS errors unwrapped, and `pkg/executor/handlers/file/file.go:290` hands them to the console as the command result. A link entry whose name collides with a non-empty directory in the same archive produces:

```
remove /private/var/folders/.../out/d: directory not empty
```

No archive, no entry name, and an internal absolute path. The golang stack rules ask for `fmt.Errorf("...: %w", err)`; at minimum the entry name belongs in there.

**6. The absolute-target check is written twice** — `pkg/utils/archive.go:508`

**7. `UnzipReader` takes `*zip.ReadCloser` but never closes it** — `pkg/utils/archive.go:306`

**8. Two table tests iterate a map** — `pkg/utils/archive_test.go:476`

Also two coverage gaps: nothing asserts what is left in destDir after a rejection, which is why Critical 1 passes CI, and nothing covers directory permissions.

---

### Dead and duplicate code

Nothing unused in the new code. `errNotRegular`, `isInside`, `symlinkAttempts`, `extractedLink`, and `symlinkUnsupported` are all reached. `go vet ./pkg/utils/`, `gofmt -l pkg/utils/`, and `GOOS=windows go build ./pkg/utils/` are all clean.

The pre-check in `extractSymlink` and the re-check in `recheckLinks` look redundant but are not. The first narrows the window in which an escaping link exists at all; the second closes entry-order dependence. Both earn their place.

---

### Highlights

- The two-pass structure is the right shape for the ordering problem, and it happens to give Windows what it needs to pick a file or directory link.
- Component-by-component resolution over `filepath.Rel`. The observation that `filepath.Join` folds `a/link/..` into `a` while the kernel follows the link first is exactly the point, and the note on why `filepath.EvalSymlinks` is unusable here is right.
- `%q` on untrusted values, with a test. Tracing an archive byte all the way to live terminal input is a good catch.
- `mkdirAllInside` plus the `Lstat` in `extractFile` cover the parent chain and the leaf component with no gap between them.
- The concurrency regression was confirmed against a merge-base build rather than assumed, and `-race -count=20` is clean.
- Verified against archives written by the real `zip --symlinks` CLI, not only by `CreateZip`.

---

Critical 1 sits on the invariant this PR exists to establish, so it should land before merge; the repros above drop straight into `archive_test.go`. Warnings 2 and 3 are worth taking in the same pass. The rest is optional.

**Checklist**
- [x] Code follows stack conventions (error wrapping aside, see 5)
- [x] Tests included and passing, though nothing covers the post-rejection state
- [ ] No security vulnerabilities: Critical 1, Warning 3
- [x] No documentation changes needed (no exported signature change)

pkg/utils/archive.go
jisung-02: Fixed in 262e78b0, taken as written. You are right that the return skips the sweep entirely, and right about which half matters: nothing lands outside, but the leftover link fails every later extraction into that directory at `mkdirAllInside`, so poisoning the path once keeps uploads to it broken.

The loop keeps the creation error and the sweep runs regardless, as you suggested. One addition on top of your snippet: `recheckLinks` now returns the rejection and the removal failure separately, and the caller reports a link it could not remove ahead of both the creation error and the rejection, since that is the only case where destDir still holds an escape when the call returns.

`TestUnzip_RejectionLeavesNoEscapingLink` carries your repro as its second case and asserts `escape` is gone. It failed before the fix with the link still present.

pkg/utils/archive.go
jisung-02: Fixed in the same commit, 262e78b0. The reasoning is the part I had missed: those links passed validation at creation time only because the later entries did not exist yet, so the sweep is the only barrier they ever get, and stopping at the first offender leaves the rest with none.

The sweep now visits every link. Two changes to your snippet. `os.Remove` failing no longer returns immediately, because that would reintroduce the same early exit one level down; the first failure is kept as `left` and the loop continues. And `os.IsNotExist` is not treated as a failure at all, since a concurrent extraction into the same destination may have swept the link already.

`TestUnzip_RejectionLeavesNoEscapingLink` runs your two-offender case and asserts both `escape1` and `escape2` are gone; it failed before the fix with `escape2` surviving.

pkg/utils/archive.go
jisung-02: Fixed in 8fd0e3cf. You are right on both counts: the counter incremented per component, and the comment claiming ELOOP semantics was describing something the code did not do. A deep tree round-tripped through WebFTP failed on data that was never malformed.

Counting hops as you wrote it, with the constant comment rewritten to say what is actually bounded. `TestUnzip_DeepLinkTargetWithoutLinksIsAccepted` covers both of your cases, the deep tree and the all-`.` target, and failed before the fix with `illegal link target in zip`.

Your termination argument is where I ended up disagreeing, and 561ed33a is the follow-up. `maxLinkResolution` expansions times `maxSymlinkTarget` components is a bound, but it is a large one: chained links whose targets are padding held a worker for tens of seconds while staying under 64 hops. `maxLinkWalk` caps the components visited across the whole walk, far above any real target. `TestUnzip_LinkWalkPastTheComponentBoundIsRejected` builds that archive.

pkg/utils/archive.go
jisung-02: Fixed in a4700d42. The framing is what made this land: this PR is what taught the archive to carry the mode, so this PR is what made the widening visible, and on a root install a private directory coming back world-readable after a round trip is the outcome.

Applied differently from your suggestion, on one point. "Directory entries are only ever written for empty directories" holds for what `CreateZip` writes, but not for `zip -r`, which writes an entry for every directory before the entries below it. `AllowUnzip` extracts whatever the user uploaded, so a mode landing at mkdir time would block the children of any read-only directory in such an archive. The mode is applied after every entry instead, deepest first, since a parent with no search permission blocks the chmod under it.

Two things came out of that. `os.Chmod` follows a link, so a link entry that took a directory's name would have been chmod'd through: `chmodDir` opens with `O_NOFOLLOW|O_DIRECTORY` and calls `fchmod` on the handle, and the path is `Lstat`ed first so that case is refused rather than silently skipped. And mode 0 cannot be read as "no mode" for a non-Unix creator, where `archive/zip` derives 0666 from FAT attributes; 5052ef3f gates on `CreatorVersion`, without which a PowerShell `Compress-Archive` directory came back `drw-rw-rw-`.

Tests: `TestCreateZipAndUnzip_EmptyDirectoryModeIsPreserved`, `TestUnzip_ReadOnlyDirectoryEntryDoesNotBlockItsChildren`, `TestUnzip_LinkReplacingADirectoryEntryIsRejected`, `TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault` and `TestUnzip_DirectoryModeFromANonUnixArchiveIsIgnored`.

pkg/utils/archive.go
jisung-02: Applied in e0508c24, as written. The measurement is what settled it: `filepath.Walk` calls this once per directory, so it sits on the whole download path.

One follow-up on top, 862f7d71. Answering "does it have a child" is not quite the question the archive asks. A directory holding only a FIFO or a socket read as non-empty, so it got no entry, and the walk then dropped the FIFO for not being a regular file without recording a skip: the directory and its contents left no trace anywhere. `isEmptyDir` now stops at the first child that will actually be archived, which keeps your early exit for the common case. `TestCreateZip_DirectoryHoldingOnlySpecialFilesKeepsItsEntry` covers it, and failed before with 0 entries and 0 skips.

Also taking the measurement you ran in the other direction: the per-component `Lstat` memoization in `mkdirAllInside` is left alone.

pkg/utils/archive.go
jisung-02: Fixed in 0a92458d. The entry name is the part that makes it actionable, and that was the gap: the message reaching the console had no archive, no entry name, and an absolute path off the host.

`mkdirAllInside`, `extractSymlink`, `extractFile` and the `UnzipReader` call site all wrap with the name now. `TestUnzip_ErrorNamesTheEntry` pins your exact collision and failed before with the bare `remove ...: directory not empty`.

The repro also turned up something else, fixed in 27105e16. An entry name or link target carrying control bytes can reach an OS error, and the raw path inside a wrapped `*PathError` is somewhere `%q` cannot reach, so it goes to the console unescaped anyway. `TestUnzip_RejectionMessageEscapesControlBytes` did not catch that because it only covered the messages this code writes itself. Both are now rejected at the gate.

pkg/utils/archive.go
jisung-02: Applied in ab09047a, your helper as written. The reason is the one that matters: a future tightening of the rule would land on one of the two and leave the other, which is the harder half to notice.

cc2c90f5 tightened it once already, and having the single helper is what made that safe: link targets are now normalized with `filepath.ToSlash` on the way into the archive and on the way out, so a backslash target archived on Windows no longer arrives on Unix as one opaque component.

pkg/utils/archive.go
jisung-02: Applied in bbc25c04, not deferred. The argument is the one the PR body already makes about the destination directory, so leaving the signature asking for more than the work needs would have been inconsistent with the rest of the change.

Both call sites pass `&rc.Reader` and keep closing the handle themselves. `TestUnzipReader_ExtractsAnInMemoryArchive` is the caller the old signature refused, building the archive in a `bytes.Buffer` and reading it straight back with no file involved.

Flagging this as the one breaking change in the PR: it is noted as such in the commit and in the body, since `UnzipReader` is exported.

pkg/utils/archive_test.go
jisung-02: Applied in eb7815a0. Both tables are slices of structs now, so the subtest order is fixed and two CI logs of the same failure diff cleanly. The same commit removes `writeZip`, which built the same archive as `writeEntryZip` minus modes and links: two helpers for one job means the next test picks whichever it sees first.

Both coverage gaps are closed, and you were right that the first one is why the Critical passed CI.

`TestUnzip_RejectionLeavesNoEscapingLink` asserts what is left in destDir after a rejection, over both of your repros. `TestRecheckLinks_OffenderLeftOnDiskOutranksTheOnesRemoved` covers the other half, a link the sweep cannot remove; it drives `recheckLinks` directly, because directory modes are applied after the sweep, so no archive can leave a link's parent read-only in time.

Directory permissions across a round trip: `TestCreateZipAndUnzip_EmptyDirectoryModeIsPreserved`. Its name is narrow on purpose. Only empty directories get an entry of their own, so only they carry a mode through a `CreateZip` round trip; a directory holding files still comes back with the default. `applyDirModes` restores the mode for `zip -r` archives, which do carry an entry per directory, and `TestUnzip_ReadOnlyDirectoryEntryDoesNotBlockItsChildren` covers that shape.

pkg/utils/archive_unix.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"os"
+	"syscall"
+)
+
+// symlinkUnsupported reports whether os.Symlink failed because the host makes no
+// links at all. Nothing on Unix says only that: EPERM covers a filesystem without
+// link support and a plain permission denial alike.
+func symlinkUnsupported(error) bool { return false }
```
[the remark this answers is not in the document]
jisung-02: Not taking this one: nothing is shadowed. In `func symlinkUnsupported(error) bool` the `error` is the parameter type, not a parameter name. A Go parameter list is either all named or all unnamed, and a lone `error` with no identifier in front of it is a type, so the parameter is anonymous and the predeclared identifier is untouched.

Checked rather than assumed, with the same declaration form:

```go
func f(error) bool {
    var e error = nil // would not compile if the name were shadowed
    _ = e
    return false
}
// compiles
```

The linter claim does not hold here either: `golangci-lint run ./pkg/utils/` reports `0 issues.` and `go vet ./...` is clean on darwin and with `GOOS=windows`.

The asymmetry with the Windows file is deliberate. `symlinkUnsupported(err error)` there names the parameter because it inspects it; the Unix one always returns false, since EPERM covers both a filesystem without link support and a plain permission denial, so there is nothing to inspect and nothing to name.

pkg/utils/archive.go
```diff
@@ -237,50 +331,406 @@ func Unzip(src, destDir string) error {
...
+			}
+			fi, err = os.Lstat(current)
+		}
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to create directory for %q: %w", name, err)
+		}
+		if fi.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("illegal link target in zip: %q is written through a symlink", name)
+		}
```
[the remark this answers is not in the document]
jisung-02: Correct, and fixed in 8ae3b0a3. The entry this refuses is usually not a link at all, so naming a link target was wrong twice over: a plain file entry read

    illegal link target in zip: "a/passwd" is written through a symlink

on an archive that carries no link. What is a link is a component of the path the entry would be written through, which the rest of the sentence already says.

You also caught that it disagreed with the check on the very next line. That one is new in this PR (be9cbde3), so the inconsistency arrived with it rather than being pre-existing. Both say `illegal entry in zip` now.

One test asserted the old prefix. `TestUnzip_EntryWrittenThroughExistingLinkIsRejected` extracts a plain file entry through a link already sitting in the destination, so it now asserts the whole sentence instead of the prefix alone, which is the part that says why the entry was refused. The other tests on this path assert `written through a symlink`, which is unchanged.

pkg/utils/archive.go
```diff
@@ -237,50 +331,408 @@ func Unzip(src, destDir string) error {
...
+	if fi, err := os.Lstat(fpath); err == nil && fi.Mode()&os.ModeSymlink != 0 {
+		if err := os.Remove(fpath); err != nil && !os.IsNotExist(err) {
+			return err
+		}
+	}
+
+	// Perm only: extraction runs as root on a normal install, so setuid, setgid
+	// or sticky off an entry would land on a root-owned file. newZipEntry drops
+	// the same bits on the way in.
+	outFile, err := createFile(fpath, f.Mode().Perm())
```
geunwoonoh, line 722: Observation (not blocking, and pre-existing): an entry from a Unix creator with a zero external attribute extracts as a 0000 file. I confirmed it by hand-building such an archive:

```
P3 unzip=<nil>  resulting file mode=----------
```

`applyDirModes` guards exactly this case a few lines up, with the reasoning that "Zero is an archive that carried no mode at all", but the file path has no equivalent guard, so the two disagree. `extractFile` on `main` already passed `f.Mode()` through unchanged, so this is not a regression from this PR - only an inconsistency the new directory guard makes visible. Worth a follow-up rather than a change here.

pkg/utils/archive.go
```diff
@@ -237,50 +331,408 @@ func Unzip(src, destDir string) error {
...
 
-		_, err = io.Copy(outFile, rc)
-		_ = rc.Close()
-		_ = outFile.Close()
+	return rejected, left
+}
+
+// hasControlBytes reports whether s carries a byte a terminal would act on.
+func hasControlBytes(s string) bool {
+	return strings.ContainsFunc(s, func(r rune) bool { return r < 0x20 || r == 0x7f })
```
geunwoonoh, line 512: Observation (not blocking): this covers C0 and DEL, but a raw 0x9b byte (C1 CSI) passes:

```
hasControlBytes(ESC 0x1b)      = true
hasControlBytes(DEL 0x7f)      = true
hasControlBytes(raw byte 0x9b) = false
```

A raw 0x9b is invalid UTF-8, so `strings.ContainsFunc` sees U+FFFD and lets it through, yet the byte itself still reaches the console inside a wrapped OS error. Few UTF-8 terminals act on C1 these days, so the value is small - noting it only because the function's stated job is "a byte a terminal would act on".

geunwoonoh (approved):
## Follow-up review: all nine findings addressed

**Stack**: golang · **Changes**: +1484 -48 across 7 files · 20 commits since the previous review at `964f0d8`

---

### Summary

Every finding from the previous review is fixed, and each one has a regression test behind it. I verified this by running the code rather than only reading it.

| # | Severity | Finding | How it was fixed |
|---|---|---|---|
| 1a | Critical | `recheckLinks` returned on the first offender, leaving later escaping links on disk | Returns `(rejected, left error)` and sweeps every link |
| 1b | Critical | An `extractSymlink` failure skipped `recheckLinks` entirely | Failure is held in `failed` and the loop breaks; the sweep always runs |
| 2 | Warning | `maxLinkResolution` counted path components, not symlink hops | `hops` counts only real links; components get their own bound in `maxLinkWalk` |
| 3 | Warning | Directory entry modes were flattened to 0755 | `applyDirModes` lands each mode last, deepest first |
| 4 | Suggestion | `isEmptyDir` read and sorted the whole directory | `f.ReadDir(64)` with an early return |
| 5 | Suggestion | Bare OS errors reached the console | Call sites wrap them with the entry name |
| 6 | Suggestion | The absolute-target test appeared twice | Folded into `isAbsTarget` |
| 7 | Suggestion | `*zip.ReadCloser` asked for more than the function used | Signature takes `*zip.Reader`, caller updated |
| 8 | Suggestion | Map-iteration table tests | Replaced with slices, and the two zip helpers folded into one |

`TestUnzip_RejectionLeavesNoEscapingLink` reproduces both paths of the critical finding, and `TestRecheckLinks_OffenderLeftOnDiskOutranksTheOnesRemoved` drives the two-error contract directly. The only thing removed from the test file is the duplicate `writeZip` helper, so no coverage was lost.

The round went past what was asked: control-byte rejection, `O_NOFOLLOW`, and detection of a link that took a directory's name all arrived alongside. The two open Copilot threads are resolved by these changes as well.

### Verification

```
go test ./... -p 1                     46 packages pass, 0 failures
go test -race ./pkg/utils              pass
go build (linux/darwin/windows amd64)  all succeed
gofmt / go vet                         clean
```

I measured the two new judgment calls rather than taking them on the comment alone.

- Returning `true` once `hops` passes `maxLinkResolution` is sound. Linux follows at most 40 links and darwin 32, so a chain past 64 is inert. A 69-hop chain built to test this answered `too many levels of symbolic links` on access, and the escaping tail link was refused on its own.
- `maxLinkWalk` does bound the cost. An archive of 20,000 Lstat-forcing link entries (2.6 MB) took 15.7s against 1.7s for the same count of plain files, so roughly 9x. That is a constant factor, not an amplification worth blocking on.

### Blocking before merge (1)

1. **The CodeQL check is red with 12 new high-severity alerts.** It was green on `964f0d8`, the commit I reviewed previously.
   - Eleven are `go/path-injection` (`archive.go:450,486,496,519,549,552,620,620,630,668` and `archive_unix.go:22`) plus one zip-slip. All of them are new instances created by moving and adding filesystem calls during the refactor.
   - **This is not a code defect.** Every one of those calls sits behind `isInside` and `mkdirAllInside`, and the tests in this PR plus my own probes confirm escapes are refused. CodeQL simply does not recognize a `filepath.Rel`-based check as a sanitizer.
   - This repository has already dismissed about twenty alerts of exactly this class, `archive.go` among them (alerts 123, 124, 125). The same triage in the security tab clears it. It is a merge chore, not a code change.

### Observations (3, none blocking)

1. **`archive.go:722`**: an entry from a Unix creator carrying mode 0 extracts as a 0000 file, which I confirmed by running it. `applyDirModes` guards the identical case with `if mode == 0 { continue }`, so the two paths disagree. `extractFile` on `main` already passed `f.Mode()` straight through, so this is pre-existing and not a regression from this PR.
2. **`archive.go:512`**: `hasControlBytes` covers C0 and DEL. A raw 0x9b byte (C1 CSI) passes. UTF-8 terminals rarely act on it, so the value here is small.
3. The `zipCreatorUnix` / `zipCreatorMacOSX` gating, the deepest-first chmod ordering, and pinning the handle with `O_NOFOLLOW|O_DIRECTORY` in `chmodDir` are all correct.

### Highlights

- The fixes rebuilt the contract instead of patching the symptom. Having `recheckLinks` return two errors and having the caller report `left` first means "an escaping link is still on disk" can never be hidden behind another failure.
- The comments record why a decision was made, not just what it does. The kernel resolution limit as the reason for accepting a link past the hop bound is the clearest example.
- Tests assert on behavior rather than implementation, down to `TestUnzip_LinkCycleIsKeptButUnresolvable` checking for the ELOOP message itself.
- The `mkdirAllInside` component walk was left alone, as suggested.

---

**Verdict**: APPROVED, to merge once the CodeQL alerts are triaged.

The code is in good shape. The 12 CodeQL alerts are the same false-positive class this repository has closed repeatedly, so clearing them in the security tab is all that stands between this and merge.

**Checklist**:
- [x] Tests included and passing (46 packages, race detector included)
- [x] All nine previous findings fixed, each with a regression test
- [x] No exploitable vulnerabilities (verified by running the code)
- [ ] Triage the 12 new CodeQL alerts — required before merge
- [x] Documentation (comments) updated

pkg/utils/archive.go
jisung-02: Confirmed, and filed as #440 rather than changed here.

You read the disagreement the way I would: `applyDirModes` says a zero mode is an archive that carried no mode at all, and `extractFile` says it is a file nobody may open. Only one of those can be right, and the directory half is the one that got it right.

Leaving it out of this PR for the reason you gave. `extractFile` on `main` already passed `f.Mode()` straight through, so the behavior predates the branch; what this PR added is the contrast that makes it visible. Folding a pre-existing file-mode change into a symlink PR would put it under a title nobody would search for later.

#440 carries your repro and the note that the fix has to land with a test next to `TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault`, since that is the directory-side case it has to agree with.

pkg/utils/archive.go
jisung-02: Confirmed, and filed as #441.

The cause is the one your output points at: `strings.ContainsFunc` decodes UTF-8, so a lone `0x9b` never reaches the callback as itself, only as U+FFFD. The function claims to reject a byte a terminal would act on and then never looks at bytes.

Not fixing it in this PR, because the byte-level rule needs a decision rather than a one-line swap. `0x80`-`0x9f` is also where the continuation bytes of every legitimate multi-byte UTF-8 sequence live, so a naive byte scan would refuse any entry name with a non-ASCII character in it. The choice is between rejecting only a lone C1 in an otherwise invalid sequence and accepting that cost, and that belongs in its own change with its own test rather than at the end of this round.

#441 records both options and your three-line repro. Agreed on the severity: the reachable path is a wrapped OS error rendered on a UTF-8 terminal, which is where C1 is nearly always inert.

Issue #284, opened 2026-04-23
ci,docs: add windows-latest build/test CI and Windows install/upgrade/uninstall runbook


geunwoonoh:
## Problem

Two foundational gaps for Windows support are independent of each other but should land together as the "Windows foundation" baseline before any UI-visible feature work (Issue 3 `install.ps1` hardening, server-side Markdown guide, etc.).

### 1. No Windows CI

The alpamon Windows binary is shipped to operators without ever being executed on a Windows host. `.goreleaser.yaml` cross-compiles `windows/amd64` on an Ubuntu runner and releases the archive directly.

- `.github/workflows/build-and-test.yml` test matrix is Ubuntu/Debian/RHEL/Fedora/Oracle Linux containers only.
- `.github/workflows/release.yml` has no Windows smoke step.
- `pkg/runner/pty_windows.go` (~315 lines of ConPTY code), `cmd/alpamon/command/register/service_windows.go` (SCM registration), and `pkg/updater/replace_windows.go` (locked-binary replacement via `MoveFileEx`) have **zero** CI coverage.

A regression in any of the Windows-specific paths would only be discovered by operators after release.

### 2. No Windows lifecycle documentation

Operators have no documented path for:

- Manual install vs `install.ps1`-based install.
- Where alpamon stores files on Windows (`%ProgramFiles%\alpamon`, `%ProgramData%\alpamon`).
- How to stop, start, and query the Windows service.
- How to upgrade (the self-updater covers most cases, but manual fallback is undocumented).
- **How to uninstall cleanly.** Today there is no documented uninstall procedure; the `system/byebye` handler exists but the Windows-specific steps are nowhere.
- Which features are unsupported on Windows (Firewall, User/Group management, Tunnel, Code-Server, privilege demotion).

## Expected behavior

- Every PR and every release artifact is compiled and minimally smoke-tested on an actual Windows runner before merge / publish.
- A single canonical document covers the full Windows install/upgrade/uninstall lifecycle plus the platform compatibility matrix.

### CI

1. `build-and-test.yml` adds a `windows-latest` job that runs:
   - `go build ./cmd/alpamon`
   - `go test ./... -p 1` on the Windows-compatible subset (any excluded packages must be gated with `//go:build !windows` and tracked in a follow-up issue — no silent blanket skip).
   - `alpamon.exe --version` smoke invocation that exits cleanly.
2. `release.yml` adds a post-goreleaser step on `windows-latest` that:
[...]

PR #444, opened 2026-09-01
fix: close the websocket even when the close handshake fails

### LINKED ISSUE #443: [BUG] Close() skips Conn.Close() when the close handshake fails, leaking a socket per reconnect
## Description

`WebsocketClient.Close()` returns early when the close handshake write fails, so `wc.Conn.Close()` is never reached and the underlying socket is never released.

https://github.com/alpacax/alpamon/blob/main/pkg/runner/client.go#L234-L242

```go
err := wc.Conn.WriteControl(
    websocket.CloseMessage,
    websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""),
    time.Now().Add(5*time.Second),
)
if err != nil {
    log.Debug().Err(err).Msg("Failed to write close message to websocket.")
    return  // <- skips wc.Conn.Close() below
}
```

`WriteControl` fails precisely when the connection is already broken, which is the same condition that drives a reconnect. `CloseAndReconnect` (`client.go:221`) calls `Close()` and then `Connect()`, and `Connect()` overwrites `wc.Conn` with the new connection. The previous connection is now unreachable and still open, so a long-lived agent on a flaky network leaks one file descriptor per reconnect.

Every sibling implementation in the same package already closes unconditionally: `pty.go:400` (fixed in #351), `control_client.go:141`, `tunnel_client.go:495`, `ftp.go:204`, `pty_windows.go:287`. `client.go` is the only one left with the early return.

## Steps to Reproduce

1. Dial a WebSocket through a `net.Conn` wrapper that records `Close()` calls and can be switched to fail all writes.
2. Make the wrapper fail writes, so `WriteControl` returns an error.
3. Call `WebsocketClient.Close()`.
4. Check whether the wrapper's `Close()` was called.

In production the same path is hit whenever the backhaul connection drops without a TCP close and `CloseAndReconnect` runs.

## Expected Behavior

`Close()` releases the connection whether or not the close handshake succeeds. A failed handshake is a best-effort courtesy to the peer, not a precondition for freeing the local file descriptor.

## Actual Behavior

`Close()` logs the write failure and returns. The socket stays open for the lifetime of the process, and each subsequent reconnect adds another leaked descriptor.

## Error Message

```
{"level":"debug","error":"simulated write failure","message":"Failed to write close message to websocket."}
```

jisung-02:
## Background

`WebsocketClient.Close()` returned on a failed close-frame write, skipping `Conn.Close()` entirely.

```go
if err != nil {
    log.Debug()...
    return                    // leaves here
}
...drain...
err = wc.Conn.Close()         // never reached
```

The write only fails when the connection is already broken, which is exactly the reconnect path. `CloseAndReconnect` calls `Close()`, gets nothing done, and `Connect()` then overwrites `wc.Conn` with a fresh connection. Nothing holds the old socket any more. The runtime finalizer on `netFD` (`net/fd_posix.go:44`) collects it eventually, but that is not a bound worth relying on: fds accumulate while reconnects repeat.

The sibling clients—`pty.go`, `control_client.go`, `tunnel_client.go`, `ftp.go`—all close unconditionally.

## Changes

Nine commits, oldest first. The last six answer review feedback.

| Commit | Change |
| --- | --- |
| `4bebacc` | move `trackedConn` above the pty recovery tests (pure move) |
| `3772e33` | the fix: `Conn.Close()` runs whether or not the close frame went out |
| `b59f3cb` | the drain and repeat-call assertions |
| `80fc471` | the pty close test reuses `dialTracked` instead of its own copy of that dialer |
| `a1c4340` | `trackedConn` and `dialTracked` move to an untagged `conn_test.go`, so the close tests run on Windows |
| `bc2e389` | the drain moves into `drainCloseReply`; the dead `IsCloseError` branch goes |
| `8ba2c25` | `Close()` captures the connection once instead of re-reading `wc.Conn` four times |
| `35cf82d` | `net.ErrClosed` no longer logs on the second close of the same connection |
| `6c87384` | a test pins that capture: `wc.Conn` is swapped mid-write and `Close()` still closes the connection it started with |

The drain that waits for the peer's close reply moved into the `else` branch. A peer only answers a frame that actually went out, so there is nothing to wait for once the write failed. `Conn.Close()` runs either way.

`wc.Conn` is deliberately left in place rather than set to nil. `ControlClient` guards its connection with a mutex; `WebsocketClient` has none, and other goroutines read `wc.Conn` directly, so nilling it would introduce the race it looks like it prevents. The `MessageTypeReconnect` handler and `CloseAndReconnect` therefore both close the same connection, which gorilla tolerates.

`Close()` now reads `wc.Conn` into a local before doing anything with it. `Connect()` assigns that field from the `RunForever` goroutine by way of `CloseAndReconnect`, while `gracefulShutdown` (`cmd/alpamon/command/root.go:231`) calls `Close()` from the main goroutine, so the field can change between two reads of it. Close would then send the close frame to one connection and close another. That field race predates this branch, but the branch widened it: `Close()` used to return before reaching `Conn.Close()` on a write error, and now every path reaches it.

## Known limitations

Two problems in this file are left open on purpose, both pre-existing and neither made worse here. Both are now filed as #452, together with the field race described above, because all three come down to `WebsocketClient` having no owner for its connection.

`Close()` and `drainCloseReply()` read the socket while `RunForever` can be blocked in `ReadMessage()` on the same connection, which gorilla does not allow—it permits one concurrent reader. The drain was already on the success path before this branch and its exposure is unchanged. Fixing it means giving reads a single owner rather than patching the close path.

`ShutDown()` calls `close(wc.ShutDownChan)` with no guard, and it is reachable from the worker pool (`pkg/executor/handlers/system/system.go:574,644,648,704`), from `CommandRequestHandler` inline in the `RunForever` goroutine, and from `wirePendingMigration`. Two of those firing together panics the process. This branch does not touch `ShutDown()`.

The capture in `8ba2c25` narrows the field race rather than closing it: `conn := wc.Conn` is still an unsynchronized read racing `Connect()`'s write at `client.go:209`. Closing it needs a mutex around every read and write of the field, which is #452's first item.

## Testing

Four tests in `pkg/runner/client_close_test.go`, and they now run on Windows as well—`GOOS=windows go list -f '{{.TestGoFiles}}' ./pkg/runner/` lists both `client_close_test.go` and `conn_test.go`.

| Test | Asserts |
| --- | --- |
| `ClosesConnAfterWriteControlFailure` | the socket closes after a failed handshake, and the drain is skipped |
| `DrainsReplyAfterSuccessfulHandshake` | the peer receives a normal-closure frame, the drain runs, then the socket closes |
| `IsSafeToCallTwice` | closing the same connection twice is harmless |
| `ClosesTheConnItStartedWith` | a `wc.Conn` swap from inside the close-frame write does not move which connection gets closed |

Whether the drain ran is read from the `SetReadDeadline` call count, not from how long `Close()` took, so the assertion does not depend on the clock and will not flake on a loaded runner.

The regression test was seen failing again after the refactoring commits, by restoring the early return in `Close()`:

```
--- FAIL: TestWebsocketClientClose_ClosesConnAfterWriteControlFailure (0.00s)
    client_close_test.go:64:
        	Error:      	Should be true
        	Messages:   	Close() did not close the websocket connection after WriteControl failure
```

`ClosesTheConnItStartedWith` was seen red the same way, by restoring the four separate reads of `wc.Conn` in `Close()`:

```
--- FAIL: TestWebsocketClientClose_ClosesTheConnItStartedWith (5.00s)
    client_close_test.go:106:
        	Error:      	Should be true
        	Messages:   	Close() left the connection it started with open
```

It takes five seconds in that state because the drain then runs against a connection the peer sent no close frame to, and only the read deadline ends it. The test that swaps the connection reads both sides, so the mirror assertion—the connection that arrived mid-call stays open—fails there too.

`IsSafeToCallTwice` was never seen red. It pins the assumption that gorilla's `Conn.Close` is idempotent rather than guarding a regression. The drain assertion was checked the other way, by deleting the drain from `client.go` and watching it fail.

| Check | Result |
| --- | --- |
| `go test ./pkg/runner/ -race -count=2` | ok, 76.956s |
| `go vet ./pkg/runner/` | clean |
| `GOOS=windows go vet ./pkg/runner/` | clean |
| `gofmt -l pkg/runner/` | clean |

Closes #443

https://claude.ai/code/session_01PTAG5uCwFCRxF6tmdTy8AA

### DIFF
--- a/pkg/runner/client.go
+++ b/pkg/runner/client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"sync/atomic"
@@ -244,8 +245,8 @@ func (wc *WebsocketClient) Close() {
 		drainCloseReply(conn)
 	}
 
-	err = conn.Close() // a broken connection must not leak its fd across reconnects
-	if err != nil {
+	// Close unconditionally so a broken connection cannot leak its fd; net.ErrClosed only means the reconnect path already closed it.
+	if err = conn.Close(); err != nil && !errors.Is(err, net.ErrClosed) {
 		log.Debug().Err(err).Msg("Failed to close websocket connection.")
 	}
 }
--- a/pkg/runner/client.go
+++ b/pkg/runner/client.go
@@ -223,15 +223,17 @@ func (wc *WebsocketClient) CloseAndReconnect(ctx context.Context) {
 	wc.Connect()
 }
 
-// Cleanly close the websocket connection by sending a close message
+// Cleanly close the websocket connection by sending a close message, waiting for
+// the peer's reply when that frame went out, and closing the connection either way.
 // Do not close quitChan, as the purpose here is to disconnect the WebSocket,
 // not to terminate RunForever.
 func (wc *WebsocketClient) Close() {
-	if wc.Conn == nil {
+	conn := wc.Conn // Connect() may swap wc.Conn from the RunForever goroutine mid-close; every step below must act on the one connection we started with.
+	if conn == nil {
 		return
 	}
 
-	err := wc.Conn.WriteControl(
+	err := conn.WriteControl(
 		websocket.CloseMessage,
 		websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""),
 		time.Now().Add(5*time.Second),
@@ -239,19 +241,19 @@ func (wc *WebsocketClient) Close() {
 	if err != nil {
 		log.Debug().Err(err).Msg("Failed to write close message to websocket.")
 	} else {
-		wc.drainCloseReply()
+		drainCloseReply(conn)
 	}
 
-	err = wc.Conn.Close() // a broken connection must not leak its fd across reconnects
+	err = conn.Close() // a broken connection must not leak its fd across reconnects
 	if err != nil {
 		log.Debug().Err(err).Msg("Failed to close websocket connection.")
 	}
 }
 
-func (wc *WebsocketClient) drainCloseReply() {
-	_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
+func drainCloseReply(conn *websocket.Conn) {
+	_ = conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
 	for {
-		if _, _, err := wc.Conn.NextReader(); err != nil {
+		if _, _, err := conn.NextReader(); err != nil {
 			break
 		}
 	}
--- a/pkg/runner/client.go
+++ b/pkg/runner/client.go
@@ -238,21 +238,20 @@ func (wc *WebsocketClient) Close() {
 	)
 	if err != nil {
 		log.Debug().Err(err).Msg("Failed to write close message to websocket.")
-		return
+	} else {
+		_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
+		for {
+			_, _, err = wc.Conn.NextReader()
+			if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
+				break
+			}
+			if err != nil {
+				break
+			}
+		}
 	}
 
-	_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second))
-	for {
-		_, _, err = wc.Conn.NextReader()
-		if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
-			break
-		}
-		if err != nil {
-			break
-		}
-	}
-
-	err = wc.Conn.Close()
+	err = wc.Conn.Close() // a broken connection must not leak its fd across reconnects
 	if err != nil {
 		log.Debug().Err(err).Msg("Failed to close websocket connection.")
 	}
--- a/pkg/runner/client.go
+++ b/pkg/runner/client.go
@@ -239,21 +239,21 @@ func (wc *WebsocketClient) Close() {
 	if err != nil {
 		log.Debug().Err(err).Msg("Failed to write close message to websocket.")
 	} else {
-		_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
-		for {
-			_, _, err = wc.Conn.NextReader()
-			if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
-				break
-			}
-			if err != nil {
-				break
-			}
-		}
+		wc.drainCloseReply()
 	}
 
 	err = wc.Conn.Close() // a broken connection must not leak its fd across reconnects
 	if err != nil {
 		log.Debug().Err(err).Msg("Failed to close websocket connection.")
+	}
+}
+
+func (wc *WebsocketClient) drainCloseReply() {
+	_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
+	for {
+		if _, _, err := wc.Conn.NextReader(); err != nil {
+			break
+		}
 	}
 }
 
--- a/pkg/runner/client_close_test.go
+++ b/pkg/runner/client_close_test.go
@@ -6,11 +6,53 @@ package runner
 
 import (
 	"net"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/require"
 )
+
+// closeReplyServer answers the client's close frame; the pty test double never reads, so a drain against it would only end on the read deadline.
+type closeReplyServer struct {
+	url       string
+	closeCode atomic.Int32
+}
+
+func newCloseReplyServer(t *testing.T) *closeReplyServer {
+	t.Helper()
+	s := &closeReplyServer{}
+	upgrader := websocket.Upgrader{CheckOrigin: func(*http.Request) bool { return true }}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		c, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			return
+		}
+		defer func() { _ = c.Close() }()
+
+		c.SetCloseHandler(func(code int, text string) error {
+			s.closeCode.Store(int32(code))
+			// SetCloseHandler replaces gorilla's default reply; without this the client drains until its deadline.
+			_ = c.WriteControl(websocket.CloseMessage, websocket.FormatCloseMessage(code, ""), time.Now().Add(time.Second))
+			return nil
+		})
+
+		for {
+			if _, _, err := c.ReadMessage(); err != nil {
+				return
+			}
+		}
+	}))
+	t.Cleanup(ts.Close)
+
+	s.url = strings.Replace(ts.URL, "http", "ws", 1)
+	return s
+}
 
 func dialTracked(t *testing.T, url string) (*websocket.Conn, *trackedConn) {
 	t.Helper()
@@ -45,4 +87,30 @@ func TestWebsocketClientClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
 
 	require.True(t, tracked.closed.Load(), "Close() did not close the websocket connection after WriteControl failure")
 	require.Equal(t, deadlines, tracked.readDeadlines.Load(), "Close() waited for a close reply the peer can never send, because the close frame never went out")
+}
+
+func TestWebsocketClientClose_DrainsReplyAfterSuccessfulHandshake(t *testing.T) {
+	s := newCloseReplyServer(t)
+	conn, tracked := dialTracked(t, s.url)
+
+	wc := &WebsocketClient{Conn: conn}
+
+	deadlines := tracked.readDeadlines.Load()
+	wc.Close()
+
+	require.Equal(t, int32(websocket.CloseNormalClosure), s.closeCode.Load(), "the peer did not receive a normal-closure close frame")
+	require.Greater(t, tracked.readDeadlines.Load(), deadlines, "Close() skipped the drain even though the close frame went out")
+	require.True(t, tracked.closed.Load(), "Close() did not close the websocket connection after a successful handshake")
+}
+
+func TestWebsocketClientClose_IsSafeToCallTwice(t *testing.T) {
+	// Close() deliberately leaves wc.Conn in place: the reconnect handler closes it, then RunForever's next read failure sends CloseAndReconnect at the same conn.
+	s := newCloseReplyServer(t)
+	conn, tracked := dialTracked(t, s.url)
+
+	wc := &WebsocketClient{Conn: conn}
+
+	wc.Close()
+	require.True(t, tracked.closed.Load(), "the first Close() did not close the websocket connection")
+	require.NotPanics(t, wc.Close, "the second Close() on an already closed connection panicked")
 }
--- a/pkg/runner/client_close_test.go
+++ b/pkg/runner/client_close_test.go
@@ -90,3 +90,19 @@ func TestWebsocketClientClose_IsSafeToCallTwice(t *testing.T) {
 	require.True(t, tracked.closed.Load(), "the first Close() did not close the websocket connection")
 	require.NotPanics(t, wc.Close, "the second Close() on an already closed connection panicked")
 }
+
+func TestWebsocketClientClose_ClosesTheConnItStartedWith(t *testing.T) {
+	// Connect() writes wc.Conn from the RunForever goroutine while Close() runs on another; the field swap is played out here from inside the close-frame write.
+	s := newCloseReplyServer(t)
+	first, firstTracked := dialTracked(t, s.url)
+	second, secondTracked := dialTracked(t, s.url)
+	t.Cleanup(func() { _ = second.Close() })
+
+	wc := &WebsocketClient{Conn: first}
+	firstTracked.onWrite = func() { wc.Conn = second }
+
+	wc.Close()
+
+	require.True(t, firstTracked.closed.Load(), "Close() left the connection it started with open")
+	require.False(t, secondTracked.closed.Load(), "Close() closed the connection Connect() had just swapped in")
+}
--- a/pkg/runner/client_close_test.go
+++ b/pkg/runner/client_close_test.go
@@ -1,11 +1,8 @@
-//go:build !windows
-
 package runner
 
 // The backhaul mirror of the pty client's close tests in pty_recovery_test.go; keep the two in step.
 
 import (
-	"net"
 	"net/http"
 	"net/http/httptest"
 	"strings"
@@ -54,30 +51,9 @@ func newCloseReplyServer(t *testing.T) *closeReplyServer {
 	return s
 }
 
-func dialTracked(t *testing.T, url string) (*websocket.Conn, *trackedConn) {
-	t.Helper()
-
-	var tracked *trackedConn
-	dialer := websocket.Dialer{
-		NetDial: func(network, addr string) (net.Conn, error) {
-			c, err := net.Dial(network, addr)
-			if err != nil {
-				return nil, err
-			}
-			tracked = &trackedConn{Conn: c}
-			return tracked, nil
-		},
-	}
-	conn, _, err := dialer.Dial(url, nil)
-	require.NoError(t, err)
-	require.NotNil(t, tracked, "the dialer never went through NetDial, so nothing is being tracked")
-
-	return conn, tracked
-}
-
 func TestWebsocketClientClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
-	s := newWshServer(t)
-	conn, tracked := dialTracked(t, s.wsURL())
+	s := newCloseReplyServer(t)
+	conn, tracked := dialTracked(t, s.url)
 
 	wc := &WebsocketClient{Conn: conn}
 
--- a/pkg/runner/client_close_test.go
+++ b/pkg/runner/client_close_test.go
@@ -1,0 +1,48 @@
+//go:build !windows
+
+package runner
+
+// The backhaul mirror of the pty client's close tests in pty_recovery_test.go; keep the two in step.
+
+import (
+	"net"
+	"testing"
+
+	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/require"
+)
+
+func dialTracked(t *testing.T, url string) (*websocket.Conn, *trackedConn) {
+	t.Helper()
+
+	var tracked *trackedConn
+	dialer := websocket.Dialer{
+		NetDial: func(network, addr string) (net.Conn, error) {
+			c, err := net.Dial(network, addr)
+			if err != nil {
+				return nil, err
+			}
+			tracked = &trackedConn{Conn: c}
+			return tracked, nil
+		},
+	}
+	conn, _, err := dialer.Dial(url, nil)
+	require.NoError(t, err)
+	require.NotNil(t, tracked, "the dialer never went through NetDial, so nothing is being tracked")
+
+	return conn, tracked
+}
+
+func TestWebsocketClientClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
+	s := newWshServer(t)
+	conn, tracked := dialTracked(t, s.wsURL())
+
+	wc := &WebsocketClient{Conn: conn}
+
+	deadlines := tracked.readDeadlines.Load()
+	tracked.failWrites.Store(true)
+	wc.Close()
+
+	require.True(t, tracked.closed.Load(), "Close() did not close the websocket connection after WriteControl failure")
+	require.Equal(t, deadlines, tracked.readDeadlines.Load(), "Close() waited for a close reply the peer can never send, because the close frame never went out")
+}
--- a/pkg/runner/pty_recovery_test.go
+++ b/pkg/runner/pty_recovery_test.go
@@ -6,8 +6,6 @@ package runner
 
 import (
 	"context"
-	"errors"
-	"net"
 	"net/http"
 	"net/http/httptest"
 	"runtime"
@@ -22,31 +20,6 @@ import (
 	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/require"
 )
-
-// trackedConn fails writes on demand, so a test can play out a broken connection without breaking a real one.
-type trackedConn struct {
-	net.Conn
-	failWrites    atomic.Bool
-	closed        atomic.Bool
-	readDeadlines atomic.Int32
-}
-
-func (c *trackedConn) SetReadDeadline(t time.Time) error {
-	c.readDeadlines.Add(1)
-	return c.Conn.SetReadDeadline(t)
-}
-
-func (c *trackedConn) Write(b []byte) (int, error) {
-	if c.failWrites.Load() {
-		return 0, errors.New("simulated write failure")
-	}
-	return c.Conn.Write(b)
-}
-
-func (c *trackedConn) Close() error {
-	c.closed.Store(true)
-	return c.Conn.Close()
-}
 
 // wshServer is a local Websh test double: a WebSocket endpoint plus the pty-channels recovery API, with server-side handles to kill connections.
 type wshServer struct {
--- a/pkg/runner/pty_recovery_test.go
+++ b/pkg/runner/pty_recovery_test.go
@@ -22,6 +22,25 @@ import (
 	"github.com/gorilla/websocket"
 )
 
+// trackedConn fails writes on demand, so a test can play out a broken connection without breaking a real one.
+type trackedConn struct {
+	net.Conn
+	failWrites atomic.Bool
+	closed     atomic.Bool
+}
+
+func (c *trackedConn) Write(b []byte) (int, error) {
+	if c.failWrites.Load() {
+		return 0, errors.New("simulated write failure")
+	}
+	return c.Conn.Write(b)
+}
+
+func (c *trackedConn) Close() error {
+	c.closed.Store(true)
+	return c.Conn.Close()
+}
+
 // wshServer is a local Websh test double: a WebSocket endpoint plus the pty-channels recovery API, with server-side handles to kill connections.
 type wshServer struct {
 	ts            *httptest.Server
@@ -221,24 +240,6 @@ func TestPtyRecovery_StormNoGoroutineLeak(t *testing.T) {
 		n := runtime.Stack(buf, true)
 		t.Fatalf("recovery goroutines did not exit after reconnect storm:\n%s", buf[:n])
 	}
-}
-
-type trackedConn struct {
-	net.Conn
-	failWrites atomic.Bool
-	closed     atomic.Bool
-}
-
-func (c *trackedConn) Write(b []byte) (int, error) {
-	if c.failWrites.Load() {
-		return 0, errors.New("simulated write failure")
-	}
-	return c.Conn.Write(b)
-}
-
-func (c *trackedConn) Close() error {
-	c.closed.Store(true)
-	return c.Conn.Close()
 }
 
 func TestPtyClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
--- a/pkg/runner/pty_recovery_test.go
+++ b/pkg/runner/pty_recovery_test.go
@@ -25,8 +25,14 @@ import (
 // trackedConn fails writes on demand, so a test can play out a broken connection without breaking a real one.
 type trackedConn struct {
 	net.Conn
-	failWrites atomic.Bool
-	closed     atomic.Bool
+	failWrites    atomic.Bool
+	closed        atomic.Bool
+	readDeadlines atomic.Int32
+}
+
+func (c *trackedConn) SetReadDeadline(t time.Time) error {
+	c.readDeadlines.Add(1)
+	return c.Conn.SetReadDeadline(t)
 }
 
 func (c *trackedConn) Write(b []byte) (int, error) {
--- a/pkg/runner/pty_recovery_test.go
+++ b/pkg/runner/pty_recovery_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/alpacax/alpamon/v2/pkg/config"
 	"github.com/alpacax/alpamon/v2/pkg/scheduler"
 	"github.com/gorilla/websocket"
+	"github.com/stretchr/testify/require"
 )
 
 // trackedConn fails writes on demand, so a test can play out a broken connection without breaking a real one.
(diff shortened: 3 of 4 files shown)


pkg/runner/client_close_test.go
```diff
@@ -0,0 +1,116 @@
+//go:build !windows
```
geunwoonoh: Warning: this `//go:build !windows` tag silently removes the new regression tests from the Windows CI job.

`WebsocketClient.Close()` has no Unix-specific behavior, and the backhaul client runs on Windows. `build-and-test.yml:218-222` states the repo's policy explicitly:

> Packages that are Unix-only by construction … are gated with `//go:build !windows` or per-test runtime.GOOS skips; **no silent blanket skip**. See alpamon issue #284 acceptance criteria.

The tag is not there because the test needs Unix—it is there only because `trackedConn` and `newWshServer` live in `pty_recovery_test.go`, which is `!windows` for pty reasons. I confirmed `GOOS=windows go vet ./pkg/runner/` passes, so the package compiles there today.

The fix inverts the dependency: put `trackedConn` and `dialTracked` in an untagged file (a new `conn_test.go`, or this file with the tag dropped). `pty_recovery_test.go` can still reference them—same package, and a `!windows` file may use symbols from an untagged one. That also makes commit `4bebacc`'s move unnecessary.

The last dependency standing is `newWshServer` on line 79; see the comment there.

pkg/runner/client_close_test.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	conn, _, err := dialer.Dial(url, nil)
+	require.NoError(t, err)
+	require.NotNil(t, tracked, "the dialer never went through NetDial, so nothing is being tracked")
+
+	return conn, tracked
+}
+
+func TestWebsocketClientClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
+	s := newWshServer(t)
```
geunwoonoh: Suggestion: the failure test pulls in `newWshServer`, which it does not need.

`newWshServer` is the Websh test double: it registers `reconnectPtyWebsocketURL` and mutates the global `config.GlobalSettings.ServerURL`. The backhaul close path uses none of that. With `failWrites` on, nothing reaches the peer at all, so the peer's behavior is irrelevant—`newCloseReplyServer` works here.

This drops the global-state mutation from the test and is the last dependency standing between the build-tag comment on line 1 and Windows coverage.

```suggestion
	s := newCloseReplyServer(t)
	conn, tracked := dialTracked(t, s.url)
```

geunwoonoh (commented):
### Summary

`WebsocketClient.Close()` returned early when the close-frame write failed, so `Conn.Close()` never ran. Since the write only fails on an already-broken connection—exactly the reconnect path—every reconnect dropped its fd on the floor for the runtime finalizer to collect. The fix moves the drain into the `else` branch and lets `Conn.Close()` run unconditionally, matching `pty.go:406-414`, `control_client.go:139-145`, and `tunnel_client.go:495-500`.

The diagnosis is right and the fix is the minimal correct one. Three regression tests, one of which was verified red against the pre-fix code. No critical issues.

I checked the concurrency claim in the PR body independently and it holds: `CommandRequestHandler` is called synchronously from `RunForever` (`client.go:152`), so the `MessageTypeReconnect` handler's `Close()` and `CloseAndReconnect()`'s `Close()` are sequential, never concurrent. Leaving `wc.Conn` non-nil is safe, and every writer of `wc.Conn` (`SendPingQuery`, `SendPongResponse`, `WriteJSON`) lives on that same goroutine.

### Critical issues (0)

None.

### Warnings (2)

**1. [`pkg/runner/client_close_test.go:1`] The `//go:build !windows` tag silently removes the new regression tests from the Windows CI job.**

`WebsocketClient.Close()` has no Unix-specific behavior, and the backhaul client runs on Windows. `build-and-test.yml:218-222` states the repo's policy explicitly:

> Packages that are Unix-only by construction … are gated with `//go:build !windows` or per-test runtime.GOOS skips; **no silent blanket skip**. See alpamon issue #284 acceptance criteria.

The tag is not there because the test needs Unix—it is there only because `trackedConn` and `newWshServer` live in `pty_recovery_test.go`, which is `!windows` for pty reasons. I confirmed `GOOS=windows go vet ./pkg/runner/` passes, so the package compiles there today.

The fix inverts the dependency: put `trackedConn` and `dialTracked` in an untagged file (a new `conn_test.go`, or `client_close_test.go` itself with the tag dropped). `pty_recovery_test.go` can still reference them—same package, and a `!windows` file may use symbols from an untagged one. That also makes commit `4bebacc`'s move unnecessary.

The remaining blocker is `newWshServer` in the failure test—see suggestion 5, which removes it.

**2. [`pkg/runner/pty_recovery_test.go:254-268`] The pty close test still hand-rolls the dialer that `dialTracked` now provides.**

`dialTracked` (`client_close_test.go:56-76`) is character-for-character the block the pty test already had. The PR was editing this file anyway, so the duplicate should go:

```go
func TestPtyClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
	s := newWshServer(t)
	conn, tracked := dialTracked(t, s.wsURL())

	pc := &PtyClient{conn: conn, sessionID: "close-test", manager: NewTerminalManager()}

	tracked.failWrites.Store(true)
	pc.close()

	require.True(t, tracked.closed.Load(), "close() did not close the websocket connection after WriteControl failure")
}
```

That also converts the two bare `t.Fatal` calls to testify, which `CLAUDE.md` asks for when you are already editing the test ("convert one while you are already editing it rather than as a separate sweep").

### Suggestions (4)

**3. [`pkg/runner/client.go:245-250`] The `IsCloseError` branch is dead—both branches just `break`.**

`websocket.IsCloseError` (gorilla v1.5.3, `conn.go:151-160`) returns true only for a non-nil `*websocket.CloseError`, so whenever it fires, the `if err != nil` two lines below would break anyway. The loop collapses to:

```go
		_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
		for {
			if _, _, err := wc.Conn.NextReader(); err != nil {
				break
			}
		}
```

Using `:=` also stops the drain from clobbering the outer `err`, which line 254 immediately reassigns—the current form makes a reader check whether that matters.

Pre-existing redundancy, but the PR re-indented every line of it, so it is cheap to clear now.

**4. [`pkg/runner/client.go:239-252`] `if err != nil { log } else { drain }` indents the normal path.**

Go Code Review Comments' "indent error flow" points the other way. Extracting the drain keeps `Close()` flat and gives the block a name:

```go
	if err != nil {
		log.Debug().Err(err).Msg("Failed to write close message to websocket.")
	} else {
		wc.drainCloseReply()
	}
```

Minor—the current form is readable and the `else` is what makes the fix obvious in the diff. Take it or leave it.

**5. [`pkg/runner/client_close_test.go:79`] The failure test pulls in `newWshServer`, which it does not need.**

`newWshServer` is the Websh test double: it registers `reconnectPtyWebsocketURL` and mutates the global `config.GlobalSettings.ServerURL`. The backhaul close path uses none of that. With `failWrites` on, nothing reaches the peer at all, so the peer's behavior is irrelevant—`newCloseReplyServer` works:

```go
	s := newCloseReplyServer(t)
	conn, tracked := dialTracked(t, s.url)
```

This drops the global-state mutation from the test and is the last dependency standing between warning 1 and Windows coverage.

**6. [`pkg/runner/client.go:254-257`] One extra Debug line per reconnect-message close.**

On the `MessageTypeReconnect` path the connection is now closed twice in sequence (handler `Close()`, then the read failure → `CloseAndReconnect` → `Close()`), so both "Failed to write close message" and "Failed to close websocket connection." now log where only the first did before. Harmless at Debug level; if it grates, filter the expected case:

```go
	if err = wc.Conn.Close(); err != nil && !errors.Is(err, net.ErrClosed) {
```

`errors` is already imported; `net` is not.

### Performance

No delta. The drain's timing is unchanged in both branches—the old code never drained on write failure either, it just returned early. Only the placement of `Conn.Close()` moved. The 5s worst-case drain against a peer that keeps sending data frames instead of a close frame is pre-existing and bounded by the single `SetReadDeadline`.

The actual win is the one the PR is named for: fds no longer accumulate across reconnect loops on long-lived agents.

### Duplicate / unused code

- Duplicate: the dialer block in warning 2.
- Dead: the `IsCloseError` branch in suggestion 3.
- Unused: none. Every field added to `trackedConn` (`readDeadlines`) and `closeReplyServer` (`url`, `closeCode`) is read by a test.

### Test coverage

Three tests covering the failure path, the success path, and idempotency. Not covered: the `wc.Conn == nil` early return (trivial) and a drain that ends on the read deadline rather than a close reply (would cost 5s of wall clock, not worth it). Good as is.

The assertions are honest ones. Reading whether the drain ran from the `SetReadDeadline` call count instead of from elapsed time is the right call—it cannot flake on a loaded runner. The PR body says the regression test was seen red against pre-fix code and the drain assertion was checked by deleting the drain; that is the level of rigor these tests should carry, and the body says plainly that `IsSafeToCallTwice` was never seen red and pins an assumption rather than guarding a regression.

### Highlights

- The root cause is stated as a decision, not a symptom, and the sibling clients were checked for the same bug before writing the fix.
- The choice *not* to nil `wc.Conn` is argued from the actual synchronization in the file rather than copied from `ControlClient`, and it is correct.
- Comments say why, not what: "the peer only answers a close frame that actually went out" is exactly the invariant that makes the `else` right.
- Commits are split cleanly: pure move, fix plus its regression test, remaining assertions.

---

**Verdict**: COMMENTED

The fix itself is correct and I would merge it. Warning 1 is what I would want addressed first: the repo has an explicit, documented policy against exactly this kind of silent Windows skip, and suggestion 5 plus the untagged-helper move clears it in a few lines.

**Checklist**:
- [x] Code follows stack conventions (`gofmt` clean, `go vet` clean, both run locally)
- [x] Tests included; quality verified by reading them
- [ ] Tests run on every supported platform: excluded from Windows CI
- [x] No documentation change needed

pkg/runner/pty_recovery_test.go
```diff
@@ -223,24 +248,6 @@ func TestPtyRecovery_StormNoGoroutineLeak(t *testing.T) {
...
-	return c.Conn.Write(b)
-}
-
-func (c *trackedConn) Close() error {
-	c.closed.Store(true)
-	return c.Conn.Close()
-}
-
 func TestPtyClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
 	s := newWshServer(t)
```
geunwoonoh, line 226: Warning: this test still hand-rolls the dialer that `dialTracked` now provides.

`dialTracked` (`client_close_test.go:56-76`) is character-for-character the block this test already had. The PR was editing this file anyway, so the duplicate should go:

```go
func TestPtyClose_ClosesConnAfterWriteControlFailure(t *testing.T) {
	s := newWshServer(t)
	conn, tracked := dialTracked(t, s.wsURL())

	pc := &PtyClient{conn: conn, sessionID: "close-test", manager: NewTerminalManager()}

	tracked.failWrites.Store(true)
	pc.close()

	require.True(t, tracked.closed.Load(), "close() did not close the websocket connection after WriteControl failure")
}
```

That also converts the two bare `t.Fatal` calls to testify, which `CLAUDE.md` asks for when you are already editing the test ("convert one while you are already editing it rather than as a separate sweep").

pkg/runner/client.go
```diff
@@ -238,21 +238,20 @@ func (wc *WebsocketClient) Close() {
...
-
-	_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second))
-	for {
-		_, _, err = wc.Conn.NextReader()
-		if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
-			break
-		}
-		if err != nil {
-			break
+	} else {
```
geunwoonoh, line 244: Suggestion: `if err != nil { log } else { drain }` indents the normal path.

Go Code Review Comments' "indent error flow" points the other way. Extracting the drain keeps `Close()` flat and gives the block a name:

```go
	if err != nil {
		log.Debug().Err(err).Msg("Failed to write close message to websocket.")
	} else {
		wc.drainCloseReply()
	}
```

Minor—the current form is readable and the `else` is what makes the fix obvious in the diff. Take it or leave it.

pkg/runner/client.go
```diff
@@ -238,21 +238,20 @@ func (wc *WebsocketClient) Close() {
...
+		_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
+		for {
+			_, _, err = wc.Conn.NextReader()
+			if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseGoingAway) {
```
geunwoonoh: Suggestion: this `IsCloseError` branch is dead—both branches just `break`.

`websocket.IsCloseError` (gorilla v1.5.3, `conn.go:151-160`) returns true only for a non-nil `*websocket.CloseError`, so whenever it fires, the `if err != nil` two lines below would break anyway. The loop collapses to:

```go
		_ = wc.Conn.SetReadDeadline(time.Now().Add(5 * time.Second)) // the peer only answers a close frame that actually went out
		for {
			if _, _, err := wc.Conn.NextReader(); err != nil {
				break
			}
		}
```

Using `:=` also stops the drain from clobbering the outer `err`, which line 254 immediately reassigns—the current form makes a reader check whether that matters.

Pre-existing redundancy, but the PR re-indented every line of it, so it is cheap to clear now.

pkg/runner/client.go
```diff
@@ -238,21 +238,20 @@ func (wc *WebsocketClient) Close() {
...
+				break
+			}
+			if err != nil {
+				break
+			}
 		}
 	}
 
-	err = wc.Conn.Close()
+	err = wc.Conn.Close() // a broken connection must not leak its fd across reconnects
```
geunwoonoh: Suggestion: one extra Debug line per reconnect-message close.

On the `MessageTypeReconnect` path the connection is now closed twice in sequence (handler `Close()`, then the read failure → `CloseAndReconnect` → `Close()`), so both "Failed to write close message" and "Failed to close websocket connection." now log where only the first did before. Harmless at Debug level; if it grates, filter the expected case:

```go
	if err = wc.Conn.Close(); err != nil && !errors.Is(err, net.ErrClosed) {
```

`errors` is already imported; `net` is not.

pkg/runner/client_close_test.go
jisung-02: Fixed in a1c4340. You were right that the tag was about where the helpers lived, not about the code under test. `trackedConn` and `dialTracked` moved to an untagged `pkg/runner/conn_test.go`, and `client_close_test.go` dropped the tag. `pty_recovery_test.go` keeps its `!windows` tag and still uses both helpers.

`GOOS=windows go list -f "{{.TestGoFiles}}" ./pkg/runner/` now lists `client_close_test.go` and `conn_test.go`, and `GOOS=windows go vet ./pkg/runner/` is clean.

pkg/runner/client_close_test.go
jisung-02: Applied in a1c4340, in the same commit as the build tag since it was the blocker for it. The failure test now dials `newCloseReplyServer`, so the global `config.GlobalSettings.ServerURL` mutation is gone from it. As you noted, with `failWrites` on nothing reaches the peer, so the Websh double was doing no work there.

pkg/runner/pty_recovery_test.go
jisung-02: Fixed in 80fc471. The block was indeed identical to `dialTracked`, and the two bare `t.Fatal` calls became `require` in the same commit, per the CLAUDE.md line you quoted.

pkg/runner/client.go
jisung-02: Taken, in bc2e389. The drain now lives in `drainCloseReply` and `Close()` reads as three steps: write the close frame, drain the reply when it went out, close the connection.

One difference from your snippet: 8ba2c25 then made it a plain function taking the connection, `drainCloseReply(conn)`, rather than a method reading `wc.Conn`. See the reply on the `client.go:254` thread for why.

pkg/runner/client.go
jisung-02: Fixed in bc2e389, exactly as you wrote it. `websocket.IsCloseError` returns true only for a non-nil `*websocket.CloseError`, so the plain `err != nil` below already covered every case it caught, and both branches only break. Reading the error with `:=` also stops the drain from clobbering the outer `err`.

pkg/runner/client.go
jisung-02: Applied in 35cf82d, `!errors.Is(err, net.ErrClosed)` as you suggested; `net` is now imported. `TestWebsocketClientClose_IsSafeToCallTwice` used to print both "Failed to write close message to websocket." and "Failed to close websocket connection." and now prints only the first.

Related, from a second pass over the same function: `Close()` was reading `wc.Conn` four separate times, and `Connect()` assigns that field from the `RunForever` goroutine while `gracefulShutdown` (`cmd/alpamon/command/root.go:231`) calls `Close()` from the main goroutine. The field could change mid-call, so `Close()` could send the close frame to one connection and close another. 8ba2c25 captures it once into a local. The field race itself predates this branch, but this branch widened it: with the early return gone, every path now reaches `Conn.Close()`.

Two things I left alone, both noted in the PR body under Known limitations. The drain reads the socket while `RunForever` may be blocked in `ReadMessage()`, which gorilla does not allow, but that was already true on the success path before this branch and fixing it means giving reads one owner. And `ShutDown()` closes `ShutDownChan` with no guard while being reachable from the worker pool and from `RunForever`, which panics if two fire together; this branch does not touch it. No issues filed for either yet.

pkg/runner/client.go
```diff
@@ -223,39 +224,40 @@ func (wc *WebsocketClient) CloseAndReconnect(ctx context.Context) {
...
 }
 
-// Cleanly close the websocket connection by sending a close message
+// Cleanly close the websocket connection by sending a close message, waiting for
+// the peer's reply when that frame went out, and closing the connection either way.
 // Do not close quitChan, as the purpose here is to disconnect the WebSocket,
 // not to terminate RunForever.
 func (wc *WebsocketClient) Close() {
-	if wc.Conn == nil {
+	conn := wc.Conn // Connect() may swap wc.Conn from the RunForever goroutine mid-close; every step below must act on the one connection we started with.
```
geunwoonoh, line 232: Suggestion (follow-up, not blocking): this capture narrows the window rather than closing it.

Good catch on the swap—I had this wrong in my first review, where I claimed every `wc.Conn` access lived on the `RunForever` goroutine. It does not: `gracefulShutdown` (`cmd/alpamon/command/root.go:231`) calls `Close()` from the main goroutine while `root.go:187` runs `RunForever` in its own, and `pkg/plugin/plugin.go:207` and `pkg/pluginclient/client.go:197` do the same.

What remains: this read still races `Connect()`'s write at `client.go:209`. On the platforms Alpamon ships a pointer read is atomic, so you get old-or-new and never garbage, but `-race` would flag it if a test ever drove both paths concurrently, and `gracefulShutdown` can still close connection A while `RunForever` has already moved to B. The complete fix is what `ControlClient` does: guard the field with a mutex.

Same root cause as the first item under Known limitations (the drain reading while `RunForever` may be blocked in `ReadMessage()`). Both are "`wc.Conn` needs a single owner," and both are correctly out of scope for an fd-leak fix. Worth one follow-up issue covering the field, the reader, and unguarded `ShutDown()`.

pkg/runner/client_close_test.go
```diff
@@ -0,0 +1,92 @@
...
+	// Close() deliberately leaves wc.Conn in place: the reconnect handler closes it, then RunForever's next read failure sends CloseAndReconnect at the same conn.
+	s := newCloseReplyServer(t)
+	conn, tracked := dialTracked(t, s.url)
+
+	wc := &WebsocketClient{Conn: conn}
+
+	wc.Close()
+	require.True(t, tracked.closed.Load(), "the first Close() did not close the websocket connection")
+	require.NotPanics(t, wc.Close, "the second Close() on an already closed connection panicked")
+}
```
geunwoonoh, line 92: Suggestion (follow-up, not blocking): `8ba2c25` is the one change in this update with no assertion behind it.

A fourth test could pin it: give `trackedConn` a hook that swaps `wc.Conn` to a second connection from inside `Write`, then assert the first is closed and the second is not.

Whether that is worth pinning an internal detail is your call—the comment on `client.go:232` carries the intent either way, and I would not hold the merge for it.

geunwoonoh (dismissed):
### Review application status

All six findings applied. I checked each against the code, not against the reply.

| # | Finding | Commit | Verified |
| --- | --- | --- | --- |
| W1 | `!windows` tag hides tests from Windows CI | `a1c4340` | `GOOS=windows go list -f '{{.TestGoFiles}}'` now lists `client_close_test.go` and `conn_test.go`; `Build and Test (windows-amd64)` passes |
| W2 | Duplicate dialer in the pty close test | `80fc471` | 15-line block replaced by `dialTracked`; both `t.Fatal` calls became `require.True` |
| S3 | Dead `IsCloseError` branch | `bc2e389` | Loop is now `if _, _, err := conn.NextReader(); err != nil { break }` |
| S4 | `else` branch indents the normal path | `bc2e389` | Drain extracted to `drainCloseReply` |
| S5 | Failure test pulls in `newWshServer` | `a1c4340` | Dials `newCloseReplyServer`; the `config.GlobalSettings.ServerURL` mutation is gone |
| S6 | Extra Debug line per reconnect close | `35cf82d` | `!errors.Is(err, net.ErrClosed)` guard, `net` imported |

The helper placement is better than what I proposed: `trackedConn` and `dialTracked` went to a new untagged `conn_test.go` with a comment saying why it is untagged, rather than being folded into `client_close_test.go`. `pty_recovery_test.go` keeps its `!windows` tag and reuses both.

I re-ran `go vet ./pkg/runner/`, `GOOS=windows go vet ./pkg/runner/`, and `gofmt -l pkg/runner/` locally—all clean. All 25 CI checks pass.

### Correction to my previous review

I got one thing wrong and it is worth stating plainly, because I posted it as a reason the design was safe:

> Leaving `wc.Conn` non-nil is safe, and every writer of `wc.Conn` (`SendPingQuery`, `SendPongResponse`, `WriteJSON`) lives on that same goroutine.

That is false. I only enumerated call sites inside `pkg/runner`. `Close()` is also called from outside it, off the `RunForever` goroutine:

- `cmd/alpamon/command/root.go:231` (`gracefulShutdown`, main goroutine) against `root.go:187` (`go wsClient.RunForever(ctx)`)
- `pkg/plugin/plugin.go:207`
- `pkg/pluginclient/client.go:197`

So `Connect()` can assign `wc.Conn` while `Close()` is mid-call, and the pre-fix `Close()` read the field four separate times. @jisung-02 found this independently and fixed it in `8ba2c25` by capturing `conn := wc.Conn` once. Without that capture this PR would have widened the window: with the early return removed, every path now reaches `Conn.Close()`, so a swapped field could close connection B after writing the close frame to connection A.

That commit is the most valuable one in the update and it came from the author's own second pass, not from my review.

### Critical issues (0)

None.

### Warnings (0)

None.

### Suggestions (2, both follow-up, neither blocking)

**1. [`pkg/runner/client.go:232`] `wc.Conn` still has no owner; the capture narrows the window rather than closing it.**

`conn := wc.Conn` is still an unsynchronized read racing `Connect()`'s write at `client.go:209`. On the platforms Alpamon ships, a pointer read is atomic, so the observed value is old-or-new and never garbage—but `-race` would flag it if a test ever drove both paths concurrently, and `gracefulShutdown` can still close connection A while `RunForever` has already moved to B. The complete fix is what `ControlClient` does: guard the field with a mutex.

That is the same root cause as the first item under Known limitations (the drain reading the socket while `RunForever` may be blocked in `ReadMessage()`—gorilla permits one concurrent reader). Both are "`wc.Conn` needs a single owner," and both are correctly out of scope here. Worth one follow-up issue covering the field, the reader, and unguarded `ShutDown()`; the PR body notes none is filed yet.

**2. [`pkg/runner/client_close_test.go`] The "close the connection `Close()` started with" behavior has no test.**

`8ba2c25` is the one change in this update with no assertion behind it. It is testable: give `trackedConn` a hook that swaps `wc.Conn` to a second connection from inside `Write`, then assert the first is closed and the second is not. Whether that is worth pinning an internal detail is a judgment call—the comment on line 232 carries the intent either way.

### Test coverage

Unchanged in substance and now running on one more platform. The three tests still cover the failure path, the success path, and idempotency, and the drain assertion still reads the `SetReadDeadline` count rather than the clock.

The PR body's testing claims hold up where I could check them: the regression test was re-verified red *after* the refactoring commits by restoring the early return, which is the right thing to do when a fix gets moved around. The body also still says plainly that `IsSafeToCallTwice` was never seen red and pins an assumption rather than guarding a regression.

### Highlights

- The response went past the review. `8ba2c25` fixes a real concurrency bug that neither Copilot nor I caught, and it was found by re-reading the function rather than by working the comment list.
- Known limitations are documented in the PR body with file and line references, scoped as pre-existing, and honestly marked "no issue filed yet" instead of implying they are handled.
- Each reply names the commit that applied it and says where it diverged from the suggestion and why (`drainCloseReply` taking the conn rather than reading `wc.Conn`).
- Commits stayed atomic through five rounds of change: test move, test reuse, refactor, fix, log cleanup.

---

**Verdict**: APPROVED

Every finding is applied, the update fixes a bug beyond the review, 25/25 checks pass including Windows, and the two open items are pre-existing, documented, and out of scope for an fd-leak fix.

**Checklist**:
- [x] Code follows stack conventions (`gofmt`, `go vet`, `GOOS=windows go vet`, `golangci-lint` all clean)
- [x] Tests included and passing on every supported platform
- [x] No documentation change needed

pkg/runner/client.go
jisung-02: Agreed, and filed as #452. You are right that the capture only narrows the window: `conn := wc.Conn` is still an unsynchronized read racing `Connect()`'s write at `client.go:209`, and `gracefulShutdown` closing A while `RunForever` has moved to B is still reachable.

The issue covers all three items as one change rather than three patches, since they share the cause you named—`wc.Conn` has no owner:

1. the field itself: writer at `client.go:209`, unsynchronized readers at `:136`, `:178`, `:232`, `:357`, with `Close()` reached from `root.go:231`, `plugin/plugin.go:207`, and `pluginclient/client.go:197` off the `RunForever` goroutine
2. the concurrent reader: `drainCloseReply` calling `NextReader` at `client.go:255` while `RunForever` may sit in `ReadMessage` at `:178`
3. `ShutDown()` closing `ShutDownChan` unguarded at `client.go:263`, reachable from six call sites

`ControlClient`'s mutex at `control_client.go:29` is named there as the pattern to follow. Kept out of this PR: an fd-leak fix is the wrong place to change the concurrency model.

pkg/runner/client_close_test.go
jisung-02: Added in `6c87384`, and you were right that it was worth pinning—it was the one change on this branch with nothing asserting it, so inlining `wc.Conn` back into the three reads would have left every test green.

Built the way you described. `trackedConn` grows an `onWrite` hook, and the test swaps `wc.Conn` to a second connection from inside the close frame's write:

```go
wc := &WebsocketClient{Conn: first}
firstTracked.onWrite = func() { wc.Conn = second }

wc.Close()

require.True(t, firstTracked.closed.Load(), "Close() left the connection it started with open")
require.False(t, secondTracked.closed.Load(), "Close() closed the connection Connect() had just swapped in")
```

Seen red by restoring the old reads:

```
--- FAIL: TestWebsocketClientClose_ClosesTheConnItStartedWith (5.00s)
    client_close_test.go:106:
        	Error:      	Should be true
        	Messages:   	Close() left the connection it started with open
```

The five seconds are the drain running against a connection the peer sent no close frame to, ending on the read deadline—which is the second half of the same bug.

One deliberate divergence: the hook is a plain field, not an atomic, because the test sets it and the write reads it on the same goroutine. The rest of `trackedConn` is atomics, so there is a comment saying why this one is not. Once #452 puts a mutex on the field, this test is the place to drive the two paths from separate goroutines under `-race`.

geunwoonoh (approved):
### Review application status

Both follow-up suggestions resolved, each in the way I said would be appropriate.

| # | Suggestion | Disposition | Verified |
| --- | --- | --- | --- |
| 1 | `wc.Conn` still has no owner; capture only narrows the window | Filed as #452 (open) | Issue covers all three items as one change: the unsynchronized field, the concurrent reader, unguarded `ShutDown()`; names `ControlClient`'s mutex as the pattern |
| 2 | `8ba2c25` had no test behind it | Implemented in `6c87384` | Reproduced red locally, see below |

#452 is the right disposition for suggestion 1—I said the mutex was out of scope for an fd-leak fix, and it stayed out. The issue enumerates the exact reader sites (`client.go:136`, `:178`, `:232`, `:357`) against the writer at `:209`, and the three off-goroutine `Close()` callers. Keeping the concurrency-model change out of this PR is the correct call.

### Independent verification of `6c87384`

I did not rely on the failure output in the PR body. I reverted the capture in `client.go` (inlining the four `wc.Conn` reads back), ran the test, then restored:

```
--- FAIL: TestWebsocketClientClose_ClosesTheConnItStartedWith (5.00s)
    client_close_test.go:106:
        	Error:      	Should be true
        	Messages:   	Close() left the connection it started with open
```

Identical to what the PR body reports, down to the 5.00s. The failure lands on the root-cause assertion—the connection `Close()` started with was left open—not on a symptom that a wrapper could hide. The five seconds are the drain running against the swapped-in connection until its read deadline, which is the second half of the same bug, so the test would have caught it either way.

With the capture restored, all four close tests pass under `-race -count=2`, each at 0.00s:

```
--- PASS: TestWebsocketClientClose_ClosesConnAfterWriteControlFailure (0.00s)
--- PASS: TestWebsocketClientClose_DrainsReplyAfterSuccessfulHandshake (0.00s)
--- PASS: TestWebsocketClientClose_IsSafeToCallTwice (0.00s)
--- PASS: TestWebsocketClientClose_ClosesTheConnItStartedWith (0.00s)
ok  	github.com/alpacax/alpamon/v2/pkg/runner	1.472s
```

No clock dependence, no race-detector findings. `go vet`, `GOOS=windows go vet`, and `gofmt -l` are clean, and `GOOS=windows go list -f '{{.TestGoFiles}}'` still lists both new test files. All 25 CI checks pass.

### Critical issues (0) · Warnings (0) · Suggestions (0)

Nothing new. I looked specifically at the two things that could have gone wrong in this commit.

**The `onWrite` hook is a plain field among atomics** (`conn_test.go:22`). Correct as written: the test sets it and `Write` reads it on the same goroutine, so there is nothing to synchronize, and an `atomic.Pointer` here would be cargo-culted. The field carries a comment stating the constraint, which is what keeps the next person from setting it from a pty test where `Write` runs on another goroutine. No other test sets it today.

**The hook fires on every `Write`, not only the close frame.** Harmless here—it is installed after the dial, so the close frame is the only write that reaches it, and the swap is idempotent anyway.

`t.Cleanup(func() { _ = second.Close() })` is registered before `wc.Close()` runs, so the unclosed second connection is not leaked by the test that asserts it stays open.

### Highlights

- The test pins the behavior by driving the real failure mode—a field swap from inside the write—rather than by asserting on an internal that could be satisfied trivially.
- Both mirror assertions are present: the first connection closes *and* the second one does not. Either half alone would pass against a `Close()` that closed both.
- The divergence from my suggestion (plain field, not atomic) is called out explicitly with its reason, and points at where it will need to change once #452 lands.
- #452 was filed as one issue for one root cause instead of three patches, with call sites enumerated.

---

**Verdict**: APPROVED

Every finding from both rounds is applied or correctly deferred with an issue behind it. The last change is test-only, verified red without the fix and green with it, and `client.go` has not moved since my previous approval.

**Checklist**:
- [x] Code follows stack conventions (`gofmt`, `go vet`, `GOOS=windows go vet`, `golangci-lint` all clean)
- [x] Tests included and passing on every supported platform (25/25 checks)
- [x] No documentation change needed

Issue #446, opened 2026-09-01
Make InitDB testable and move InitTestDB out of the production file


jisung-02:
## Overview

`InitDB` in `pkg/db` cannot be tested as it stands. Three things block it, and the test-only helper `InitTestDB` sits in the same production file. This issue covers removing the three blockers and moving the helper into its own package.

## Background

`InitDB` (`pkg/db/db.go:19`) creates the database file, runs the migrations, and returns the ent client. It sits on the agent's startup path and has no tests today.

`InitTestDB` (`pkg/db/db.go:57`) is a test helper living in a production file. It is called from 15 test files in other packages, all under `pkg/collector/check/` (the realtime, hourly, and daily suites).

### 1. The path cannot be injected

`InitDB` decides where to store the file by calling `utils.DataDir()` (`pkg/db/db.go:20`), which returns a hardcoded `/var/lib/alpamon`, falling back to the working directory when that directory is absent. A test would therefore take a different code path depending on the machine, and would write to a real system path on any machine where the directory exists.

### 2. Failures kill the process

Every error path calls `os.Exit(1)` (`pkg/db/db.go:30`, `44`, `51`). That ends the test binary itself, so no failure path can be asserted without re-executing the test as a subprocess.

### 3. The driver is registered inside the function

`sql.Register("sqlite3", &sqlite.Driver{})` is called by both `InitDB` (`pkg/db/db.go:36`) and `InitTestDB` (`pkg/db/db.go:72`). `database/sql` panics when the same driver name is registered twice, so no test can call either function more than once, or call both in one process.

## Proposed change

- Extract an internal function that takes the path as an argument; `InitDB` becomes a thin wrapper that resolves the path through `utils.DataDir()` and calls it.
- Have that internal function return an `error`, and keep `os.Exit(1)` in `InitDB` alone.
- Register the sqlite driver once, at package level.
- Move `InitTestDB` into a `pkg/db/dbtest` package and update the imports and call sites in the 15 test files.

## Scope

- `pkg/db/db.go`: the three structural changes above
- `pkg/db/dbtest/`: new package holding `InitTestDB`
- The 15 test files under `pkg/collector/check/`: import and call-site updates
- `pkg/db/db_test.go`: new tests covering the success path against a temp directory, a second call against an existing file, and a migration failure surfacing as an error

`pkg/db/client_test.go` and `pkg/db/migrate_test.go` open their databases directly with `sql.Open`, so they are unaffected.

## Open question

[...]

PR #447, opened 2026-09-01
fix: close the db file handle InitDB only opened to create it

### LINKED ISSUE #445: Close the DB file handle in InitDB after the file is created
## Overview

`InitDB` opens the database file with `os.OpenFile` and never closes the handle. `pkg/db/db.go:26`

## Details

The only reason to call `os.OpenFile` here is to create the file when it is absent. Everything downstream takes the path, not the handle: `RunMigration` and the ent client each open their own connection to the file. Once `InitDB` has the file on disk, the handle it holds has no remaining reader, yet it stays open for the lifetime of the process.

`InitTestDB`, directly below, closes the handle at the same point and documents why. The two functions disagree on the same step.

## Impact

`InitDB` runs once per process, so this is a single descriptor and nothing breaks today. It matters because an open handle blocks file deletion on Windows, which is what motivated the close in `InitTestDB`, and because leaving the two siblings inconsistent invites the same pattern to spread.

## Proposed change

Close the handle right after the file is created, the way `InitTestDB` does, and use the `fileName` variable already in scope for the migration and the client manager instead of `dbFile.Name()`.

```go
dbFile, err := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
if err != nil {
	...
}
// OpenFile only creates the file when it is absent; the migration and
// the ent client take the path and open their own handles.
_ = dbFile.Close()
```

jisung-02:
## Background

`InitDB` (`pkg/db/db.go:19`) opens the database file with `os.OpenFile` and never closes the handle. Nothing downstream wants that handle: `RunMigration` takes the path and opens its own `database/sql` connection, and the ent client does the same through the sqlite driver. From the moment the file exists, the handle has no reader, but it stays open for the life of the process.

What the `os.OpenFile` call does contribute is its `0750` mode argument, and this PR originally described that wrongly. An earlier revision of this body said `O_CREATE` was the only reason for the call. It is not: the driver opens with `SQLITE_OPEN_READWRITE|SQLITE_OPEN_CREATE` (`glebarez/go-sqlite@v1.23.0` `sqlite.go:816`), so `RunMigration`'s `sql.Open` would create the file on its own. Opening through the driver alone leaves the file at `0644` minus umask, measured as `0644` under umask `022` on darwin/arm64, so the explicit `0750` is what keeps the metrics DB off other users' read path on a fresh install. Raised by @geunwoonoh in review; the correction is in this body and in a code comment, since the commit that carried the wrong claim is already pushed under review and is not being rewritten.

`InitTestDB`, directly below in the same file, already closes at that point and says why in a comment. It has to, because an open handle blocks `os.Remove` on Windows and the test teardown removes the file. `InitDB` never removes it, so what is left is one leaked descriptor rather than a failure. The reason to fix it is that the two siblings disagree on the same step, which invites the pattern to spread.

## Changes

Five commits, none of them more than one line of code.

`InitDB` closes the handle right after the file is created:

```go
_ = dbFile.Close() // The migration and the ent client open the path themselves.
```

`RunMigration` and `NewDBClientManager` now take the path variable already in scope instead of `dbFile.Name()`. The value is identical: `os/file.go:63` returns the string stored at open time, and on both unix (`os/file_unix.go:279`) and Windows (`os/file_windows.go:163`) that string is the argument passed to `OpenFile`, unnormalized and untouched by `Close`. The change is for the reader, since calling a method on a handle closed two lines above makes anyone reviewing this stop and check.

The second commit renames that local from `fileName` to `dbPath` in both `InitDB` and `InitTestDB`. It holds a path, and it sat two lines below the `dbFileName` constant, which holds the bare `"alpamon.db"`—two similar names for two different kinds of value. `dbPath` is what the package already calls this value in `client_test.go` and `migrate_test.go`. Local variables only; no signature or caller changes.

The third commit builds that path with `filepath.Join` instead of `fmt.Sprintf("%s/%s", …)`:

```go
dbPath := filepath.Join(dataDir, dbFileName)
```

The hard-coded separator was not wrong in the sense of failing. On Windows `utils.DataDir()` returns `C:\ProgramData\alpamon\data`, so the old expression produced `C:\ProgramData\alpamon\data/alpamon.db`, and both the Windows API and Go's `filepath` accept `/` as a separator there. What it produced was a path that did not match how the rest of the package writes one. `filepath.Join` picks the platform separator and cleans the result.

The fourth commit records in the code why the `os.OpenFile` call is still there at all:

```go
// O_CREATE at 0750 is the point: the sqlite driver would create the file at 0644.
```

Before the first commit, the `dbFile.Name()` call two lines down made the handle look load-bearing. Now `Close` is its only use, and without this line the block reads as ceremony that a later cleanup could delete, dropping the permission with it. The mode argument only carries weight on Unix; Windows ignores the permission bits.

Two of the five commits go past what #445 describes, which is the unclosed handle alone. That is deliberate: the rename came out of reading the same four lines, the `filepath.Join` change came out of review on this PR, and each is a single line in the function already being edited. Splitting either into its own PR would cost more review attention than it saves. Nothing else in the file was touched—in particular the ignored `filepath.Abs` error on lines 23 and 57 is left as it is, since handling it means adding an error branch, and that is a change of a different size.

The fifth commit trims `InitTestDB`'s close comment from five lines to one:

```go
_ = dbFile.Close() // On Windows an open handle blocks TearDownSuite's os.Remove.
```

Its first three sentences said what the fourth commit now states once, thirty lines up, for both functions. What is left is the part specific to the test path, in the same trailing form `InitDB` uses. The claim in it holds: `TearDownSuite` calls `os.Remove` on the db file and asserts the result (`pkg/collector/check/realtime/cpu/cpu_test.go:48`).

Two further suggestions from review are deferred, not dropped. The duplication between `InitDB` and `InitTestDB`, which this PR made fully visible, and the duplicate `sql.Register("sqlite3", …)` that `database/sql` would panic on if both functions ever ran in one process, are both recorded on #446 (https://github.com/alpacax/alpamon/issues/446#issuecomment-5487965780).

## Testing

Run on the pushed head `3960ba98`.

| Check | Result |
| --- | --- |
| `go test ./... -p 1 -count=1` (darwin/arm64) | 46 packages ok, 0 FAIL |
| `go build ./...` for linux/amd64, darwin/arm64, windows/amd64 | ok |
| `go vet ./...` | clean |
| `gofmt -l pkg/db/db.go` | clean |

<sub>No test covers these changes, and that is a known gap. `InitDB` cannot be called from a test as it stands: it resolves its path through `utils.DataDir()` rather than taking one, calls `os.Exit(1)` on every error, and registers the sqlite driver inside the function, which panics on a second call. Rather than widen this PR, the work of making it testable—and of moving `InitTestDB` out of the production file—is filed separately as #446.</sub>

Closes #445

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -18,7 +18,7 @@ const dbFileName = "alpamon.db"
 
 func InitDB() *ent.Client {
 	dataDir := utils.DataDir()
-	dbPath := fmt.Sprintf("%s/%s", dataDir, dbFileName)
+	dbPath := filepath.Join(dataDir, dbFileName)
 	if _, err := os.Stat(dataDir); os.IsNotExist(err) {
 		dbPath, _ = filepath.Abs(dbFileName)
 	}
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -61,12 +61,7 @@ func InitTestDB(path string) *ent.Client {
 		_, _ = fmt.Fprintf(os.Stderr, "Failed to open test db file: %v\n", err)
 		os.Exit(1)
 	}
-	// We only needed OpenFile to create-on-absence; the filename (and
-	// the sqlite driver, which opens its own handle) is what the
-	// migration and ent client consume. Keeping this handle open would
-	// prevent the test's TearDownSuite from os.Remove'ing the file on
-	// Windows, where open handles block deletion.
-	_ = dbFile.Close()
+	_ = dbFile.Close() // On Windows an open handle blocks TearDownSuite's os.Remove.
 
 	sql.Register("sqlite3", &sqlite.Driver{})
 
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -18,12 +18,12 @@ const dbFileName = "alpamon.db"
 
 func InitDB() *ent.Client {
 	dataDir := utils.DataDir()
-	fileName := fmt.Sprintf("%s/%s", dataDir, dbFileName)
+	dbPath := fmt.Sprintf("%s/%s", dataDir, dbFileName)
 	if _, err := os.Stat(dataDir); os.IsNotExist(err) {
-		fileName, _ = filepath.Abs(dbFileName)
+		dbPath, _ = filepath.Abs(dbFileName)
 	}
 
-	dbFile, err := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
+	dbFile, err := os.OpenFile(dbPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to open db file: %v.", err)
 		_, _ = fmt.Fprintf(os.Stderr, "Failed to open db file: %v\n", err)
@@ -36,13 +36,13 @@ func InitDB() *ent.Client {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
 
-	err = RunMigration(fileName, ctx)
+	err = RunMigration(dbPath, ctx)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to migrate db: %v.", err)
 		os.Exit(1)
 	}
 
-	dbManager := NewDBClientManager(fileName)
+	dbManager := NewDBClientManager(dbPath)
 	client, err := dbManager.GetClient()
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to get db client: %v.", err)
@@ -53,8 +53,8 @@ func InitDB() *ent.Client {
 }
 
 func InitTestDB(path string) *ent.Client {
-	fileName, _ := filepath.Abs(path)
-	dbFile, err := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
+	dbPath, _ := filepath.Abs(path)
+	dbFile, err := os.OpenFile(dbPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to open test db file: %v.", err)
 		_, _ = fmt.Fprintf(os.Stderr, "Failed to open test db file: %v\n", err)
@@ -72,13 +72,13 @@ func InitTestDB(path string) *ent.Client {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
 
-	err = RunMigration(fileName, ctx)
+	err = RunMigration(dbPath, ctx)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to migrate test db: %v.", err)
 		os.Exit(1)
 	}
 
-	dbManager := NewDBClientManager(fileName)
+	dbManager := NewDBClientManager(dbPath)
 	client, err := dbManager.GetClient()
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to get db client: %v.", err)
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -29,19 +29,20 @@ func InitDB() *ent.Client {
 		_, _ = fmt.Fprintf(os.Stderr, "Failed to open db file: %v\n", err)
 		os.Exit(1)
 	}
+	_ = dbFile.Close() // The migration and the ent client open the path themselves.
 
 	sql.Register("sqlite3", &sqlite.Driver{})
 
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
 	defer cancel()
 
-	err = RunMigration(dbFile.Name(), ctx)
+	err = RunMigration(fileName, ctx)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to migrate db: %v.", err)
 		os.Exit(1)
 	}
 
-	dbManager := NewDBClientManager(dbFile.Name())
+	dbManager := NewDBClientManager(fileName)
 	client, err := dbManager.GetClient()
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to get db client: %v.", err)
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -23,6 +23,7 @@ func InitDB() *ent.Client {
 		dbPath, _ = filepath.Abs(dbFileName)
 	}
 
+	// O_CREATE at 0750 is the point: the sqlite driver would create the file at 0644.
 	dbFile, err := os.OpenFile(dbPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to open db file: %v.", err)


pkg/db/db.go
```diff
@@ -18,30 +18,31 @@ const dbFileName = "alpamon.db"
 
 func InitDB() *ent.Client {
 	dataDir := utils.DataDir()
-	fileName := fmt.Sprintf("%s/%s", dataDir, dbFileName)
+	dbPath := fmt.Sprintf("%s/%s", dataDir, dbFileName)
 	if _, err := os.Stat(dataDir); os.IsNotExist(err) {
-		fileName, _ = filepath.Abs(dbFileName)
+		dbPath, _ = filepath.Abs(dbFileName)
 	}
```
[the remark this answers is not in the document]
jisung-02: Half applied, in 6a42b94e.

The separator point is right, so line 21 now reads `dbPath := filepath.Join(dataDir, dbFileName)`. Nothing was failing—Windows and Go's `filepath` both accept `/` there—but on Windows `utils.DataDir()` returns `C:\ProgramData\alpamon\data`, so the old `fmt.Sprintf` produced a path with mixed separators, unlike every other path this package builds. The trailing-separator case you raise cannot occur with the current `DataDir()`: all three platform implementations return either a constant with no trailing separator or a `filepath.Join` result, which is already cleaned.

The `filepath.Abs` error is left ignored, here and on line 56. When `Abs` fails, `dbPath` becomes `""`, `os.OpenFile("")` fails immediately, and the existing branch logs and exits—so the outcome is a less actionable message rather than a wrong path. Fixing that means adding an error branch, which is wider than a PR about closing one handle. It belongs with the testability work already filed as #446, where it is now recorded.

pkg/db/db.go
```diff
@@ -18,30 +18,31 @@ const dbFileName = "alpamon.db"
...
 
-	dbFile, err := os.OpenFile(fileName, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
+	dbFile, err := os.OpenFile(dbPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to open db file: %v.", err)
 		_, _ = fmt.Fprintf(os.Stderr, "Failed to open db file: %v\n", err)
 		os.Exit(1)
 	}
+	_ = dbFile.Close() // The migration and the ent client open the path themselves.
```
geunwoonoh, line 33: Warning: the surviving reason for `os.OpenFile` is now undocumented, and the PR body states it incorrectly.

The body says "The only reason for that call is `O_CREATE`, which creates the file when it is absent." That is not the only reason, and after this PR it is not even a load-bearing one. The driver opens with `SQLITE_OPEN_READWRITE|SQLITE_OPEN_CREATE` (`glebarez/go-sqlite@v1.23.0/sqlite.go:816`), so `RunMigration`'s `sql.Open` would create the file on its own.

What `os.OpenFile` still contributes is the `0750` perm argument: SQLite creates its database at `0644` minus umask, so dropping this block would silently make the metrics DB world-readable on every fresh install. The early, explicit `failed to open db file` error is the secondary benefit.

Nothing in the code says this any more. Before the PR, `dbFile.Name()` two calls down at least made the handle look load-bearing; now the block reads as pure ceremony, and the new comment only explains why closing is safe, not why opening happens at all. Together with the PR body's claim, this invites a future cleanup to delete these lines and loosen the permission without noticing.

```suggestion
	// O_CREATE with 0750 is the point: the sqlite driver would create the
	// file itself, but at its own 0644, leaving the metrics DB world-readable.
	dbFile, err := os.OpenFile(dbPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
	if err != nil {
		log.Error().Err(err).Msgf("failed to open db file: %v.", err)
		_, _ = fmt.Fprintf(os.Stderr, "Failed to open db file: %v\n", err)
		os.Exit(1)
	}
	_ = dbFile.Close() // The migration and the ent client open the path themselves.
```

pkg/db/db.go
```diff
@@ -18,30 +18,31 @@ const dbFileName = "alpamon.db"
...
 
 	sql.Register("sqlite3", &sqlite.Driver{})
```
geunwoonoh, line 35: Suggestion: two names for one driver, and a `Register` that panics on a second call.

Pre-existing and out of scope for this PR, but adjacent enough to flag for #446.

`sql.Register("sqlite3", …)` appears in both `InitDB` and `InitTestDB`, and `database/sql.Register` panics on a duplicate name. Nothing panics today only because each test package is its own binary and each calls `InitTestDB` once. Two suites in one package, or any process calling both functions, panics at startup.

The same line is also why this package refers to one driver by two names: `migrate.go:37` opens `"sqlite"` (registered by the blank import's `init`) while ent opens `"sqlite3"` (the alias registered here, which ent's sqlite3 dialect requires). Both are needed, but a one-line comment here would save the next reader the lookup.

pkg/db/db.go
```diff
@@ -18,30 +18,31 @@ const dbFileName = "alpamon.db"
...
 
-	err = RunMigration(dbFile.Name(), ctx)
+	err = RunMigration(dbPath, ctx)
 	if err != nil {
 		log.Error().Err(err).Msgf("failed to migrate db: %v.", err)
 		os.Exit(1)
 	}
 
-	dbManager := NewDBClientManager(dbFile.Name())
+	dbManager := NewDBClientManager(dbPath)
```
geunwoonoh, line 46: Suggestion: `InitDB` and `InitTestDB` are now near-identical, and this PR made them more so.

After the rename they differ in exactly three things: how `dbPath` is resolved, four words of log wording, and the length of the close comment. Everything from `sql.Register` down (lines 34-52 vs. 70-88) is character-for-character the same.

This is the largest duplication the diff touches, and converging the two was the stated motivation. Extracting `func openDB(dbPath string) *ent.Client` would remove about 18 duplicated lines. Correctly out of scope here under surgical-change discipline, and #446 already covers it, so this is just a note that the duplication is now fully visible and worth closing out there.

geunwoonoh (dismissed):
## Summary

`InitDB` opened the SQLite file with `os.OpenFile` and never closed it. The handle had no reader once the file existed, so the process held a dead descriptor for its lifetime. This PR closes it right after creation, switches the two downstream calls from `dbFile.Name()` to the in-scope path variable, renames that variable `fileName` → `dbPath` in both `InitDB` and `InitTestDB`, and builds the path with `filepath.Join` instead of `fmt.Sprintf("%s/%s", …)`.

The fix is correct, and the reasoning in the PR body holds up under checking:

- `File.Name()` returns the string passed to `OpenFile` verbatim and `Close` does not touch it, so `dbFile.Name()` → `dbPath` is a pure identity substitution.
- `DataDir()` returns no trailing separator in any of the three platform implementations (`paths_linux.go:7`, `paths_darwin.go:7`, `paths_windows.go:19`), so `filepath.Join` cannot collapse a segment the old `Sprintf` preserved. The change is separator correctness on Windows plus `Clean`, nothing more.
- Closing before `RunMigration` is what makes the Windows delete path safe, matching the already-correct `InitTestDB`.

## Critical issues (0)

None.

## Warnings (1)

**1. [`pkg/db/db.go:26-32`] The surviving reason for `os.OpenFile` is now undocumented, and the PR body states it incorrectly.**

The body says "The only reason for that call is `O_CREATE`, which creates the file when it is absent." That is not the only reason, and after this PR it is not even a load-bearing one. The driver opens with `SQLITE_OPEN_READWRITE|SQLITE_OPEN_CREATE` (`glebarez/go-sqlite@v1.23.0/sqlite.go:816`), so `RunMigration`'s `sql.Open` would create the file on its own. What `os.OpenFile` still contributes is the `0750` perm argument: SQLite creates its database at `0644` minus umask, so dropping this block would silently make the metrics DB world-readable on every fresh install. The early, explicit `failed to open db file` error is the secondary benefit.

Nothing in the code says this any more. Before the PR, `dbFile.Name()` two calls down at least made the handle look load-bearing; now the block reads as pure ceremony, and the comment on line 32 only explains why closing is safe, not why opening happens at all. Together with the PR body's claim, this invites a future cleanup to delete lines 26-32 and loosen the permission without noticing.

```go
	// O_CREATE with 0750 is the point: the sqlite driver would create the
	// file itself, but at its own 0644, leaving the metrics DB world-readable.
	dbFile, err := os.OpenFile(dbPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0750)
```

## Suggestions (3)

**1. [`pkg/db/db.go:19-89`] `InitDB` and `InitTestDB` are now near-identical, and this PR made them more so.**

After the rename they differ in exactly three things: how `dbPath` is resolved, four words of log wording, and the length of the close comment. Everything from `sql.Register` down (lines 34-52 vs. 70-88) is character-for-character the same. This is the largest duplication the diff touches, and converging the two was the stated motivation. Extracting `func openDB(dbPath string) *ent.Client` would remove about 18 duplicated lines. Correctly out of scope here under surgical-change discipline, and #446 already covers it, so this is a note that the duplication is now fully visible and worth closing out there.

**2. [`pkg/db/db.go:34`] Two names for one driver, and a `Register` that panics on a second call.**

Pre-existing, but adjacent enough to flag for #446: `sql.Register("sqlite3", …)` appears in both functions, and `database/sql.Register` panics on a duplicate name. Nothing panics today only because each test package is its own binary and each calls `InitTestDB` once. Two suites in one package, or any process calling both functions, panics at startup. The same block is also why the package refers to one driver by two names: `migrate.go:37` opens `"sqlite"` (registered by the blank import's `init`) while ent opens `"sqlite3"` (the alias registered here, which ent's sqlite3 dialect requires). Both are needed, but a one-line comment at line 34 would save the next reader the lookup.

**3. [`pkg/db/db.go:63-68`] The two close comments now say the same thing at very different lengths.**

Five lines in `InitTestDB`, one in `InitDB`. Only the last sentence of the long one is specific to the test path (Windows blocking `os.Remove` in `TearDownSuite`). Trimming it to that sentence would leave the shared rationale stated once, in the shorter form the new comment already uses. Cosmetic.

## Convention, dead code, and performance

**Go convention**: clean. `gofmt -l pkg/db/db.go` is empty, `go vet ./pkg/db/...` is clean, `golangci-lint` passes in CI. `_ = dbFile.Close()` discards an error, which the stack rules generally forbid, but it is right here and consistent with the sibling: the file was opened write-only and never written, so there is no deferred flush error to lose, and a failed `Close` leaves nothing to recover. `filepath.Join` over a hand-built separator is the idiomatic form.

**Dead or unused code**: none introduced. `fmt` is still used by the two `Fprintf` calls at lines 29 and 60, so the `Sprintf` removal orphans no import, confirmed by clean builds for linux/amd64, windows/amd64 and darwin/arm64. `dbFile` is now referenced only by its `Close`, which is the intended shape. The one pre-existing item nearby is `InitTestDB` sitting in a production file and shipping in the binary; it adds no imports, so the cost is a few hundred bytes of text, and #446 owns moving it.

**Performance**: net positive, all of it negligible in magnitude. The real win is the one the PR is named for, one descriptor released instead of held for the process lifetime. `filepath.Join` is also cheaper than `fmt.Sprintf`, which pays for format parsing and interface boxing, but this runs once at startup so the difference is unmeasurable. One micro-observation, not worth acting on: line 21 builds and then discards the joined path when `dataDir` is missing (line 23). Restructuring into if/else to save one cold-path string allocation would cost more readability than it buys. No regressions anywhere in the diff.

## Test coverage

No tests added, and the PR body says so plainly rather than glossing over it. The stated reason checks out against the code: `InitDB` takes no path, resolves through `utils.DataDir()`, calls `os.Exit(1)` on every error branch, and registers the driver inside the function. It is not callable from a test as written, and making it callable is a larger change than this fix. Filing that as #446 instead of widening the PR is the right call. The existing `pkg/db` suite covers `RunMigration` and `sqliteDSN` and is unaffected.

## Verification

| Check | Result |
| --- | --- |
| CI on `6a42b94` | 25/25 pass (20 distro builds, windows-amd64, macos-arm64, CodeQL, golangci-lint, govulncheck, verify-tidy) |
| `go test ./pkg/db/ -p 1` (darwin/arm64) | ok |
| `go build ./...` for linux/amd64, windows/amd64, darwin/arm64 | ok |
| `go vet ./pkg/db/...` | clean |
| `gofmt -l pkg/db/db.go` | clean |

## Highlights

- The commit split is exemplary: behavior fix, rename and path construction are three commits of one line each, so each is independently revertable and reviewable.
- The PR body verifies rather than asserts. Citing `os/file.go:63` and both platform `Name()` implementations to prove `dbFile.Name() == dbPath` is exactly the evidence a reviewer would otherwise have to gather.
- Scope creep is disclosed, not hidden. The body names which two commits go past #445 and argues the cost of splitting them, which is the honest way to handle it.
- The Copilot round was answered item by item with a decision and a reason for each, including a documented "not applicable" backed by checking all three platform implementations.

---

**Verdict**: APPROVED

The single warning is a comment and PR-body accuracy issue, not a defect in the change. Worth fixing before merge, because the incorrect claim about `O_CREATE` will outlive the PR in the git log, but it does not block.

**Checklist**:
- [x] Code follows stack conventions (gofmt, vet, golangci-lint all clean)
- [x] Tests: none added, gap disclosed and filed as #446; existing suite passes, CI green
- [x] No security vulnerabilities introduced (see warning: the `0750` mode must survive future cleanup)
- [x] Documentation: no user-facing docs affected

pkg/db/db.go
jisung-02: Applied in 68b4d87b, as one line rather than two.

You are right on both halves, and I checked each before writing it. The driver does open with `SQLITE_OPEN_READWRITE|SQLITE_OPEN_CREATE` at `glebarez/go-sqlite@v1.23.0` `sqlite.go:816`, so `O_CREATE` is not load-bearing. And the mode does matter: a throwaway test that opened a fresh path through `sql.Open("sqlite", …)` and stat'd it printed `0644` under umask `022` on darwin/arm64, against `0750` for the file this block creates.

The comment landed shorter than your suggestion:

```go
// O_CREATE at 0750 is the point: the sqlite driver would create the file at 0644.
```

One line keeps it level with the `Close` comment right below it, and the pair then reads as why-open followed by why-close. The commit body carries the driver citation, the measured mode, and the note that the mode argument is Unix-only since Windows ignores the permission bits.

The PR body is rewritten too. It now states the `0750` reason and says outright that the earlier `O_CREATE` claim was wrong. The commit that carried that claim, 650dadfe, is pushed and this branch is under review, so I left the git log alone rather than force-push out from under your inline comments.

pkg/db/db.go
jisung-02: Not applied here; recorded on #446 instead, at https://github.com/alpacax/alpamon/issues/446#issuecomment-5487965780.

Agreed on both points. The panic is real and only latent because each test package is its own binary calling `InitTestDB` once, and the `"sqlite"` / `"sqlite3"` split does cost the next reader a lookup. But where the registration ends up—a package-level `init`, a `sync.Once`, or inside the extracted helper—is decided by the same restructure that makes `InitDB` testable and moves `InitTestDB` out of this file, so writing a comment now would likely describe a shape that changes. The issue comment carries the panic reasoning, the `migrate.go:37` vs. ent naming detail, and the note that whatever form it takes should carry that one-line comment.

pkg/db/db.go
jisung-02: Not applied here; recorded on #446 instead, at https://github.com/alpacax/alpamon/issues/446#issuecomment-5487965780.

Your reading matches mine: after the rename the two functions differ only in how `dbPath` is resolved, four words of log wording, and the length of the close comment. The issue comment names `func openDB(dbPath string) *ent.Client` as the extraction and the roughly 18 duplicated lines it removes, and ties it to the testability work already scoped there—once `InitDB` stops resolving its own path, the extracted helper is the piece a test can call directly.

geunwoonoh (approved):
Round 2, on `3960ba9`. All four findings from [my previous review](https://github.com/alpacax/alpamon/pull/447#pullrequestreview-5073408018) are closed out, and the diff got three lines smaller in the process.

## Status of the round-1 findings

| # | Finding | Status | Where |
| --- | --- | --- | --- |
| Warning 1 | `os.OpenFile`'s surviving reason undocumented; PR body claimed `O_CREATE` was the only reason | **Applied** | `68b4d87` — comment at `db.go:26`, PR body corrected |
| Suggestion 1 | `InitDB` and `InitTestDB` near-identical | **Deferred, properly** | #446, with the `openDB` extraction spelled out |
| Suggestion 2 | `sql.Register` panics on a second call; one driver, two names | **Deferred, properly** | #446, including the `migrate.go:37` naming detail |
| Suggestion 3 | The two close comments say the same thing at very different lengths | **Applied** | `3960ba9` — trimmed to the one test-specific sentence |

Nothing was silently dropped, and every inline thread got a reply naming where the item landed.

## Verification of the applied fixes

I checked the load-bearing claims rather than reading the commit messages and moving on.

**The `0750` claim is correct, and I measured it rather than assuming it.** A throwaway comparison of two fresh databases under umask `022` on darwin/arm64, one created by `sql.Open("sqlite", …)` alone and one pre-created the way `InitDB` does it:

```
driver-created mode = 0644   world-readable? true
InitDB-created mode = 0750   world-readable? false
```

That matches your stat exactly, so the new comment at `db.go:26` states the situation accurately: without that `os.OpenFile`, the metrics DB ships world-readable on a fresh install. The `SQLITE_OPEN_CREATE` half was already confirmed at `glebarez/go-sqlite@v1.23.0` `sqlite.go:816`.

**The trimmed comment's surviving claim is correct.** `TearDownSuite` does call `os.Remove` on the db file and assert the result (`pkg/collector/check/realtime/cpu/cpu_test.go:48`), and its own comment there records the Windows sharing violation an open handle causes. The sentence you kept is the one that was both true and specific to the test path.

**Both new commits are comment-only.** The cumulative diff against `main` carries no behavior change beyond what round 1 already had: the `Close`, the `dbPath` rename, and `filepath.Join`.

## New issues (0)

None. Two observations, neither worth acting on:

- `InitTestDB:58` has the identical `os.OpenFile(…, 0750)` with no `0750` comment, so the trimmed comment there leans on the reader having seen `InitDB` thirty lines up. That is the deliberate trade — state the shared rationale once rather than twice at two lengths — and it is the right call for a file #446 will restructure anyway.
- The one-liner says `0644` without "minus umask", and the rationale is Unix-only since Windows ignores the mode bits. Both caveats live in `68b4d87`'s commit body, which is the right home for them; a one-line comment should not carry them.

## Verification

| Check | Result |
| --- | --- |
| CI check-runs on `3960ba9` | 26/26 success |
| `go test ./pkg/db/... ./pkg/collector/... -p 1 -count=1` (darwin/arm64) | 16 packages ok, 0 FAIL |
| `go build ./...` for linux/amd64, windows/amd64, darwin/arm64 | ok |
| `go vet ./pkg/db/...` | clean |
| `gofmt -l pkg/db/db.go` | clean |

## Highlights

- The `0750` fact was measured with a throwaway stat rather than reasoned about, and the measurement is recorded in the commit body where the next reader will find it.
- `68b4d87` handles the already-pushed wrong claim honestly: no force-push over `650dadf`, which would have orphaned the inline threads, and a plain statement in both the commit body and the PR body that the earlier claim was wrong.
- Suggestion 3 was held back with a stated reason, then applied when that reason did not survive a second read, and the reversal is written down rather than quietly performed.
- The deferrals to #446 are real deferrals. The issue now carries the fix shape for both items, not just a pointer back to this PR.

---

**Verdict**: APPROVED

Re-approving because my round-1 approval was auto-dismissed when these two commits landed, leaving the PR `BLOCKED` on review with CI already fully green.

**Checklist**:
- [x] Code follows stack conventions (gofmt, vet, golangci-lint all clean)
- [x] Tests: none added, gap disclosed and filed as #446; existing suite passes, CI green
- [x] No security vulnerabilities; the `0750` mode is now documented so a future cleanup will not drop it
- [x] Documentation: no user-facing docs affected

PR #456, opened 2026-09-03
fix(runner): close the auth socket on every request handler exit

### LINKED ISSUE #451: auth.sock connection is not closed when the first read fails
## Overview

In `handleSudoRequest`, the Unix socket handler that serves PAM module requests over `auth.sock`, the branch that handles a failed first `Read` returns without closing the connection. Every other branch in the function closes it.

## Location

`pkg/runner/auth_manager.go:384`

```go
n, err := unixConn.Read(buf)
if err != nil {
    log.Warn().Err(err).Msg("Failed to read sudo request")
    am.sendIsAlpconResponse(unixConn, "", "", 0, 0, false)
    return          // no Close here
}
```

## Evidence

- The caller at `pkg/runner/auth_manager.go:300` hands the connection off with `go am.handleSudoRequest(unixConn)` and never closes it itself.
- The deferred function at the top of `handleSudoRequest` closes the connection only when `recover()` returns a value, so a normal return does not reach it.
- Every other branch of the same function closes explicitly, at lines 379, 394, 400, 410, 422, 428, and 438.

## Impact

Each time the PAM module connects and drops the connection without sending anything, one socket file descriptor stays open. Go's runtime finalizer on `net.Conn` eventually closes it, but with no guarantee about when, so on a host with frequent logins the descriptors can pile up until `accept` starts failing. It does not show up during normal operation; it shows up when a client dies mid-request or when something else connects to the socket by mistake.

## Reproduction

1. Run alpamon so that `auth.sock` is listening.
2. Connect to the socket and close it immediately without writing anything: `python3 -c "import socket; s=socket.socket(socket.AF_UNIX); s.connect('/run/alpamon/auth.sock'); s.close()"`. The socket lives under `utils.RunDir()` (`pkg/runner/auth_manager.go:253`), which is `/run/alpamon` on Linux, `/var/run/alpamon` on macOS, and `/tmp/alpamon` when alpamon runs as a non-root user.
3. `Failed to read sudo request` appears in the log, and `lsof -p <alpamon pid> | grep auth.sock` shows one more socket entry after every repetition.

## Suggested fix

Replace the scattered closes with a single `defer func() { _ = unixConn.Close() }()` at the top of `handleSudoRequest` and drop the per-branch `Close` calls, including the one inside the panic recovery defer. `handleSessionEvent` in `pkg/runner/auth_session_event.go:262` already follows this shape, and it keeps a newly added branch from repeating the same omission.

Adding `_ = unixConn.Close()` to the one branch also fixes the leak, but it leaves the root cause—closing by hand in every branch—in place.

jisung-02:
## Background

`startSocketListener` hands every accepted `auth.sock` connection to `handleSudoRequest` on its own goroutine, and that handler owns the connection's lifetime. Its close calls sat one per branch, and the first-`Read`-failure branch had none: it logged the error, answered with `is_alpacon_user: false` so PAM fails open, and returned. The descriptor then stayed open until the Go runtime's finalizer reclaimed it, at a moment nothing controls.

A PAM client that hangs up before writing takes exactly that branch, so a login interrupted at the wrong instant leaks a descriptor every time. The panic-recovery branch did close, but only on a panic.

## Changes

The close moves into the deferred function that already runs the panic recovery, so it happens on every exit: normal return, error return, and panic. The per-branch calls this makes redundant are removed—six in `handleSudoRequest`, four in `handleSudoApprovalRequest`, and the separate `defer` in `handleSessionEvent`, whose only caller is `handleSudoRequest`.

Two closes stay. `HandleSudoApprovalResponse` and `denyPendingRequests` run on a different goroutine, on a request already taken out of the session map, and they answer it before releasing the goroutine parked in `handleSudoApprovalRequest`. `handleSudoRequest` closes again once that waiter returns; `net.Conn.Close` is idempotent, so the second call only returns `net.ErrClosed`.

Moving the close to one place opened a window the first review round caught. `HandleSudoApprovalResponse` deregisters a request before it writes, so a 30s timeout or a cancelled context landing in that window finds nothing pending; the waiter used to return there having written nothing, and the deferred close then cut a response that was still going out. The waiter now parks on the completion signal both takers send after writing and closing, bounded by `authSocketHandoverTimeout` so a taker that died before signaling cannot strand the goroutine and the descriptor with it. `Stop()` only cancels the context, so shutdown is the path that puts every in-flight request through this branch at once.

The second round found that the bound this handover rests on was not the one the comment claimed. `denyPendingRequests` answers a whole session's detached requests and did so in sequence, so request k's completion signal sat behind every write before it: one peer that stopped reading pushes the later waiters past `authSocketHandoverTimeout`, and they close connections the loop has yet to write to. Each request now gets its own goroutine, so the bound only ever has to outlive one bounded write. A `sync.WaitGroup` keeps the function synchronous for its callers, which is the ordering `finalizeRequest` depends on, and the goroutine carries its own recover—on that path the work used to run under `handleSudoRequest`'s recover, and off that stack an unrecovered panic would take the whole agent down instead of costing one request.

`awaitTakenRequest` now takes the completion channel from its caller instead of re-reading the map. A lookup that missed would have returned a nil channel and parked for the full bound rather than failing, and `handleSudoApprovalRequest` already holds the channel from its registration. `completionChannel()` had no other caller and is gone.

The third round found the twin of that window one step earlier, in the code this PR rewrote. `finalizeRequest` returns without writing when a taker deregistered the request in the gap between the caller's pending check and its own lock, and the waiter then left the select with no handover wait at all—the same truncated response, reached before `awaitTakenRequest` could run. `finalizeRequest` now reports whether it took the request, and every caller holding the connection awaits when it did not. That answer is the only check the select needs: the take under `am.mu` already settles the race with the response path, so a pending check ahead of it only reopens the gap between two locks, and `isRequestPending` has no remaining caller. The retry-failure path never awaited on any branch and is the widest of the three, because a session teardown can take the request across the whole retry span.

The fourth round found the same shape twice more, both on the completion signal itself. The recover added to `denyPendingRequests`' goroutine skipped `signalCompletion` along with everything else when a write panicked, so the waiter parked the full `authSocketHandoverTimeout` and PAM waited six seconds for a body that was never coming—failing open late, where the rest of the design fails open at once. `HandleSudoApprovalResponse` has the same gap one step out: it takes the request and then returns on a marshal error without writing, closing, or signaling. Both signals are deferred now, registered after the recover in the first case so it runs first on the way out, and right after the take in the second so it also covers any early return added later between the take and the write. The marshal error is not reachable today, since `SudoApprovalResponse` carries strings, ints and bools only; it is closed because it is the same shape as the bug this PR exists to fix, a take without a guaranteed answer.

This also lines the handler up with its two siblings, `LogServer.handleConnection` in `pkg/logger/server.go:104` and `handleDaemonConnection` in `pkg/runner/tunnel_daemon.go:157`, which both open with a deferred close.

Comments describing the old ownership are updated in the same diff. `HandleSudoApprovalResponse`, `finalizeRequest`, `takeRequestLocked`, and `takePendingRequestsLocked` each claimed a sole owner for the close, which is no longer true. The invariant those comments protect is the response, not the close: the PAM module parses everything it reads up to EOF as a single JSON document, so a second response breaks it. `awaitTakenRequest` said takers signal "after writing and closing", which the panic path no longer satisfies, and `HandleSudoApprovalResponse` now states that every exit releases the waiter.

## Testing

Eight tests in `pkg/runner/auth_manager_cleanup_test.go`:

| Test | What it pins |
| --- | --- |
| `TestHandleSudoRequest_ClosesConnectionWhenReadFails` | the reported leak: the peer hangs up before writing |
| `TestHandleSudoRequest_ClosesConnectionOnEveryPath` | seven request shapes, each asserting EOF, the response, and the close |
| `TestHandleSudoRequest_AnswersBeforeClosingWhenReadFails` | the read-failure path still answers, so PAM does not wait out its own timeout |
| `TestHandleSudoRequest_ClosesConnectionOnPanic` | a panicking handler still closes |
| `TestHandleSudoApprovalRequest_ShutdownWaitsForTakenResponse` | a shutdown during a taken response still delivers it to PAM |
| `TestDenyPendingRequests_StalledPeerDoesNotDelayTheRest` | a peer that stopped reading holds up its own request only, not the rest of the batch |
| `TestHandleSudoApprovalRequest_RetryFailureWaitsForTakenResponse` | a teardown that takes the request during the retry span still delivers its response to PAM |
| `TestDenyPendingRequests_SignalsWhenAnsweringPanics` | a panicking write still releases the waiter promptly, rather than after the full handover bound |

Each one was seen red before being trusted:

- putting the close back inside the `recover()` branch fails all seven table cases and the other three tests
- deleting `sendIsAlpconResponse` from the read-failure path fails `TestHandleSudoRequest_AnswersBeforeClosingWhenReadFails`
- deleting the deferred close fails `TestHandleSudoRequest_ClosesConnectionOnPanic`
- removing the `awaitTakenRequest` calls fails `TestHandleSudoApprovalRequest_ShutdownWaitsForTakenResponse` on all three runs, with the handler logging `io: read/write on closed pipe`
- restoring the sequential loop in `denyPendingRequests` fails `TestDenyPendingRequests_StalledPeerDoesNotDelayTheRest` with "the second request's waiter was held behind the stalled peer's write"
- deleting the `awaitTakenRequest` call from the `ctx.Done` branch fails the `assert.False` that now backs the 100ms wait in `TestHandleSudoApprovalRequest_ShutdownWaitsForTakenResponse`, at 0.10s and on the wait itself rather than on the decode below it
- restoring the bare `finalizeRequest` call on the retry-failure path fails `TestHandleSudoApprovalRequest_RetryFailureWaitsForTakenResponse`, with the handler logging `No pending sudo request to finalize` and then `io: read/write on closed pipe`, and the PAM client reading EOF with no body
- moving the signal back after the work in `denyPendingRequests`' goroutine fails `TestDenyPendingRequests_SignalsWhenAnsweringPanics` at 2.00s, with "a panicking write left the waiter unsignaled, so it holds the connection until its handover bound expires"

Both handover tests take their decode goroutine from `readResponseAsync`: `net.Pipe` gives a taken write nowhere to land while a test drives the waiter out, so each of them has to keep reading. `TestHandleSudoRequest_ClosesConnectionOnPanic` asserts only the close, and now says why—`recover` runs wherever the panic landed, which may be after a response already went out, so answering from there could hand PAM a second one.

The marshal-error return in `HandleSudoApprovalResponse` has no test: the branch cannot be entered without a hook in the production path, and adding one for an unreachable return costs more than it pins.

The six tests that call `handleSessionEvent` directly now take their pipe from a helper that closes both ends on cleanup. Nothing leaked before—`net.Pipe` holds no descriptor—but the helper is what holds up the contract the docstring states.

`go test ./... -p 1` passes on the pushed head, 35 packages `ok` and no failures. `go test ./pkg/runner/ -race -p 1` passes at `ok github.com/alpacax/alpamon/v2/pkg/runner 47.071s`. `gofmt -l` and `go vet ./...` print nothing.

Closes #451


pkg/runner/auth_session_event.go
```diff
@@ -259,8 +259,6 @@ func (am *AuthManager) resolveSessionEvent(req SessionEventRequest) (NonAlpaconA
 // thus sshd) never waits on emission; the POST runs on its own
 // goroutine. Fail-open: every path answers the socket.
 func (am *AuthManager) handleSessionEvent(data []byte, unixConn net.Conn) {
-	defer func() { _ = unixConn.Close() }()
-
 	var req SessionEventRequest
 	if err := json.Unmarshal(data, &req); err != nil {
```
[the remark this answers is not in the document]
jisung-02: Documented the contract instead of restoring the close, in a6f41b0c.

The descriptor-leak half of this does not hold. `handleSessionEvent` has exactly one production caller, `handleSudoRequest` at `pkg/runner/auth_manager.go:425`, and that caller's deferred close runs on every exit—normal return, error return, and panic. The direct callers you spotted are the tests in `auth_session_event_test.go`, and they hand it a `net.Pipe()` end, which is an in-memory pipe holding no file descriptor at all.

The contract half does hold, so it is fixed. Re-adding a deferred close would put the connection back under two owners, which is precisely what this PR removes and what makes the handler match its two siblings, `LogServer.handleConnection` and `handleDaemonConnection`. Both close in the accept-loop handler and leave their helpers out of it. So the ownership is now stated where a future caller will read it: `handleSessionEvent` and `handleSudoApprovalRequest` each say closing `unixConn` is the caller's job, matching `sendSudoApprovalResponse`, which already said the same thing.

pkg/runner/auth_manager.go
```diff
@@ -435,16 +426,17 @@ func (am *AuthManager) handleSudoRequest(unixConn net.Conn) {
...
 	default:
 		log.Warn().Str("type", baseReq.Type).Msg("Unknown request type")
-		_ = unixConn.Close()
 	}
 }
 
+// handleSudoApprovalRequest answers one sudo_approval request, parking the
+// caller's goroutine until the server replies or the wait times out. Closing
+// unixConn is the caller's job.
 func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn) {
```
junho226, line 439: Warning: the two "already handled" branches below can now close the connection out from under the goroutine that is mid-write.

`HandleSudoApprovalResponse` removes the request under `am.mu` (`auth_manager.go:611-613`) and only then marshals, sets the write deadline, writes, and closes. If the 30s timer (`:516`) or `am.ctx.Done()` (`:525`) fires inside that window, `isRequestPending` returns false, the waiter returns having written nothing, and `handleSudoRequest`'s new deferred close fires—so the response write lands on a closed socket and PAM reads EOF with an empty body.

Before this PR the waiter closed nothing on that branch, so the write always landed. `Stop()` (`:864`) only cancels the context, so shutdown is exactly the path that fires `ctx.Done()` for every in-flight request at once.

The window is sub-millisecond and it fails closed (a sudo the server approved is denied), so this is not urgent—but it is new, and it is the one ownership hole the PR's reasoning does not cover. Minimal fix: in both branches, when `isRequestPending` is false, wait on `completionChan` before returning. Whoever took the request always signals after writing, so that is exactly the handover the close needs.

pkg/runner/auth_session_event.go
```diff
@@ -257,10 +257,9 @@ func (am *AuthManager) resolveSessionEvent(req SessionEventRequest) (NonAlpaconA
 // handleSessionEvent processes a session_event from the PAM session
 // hook. The ack is written before any server round-trip so PAM (and
 // thus sshd) never waits on emission; the POST runs on its own
-// goroutine. Fail-open: every path answers the socket.
+// goroutine. Fail-open: every path answers the socket. Closing
+// unixConn is the caller's job.
 func (am *AuthManager) handleSessionEvent(data []byte, unixConn net.Conn) {
```
junho226, line 262: Suggestion: six tests call this directly (`auth_session_event_test.go:332, 359, 382, 412, 424, 445`) and none of them close their pipe. The new docstring states the contract, which is the right call, but nothing enforces it.

No leak in practice—`net.Pipe` holds no descriptor, and those tests read a fixed buffer via `readSessionEventAck` rather than to EOF, so they still pass. Either close in the tests, or drive them through `handleSudoRequest` the way `TestHandleSudoApprovalRequest_LocalSudoRegistersNothing` now does.

pkg/runner/auth_manager.go
```diff
@@ -367,17 +367,13 @@ func (am *AuthManager) lookupSessionLocked(sid int, sidOK bool, parentPID int) (
...
 
 func (am *AuthManager) handleSudoRequest(unixConn net.Conn) {
-	// Panic recovery per the repo convention (internal/pool/pool.go,
-	// pkg/executor/executor.go). This handler used to run only on sudo
-	// invocations; with session events it runs on every PAM session open, so
-	// an unrecovered panic here would take the whole agent down from the login
-	// path. Close the socket on the way out so the PAM module reads EOF and
-	// fails open immediately instead of waiting out its own timeout.
+	// This runs on every PAM session open, so an unrecovered panic here would
+	// take the whole agent down from the login path.
```
junho226: Suggestion: the trim dropped the pointer to the repo convention (`internal/pool/pool.go`, `pkg/executor/executor.go`) that explains why the recover is here at all. The surviving half says what the panic would cost; the removed half said where the pattern comes from. Worth keeping both.

junho226 (dismissed):
## Summary

Moves `unixConn.Close()` out of the per-branch calls in `handleSudoRequest` and into the deferred function that already runs the panic recovery, so every exit path closes—including the first-`Read`-failure branch that had no close at all and leaked a descriptor whenever a PAM client hung up before writing. Ten redundant per-branch calls and `handleSessionEvent`'s own `defer` are removed, and the "single owner of the close" comments are reconciled to what is now true (the invariant is the *response*, not the close).

I verified the ordering claim the fix rests on: both `HandleSudoApprovalResponse` (`auth_manager.go:627-633`) and `denyPendingRequests` write, then close, then `signalCompletion`. Since the waiter in `handleSudoApprovalRequest` is released only by that signal, its return-and-close can never preempt the write. That part is sound.

## Critical issues (0)

## Warnings (1)

1. **[pkg/runner/auth_manager.go:435]** The two "already handled" branches in `handleSudoApprovalRequest` can now close the connection out from under the goroutine that is mid-write. Sub-millisecond window, fails closed, but new—see the inline comment.

## Suggestions (2)

1. **[pkg/runner/auth_session_event.go:262]** `handleSessionEvent` is called directly by six tests, none of which close their pipe. No leak in practice (`net.Pipe` holds no descriptor, and those tests read a fixed buffer rather than to EOF), but the new contract is unenforced.
2. **[pkg/runner/auth_manager.go:370]** The trimmed comment dropped the pointer to the repo convention that explains why the recover is there at all.

## Highlights

- Every new test was seen red first, and the three falsification checks in the PR body are the right ones (close back inside `recover()`, `sendIsAlpconResponse` deleted, deferred close deleted).
- `TestHandleSudoRequest_ClosesConnectionOnEveryPath` covers all seven returns as a table and asserts EOF *and* the response *and* the close—the EOF assertion is what actually pins the leak rather than proxying for it.
- The read deadlines on the client side mean a regression fails the test instead of hanging until the suite-wide panic.
- Lining the handler up with `LogServer.handleConnection` and `handleDaemonConnection` is the right consistency argument.

## Verdict: approved

- [x] Tests included
- [x] Comments reconciled with the new ownership

pkg/runner/auth_manager.go
jisung-02: Fixed in ab8a0826, with the test in a7e61a45. You were right that this branch introduced the window: before it, the waiter closed nothing on that path, so the taken response always landed.

The waiter now parks on `completionChan` when `isRequestPending` comes back false, which is the handover you suggested. Both takers signal only after writing and closing, so that signal is exactly the point where the deferred close becomes safe.

One thing added on top of the minimal fix: the wait is bounded by `authSocketHandoverTimeout` (`authSocketWriteTimeout + 1s`). Every taker writes under a 5s deadline and signals right after, so a longer silence means the signal is never coming—and an unbounded wait there would hold the goroutine and leak the descriptor this PR exists to free.

The test drives the real window rather than the symptom: a conn wrapper parks the first `Write`, the context is cancelled while the approved response is in flight, and the assertion is that the PAM client still reads `approved: true`. Reverted the fix to see it red—three runs, each failing with "the approved response never reached the PAM client" after the handler logged `io: read/write on closed pipe`.

pkg/runner/auth_session_event.go
jisung-02: Fixed in a618228c. The contract had nothing holding it up, which is the part that mattered—the six tests were callers too, and none of them closed.

Went with closing in the tests: a helper hands out both ends and closes them on `t.Cleanup`, so a test added later inherits the contract instead of having to remember it. Driving them through `handleSudoRequest` would enforce it through the production path, but it rewrites tests whose subject is `handleSessionEvent` itself.

No red to see here, since the open pipes were never why a test failed. What was checked instead is that the cleanup fires: a throwaway test called the helper inside a subtest and confirmed both ends return `io.ErrClosedPipe` on a write from the parent.

pkg/runner/auth_manager.go
jisung-02: Fixed in b9e6fbdc. Both halves are back:

```go
// Panic recovery per the repo convention (internal/pool/pool.go,
// pkg/executor/executor.go). This runs on every PAM session open, so an
// unrecovered panic here would take the whole agent down from the login path.
```

Checked both pointers still hold before writing them down—`internal/pool/pool.go:92` and `pkg/executor/executor.go:166` guard their handlers the same way.

pkg/runner/auth_manager.go
```diff
@@ -170,6 +170,9 @@ const (
 	// Command registration deny leftover requests inline, so a peer that stopped
 	// reading must not stall them.
 	authSocketWriteTimeout = 5 * time.Second
+	// authSocketHandoverTimeout outlives one bounded write, which is all a taker
+	// needs before it signals. A longer silence means the signal is never coming.
+	authSocketHandoverTimeout = authSocketWriteTimeout + time.Second
```
junho226, line 175: Warning: the comment above says this bound "outlives one bounded write, which is all a taker needs before it signals." That holds for `HandleSudoApprovalResponse`, and for `finalizeRequest`, which denies exactly one request. It does not hold for `denyPendingRequests`, which serves a whole session's detached requests in a loop and writes each one before it signals:

```go
for _, req := range pending {
    am.sendSudoApprovalResponse(req.Connection, req.Request, false, reason) // deadline: authSocketWriteTimeout
    _ = req.Connection.Close()
    am.signalCompletion(req.Request.RequestID)
}
```

Request *k*'s `signalCompletion` sits behind up to (k−1) × `authSocketWriteTimeout`. With a peer that stopped reading on an earlier request in the batch, request 3's waiter reaches the 6 s handover bound at T+6 while the loop only reaches request 3 at T+10: the waiter returns, `handleSudoRequest`'s deferred close fires, and the loop then writes to a closed connection. PAM reads EOF with no body—the failure `awaitTakenRequest` exists to prevent.

Reachability is narrow. It needs a multi-request session going through `takePendingRequestsLocked` (lines 700, 713, 772—session replaced or ended), a waiter already at the edge of its 30 s timeout or a shutdown, and a stalled peer. Worth flagging anyway because the interleaving is **new**: on `main` the waiter's early return closed nothing, since the taker owned the close. The shutdown path stays at N=1, because each waiter's `finalizeRequest` takes only its own request.

Two ways out, your call:

- Give each request its own goroutine in `denyPendingRequests`. It is already documented as running without `am.mu` held, so the per-request write and signal are independent.
- Or leave the behavior and amend this comment to say the bound covers a single-request take, and that a batch relies on the peers ahead of it draining. That at least stops the next reader trusting a property that is not there.

pkg/runner/auth_manager.go
```diff
@@ -532,15 +525,31 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
...
 }
 
+// awaitTakenRequest parks until whoever took this request has answered it. Takers
+// signal only after writing and closing, so an earlier return would let the caller's
+// deferred close cut a response mid-write and leave PAM an empty body to parse. The
+// bound is there because a taker that died before signaling must not strand this
+// goroutine with the descriptor open.
+func (am *AuthManager) awaitTakenRequest(requestID string) {
+	select {
+	case <-am.completionChannel(requestID):
```
junho226: Suggestion: take the channel as a parameter instead of re-reading the map.

`completionChannel()` returns the zero value for a missing key, and a receive on a nil channel blocks forever—so a missing entry does not fail, it degrades into a silent 6 s stall plus a warning that blames the taker for something the map did. Today that cannot happen: the only remover is `handleSudoApprovalRequest`'s own `defer` at line 497, which runs after this. But nothing in the code states that, and it is the kind of invariant a later change breaks without noticing.

`handleSudoApprovalRequest` already holds `completionChan` from line 495, so passing it makes the hazard structurally impossible and drops an `RLock` from the path:

```go
func (am *AuthManager) awaitTakenRequest(requestID string, completionChan <-chan struct{}) {
	select {
	case <-completionChan:
	case <-time.After(authSocketHandoverTimeout):
		log.Warn().Str("request_id", requestID).Msg("Taken sudo_approval request never signaled completion")
	}
}
```

With that, `completionChannel()` has no remaining caller and can go too.

pkg/runner/auth_manager_cleanup_test.go
```diff
@@ -376,3 +377,239 @@ func TestRemovePIDCommandMapping_DeniesPendingRequests(t *testing.T) {
...
+	}()
+
+	select {
+	case <-conn.entered:
+	case <-time.After(5 * time.Second):
+		t.Fatal("the response write never started")
+	}
+
+	cancel()
+	time.Sleep(100 * time.Millisecond) // Long enough for the waiter to close, if it is going to.
```
junho226: Suggestion: make the sleep carry an assertion.

The comment is honest about what the wait is for, and the test does fail loudly today if the waiter closes—you saw that with `io: read/write on closed pipe`. The risk is the other direction: on a loaded box, or under `-race`, 100 ms may elapse before the waiter has reached its close, and the test then passes without having exercised the window at all. A vacuous pass is worse than a flake, because nothing tells you it stopped testing.

Composing `blockingWriteConn` over `closeTrackingConn` lets the wait assert instead of hope, and puts the diagnosis on this line rather than on the decode further down:

```go
cancel()
time.Sleep(100 * time.Millisecond) // Long enough for the waiter to close, if it is going to.
assert.False(t, conn.isClosed(), "the waiter closed a connection whose response was still going out")
close(conn.release)
```

junho226 (dismissed):
## Summary

`handleSudoRequest` had its close calls one per branch and the first-`Read`-failure branch had none, so a PAM client that hangs up before writing leaked a descriptor until the runtime finalizer reclaimed it. The close moves into the deferred function that already runs the panic recovery, the eleven now-redundant per-branch calls come out, and `awaitTakenRequest` closes the window that move opened—`HandleSudoApprovalResponse` deregisters before it writes, so a waiter released in that window used to cut a response still going out.

I checked the surrounding invariants rather than taking them from the description:

- `handleSessionEvent`'s emit goroutine never touches `unixConn`—the ack is written synchronously before the round-trip—so moving its close up to the caller is strictly later and safe.
- The only remover of a completion channel is `handleSudoApprovalRequest`'s own `defer`, so `completionChannel()` cannot return nil while `awaitTakenRequest` is running.
- testify is already a dependency (`go.mod`) used across `pkg/runner`, so the new `assert`/`require` imports add nothing.

## Critical issues (0)

None.

## Warnings (1)

**1. [`pkg/runner/auth_manager.go:175`] `authSocketHandoverTimeout` assumes a taker that does one bounded write; `denyPendingRequests` does N of them, serially.** Request *k*'s signal sits behind up to (k−1) × `authSocketWriteTimeout`, so with a stalled peer earlier in the batch a waiter can hit the 6 s bound and close before the loop reaches its request. See the inline comment—this specific interleaving is new, since on `main` the waiter's early return closed nothing.

## Suggestions (2)

**1. [`pkg/runner/auth_manager.go:547`] Pass the channel the waiter already holds instead of re-reading the map.** Makes a nil-channel stall structurally impossible rather than accidentally absent, and drops an `RLock`.

**2. [`pkg/runner/auth_manager_cleanup_test.go:609`] The 100 ms sleep can go vacuous rather than flaky**—a false green, which is the worse direction. One assertion turns the wait into a check.

## Highlights

- The first review round's finding is fixed by **narrowing the wait**, not by widening a timeout or reordering until the symptom stops. `awaitTakenRequest` names the exact invariant it protects (takers signal only after writing and closing) and says why it is bounded at all.
- Every test was seen red first, with the specific mutation named per test—including the one that only fails on all three runs.
- The comments that stopped being true are corrected in the same diff. `HandleSudoApprovalResponse`, `finalizeRequest`, `takeRequestLocked` and `takePendingRequestsLocked` each claimed a sole owner for the *close*; the invariant that actually survives is one *response*, and that is now what they say. Leaving those stale would have been the expensive part of this change.
- Keeping the two remaining closes and explaining why (`net.Conn.Close` is idempotent, the second call returns `net.ErrClosed`) rather than chasing symmetry.
- `newSessionEventPipe` holds up a contract that changed even though nothing leaked—`net.Pipe` holds no descriptor.
- The handler now matches its two siblings, `LogServer.handleConnection` (`pkg/logger/server.go:104`) and `handleDaemonConnection` (`pkg/runner/tunnel_daemon.go:157`).

## Verdict: APPROVED

- [x] Tests included (CI is the pass gate)
- [x] No security vulnerabilities introduced
- [x] Documentation: comments corrected in the same diff; one comment overstates its bound (Warning 1)

pkg/runner/auth_manager.go
jisung-02: Fixed in c90d1b15, with the test in 1df81d8a. Took the first of your two options: `denyPendingRequests` now answers each request on its own goroutine, so request k's signal no longer sits behind the writes before it and the bound covers what the comment says it covers.

Amending the comment was the other way out, but it leaves the failure standing: a waiter that returns at T+6 while the loop reaches its request at T+10 hands PAM an EOF with no body, which is exactly what `awaitTakenRequest` exists to stop. Making the property real seemed better than documenting its absence.

A `sync.WaitGroup` keeps the function synchronous for its callers, since `finalizeRequest` returning after the response is written is what the waiter's deferred close depends on.

One thing added on top: the goroutine carries its own recover. On the `finalizeRequest` path this work used to run on the waiter's stack, under `handleSudoRequest`'s recover; off that stack an unrecovered panic takes the whole agent down instead of costing one request. The `pty.go` teardown paths never had that cover, so this only restores what the move took away.

The test parks the first request's write and asserts the second waiter is released within 2s. Reverted to the sequential loop to see it red: "the second request's waiter was held behind the stalled peer's write".

I did not add a semaphore around the fan-out. Past any cap the later requests queue behind a stalled peer again, which is the failure being fixed, and a batch is one session's pending sudo prompts rather than an open-ended stream.

pkg/runner/auth_manager.go
jisung-02: Applied as asked, in 24fe80b5. You were right that the hazard is not the miss itself but what a miss turns into: a nil channel does not fail, it degrades into a silent wait for the full bound plus a warning blaming the taker for something the map did.

`awaitTakenRequest` now takes `completionChan` as a parameter, and `completionChannel()` is deleted along with its `RLock`, exactly as you sketched.

pkg/runner/auth_manager_cleanup_test.go
jisung-02: Fixed in 2866c269. Composed `blockingWriteConn` over `closeTrackingConn` and put the assertion on the wait:

```go
cancel()
time.Sleep(100 * time.Millisecond)
assert.False(t, tracked.isClosed(), "the waiter closed a connection whose response was still going out")
close(conn.release)
```

Your point about the direction of the risk is the one that mattered: the test already failed loudly on a regression, but nothing stopped it from passing without entering the window at all.

Checked the assertion catches what the decode used to: deleting the `awaitTakenRequest` call from the `ctx.Done` branch now fails at 0.10s on this line rather than further down.

pkg/runner/auth_manager.go
```diff
@@ -532,15 +525,33 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
 			am.finalizeRequest(sudoApprovalReq.RequestID, "Response timeout")
```
junho226: **Critical:** the `req == nil` early return in `finalizeRequest` leaves the twin of the window this PR closes, and it is in the code the PR just rewrote.

Both select branches read:

```go
if am.isRequestPending(sudoApprovalReq.RequestID) {
    am.finalizeRequest(sudoApprovalReq.RequestID, "…")
} else {
    am.awaitTakenRequest(sudoApprovalReq.RequestID, completionChan)
}
```

`isRequestPending` releases `am.mu` before `finalizeRequest` re-acquires it. A taker that deregisters in that gap makes `finalizeRequest` find `nil`, log `No pending sudo request to finalize`, and return — and the waiter then falls out of the select with **no** `awaitTakenRequest`, so `handleSudoRequest`'s deferred close cuts the taker's in-flight write. That is the same failure `TestHandleSudoApprovalRequest_ShutdownWaitsForTakenResponse` was written for, reached one step earlier.

**Three call sites, not two.** `auth_manager.go:508` — the `sendSudoRequestWithRetry` failure — is the widest of them: a session teardown can take the request across the whole retry span, and that path has never awaited.

The fix is cheap because the channel is capacity 1 and `denyPendingRequests` signals before `wg.Wait()` returns, so an unconditional await is a no-op on the ordinary path:

```suggestion
		if am.isRequestPending(sudoApprovalReq.RequestID) {
			log.Warn().Msg("sudo_approval response timeout")
			am.finalizeRequest(sudoApprovalReq.RequestID, "Response timeout")
		} else {
			log.Debug().Msgf("sudo_approval timeout triggered but request already handled: %s", sudoApprovalReq.RequestID)
		}
		// Unconditional: finalizeRequest finds nothing when a taker won the gap
		// between isRequestPending and its own lock, and that is the same window
		// awaitTakenRequest exists for.
		am.awaitTakenRequest(sudoApprovalReq.RequestID, completionChan)
```

Same shape for `ctx.Done()`, plus an await after the `"Communication error"` finalize at line 508.

Alternatively have `finalizeRequest` return whether it took the request, and await when it did not. That spells the invariant out instead of leaning on the buffer, which matters because the buffer is what makes the unconditional version cheap and a future change to `signalCompletion` would silently turn the await into a 6s park.

Honest on probability: the taker only has to reach `Write` while the waiter acquires `am.mu` twice and unwinds two frames, so the taker usually wins. It is a race, not an ordering guarantee, and the payoff for losing it is the truncated body this PR exists to prevent.

pkg/runner/auth_manager.go
```diff
@@ -532,15 +525,33 @@ func (am *AuthManager) handleSudoApprovalRequest(data []byte, unixConn net.Conn)
...
 }
 
+// awaitTakenRequest parks until whoever took this request has answered it. Takers
+// signal only after writing and closing, so an earlier return would let the caller's
+// deferred close cut a response mid-write and leave PAM an empty body to parse. The
+// bound is there because a taker that died before signaling must not strand this
+// goroutine with the descriptor open. The channel comes from the caller rather than
+// the map: a lookup that missed would return nil and park here for the full bound
+// instead of failing.
+func (am *AuthManager) awaitTakenRequest(requestID string, completionChan <-chan struct{}) {
```
junho226, line 548: **Warning** (about the test that backs this function, not about the function).

`TestHandleSudoApprovalRequest_ShutdownWaitsForTakenResponse` rests its handover claim on `time.Sleep(100 * time.Millisecond)` followed by `assert.False(tracked.isClosed())`. The comment says the assertion is what keeps the wait from going vacuous, but it is the assertion that can pass vacuously: a waiter that has not reached its close yet within 100ms looks identical to one correctly parked here.

What actually carries the test is the `require.NoError(got.err)` / `assert.True(got.resp.Approved)` decode at the end — the response arriving whole is the property, and it is unconditional. Worth saying that in the comment so a later reader does not tighten the sleep and think they strengthened the test, or lengthen it and think they made it more reliable.

No change requested to the code here; `awaitTakenRequest` itself is right, including taking the channel from the caller rather than re-reading the map.

junho226 (changes requested):
## Summary

Moves `unixConn.Close()` into the deferred function that already carries the panic recovery, so it runs on every exit of `handleSudoRequest`, and removes the eleven per-branch calls that makes redundant. Adds `awaitTakenRequest` so a timeout or a cancelled context landing after a taker deregistered the request cannot close a response still being written, and makes `denyPendingRequests` answer each request on its own goroutine so `authSocketHandoverTimeout` only ever has to outlive one bounded write.

The core reasoning holds, and I verified the parts it rests on:

- `go.mod` declares `go 1.25.12`, so the parameterless `go func(){ … req … }()` in `denyPendingRequests` captures a per-iteration variable and is safe.
- `newCompletionChannelLocked` makes a capacity-1 channel, so a signal that beats the waiter's select is buffered rather than dropped.
- `HandleSudoApprovalResponse` writes, closes, and *then* signals (`auth_manager.go:653-655`), which is exactly the ordering `awaitTakenRequest` depends on.
- `handleSessionEvent` writes its ack synchronously before spawning the POST goroutine, so the caller closing on return cannot cut it.
- Deferred ordering in the new deny goroutine is right: `wg.Done()` is registered first, so it still runs after the recover.

One thing blocks it.

## Critical issues (1)

1. **`pkg/runner/auth_manager.go:525`** — the `req == nil` early return in `finalizeRequest` leaves the twin of the window this PR closes, on three call sites. Details inline.

## Warnings (1)

1. **`pkg/runner/auth_manager.go:547`** — `TestHandleSudoApprovalRequest_ShutdownWaitsForTakenResponse` rests its handover claim on a 100ms sleep plus `assert.False(isClosed())`, and that assertion is the one that can pass vacuously. Details inline.

## Suggestions (0)

None.

## Highlights

- Every test was seen red before being trusted, and the negative-control list names the exact mutation and the exact failure for each — including `deleting the awaitTakenRequest call from the ctx.Done branch fails … at 0.10s and on the wait itself rather than on the decode below it`, which is the distinction that separates a real pin from a coincidental one.
- `TestDenyPendingRequests_StalledPeerDoesNotDelayTheRest` pins a bound the comment used to claim without holding. Request *k*'s completion signal sitting behind every write before it only shows up under one stalled peer, and the test builds exactly that peer.
- The added `recover` in the deny goroutine is the right call and the reasoning is stated: off `handleSudoRequest`'s stack an unrecovered panic takes the agent down instead of costing one request.
- Updating the four ownership comments (`HandleSudoApprovalResponse`, `finalizeRequest`, `takeRequestLocked`, `takePendingRequestsLocked`) in the same diff, and naming the invariant they actually protect — the response, not the close — is what keeps the next reader from restoring a close that is no longer theirs.
- `TestHandleSudoRequest_ClosesConnectionOnEveryPath` asserting EOF, the response *and* the close per shape is the right table: a close-only assertion would have passed while PAM waited out its own timeout.

## Checklist

- [x] `go mod tidy` — no dependency change; `stretchr/testify v1.11.1` is already direct and already used across `pkg/runner`
- [x] No goroutine leaks — `denyPendingRequests` joins with `sync.WaitGroup`; `awaitTakenRequest` is bounded
- [x] Errors handled — writes keep their existing deadline and log level helper
- [x] Tests include edge cases — seven request shapes, panic, shutdown, stalled peer
- [ ] **Blocked**: the `finalizeRequest` nil path leaves the close/write race open on three sites

**Verdict: REQUEST CHANGES**

pkg/runner/auth_manager.go
jisung-02: Fixed in 3b935ae2, with the test in d15cf8e4. You were right that the twin was still there, and in the code this PR rewrote.

Took your second option. `finalizeRequest` now reports whether it took the request, and every caller holding the connection awaits when it did not. The unconditional version is cheap only because the channel is capacity 1, and you named the reason not to lean on that: a later change to `signalCompletion` would turn the await into a silent 6s park with nothing pointing at what changed.

One thing on top of the three call sites. The select's own `isRequestPending` checks are gone: the take under `am.mu` already settles the race with the response path, so a separate check ahead of `finalizeRequest` only reopens the gap between the two locks. Both branches now read the take's own answer, and `isRequestPending` has no remaining caller.

```go
case <-am.ctx.Done():
	log.Debug().Msg("Context cancelled, cleaning up sudo_approval connection")
	if !am.finalizeRequest(sudoApprovalReq.RequestID, "Service shutdown") {
		am.awaitTakenRequest(sudoApprovalReq.RequestID, completionChan)
	}
```

The test drives `:508`, the site you called the widest, because it is the one that can be made deterministic: a 500 holds the waiter inside its retry span, a session teardown takes the request and parks mid-write, and the context is then cancelled so the waiter reaches its finalize with nothing left to take. Reverted the fix to see it red, and the log is the failure you described:

```
"No pending sudo request to finalize"
"io: read/write on closed pipe"
--- FAIL: the teardown's response never reached the PAM client (EOF)
```

The two select branches have no test of their own. Their window needs the waiter paused between two lock acquisitions, and I did not want a test hook in the production path for it. `TestHandleSudoApprovalRequest_ShutdownWaitsForTakenResponse` still covers the `ctx.Done` branch reaching `awaitTakenRequest`, just not through this gap.

pkg/runner/auth_manager.go
jisung-02: Amended in a5335313—no code change, as you said.

```go
cancel()
// A cheap check that fails on this line rather than on the decode below. It can
// still pass without entering the window—a waiter that has not reached its close
// yet looks the same as one correctly parked—so neither tightening nor lengthening
// this sleep strengthens the test. What carries it is the decode at the end: the
// response arriving whole is the property, and that assertion is unconditional.
time.Sleep(100 * time.Millisecond)
```

The previous comment had it backwards: it credited the assertion with keeping the wait from going vacuous, when the assertion is the half that can pass vacuously. Wrote the same note into the new retry-failure test, since it copies this shape.

pkg/runner/auth_manager.go
```diff
@@ -923,14 +926,33 @@ func takePendingRequestsLocked(session *SessionInfo) []*SudoRequest {
...
+					log.Error().Interface("panic", r).
+						Str("request_id", req.Request.RequestID).
+						Msg("Denying a pending sudo request panicked")
+				}
+			}()
+			if req.Connection != nil {
+				am.sendSudoApprovalResponse(req.Connection, req.Request, false, reason)
+				_ = req.Connection.Close()
+			}
+			am.signalCompletion(req.Request.RequestID)
```
junho226: **Warning:** the new recover leaves `signalCompletion` unreached.

The recover keeps a panic in the write from taking the agent down, which is the right call. But `am.signalCompletion` sits *after* the work, so a panic skips it. The waiter then gets no signal, parks the full `authSocketHandoverTimeout`, and closes six seconds later — the PAM client waits six seconds for a body that is never coming, on exactly the path this PR exists to keep prompt. Failing open after 6s is worse than failing open at once, which is the reasoning `handleSudoRequest`'s own deferred close rests on.

A deferred signal makes the ordering structural instead of positional:

```go
		go func() {
			defer wg.Done()
			defer func() {
				if r := recover(); r != nil {
					log.Error().Interface("panic", r).
						Str("request_id", req.Request.RequestID).
						Msg("Denying a pending sudo request panicked")
				}
			}()
			defer am.signalCompletion(req.Request.RequestID)
			if req.Connection != nil {
				am.sendSudoApprovalResponse(req.Connection, req.Request, false, reason)
				_ = req.Connection.Close()
			}
		}()
```

Registered last, it runs first on the way out: still after the write and the close on the normal path, and during unwinding on the panic path, where the write has already failed and an early signal is the correct answer rather than a truncation risk.

pkg/runner/auth_manager.go
```diff
@@ -613,8 +624,9 @@ func logSudoResponseWriteError(err error, requestID string) {
 	log.WithLevel(authSocketWriteLevel(err)).Err(err).Str("request_id", requestID).Msg("Failed to send sudo_approval_response")
 }
 
-// HandleSudoApprovalResponse owns the response and close for the request it
-// takes, the way finalizeRequest does for the paths alpamon answers itself.
+// HandleSudoApprovalResponse answers and closes the request it takes, the way
+// finalizeRequest does for the paths alpamon answers itself; handleSudoRequest
+// closes again once the released waiter returns, which is harmless.
```
junho226: **Suggestion:** the same gap one step out. (About lines 644-648, outside the diff, so left on the comment this PR rewrote.)

`HandleSudoApprovalResponse` takes the request out of the map at `:634` and then returns at `:647` on a marshal error, without writing, closing, or signaling. The waiter's `finalizeRequest` finds nothing left to take, calls `awaitTakenRequest`, and times out — the same six-second stall, and PAM reads EOF with no body.

`json.Marshal` on `SudoApprovalResponse` cannot fail today: strings, ints and bools only, no `chan`, `func`, or `json.Marshaler`. So this is not reachable, and I would not block on it. It is worth closing because it is the same shape as the bug this PR fixes — a take without a guaranteed answer — and because the fix is one line in the same place as the warning above:

```go
	am.mu.Lock()
	sudoRequest := am.takeRequestLocked(response.RequestID)
	am.mu.Unlock()

	if sudoRequest == nil {
		…
	}

	defer am.signalCompletion(response.RequestID)
```

That also covers any future early return added between the take and the write, which is where this class of bug keeps coming back.

junho226 (dismissed):
## Summary

The close moves into the deferred function that already carries the panic recovery, so it runs on every exit, and `awaitTakenRequest` keeps that close from cutting a response a taker is still writing.

@junho226's standing CHANGES_REQUESTED blocked on the `finalizeRequest` nil path leaving the same window on three call sites. `3b935ae2` landed it: `finalizeRequest` now reports whether it took the request and every caller holding the connection awaits when it did not. Verified at head.

## Verified

- `go.mod` is `go 1.25.13`, so the parameterless `go func(){ … req … }()` in `denyPendingRequests` captures a per-iteration variable
- `newCompletionChannelLocked` makes a capacity-1 channel and `signalCompletion` sends non-blocking, so a signal that beats the waiter's select is buffered and a second one cannot panic
- both takers write, close, *then* signal (`auth_manager.go:653-655`, `:951-954`), and `authSocketHandoverTimeout` (6s) outlives the 5s deadline both writes set
- every exit of `handleSudoApprovalRequest` either never registered the request (`:444`, `:460`, `:466`, `:486`), holds a completion signal (`:518`), or awaits one (`:509`, `:530`, `:535`)
- `denyPendingRequests`' four external callers (`:692`, `:725`, `:749`, `:783`) all release `am.mu` before calling it, so the new `wg.Wait()` cannot deadlock against `signalCompletion`
- `removeCompletionChannel` running before the waiter's own close means a late signal finds nothing and no-ops rather than panicking
- `isRequestPending` has no remaining caller

## Critical issues (0)

## Warnings (1)

1. **`pkg/runner/auth_manager.go:954`** — the recover this PR adds leaves `signalCompletion` unreached on a panic, so the waiter pays the full `authSocketHandoverTimeout` before closing. Details and a one-line fix inline.

## Suggestions (1)

1. **`pkg/runner/auth_manager.go:629`** — the same gap one step out, in `HandleSudoApprovalResponse`'s marshal-error return. Unreachable today; the fix is the same line in the same place.

## Highlights

- Every test was seen red before being trusted, and the negative-control list names the mutation *and* the failure. `deleting the awaitTakenRequest call from the ctx.Done branch fails … at 0.10s and on the wait itself rather than on the decode below it` is the distinction that separates a real pin from a coincidental one.
- `TestDenyPendingRequests_StalledPeerDoesNotDelayTheRest` pins a bound the comment previously claimed without holding. Request *k*'s completion signal sitting behind every write before it only surfaces under one stalled peer, and the test builds exactly that peer.
- Updating the four ownership comments in the same diff, and naming the invariant they actually protect — the response, not the close — is what stops the next reader restoring a close that is no longer theirs.
- `TestHandleSudoRequest_ClosesConnectionOnEveryPath` asserting EOF, the response *and* the close per shape. A close-only table would have passed while PAM waited out its own timeout.
- Saying why `TestHandleSudoRequest_ClosesConnectionOnPanic` asserts no response, rather than leaving the asymmetry with its read-error twin to look like an oversight.

## Checklist

- [x] `go mod tidy` — no dependency change; `stretchr/testify` is already direct and already used across `pkg/runner`
- [x] No goroutine leaks — `denyPendingRequests` joins with `sync.WaitGroup`, `awaitTakenRequest` is bounded
- [x] Errors wrapped / handled — writes keep the existing deadline and log-level helper
- [x] Tests include edge cases — seven request shapes, panic, shutdown, stalled peer, retry failure
- [x] Context propagated — `am.ctx` unchanged; `Stop()` still only cancels

**Verdict**: APPROVED

pkg/runner/auth_manager.go
jisung-02: Fixed in be5fbfa3, with the test in 96b4585b. You were right about the direction of the cost: the recover keeps the agent up, but a signal placed after the work goes down with it, and failing open six seconds late is worse than failing open at once.

Took the deferred signal exactly as you sketched, registered after the recover so it runs first on the way out. On the ordinary path it still lands after the write and the close; on the panic path it runs during unwinding, where the write has already failed and there is no response left to truncate.

`awaitTakenRequest`'s doc comment went with it. It said takers signal "after writing and closing", which the panic path no longer satisfies, so it now states the property that actually holds—a taker signals once it is done with the connection.

The test builds the real window rather than the symptom: `panicOnWriteConn` blows up where the denied response is written, and the assertion is that the completion signal arrives within two seconds. The bound is deliberately well short of `authSocketHandoverTimeout`, since promptly is the property, not eventually. Moved the signal back after the work to see it red: `--- FAIL: TestDenyPendingRequests_SignalsWhenAnsweringPanics (2.00s)`, "a panicking write left the waiter unsignaled, so it holds the connection until its handover bound expires".

pkg/runner/auth_manager.go
jisung-02: Fixed in a6f2d78a. Closed it for the reason you gave rather than for reachability: it is the same shape as the bug this PR exists to fix, a take without a guaranteed answer, and the defer also covers any early return added later between the take and the write.

`defer am.signalCompletion(response.RequestID)` sits right after the nil check, and the inline call after the write and close is gone. The ordering the handover rests on is unchanged, since the defer runs at return, after both.

The comment above the write lost its second half—"the waiter has to be released whether the write landed or not" now sits on the defer that does it—and the function docstring says every exit releases the waiter, because the take leaves nobody else able to.

No test. `json.Marshal` on `SudoApprovalResponse` cannot fail today, as you said, so there is no way to enter the branch without a hook in the production path, and I did not want one for an unreachable return.

junho226 (approved):
## Summary

Delta review at `61e711db`, over my approving review at `e50ad719`. Both items it carried are in, and the branch has since taken two merges from `main`.

| Item | Landed |
| --- | --- |
| Warning, `auth_manager.go:954` — the deny goroutine's recover left `signalCompletion` unreached | `be5fbfa3` fix, `96b4585b` test |
| Suggestion, `auth_manager.go:629` — the same gap in `HandleSudoApprovalResponse`'s marshal-error return | `a6f2d78a` |
| Merges of `origin/main` (#457, #458, #459) | `81c211f9`, `61e711db` |

## Earlier findings

- **Critical, `finalizeRequest`'s `req == nil` path (`2866c269`)** — resolved at `3b935ae2`, test at `d15cf8e4`. Re-verified at head rather than taken on the earlier read: `finalizeRequest` reports whether it took the request (`:888-905`), and all three sites that hold the connection await when it did not (`:509` retry failure, `:530` timeout, `:535` `ctx.Done`). Taking the answer from the take itself instead of a separate `isRequestPending` check is the better half of that fix, and `isRequestPending` now has no caller anywhere in the tree.
- **Warning, the handover test's 100ms sleep (`2866c269`)** — answered at `a5335313`. Comment now says the sleep can pass without entering the window and that the decode is what carries the test. No code was owed here.
- **Warning, `denyPendingRequests` (`e50ad719`)** — fixed at `be5fbfa3` as sketched. `defer am.signalCompletion(...)` is registered after the recover (`:951-962`), so it runs first on the way out: still after the write and the close on the ordinary path, and during unwinding on the panic path. The recover below it only logs, so nothing there touches a connection the released waiter may already have closed.
- **Suggestion, `HandleSudoApprovalResponse` (`e50ad719`)** — fixed at `a6f2d78a`. The signal sits right after the take's nil check (`:648`) and the inline call after the write is gone; the ordering the handover rests on is unchanged, since the defer runs at return.

Nothing from either round is left open.

## Verified at head

- Every path that takes a request now owes and delivers a signal. Both `takeRequestLocked` callers (`:636`, `:890`) and all three `takePendingRequestsLocked` sites (`:706` via `putSessionLocked`, `:719`, `:778`) end in either the new defer or `denyPendingRequests`, whose per-request defer fires even when `Connection` is nil.
- The marshal-error exit deliberately leaves the connection open; the released waiter's deferred close in `handleSudoRequest` is what closes it, so PAM reads EOF at once rather than at `authSocketHandoverTimeout` (`authSocketWriteTimeout + 1s`, `:172-175`).
- `TestDenyPendingRequests_SignalsWhenAnsweringPanics` asserts the signal within two seconds, well short of the six-second bound, so what it pins is promptly rather than eventually — which is the whole point of the finding.
- The two merges are clean. `origin/main` is an ancestor of head, and the diff against it is the PR's own four files and nothing else, so none of #457, #458 or #459 was reverted in the resolution.
- `gofmt -l pkg/runner` is silent and `go vet ./pkg/runner/` exits 0 at head, which type-checks the test files too — so the merge left the package compiling. I did not run the suite; the run at the pushed head is recorded in the PR thread.

Agreed on leaving the marshal branch untested: `SudoApprovalResponse` carries strings, ints and bools only, and a hook in the production path costs more than an unreachable return pins.

## Critical issues (0)

## Warnings (0)

## Suggestions (0)

## Highlights

- Both fixes make the release structural rather than positional. A defer registered next to the take covers early returns that do not exist yet, which is where this class of bug kept coming back — the marshal return was itself the second instance of it.
- `awaitTakenRequest`'s doc comment moved with the code. It claimed takers signal "after writing and closing", which the panic path no longer satisfies; it now states the property that actually holds.
- `panicOnWriteConn` blows up at the write rather than simulating the symptom, and the negative control is named: moving the signal back after the work fails at 2.00s with the message the test exists to produce.

## Checklist

- [x] `go mod tidy` — the PR touches no dependency; `go.mod` moved only under the merge from `main`
- [x] No goroutine leaks — `denyPendingRequests` joins with `sync.WaitGroup`, `awaitTakenRequest` is bounded
- [x] Errors handled — writes keep the existing deadline and log-level helper
- [x] Tests include edge cases — seven request shapes, panic on write, shutdown, stalled peer, retry failure
- [x] Documentation — the two comments the new defers invalidated were corrected in the same commits

**Verdict**: APPROVED. My standing CHANGES_REQUESTED is lifted — its blocking item landed at `3b935ae2` and I re-verified it at head, and nothing in the four commits since has reopened it.

Issue #379, opened 2026-07-24
feat(editor): support web editor on Windows


jisung-02:
## Background

The editor (code-server) client type is unsupported on Windows. `startCodeServerProcess` in `tunnel_windows.go` returns `"code-server is not supported on Windows"`.

The root cause is the engine, not the agent code. code-server (Coder) does not officially support Windows natively, and the install path in `editor.go` (`installCodeServer`) assumes Unix by piping `install.sh` through `sh`.

This issue depends on Phase 1 (#378) — Windows support for cli/web tunneling — landing first.

## Prerequisite decision: editor engine

Windows support and licensing differ per engine, so the engine must be chosen before implementation.

| Engine | Windows | Notes |
|---|---|---|
| code-server | Unsupported | WSL only. Not suitable for a server agent. |
| VS Code `code serve-web` | Supported | Official MS CLI. Needs a license review around redistribution / commercial bundling. |
| openvscode-server (Gitpod) | Unverified | Need to verify whether releases ship a Windows build. |

`code serve-web` is the most realistic technically, but the final choice depends on the license review.

## Changes

Once the engine is chosen:

- Install: split the hardcoded `sh` in `installCodeServer` into a platform-specific `installEditor()`. Windows would download and unpack a zip directly, or use winget.
- Run: follow the darwin pattern — run as the current user without demotion. The chown in `setupUserDataDir` is already guarded by `os.Getuid() == 0`, so it is skipped automatically on Windows.
- Process tree management: `Setpgid` is Unix-only. Editor engines spawn multiple node child processes, so a Job Object is needed to group and kill the whole tree on shutdown (if `terminateProcess` in `process_windows.go` only kills a single process, orphaned node processes remain).
- Engine abstraction: `--idle-timeout-seconds`, `--user-data-dir`, and the config.yaml format are code-server-specific. If the engine changes, `CodeServerConfig.ToArgs`/`ToConfigYAML` must move behind an engine abstraction.
- `Username` parameter semantics: the required `username` validation for editor tunnels (`tunnel.go:88-90`) is not used for demotion on Windows, so it should be clarified at the protocol level — used only to resolve the home directory, or ignored.

## References

- `pkg/runner/editor.go`
- `pkg/runner/tunnel_windows.go`
- `pkg/runner/tunnel_linux.go` / `tunnel_darwin.go`
- `pkg/runner/process_windows.go`
- Depends on: Phase 1 issue

PR #457, opened 2026-09-03
fix(runner): resolve the editor user data dir one component at a time

jisung-02:
## Background

`setupUserDataDir` builds `$HOME/.alpamon-editor` for code-server: `config.yaml`, `User/settings.json`, and, when alpamon runs as root, an ownership pass so the demoted user can edit them afterwards. It did all three with `os.MkdirAll`, `os.WriteFile`, and a `filepath.Walk` of `os.Chown`.

Every one of those resolves a whole path, and resolving a path follows the symbolic links in it. The home directory belongs to the user rather than to alpamon, so its contents are not something this function gets to assume: a link at `.alpamon-editor/User/settings.json` sends the write to wherever the link points, and the `Walk` hands the chown to the link target instead of the link. The function manages one directory tree and should act on nothing outside it, whatever the tree happens to contain when it runs.

Two smaller things came out of the same reading. `UserDataDirName` reaches the join from `CodeServerConfig` with nothing checking its shape, and the writes used `O_TRUNC`, which empties the file before it refills it.

### One component at a time

`setupUserDataFiles` and `chownTreeNoFollow` replace the path-based calls (`pkg/runner/editor_unix.go`). Both hold a directory file descriptor and resolve a single name against it at each step, with `O_NOFOLLOW` on the open, so a link is refused rather than followed. Nothing is resolved by name twice.

The home directory's own entry is the one exception. It lives in a directory the user does not own, so it is not theirs to swap, and admins do point home directories at another filesystem with a symlink. That open gets its own flag set without `O_NOFOLLOW`; every step under it keeps the guard.

`O_NOFOLLOW` covers only the last component of the name it is handed, so every name these helpers take has to be one plain component. Three of them are literals, now named in `editor.go`; `ToArgs` reads `config.yaml` back for the `--config` flag. The fourth, `UserDataDirName`, is configuration, so `validateUserDataDirName` rejects it up front when it is not a single usable component. `.` needs its own clause: `filepath.IsLocal` accepts it, and joining it gives back the home directory.

### Nothing outside the tree

The walk re-owns what the tree is the only name for, and nothing else. Each entry is opened once and the stat, the chown and the descent all go through that descriptor, so the inode a check clears is the inode the chown reaches. Resolving the name a second time to act on it is what would let the user rename something else onto it in between, and `rename` inside their own directory is atomic, needs no privilege, and can be retried against every editor start.

A hard link is not a link to a file: it is a second name for the same inode. A user who plants `.alpamon-editor/x` as a hard link to a file they can read would have had root hand them its ownership, so an entry whose link count is above one is skipped. Directories are exempt, since `.` always gives them a second link. Every other type is not: `link` does not care what the inode is, and `fs.protected_hardlinks` is 0 unless something sets it, so a root-owned socket or FIFO is reachable the same way as a regular file.

Two kinds of entry drop out of being opened before being acted on. A symlink cannot be opened under `O_NOFOLLOW`, so it is skipped rather than re-owned; its own ownership decides nothing inside a directory the user already controls, and the target is what mattered. A socket cannot be opened at all, which is the safe direction. Those two and an entry a live code-server has just removed are the three the walk skips on—`ELOOP`, `ENXIO`, `ENOENT`—because abandoning the walk leaves whatever it had not reached root-owned, and `setupUserDataDir` only logs a chown failure. Every other errno fails the walk instead. `EMFILE` and `ENFILE` say nothing about the entry, so skipping on them skips every entry in turn and reports success over a tree that is still root-owned end to end, and the agent shares one descriptor table with the WebSocket connection and the FTP sessions.

The recursion holds one directory file descriptor per level, and what lives under `.alpamon-editor` is whatever the user puts there. `configs/alpamon.service` sets no `LimitNOFILE`, so the systemd default governs the whole agent, and a deep enough chain would not merely fail the chown but starve the WebSocket connection and the FTP sessions of descriptors. `filepath.Walk` held no descriptors, so this cost is new here. `maxChownDepth` bounds it at 64, checked immediately before the child open so the descriptor the budget protects is never the one that overruns it, and an over-deep subtree is skipped with a warning rather than failing the walk: readdir order on ext4 is hash order, so a subtree nested deep enough could otherwise be met before `User/` and leave it root-owned.

The tree is also the user's to fill sideways, so reading a directory whole would put a slice of however many entries they chose inside the root agent. `eachName` reads names a batch at a time and checks the context between them, and the context comes from the `CodeServerManager`, so closing an editor session ends the walk instead of leaving one running per open and close.

The directory and the two files are created `0700` and `0600` rather than `0755` and `0644`. code-server fills this tree as the user, and what it puts there is not neutral: `User/History` holds VS Code's local history, a snapshot of every file opened in the editor, and `User/globalStorage` holds extension state and token stores. `mkdirOpenAt` also `fchmod`s a directory that was already there, so a tree left over from an earlier alpamon is tightened on the next editor start rather than staying open. A file that exists keeps the mode it had, which is what lets an admin tighten `settings.json` further; only the mode a new file is created with moves. A `config.yaml` an older install left `0644` therefore stays `0644`, and what keeps it out of other accounts is the `0700` directory around it rather than its own mode. The two policies are opposite on purpose: a directory here holds only what code-server writes on the user's behalf, so there is no admin intent in its mode to preserve.

### Atomic replacement

`RegisterTunnel` deduplicates tunnels by session ID, not by user (`pkg/runner/tunnel_client.go`), so one user can hold two editor tunnels at once. Each gets its own `CodeServerManager` and each runs `setupUserDataDir`, which puts two writers on the same `settings.json` while a running code-server watches that file. The `O_TRUNC` write left a window in which a reader saw an empty file.

The bytes now go to a sibling temp file that `renameat` swaps into place, so a reader gets the old file or the new one. The temp name carries a random suffix rather than a pid, since the two writers share a process.

`renameat` replaces a symlink sitting at the target instead of following it, so the swap alone would already stay inside the tree, but it does so silently. A stat with `AT_SYMLINK_NOFOLLOW` runs first so a planted link is reported rather than replaced. It only reports: the rename stays inside the tree whatever appears after the check, which is why a stat is enough here and would not be enough anywhere else in the file. It also creates nothing, so a failure in any later step leaves the target absent rather than empty.

A replaced file arrives with the mode its temp file was created with, where truncating in place left the old mode alone. The same stat reports it, so a regular file passes its own mode to the temp create. `O_CREAT` applies that mode as `mode &^ umask`, which would narrow it rather than preserve it, so an `fchmod` follows the create; `fchmod` is not subject to the umask. Ownership is left alone: it cannot widen access the way a mode can, `chownUserDataDir` resets it on the path where it matters, and elsewhere the process already runs as the user who owns the tree.

What the temp file costs is a leftover. A process that dies between the create and the rename leaves one behind, where an in-place truncate left no residue, and the directory belongs to the user, so nothing else would ever collect it. `setupUserDataFiles` sweeps each directory before it writes, taking only entries whose name a `writeFileAt` of that file could have produced and whose age is over an hour, so a temp another session is filling right now is left alone. The age is read through a descriptor rather than off the name, so the entry that is measured is the entry that is unlinked. That open carries `O_NONBLOCK` as well as `O_NOFOLLOW`: the temp name is predictable and `mkfifo` needs no privilege, and `O_RDONLY` on a FIFO waits for a writer that never comes, which would hang the setup inside the root agent with no signal reaching it. A sweep that fails is logged and ignored.

### Windows

`setupUserDataFiles` and `chownTreeNoFollow` return an error on Windows rather than keeping the path-based implementation the unix side replaced. Nothing reachable changes: `editorTunnelSupported` rejects editor tunnels there before any of this runs (`pkg/executor/handlers/tunnel/tunnel.go`), and `chownTreeNoFollow` sat behind `os.Getuid() == 0`, which Windows never satisfies. What changes is the day someone lifts that rejection. The agent runs as LocalSystem, and a directory junction, which any user can create without the privilege a symlink requires, redirects a path-based write the way a symlink does on unix, so `editor_windows.go` needs a junction-safe implementation of its own before that (#379). `editor_windows_test.go` says so as a test instead of a comment.

`TestSetupUserDataDir` and `TestSetupUserDataDirIdempotent` move to `editor_unix_test.go` with that, since they assert files landing on disk and `editor_test.go` carries no build tag. The name check runs before either platform implementation, so its tests stay in `editor_test.go`.

## Testing

`go test ./... -p 1 -count=1` passes:

```
ok  	github.com/alpacax/alpamon/v2/pkg/runner	35.539s
```

`go vet ./...`, `GOOS=windows go vet ./pkg/runner/` and `gofmt -l` are all clean. `golangci-lint run ./pkg/runner/...` reports one issue, `commit_types.go:162: field expireDate is unused`, in a file this branch does not touch.

Every target in `.goreleaser.yaml` compiles: `linux/amd64`, `linux/arm64`, `darwin/amd64`, `darwin/arm64` and `windows/amd64`.

Each new test was seen fail first, by putting back the code it covers:

| Reverted to | Fails |
| --- | --- |
| `os.MkdirAll` / `os.WriteFile` / `filepath.Walk`, name check dropped | `TestSetupUserDataDirRefusesSymlinkedFiles`, `TestSetupUserDataDirRefusesSymlinkedDirs`, `TestChownUserDataDirDoesNotFollowSymlinks`, all five cases of `TestValidateUserDataDirNameRejectsUnsafeNames` |
| `O_TRUNC` in `writeFileAt` | `TestWriteFileAtReplacesInPlaceFile` |
| the symlink check as an `O_CREAT` open | `TestWriteFileAtLeavesNothingWhenTheTempWriteFails` |
| a fixed 0644 on the temp create, or the `fchmod` dropped | `TestWriteFileAtKeepsTheModeItReplaces` |
| `O_NOFOLLOW` on the home directory open | `TestSetupUserDataDirFollowsASymlinkedHome` |
| the hard link skip dropped | `TestChownUserDataDirSkipsHardLinkedFiles` |
| that skip narrowed back to `S_IFREG` | `TestChownUserDataDirSkipsHardLinkedNonRegularFiles` |
| a name-based `lchown` for entries that fail to open | `TestChownUserDataDirDoesNotFollowSymlinks` |
| the over-deep subtree failing the walk instead of being skipped, or the recursion in `chownChildren` removed | `TestChownUserDataDirStopsDescendingAtTheLimit` |
| the per-entry context check in `eachName` dropped | `TestEachNameStopsMidDirectory` |
| the context check before the root chown dropped | `TestChownUserDataDirStopsOnACancelledContext` |
| `userDataDirMode` and `userDataFileMode` widened back to `0755` and `0644`, or the `fchmod` on an existing directory dropped | `TestSetupUserDataDirKeepsTheTreeToItsUser` |
| the mode carry-over in `writeFileAt` dropped | `TestSetupUserDataDirKeepsAnOlderFileMode` |
| the temp sweep dropped | `TestSetupUserDataDirSweepsStaleTempFiles` |
| `O_NONBLOCK` dropped from the sweep's open | `TestSetupUserDataDirSurvivesAFifoNamedLikeATemp` |
| `os.MkdirAll` creating an absent home | `TestSetupUserDataDirRefusesAMissingHome` |
| a fixed temp name, so two sessions collide | `TestSetupUserDataDirUnderConcurrentSessions` |
| the name check widened | `TestValidateUserDataDirNameAcceptsAPlainComponent` |

`TestWriteFileAtReplacesInPlaceFile` holds a second hard link to the old inode and asserts it still reads `old` afterwards, which an in-place truncate would have destroyed, plus that no temp file outlives the call. It is where atomicity from a reader's side is pinned; `TestSetupUserDataDirUnderConcurrentSessions` reads only after all eight sessions have finished, so it covers that none of them fails and the file they leave is whole, and says so rather than claiming more.

`TestWriteFileAtLeavesNothingWhenTheTempWriteFails` picks a name long enough that the derived temp name exceeds `NAME_MAX`, which fails the temp create and nothing else, and asserts `os.ErrNotExist` at the target.

`TestWriteFileAtKeepsTheModeItReplaces` runs under an explicit `umask` of `0077` against a `0644` file, so the mode and the umask disagree by construction. The earlier version used `0600`, which neither `022` nor `077` takes a bit off, and so passed with the umask bug present.

`TestChownUserDataDirDoesNotFollowSymlinks` runs on both sides of the root check: as root it plants the target in a temp directory, and as a regular user it points the link at `/etc/passwd`, where following it would fail with `EPERM`. Both the link and its target must keep their owner, and an ordinary file beside them must not, so the case cannot pass by the chown having done nothing at all. The hard-link cases carry the same witness.

`TestChownUserDataDirStopsDescendingAtTheLimit` pins both sides of the budget. A tree nested exactly `maxChownDepth` deep is still descended, the level past that is not, and the `settings.json` sitting beside the deep subtree changes hands whichever of the two readdir returns first. A witness one level inside the tree carries the descent itself: without it, an unchanged owner on the deepest entry reads the same whether the walk stopped at the limit or never entered, and turning the descent condition into `if true` left the whole package green.

`TestEachNameStopsMidDirectory` fills a directory past `readdirBatch` and cancels from inside the callback on its first call, requiring an exact count of one. Checking the context only per batch, or only before the first read, gives 512.

The `EMFILE` and `ENFILE` branch added in this round is deliberately uncovered. Reaching it means lowering `RLIMIT_NOFILE` for the whole test binary, which every other case in the package shares under `-p 1`.

`TestUserDataDirIsUnsupportedOnWindows` was not run here—it is `//go:build windows`, so the Windows CI job is the first place it executes.

Each of the six commits in this round was checked out on its own and builds and passes `go test ./pkg/runner/` there, and the rounds before it were checked the same way when they landed.

### DIFF
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -259,7 +259,7 @@ func writeFileAt(parent *os.File, name string, data []byte) error {
 	mode := uint32(userDataFileMode)
 	var st unix.Stat_t
 	switch err := unix.Fstatat(dirfd, name, &st, unix.AT_SYMLINK_NOFOLLOW); {
-	case err == unix.ENOENT:
+	case errors.Is(err, unix.ENOENT):
 	case err != nil:
 		return &os.PathError{Op: "stat", Path: path, Err: err}
 	case st.Mode&unix.S_IFMT == unix.S_IFLNK:
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -23,10 +23,16 @@ import (
 
 const dirFlags = unix.O_RDONLY | unix.O_DIRECTORY | unix.O_NOFOLLOW | unix.O_CLOEXEC
 
+// The home directory's own entry sits in a directory the user does not own, so
+// nothing there is theirs to swap and following a link is safe. Admins do point
+// home directories at another filesystem that way, and refusing them would turn
+// a supported layout into an editor that will not start.
+const homeDirFlags = unix.O_RDONLY | unix.O_DIRECTORY | unix.O_CLOEXEC
+
 // setupUserDataFiles creates dirName/User under homeDir and writes config.yaml
 // and User/settings.json, refusing to traverse or write through symlinks.
 func setupUserDataFiles(homeDir, dirName string, configData, settingsData []byte) error {
-	home, err := openDir(homeDir)
+	home, err := openDir(homeDir, homeDirFlags)
 	if err != nil {
 		return err
 	}
@@ -53,7 +59,7 @@ func setupUserDataFiles(homeDir, dirName string, configData, settingsData []byte
 // chownTreeNoFollow recursively changes ownership under root without ever
 // following a symlink: links themselves are re-owned, their targets untouched.
 func chownTreeNoFollow(root string, uid, gid int) error {
-	dir, err := openDir(root)
+	dir, err := openDir(root, dirFlags)
 	if err != nil {
 		return err
 	}
@@ -97,9 +103,10 @@ func chownChildren(dir *os.File, uid, gid int) error {
 	return nil
 }
 
-// openDir opens path as a directory fd. The final component must not be a symlink.
-func openDir(path string) (*os.File, error) {
-	fd, err := unix.Open(path, dirFlags, 0)
+// openDir opens path as a directory fd. Under dirFlags the final component must
+// not be a symlink; homeDirFlags drops that and is only for the home directory.
+func openDir(path string, flags int) (*os.File, error) {
+	fd, err := unix.Open(path, flags, 0)
 	if err != nil {
 		return nil, &os.PathError{Op: "open", Path: path, Err: err}
 	}
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -4,6 +4,7 @@ package runner
 
 import (
 	"cmp"
+	"errors"
 	"math/rand/v2"
 	"os"
 	"path/filepath"
@@ -117,7 +118,7 @@ func openDir(path string, flags int) (*os.File, error) {
 // regular file squatting on the name fails the open (ELOOP or ENOTDIR).
 func mkdirOpenAt(parent *os.File, name string) (*os.File, error) {
 	path := filepath.Join(parent.Name(), name)
-	if err := unix.Mkdirat(int(parent.Fd()), name, 0755); err != nil && err != unix.EEXIST {
+	if err := unix.Mkdirat(int(parent.Fd()), name, 0755); err != nil && !errors.Is(err, unix.EEXIST) {
 		return nil, &os.PathError{Op: "mkdir", Path: path, Err: err}
 	}
 	fd, err := unix.Openat(int(parent.Fd()), name, dirFlags, 0)
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -141,12 +141,17 @@ func chownChildren(ctx context.Context, dir *os.File, uid, gid, depth int) error
 	dirfd := int(dir.Fd())
 	return eachName(ctx, dir, func(name string) error {
 		path := filepath.Join(dir.Name(), name)
-		// An open that fails skips the entry, not the walk: ELOOP is a symlink,
-		// ENXIO a socket, ENOENT one a live code-server just removed, and none
-		// is worth leaving the settings.json elsewhere in the tree root-owned.
 		fd, err := unix.Openat(dirfd, name, entryFlags, 0)
 		if err != nil {
-			return nil
+			// ELOOP is a symlink, ENXIO a socket, ENOENT one a live code-server
+			// just removed, and none is worth leaving the settings.json elsewhere
+			// in the tree root-owned. Anything else is about the process rather
+			// than the entry—EMFILE and ENFILE would skip every one of them and
+			// still report success—so it fails the walk.
+			if errors.Is(err, unix.ELOOP) || errors.Is(err, unix.ENXIO) || errors.Is(err, unix.ENOENT) {
+				return nil
+			}
+			return &os.PathError{Op: "open", Path: path, Err: err}
 		}
 		var st unix.Stat_t
 		if err := unix.Fstat(fd, &st); err != nil {
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -4,7 +4,9 @@ package runner
 
 import (
 	"cmp"
+	"context"
 	"errors"
+	"io"
 	"math/rand/v2"
 	"os"
 	"path/filepath"
@@ -38,6 +40,10 @@ const homeDirFlags = unix.O_RDONLY | unix.O_DIRECTORY | unix.O_CLOEXEC
 // O_RDONLY on a FIFO would wait for a writer; O_NOFOLLOW makes a symlink ELOOP.
 const entryFlags = unix.O_RDONLY | unix.O_NOFOLLOW | unix.O_NONBLOCK | unix.O_CLOEXEC
 
+// Reading a user-controlled directory whole would materialize millions of
+// strings in the root agent and put the first cancellation check after all of it.
+const readdirBatch = 512
+
 // The walk holds one directory fd per level, and the tree below the user data
 // dir is whatever the user puts there. alpamon.service sets no LimitNOFILE, so
 // the systemd default applies to the whole agent: a deep enough chain would
@@ -58,7 +64,7 @@ func tempPrefix(name string) string { return "." + name + "." }
 
 // setupUserDataFiles creates dirName/User under homeDir and writes config.yaml
 // and User/settings.json, refusing to traverse or write through symlinks.
-func setupUserDataFiles(homeDir, dirName string, configData, settingsData []byte) error {
+func setupUserDataFiles(ctx context.Context, homeDir, dirName string, configData, settingsData []byte) error {
 	home, err := openDir(homeDir, homeDirFlags)
 	if err != nil {
 		return err
@@ -77,8 +83,8 @@ func setupUserDataFiles(homeDir, dirName string, configData, settingsData []byte
 	}
 	defer func() { _ = userDir.Close() }()
 
-	sweepStaleTemps(dataDir, userDataConfigFile)
-	sweepStaleTemps(userDir, userDataSettingsFile)
+	sweepStaleTemps(ctx, dataDir, userDataConfigFile)
+	sweepStaleTemps(ctx, userDir, userDataSettingsFile)
 
 	if err := writeFileAt(dataDir, userDataConfigFile, configData); err != nil {
 		return err
@@ -86,36 +92,61 @@ func setupUserDataFiles(homeDir, dirName string, configData, settingsData []byte
 	return writeFileAt(userDir, userDataSettingsFile, settingsData)
 }
 
+// eachName calls fn for every entry in dir, a batch at a time, and stops as soon
+// as ctx is done: closing the editor session has to end a walk over a directory
+// the user made as wide as they liked.
+func eachName(ctx context.Context, dir *os.File, fn func(name string) error) error {
+	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		names, err := dir.Readdirnames(readdirBatch)
+		if errors.Is(err, io.EOF) {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+		for _, name := range names {
+			if err := ctx.Err(); err != nil {
+				return err
+			}
+			if err := fn(name); err != nil {
+				return err
+			}
+		}
+	}
+}
+
 // chownTreeNoFollow re-owns the tree under root, holding every entry open before
 // acting on it. What it cannot open that way—symlinks, sockets—it leaves alone
 // rather than reaching by name.
-func chownTreeNoFollow(root string, uid, gid int) error {
+func chownTreeNoFollow(ctx context.Context, root string, uid, gid int) error {
 	dir, err := openDir(root, dirFlags)
 	if err != nil {
 		return err
 	}
 	defer func() { _ = dir.Close() }()
 
+	if err := ctx.Err(); err != nil {
+		return err
+	}
 	if err := dir.Chown(uid, gid); err != nil {
 		return err
 	}
-	return chownChildren(dir, uid, gid, maxChownDepth)
+	return chownChildren(ctx, dir, uid, gid, maxChownDepth)
 }
 
-func chownChildren(dir *os.File, uid, gid, depth int) error {
-	names, err := dir.Readdirnames(-1)
-	if err != nil {
-		return err
-	}
+func chownChildren(ctx context.Context, dir *os.File, uid, gid, depth int) error {
 	dirfd := int(dir.Fd())
-	for _, name := range names {
+	return eachName(ctx, dir, func(name string) error {
 		path := filepath.Join(dir.Name(), name)
 		// An open that fails skips the entry, not the walk: ELOOP is a symlink,
 		// ENXIO a socket, ENOENT one a live code-server just removed, and none
 		// is worth leaving the settings.json elsewhere in the tree root-owned.
 		fd, err := unix.Openat(dirfd, name, entryFlags, 0)
 		if err != nil {
-			continue
+			return nil
 		}
 		var st unix.Stat_t
 		if err := unix.Fstat(fd, &st); err != nil {
@@ -128,7 +159,7 @@ func chownChildren(dir *os.File, uid, gid, depth int) error {
 		// name in ".", so the count means this only for the other types.
 		if !isDir && st.Nlink > 1 {
 			_ = unix.Close(fd)
-			continue
+			return nil
 		}
 		if err := unix.Fchown(fd, uid, gid); err != nil {
 			_ = unix.Close(fd)
@@ -142,16 +173,13 @@ func chownChildren(dir *os.File, uid, gid, depth int) error {
 				log.Warn().Msgf("Not descending past %d levels at %s; ownership below it is left as it is.", maxChownDepth, path)
 			}
 			_ = unix.Close(fd)
-			continue
+			return nil
 		}
 		child := os.NewFile(uintptr(fd), path)
-		err = chownChildren(child, uid, gid, depth-1)
+		err = chownChildren(ctx, child, uid, gid, depth-1)
 		_ = child.Close()
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+		return err
+	})
 }
 
 // openDir opens path as a directory fd. Under dirFlags the final component must
@@ -182,32 +210,30 @@ func mkdirOpenAt(parent *os.File, name string) (*os.File, error) {
 // dying before its rename. Nothing else collects them, and the directory
 // belongs to the user, so they would sit there for the life of the account.
 // A failure here is not worth refusing to start the editor over.
-func sweepStaleTemps(parent *os.File, name string) {
-	names, err := parent.Readdirnames(-1)
-	if err != nil {
-		log.Debug().Err(err).Msgf("Failed to list %s while sweeping temp files.", parent.Name())
-		return
-	}
-
+func sweepStaleTemps(ctx context.Context, parent *os.File, name string) {
 	dirfd := int(parent.Fd())
 	prefix, cutoff := tempPrefix(name), time.Now().Add(-staleTempAge)
-	for _, entry := range names {
+	err := eachName(ctx, parent, func(entry string) error {
 		if !strings.HasPrefix(entry, prefix) || !strings.HasSuffix(entry, tempSuffix) {
-			continue
+			return nil
 		}
 		// Through an fd, so the age that decides the unlink is read off the entry
 		// the name resolved to and not off whatever replaced it since.
 		fd, err := unix.Openat(dirfd, entry, entryFlags, 0)
 		if err != nil {
-			continue
+			return nil
 		}
 		f := os.NewFile(uintptr(fd), entry)
 		info, err := f.Stat()
 		_ = f.Close()
 		if err != nil || !info.Mode().IsRegular() || info.ModTime().After(cutoff) {
-			continue
+			return nil
 		}
 		_ = unix.Unlinkat(dirfd, entry, 0)
+		return nil
+	})
+	if err != nil {
+		log.Debug().Err(err).Msgf("Stopped sweeping temp files under %s.", parent.Name())
 	}
 }
 
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -34,6 +34,11 @@ const dirFlags = unix.O_RDONLY | unix.O_DIRECTORY | unix.O_NOFOLLOW | unix.O_CLO
 // a supported layout into an editor that will not start.
 const homeDirFlags = unix.O_RDONLY | unix.O_DIRECTORY | unix.O_CLOEXEC
 
+// For an entry of unknown type, so that stat and chown act on the descriptor
+// rather than on a name the user can rename something else onto in between.
+// O_RDONLY on a FIFO would wait for a writer; O_NOFOLLOW makes a symlink ELOOP.
+const entryFlags = unix.O_RDONLY | unix.O_NOFOLLOW | unix.O_NONBLOCK | unix.O_CLOEXEC
+
 // The walk holds one directory fd per level, and the tree below the user data
 // dir is whatever the user puts there. alpamon.service sets no LimitNOFILE, so
 // the systemd default applies to the whole agent: a deep enough chain would
@@ -82,8 +87,9 @@ func setupUserDataFiles(homeDir, dirName string, configData, settingsData []byte
 	return writeFileAt(userDir, userDataSettingsFile, settingsData)
 }
 
-// chownTreeNoFollow recursively changes ownership under root without ever
-// following a symlink: links themselves are re-owned, their targets untouched.
+// chownTreeNoFollow re-owns the tree under root, holding every entry open before
+// acting on it. What it cannot open that way—symlinks, sockets—it leaves alone
+// rather than reaching by name.
 func chownTreeNoFollow(root string, uid, gid int) error {
 	dir, err := openDir(root, dirFlags)
 	if err != nil {
@@ -108,29 +114,36 @@ func chownChildren(dir *os.File, uid, gid, depth int) error {
 	}
 	dirfd := int(dir.Fd())
 	for _, name := range names {
-		// Stat before the chown, not after: AT_SYMLINK_NOFOLLOW speaks only for
-		// symlinks, and a hard link is not a link to a file but a second name
-		// for the same inode, one the user can plant pointing anywhere they can
-		// read. Handing that inode away is exactly what this walk must not do.
-		var st unix.Stat_t
-		if err := unix.Fstatat(dirfd, name, &st, unix.AT_SYMLINK_NOFOLLOW); err != nil {
-			return &os.PathError{Op: "stat", Path: filepath.Join(dir.Name(), name), Err: err}
-		}
-		if st.Mode&unix.S_IFMT == unix.S_IFREG && st.Nlink > 1 {
-			continue
-		}
-		if err := unix.Fchownat(dirfd, name, uid, gid, unix.AT_SYMLINK_NOFOLLOW); err != nil {
-			return &os.PathError{Op: "lchown", Path: filepath.Join(dir.Name(), name), Err: err}
-		}
-		if st.Mode&unix.S_IFMT != unix.S_IFDIR {
-			continue
-		}
 		path := filepath.Join(dir.Name(), name)
-		childFd, err := unix.Openat(dirfd, name, dirFlags, 0)
+		// An open that fails skips the entry, not the walk: ELOOP is a symlink,
+		// ENXIO a socket, ENOENT one a live code-server just removed, and none
+		// is worth leaving the settings.json elsewhere in the tree root-owned.
+		fd, err := unix.Openat(dirfd, name, entryFlags, 0)
 		if err != nil {
-			return &os.PathError{Op: "open", Path: path, Err: err}
+			continue
 		}
-		child := os.NewFile(uintptr(childFd), path)
+		var st unix.Stat_t
+		if err := unix.Fstat(fd, &st); err != nil {
+			_ = unix.Close(fd)
+			return &os.PathError{Op: "stat", Path: path, Err: err}
+		}
+		isDir := st.Mode&unix.S_IFMT == unix.S_IFDIR
+		// A hard link is a second name for one inode, and the user can plant one
+		// over anything link() lets them reach. Directories always carry a second
+		// name in ".", so the count means this only for the other types.
+		if !isDir && st.Nlink > 1 {
+			_ = unix.Close(fd)
+			continue
+		}
+		if err := unix.Fchown(fd, uid, gid); err != nil {
+			_ = unix.Close(fd)
+			return &os.PathError{Op: "chown", Path: path, Err: err}
+		}
+		if !isDir {
+			_ = unix.Close(fd)
+			continue
+		}
+		child := os.NewFile(uintptr(fd), path)
 		err = chownChildren(child, uid, gid, depth-1)
 		_ = child.Close()
 		if err != nil {
@@ -181,13 +194,9 @@ func sweepStaleTemps(parent *os.File, name string) {
 		if !strings.HasPrefix(entry, prefix) || !strings.HasSuffix(entry, tempSuffix) {
 			continue
 		}
-		// Stat through an fd rather than the name, so the age that decides the
-		// unlink is read off the very entry the name resolved to. O_NONBLOCK
-		// because the temp name is predictable and mkfifo needs no privilege:
-		// a FIFO parked here would hold this open until a writer arrived, and
-		// the IsRegular check that rejects it comes after the open.
-		fd, err := unix.Openat(dirfd, entry,
-			unix.O_RDONLY|unix.O_NONBLOCK|unix.O_NOFOLLOW|unix.O_CLOEXEC, 0)
+		// Through an fd, so the age that decides the unlink is read off the entry
+		// the name resolved to and not off whatever replaced it since.
+		fd, err := unix.Openat(dirfd, entry, entryFlags, 0)
 		if err != nil {
 			continue
 		}
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -52,6 +52,14 @@ const staleTempAge = time.Hour
 // from whatever else the user keeps in the directory.
 const tempSuffix = ".tmp"
 
+// code-server fills the user data dir on the user's behalf, and what lands
+// there is theirs alone: User/History holds snapshots of every file they edit
+// and User/globalStorage holds extension state. Nothing needs group or other.
+const (
+	userDataDirMode  = 0700
+	userDataFileMode = 0600
+)
+
 func tempPrefix(name string) string { return "." + name + "." }
 
 // setupUserDataFiles creates dirName/User under homeDir and writes config.yaml
@@ -188,12 +196,18 @@ func openDir(path string, flags int) (*os.File, error) {
 // regular file squatting on the name fails the open (ELOOP or ENOTDIR).
 func mkdirOpenAt(parent *os.File, name string) (*os.File, error) {
 	path := filepath.Join(parent.Name(), name)
-	if err := unix.Mkdirat(int(parent.Fd()), name, 0755); err != nil && !errors.Is(err, unix.EEXIST) {
+	if err := unix.Mkdirat(int(parent.Fd()), name, userDataDirMode); err != nil && !errors.Is(err, unix.EEXIST) {
 		return nil, &os.PathError{Op: "mkdir", Path: path, Err: err}
 	}
 	fd, err := unix.Openat(int(parent.Fd()), name, dirFlags, 0)
 	if err != nil {
 		return nil, &os.PathError{Op: "open", Path: path, Err: err}
+	}
+	// Also for a directory that was already there: mkdir under a umask, or an
+	// earlier alpamon that made it 0755, would otherwise leave it open.
+	if err := unix.Fchmod(fd, userDataDirMode); err != nil {
+		_ = unix.Close(fd)
+		return nil, &os.PathError{Op: "chmod", Path: path, Err: err}
 	}
 	return os.NewFile(uintptr(fd), path), nil
 }
@@ -237,7 +251,7 @@ func writeFileAt(parent *os.File, name string, data []byte) error {
 
 	// renameat replaces a symlink at name rather than following it, so this only
 	// turns that silence into an error; nothing below depends on it still holding.
-	mode := uint32(0644)
+	mode := uint32(userDataFileMode)
 	var st unix.Stat_t
 	switch err := unix.Fstatat(dirfd, name, &st, unix.AT_SYMLINK_NOFOLLOW); {
 	case err == unix.ENOENT:
--- a/pkg/runner/editor_unix.go
+++ b/pkg/runner/editor_unix.go
@@ -19,20 +19,15 @@ import (
 )
 
 // The user owns the home directory, so any path under it can be swapped for a
-// symlink between a check and the use of the checked path. Every operation
-// below the home directory therefore resolves one path component at a time
-// against a held directory fd with O_NOFOLLOW, which leaves no window to
-// redirect root's writes or chowns elsewhere. The home directory itself is
-// opened by full path, and every name handed to these helpers must be a single
-// component: O_NOFOLLOW guards only the last one, so a separator inside a name
-// would let the components before it be followed.
+// symlink between a check and the use of the checked path. Everything below it
+// is resolved one component at a time against a held directory fd, and every
+// name handed to these helpers must be a single component: O_NOFOLLOW guards
+// only the last one.
 
 const dirFlags = unix.O_RDONLY | unix.O_DIRECTORY | unix.O_NOFOLLOW | unix.O_CLOEXEC
 
-// The home directory's own entry sits in a directory the user does not own, so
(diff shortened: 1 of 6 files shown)


pkg/runner/editor_unix.go
```diff
@@ -0,0 +1,162 @@
...
+	path := filepath.Join(parent.Name(), name)
+
+	// renameat replaces a symlink at name rather than following it, which is
+	// safe but silent. This open refuses one first, so a planted link is an error.
+	probe, err := unix.Openat(dirfd, name,
+		unix.O_WRONLY|unix.O_CREAT|unix.O_NOFOLLOW|unix.O_CLOEXEC, 0644)
+	if err != nil {
+		return &os.PathError{Op: "open", Path: path, Err: err}
+	}
+	_ = unix.Close(probe)
```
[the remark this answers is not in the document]
jisung-02: Fixed in b3e3319. You are right on both counts, and the second one is the worse of the two: the probe was carrying O_CREAT, so on a first run it created the target before the temp file existed, which is exactly the empty-file window the rename was added to close.

The probe is now a stat with AT_SYMLINK_NOFOLLOW, which answers the same question and creates nothing. A failure in any later step now leaves the target absent rather than empty, and first-time creation goes straight from nothing to the finished file.

Worth saying why a stat is enough here when it would not be anywhere else in this file: renameat replaces a symlink at the target instead of following it, so the swap stays inside the tree whatever appears after the check. The check only turns a silent replacement into a reported error, so a race against it costs nothing.

TestWriteFileAtLeavesNothingWhenTheTempWriteFails covers it. It picks a name long enough that the derived temp name exceeds NAME_MAX, which fails the temp create and nothing else, and asserts os.ErrNotExist at the target. Against the old probe it fails with the empty file left behind.

pkg/runner/editor_unix_test.go
```diff
@@ -232,3 +237,256 @@ func TestStopKillsGroupIgnoringSIGTERM(t *testing.T) {
...
+func TestWriteFileAtLeavesNothingWhenTheTempWriteFails(t *testing.T) {
+	dir := t.TempDir()
+	parent, err := openDir(dir, dirFlags)
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = parent.Close() })
+
+	name := strings.Repeat("n", 250)
+
+	require.Error(t, writeFileAt(parent, name, []byte("x")))
+
```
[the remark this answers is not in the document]
jisung-02: Fixed in 14cbdf8. You are right that 250 was an assumption about NAME_MAX, and a filesystem that allows longer names would have made the test fail while the code was correct.

The limit is now bisected at run time. unix.Pathconf would answer it directly but x/sys only defines it on darwin, not linux, so the search runs between one character and a length no filesystem in use accepts, and it asserts that upper bound rather than trusting it: if a name that long is ever accepted the test says so instead of quietly passing on a premise that stopped holding.

The length is still the lever rather than permissions, and that is deliberate. The point of the test is that a failure after the symlink check leaves no file at the target, which only discriminates when creating the target would have succeeded and creating the temp name does not. A read-only directory fails both, so it would pass against the old O_CREAT probe too and stop guarding anything. Against that probe the reworked test still fails with the empty file left behind, which is the regression it exists to catch.

geunwoonoh (commented):
## Summary

This replaces the `os.MkdirAll` / `os.WriteFile` / `filepath.Walk(os.Chown)` calls behind `$HOME/.alpamon-editor` with fd-based operations. Every step under the home directory is opened one component at a time against a held directory fd, with `O_NOFOLLOW` on the open and `AT_SYMLINK_NOFOLLOW` on the chown; the home directory itself is the one entry allowed to be a link. File writes move from `O_TRUNC` to a sibling temp file swapped in by `renameat`, and `UserDataDirName` is validated up front as a single path component. The Windows side refuses rather than keeping the path-based implementation.

The design is right. Walking down with the openat family one component at a time is the textbook answer to this problem, and the line the exception is drawn on—whether the user can replace the entry—is the correct one. What follows are the gaps left on top of it.

---

## Critical issues (0)

None.

---

## Warnings (3)

### 1. The chown walk reaches inodes outside the tree through a hard link

**`pkg/runner/editor_unix.go:82`**

`chownChildren` chowns an entry before it knows what the entry is:

```go
if err := unix.Fchownat(dirfd, name, uid, gid, unix.AT_SYMLINK_NOFOLLOW); err != nil {  // 82: chown first
	return &os.PathError{Op: "lchown", Path: path, Err: err}
}
var st unix.Stat_t
if err := unix.Fstatat(dirfd, name, &st, unix.AT_SYMLINK_NOFOLLOW); err != nil {        // 86: stat after
```

`AT_SYMLINK_NOFOLLOW` covers symlinks only. A hard link is not a link to a file, it is another name for the same inode, so the flag has nothing to say about it. A user who plants `~/.alpamon-editor/x` as a hard link to a file outside the tree gets root to hand them ownership of that inode.

This lands squarely on the standard the PR sets for itself: "the function manages one directory tree and should act on nothing outside it." `fs.protected_hardlinks=1` raises the bar on every supported distro, but that is a value distros set rather than a kernel default, and a container can be running without it. More to the point, `Fstatat` already runs here, so the guard costs nothing.

```go
// Stat first, and skip a regular file with more than one name: it may point outside the tree.
var st unix.Stat_t
if err := unix.Fstatat(dirfd, name, &st, unix.AT_SYMLINK_NOFOLLOW); err != nil {
	return &os.PathError{Op: "stat", Path: path, Err: err}
}
isDir := st.Mode&unix.S_IFMT == unix.S_IFDIR
if st.Mode&unix.S_IFMT == unix.S_IFREG && st.Nlink > 1 {
	continue
}
if err := unix.Fchownat(dirfd, name, uid, gid, unix.AT_SYMLINK_NOFOLLOW); err != nil {
	return &os.PathError{Op: "lchown", Path: path, Err: err}
}
if !isDir {
	continue
}
```

### 2. umask defeats the mode preservation, and the test cannot see it

**`pkg/runner/editor_unix.go:161`** / **`pkg/runner/editor_unix_test.go:382`**

The mode handed to `unix.Openat` is applied as `mode &^ umask`. The property 042bc99 restored, that a replacement keeps the mode it lands on, does not hold whenever the umask is non-zero. Confirmed by running it:

```
umask 077: replaced a 0644 file, resulting mode = 0600
```

Permissions only narrow, never widen, so this is not a security problem, and nothing breaks in practice because the file is chowned to the user anyway. The issue is that `TestWriteFileAtKeepsTheModeItReplaces` picked `0600`, which neither common umask (022 or 077) takes a bit off, so it passes with the bug present. The test states a property it does not actually hold the code to.

`fchmod` is not subject to the umask, so one line after the create closes it:

```go
f := os.NewFile(uintptr(fd), tmpPath)
if err := unix.Fchmod(fd, mode); err != nil {   // unlike the O_CREAT mode, this ignores the umask
	_ = f.Close()
	_ = unix.Unlinkat(dirfd, tmp, 0)
	return &os.PathError{Op: "chmod", Path: tmpPath, Err: err}
}
```

The test needs `0644` instead of `0600`, or a `unix.Umask(0077)` around it, to catch the regression.

### 3. The recursion has no depth limit and holds one fd per level

**`pkg/runner/editor_unix.go:92-101`**

`chownChildren` keeps the child directory fd open across the recursive call, so a tree of depth D occupies D fds at once. The layout under `.alpamon-editor` is whatever the user makes it, and `configs/alpamon.service` sets no `LimitNOFILE`, so the systemd default soft limit (typically 1024) applies. A single thousand-deep directory puts the whole process into fd exhaustion mid-walk.

The chown failure itself is only a `log.Warn`, but the fd pressure lands on the WebSocket connection and FTP sessions in the same process. The old `filepath.Walk` held no fds, so this cost is new here. A depth cap that returns an error past the limit would bound it.

---

## Suggestions (7)

1. **`editor_unix.go:161-177`** A process that dies between the create and the rename leaves a `.settings.json.<rand>.tmp` behind, and nothing cleans it up. The old `O_TRUNC` write left no such residue, so a sweep of `.`+name+`.*.tmp` at startup is worth considering.

2. **`editor_unix.go:81`** `path := filepath.Join(dir.Name(), name)` is allocated for every entry but consumed only on the error paths. A code-server user-data-dir grows `logs/`, `CachedData/` and `User/globalStorage/` into thousands of entries, so this belongs inside each error branch.

3. **`editor_unix.go:120`** `err != unix.EEXIST` should be `errors.Is(err, unix.EEXIST)`. It works today because `unix.Mkdirat` returns a bare `syscall.Errno`, and it breaks silently the moment anything wraps that error.

4. **`editor.go:640`** The whole-tree chown runs on every editor start. A short circuit when the root already carries the target uid/gid removes the cost from the second start onward. The behavior predates this PR, but it is more expensive now that the walk holds fds.

5. **`editor_unix.go:35`** The behavior changed for a home directory that exists in `/etc/passwd` but not on disk (LDAP users, `useradd -M`). The old `os.MkdirAll` created it root-owned; this fails with ENOENT. The new behavior is the better one, but it is in neither the description nor a test, so a line confirming it was intended would help.

6. **`editor_test.go:77`** `TestSetupUserDataDirRejectsUnsafeDirName` mutates the singleton `GetCodeServerConfig()` returns. `t.Cleanup` makes it safe today, and it breaks quietly the first time any test in this package calls `t.Parallel()`.

7. **`editor_unix_test.go`** Nothing exercises the case that motivates the PR: two editor sessions for one user writing `settings.json` at the same time. The hard-link witness is a good proxy, but two concurrent `setupUserDataDir` calls asserting both succeed and the result parses as JSON can be written without flakiness.

---

## Highlights

- **`mode = uint32(st.Mode) & 0777` (`editor_unix.go:154`)** stops setuid/setgid/sticky bits on a planted file from carrying into the replacement. Easy to miss, and it is closed.
- **`!filepath.IsLocal(name) || name != filepath.Base(name) || name == "."` (`editor.go:622`)** I ran the combination directly: `""`, `"."`, `".."`, `"a/"`, `"a/."` and `"./x"` are all rejected while a legitimate name like `"..foo"` passes. Catching that `IsLocal` accepts `"."` is the part that matters.
- **The short-write rebuttal is correct.** `os.File.Write` sets `err = io.ErrShortWrite` whenever `n != len(b)`, before any syscall error is wrapped, so `(n < len(data), nil)` cannot occur by construction. Copilot raised this across four rounds; declining it with the source cited is the right call.
- **The comment saying the `writeFileAt` stat only reports.** The reasoning that the rename stays inside the tree regardless, so a TOCTOU on the check does not matter, is accurate, and it carries the caveat that the same argument does not transfer elsewhere.
- **The bisection in `nameLimit`.** It asserts its upper bound with `require.False(t, fits(hi))` instead of assuming it, so a test that fails to reach the failure path it targets does not pass quietly.
- **Windows is closed by a test, not a comment.** Whoever removes `editorTunnelSupported` goes red on `TestUserDataDirIsUnsupportedOnWindows`, and the Windows job runs `go test ./...`, so it has teeth.

---

**Verdict**: COMMENTED

Warnings 1 and 2 are gaps in the two invariants this PR sets for itself: act on nothing outside the tree, and keep the mode a replacement lands on. Both close in a few lines, and 2 in particular is passing right now, so it will keep passing until the test is changed.

**Checklist**:
- [x] Follows Go conventions (gofmt, vet and golangci-lint clean; one `errors.Is` nit)
- [x] No duplicate or dead code (`userDataUserDir` and `userDataSettingsFile` go unreferenced on a Windows build, but lint runs on linux only, so it is harmless)
- [x] No performance regression (`Readdirnames` drops the sort `filepath.Walk` paid for, a small improvement; the fd occupancy in Warning 3 is the new cost)
- [x] Tests included, and each was seen fail against the code it covers
- [ ] No security vulnerabilities (Warning 1: the hard-link path is still open)
- [x] No documentation change needed

pkg/runner/editor_unix.go
```diff
@@ -0,0 +1,256 @@
...
+
+	dirfd := int(parent.Fd())
+	prefix, cutoff := tempPrefix(name), time.Now().Add(-staleTempAge)
+	for _, entry := range names {
+		if !strings.HasPrefix(entry, prefix) || !strings.HasSuffix(entry, tempSuffix) {
+			continue
+		}
+		// Stat through an fd rather than the name, so the age that decides the
+		// unlink is read off the very entry the name resolved to.
+		fd, err := unix.Openat(dirfd, entry, unix.O_RDONLY|unix.O_NOFOLLOW|unix.O_CLOEXEC, 0)
```
geunwoonoh: Warning: this open blocks forever on a FIFO.

`IsRegular()` runs after the open, and opening a FIFO `O_RDONLY` blocks until a writer arrives, so the check is never reached. The name pattern is predictable (`.config.yaml.<anything>.tmp`) and `mkfifo` needs no privilege. Reproduced: `setupUserDataDir` never returns, so the editor stops opening for that user and every attempt parks another goroutine in an uninterruptible open inside the root agent. Nothing covers this point with a timeout.

This is the only open in the file exposed to it, since the others carry `O_DIRECTORY` (ENOTDIR before the FIFO handler runs) or `O_EXCL` (never opens an existing entry). `O_NONBLOCK` returns immediately on a FIFO, which lets `IsRegular()` reject it, and is ignored on a regular file. Verified that `TestSetupUserDataDirSweepsStaleTempFiles` still passes with it.

```suggestion
		fd, err := unix.Openat(dirfd, entry, unix.O_RDONLY|unix.O_NOFOLLOW|unix.O_NONBLOCK|unix.O_CLOEXEC, 0)
```

geunwoonoh (commented):
## Summary

Re-reviewed at `37a4f3e` against the round I left at `14cbdf8`. Twelve commits. Nine of the ten items landed, one was declined, and the decline is correct. I found one new problem in code this round added.

Verified locally on the pushed head: `gofmt`, `go vet` for linux, windows and darwin, and `go test ./pkg/runner/` all clean. CI is 26/26.

## What landed

I reverted each fix and confirmed the covering test actually goes red, rather than reading the diff alone.

| Item | Commit | Check |
| --- | --- | --- |
| Warning 1, hard link | `dcdda3b` | Stat moved ahead of the chown, `S_IFREG && Nlink > 1` skipped. **Reverting it fails `TestChownUserDataDirSkipsHardLinkedFiles`**: the outside inode's gid moves 20 to 12. |
| Warning 2, umask | `51cee50` | `unix.Fchmod` after the create. **Reverting it fails `TestWriteFileAtKeepsTheModeItReplaces`**: 0644 becomes 0600. |
| Warning 3, walk depth | `6aa9b0d` | `maxChownDepth = 64`, carried down as a parameter. |
| Suggestion 1, temp sweep | `507672b` | Landed, and see the warning below. |
| Suggestion 2, path allocation | `cbc7445` | `Join` now only in the error branches and the directory branch. |
| Suggestion 3, `errors.Is` | `d657c8e` | Applied at the mkdirat. |
| Suggestion 4, short-circuit the chown | declined | **The decline is right, see below.** |
| Suggestion 5, missing home directory | `4bd116c` | Confirmed the `os.ErrNotExist` chain resolves through the `PathError`. |
| Suggestion 6, config singleton | `d9602d3` | `validateUserDataDirName` extracted and tested directly, plus a positive case. |
| Suggestion 7, concurrent sessions | `47990db` | Eight goroutines. |

The part that matters is that Warnings 1 and 2 are load-bearing, not just present. Warning 2 in particular: moving the test from `0600` to `0644` under an explicit `unix.Umask(0077)` makes the mode and the umask disagree by construction instead of by luck, which is what the old test was missing.

**Suggestion 4 was wrong and declining it was right.** Under root, `writeFileAt` renames a fresh root-owned inode into each name on every start, so both files come back root-owned regardless of what the root directory says. Short-circuiting on the root's uid and gid would skip the walk immediately after replacing the `settings.json` the user has to be able to edit, which is the case `chownUserDataDir` exists for. Taking that suggestion would have shipped a bug.

Worth calling out `1087bf4` as well. On a non-root run the test asked for the current user's own ownership, so "the link itself is chowned" held whether or not any `lchown` happened. I missed that in the first round.

## Warnings (1, new)

### `sweepStaleTemps` blocks forever on a planted FIFO

**`pkg/runner/editor_unix.go:186`**

The `IsRegular()` check runs *after* the open. Opening a FIFO `O_RDONLY` blocks until a writer arrives, so the check is never reached. The name pattern is fully predictable (`.config.yaml.<anything>.tmp`) and `mkfifo` needs no privilege.

Reproduced. `setupUserDataDir` does not return:

```
--- FAIL: TestScratchSweepBlocksOnAPlantedFifo (5.00s)
    setupUserDataDir is still blocked after 5s: the sweep opened a FIFO read-only without O_NONBLOCK
```

This is not an escalation. The user plants it in their own home directory and is the one locked out. But their editor then never opens until the FIFO is removed, and every attempt leaves another goroutine parked in an uninterruptible open inside the root agent. Nothing covers this point with a timeout; `StartupTimeout` only starts at `waitForReady`.

This is the only open in the file exposed to it. Every other one carries `O_DIRECTORY`, which fails with ENOTDIR before the FIFO handler runs, or `O_EXCL`, which never opens an existing entry.

One flag closes it. I applied it and confirmed `TestSetupUserDataDirSweepsStaleTempFiles` still passes:

```go
fd, err := unix.Openat(dirfd, entry, unix.O_RDONLY|unix.O_NOFOLLOW|unix.O_NONBLOCK|unix.O_CLOEXEC, 0)
```

A FIFO then returns immediately and is rejected by `IsRegular()`, and `O_NONBLOCK` is ignored on a regular file.

## Suggestions (3, minor)

1. **`editor_unix.go:215`** `mkdirOpenAt` moved to `errors.Is`, but the `writeFileAt` switch still compares `err == unix.ENOENT`. Same bare `Errno` off `unix.Fstatat`, so the same reasoning applies. Consistency only.

2. **`editor_unix.go:184`** The comment claims "the entry measured is the entry unlinked." The stat goes through an fd but the unlink goes by name, so a swap in between removes a different entry. The conclusion still holds because the directory is the user's own, but for a different reason than the comment gives.

3. **`editor_unix_test.go:412`** `unix.Umask` is process-global. It is unavoidable for what this test pins and `t.Cleanup` restores it, but it is the same fragility that `d9602d3` just removed from the config singleton. A line in the comment would catch the next person reaching for `t.Parallel()`.

## Highlights

- **The `ordinary` control in `TestChownUserDataDirSkipsHardLinkedFiles`.** Asserting only on the hard link would have passed with the chown a total no-op. Pinning that a single-named file *is* re-owned closes that.
- **The three fixtures in `TestSetupUserDataDirSweepsStaleTempFiles`.** Aged and matching goes, aged and non-matching stays, fresh and matching stays. Both bounds of the sweep in one test.
- **`movableOwnership`.** It picks ownership the kernel will really move, so the assertion cannot pass because nothing happened.
- **Declining Suggestion 4 with the reasoning spelled out.** Taking it would have left the user unable to edit their own settings.

---

**Verdict**: COMMENTED, and an approve once the flag is in

Everything from the last round is handled correctly, and the two security-relevant fixes are provably doing work. The only thing holding the approve is the FIFO block in the newly added sweep, and `unix.O_NONBLOCK` is the whole fix. The three suggestions are not approve blockers.

**Checklist**:
- [x] Follows Go conventions (gofmt, vet and golangci-lint clean; one `errors.Is` consistency nit)
- [x] Tests included, and each fix was seen fail when reverted
- [x] No duplicate or dead code
- [x] Performance: the `Join` allocations are gone; the sweep adds one directory listing per start, which is nothing
- [ ] No security or availability issues (the FIFO block in the sweep is open)
- [x] No documentation change needed

pkg/runner/editor_unix.go
jisung-02: Fixed in `1a4eaadc` with the `O_NONBLOCK` you suggested. `TestSetupUserDataDirSurvivesAFifoNamedLikeATemp` plants an aged FIFO under the temp name and fails after 10s with `setupUserDataDir blocked opening the FIFO` when the flag is removed, so the timeout you noted was missing is there.

Reading the walk again with this in mind turned up five more; see the round comment on the PR. The one worth your attention is `139db674`: the hard-link guard was check-then-act, since `Fstatat` read the link count from the name and `Fchownat` resolved that name a second time to act on it.

pkg/runner/editor_unix_test.go
```diff
@@ -232,3 +238,587 @@ func TestStopKillsGroupIgnoringSIGTERM(t *testing.T) {
...
+	username, groupname := movableOwnership(t, sibling)
+	beforeSibling := fileOwner(t, sibling)
+	beforeDeepest := fileOwner(t, deepest)
+
+	require.NoError(t, chownUserDataDir(t.Context(), root, username, groupname))
+
+	assert.NotEqual(t, beforeSibling, fileOwner(t, sibling),
+		"a subtree too deep to enter must not cost the files beside it their owner")
+	assert.Equal(t, beforeDeepest, fileOwner(t, deepest),
+		"the walk must not descend past the descriptor budget")
```
geunwoonoh, line 694: Warning: this assertion cannot fail for the reason it names.

It checks only that the deepest entry was *not* re-owned, which is equally true if the walk never descended at all. I removed the recursion entirely (`if !isDir || depth <= 0` to `if true`) and the whole package still passes:

```
ok  github.com/alpacax/alpamon/v2/pkg/runner  29.621s
```

Descending to re-own `User/settings.json` is the reason `chownTreeNoFollow` exists, and every ownership test puts its fixtures at the top level, so nothing else covers it either. The code is right today: a scratch test with a nested witness passes on `5ab9de4` and reports `one level down re-owned (User/settings.json)` with the recursion removed.

A witness inside the tree gives the negative assertion something to stand against:

```go
nested := filepath.Join(root, "deep", userDataSettingsFile)
require.NoError(t, os.WriteFile(nested, nil, 0600))
// ... beforeNested := fileOwner(t, nested), then after the walk:
assert.NotEqual(t, beforeNested, fileOwner(t, nested),
    "the walk must re-own what it descends into, which is why it descends")
```

pkg/runner/editor_unix_test.go
```diff
@@ -232,3 +238,587 @@ func TestStopKillsGroupIgnoringSIGTERM(t *testing.T) {
...
+	require.NoError(t, os.MkdirAll(filepath.Join(stale, userDataUserDir), 0755))
+
+	userDataDir, err := setupUserDataDir(t.Context(), homeDir, "", "")
+	require.NoError(t, err)
+
+	for path, want := range map[string]os.FileMode{
+		userDataDir: userDataDirMode,
+		filepath.Join(userDataDir, userDataUserDir):                       userDataDirMode,
+		filepath.Join(userDataDir, userDataConfigFile):                    userDataFileMode,
+		filepath.Join(userDataDir, userDataUserDir, userDataSettingsFile): userDataFileMode,
```
geunwoonoh: Warning: the expectation is the same constant the implementation reads, so this test moves with the code instead of holding it.

Setting `userDataDirMode` back to `0755` and `userDataFileMode` back to `0644` keeps it green:

```
ok  github.com/alpacax/alpamon/v2/pkg/runner  0.388s
```

So the test written to pin `5ab9de4` passes against exactly the state `5ab9de4` fixes. Literals make it a guard, the same way the umask test stopped resting on a mode no common umask touches.

```suggestion
		userDataDir: 0700,
		filepath.Join(userDataDir, userDataUserDir):                       0700,
		filepath.Join(userDataDir, userDataConfigFile):                    0600,
		filepath.Join(userDataDir, userDataUserDir, userDataSettingsFile): 0600,
```

geunwoonoh (dismissed):
Approving. I could not find a defect in the shipping code this round. What is left is two gaps in the tests protecting it, and one of them hides the reason `chownTreeNoFollow` exists, so I would land both before merge.

Verified on `5ab9de4`: `gofmt`, `go vet` for linux, darwin and windows, and `go test ./pkg/runner/` all clean, CI 26/26.

## The FIFO fix

`1a4eaad`, and it is load-bearing. Taking `O_NONBLOCK` back out of `entryFlags` fails `TestSetupUserDataDirSurvivesAFifoNamedLikeATemp` after the full 10s. Hoisting it into a shared `entryFlags` was the better call than patching the sweep alone: `chownChildren` now opens every entry too, so without it the walk would have hit the same wait.

## Your five

`139db67` is the one I should have caught. The shape I suggested was check-then-act across two name resolutions, and `rename` inside one's own home directory is atomic, unprivileged, and retryable against a walk the user can trigger by opening a session. Resolving each entry once and running the stat, the chown and the descent through that descriptor is the right answer. Widening `S_IFREG` to `!isDir` is right for the same reason: `link()` does not care what the inode is, and directories carry a second name in `.` so the count only means something for the other types.

The rest hold up. `User/History` through a `0755` directory is a real exposure, and skipping an over-deep subtree instead of failing the walk is the correct trade once you notice readdir returns hash order, since the old behavior could abandon the walk before reaching `User/`.

## Warnings (2), both tests rather than code

### 1. Nothing pins that the walk descends at all

Making `chownChildren` never recurse passes the **entire** package suite:

```
# if !isDir || depth <= 0  ->  if true
ok  github.com/alpacax/alpamon/v2/pkg/runner  29.621s
```

Descending into `User/` to hand `settings.json` to the demoted user is the whole job of `chownUserDataDir`, and no assertion covers it. `TestChownUserDataDirStopsDescendingAtTheLimit` only checks that the deepest entry was *not* re-owned, which cannot tell "stopped at the limit" from "never went in," and every ownership test puts its fixtures at the top level. A regression here leaves `User/settings.json` root-owned, code-server comes up broken, and CI stays green.

The code is correct today. I checked both directions with a scratch test: it passes on `5ab9de4`, and with the recursion removed it reports `one level down re-owned (User/settings.json)` and `two levels down re-owned`. A nested witness in the existing deep-tree test is all it takes.

### 2. `TestSetupUserDataDirKeepsTheTreeToItsUser` checks the constants against themselves

The `want` map is built from `userDataDirMode` and `userDataFileMode`, the same constants the implementation uses, so the expectation moves with the code. Setting them back to `0755` and `0644` keeps the test green:

```
ok  github.com/alpacax/alpamon/v2/pkg/runner  0.388s
```

The test meant to pin `5ab9de4` passes against exactly the state `5ab9de4` fixes. Literal `0700` and `0600` make it a guard. Same shape as the umask test from two rounds ago.

## Suggestions (4)

1. **`editor_unix.go:147`** Every open failure skips the entry. The reasons in the comment (ELOOP, ENXIO, ENOENT) are per-entry, but EMFILE and ENFILE are not: under descriptor pressure every entry is skipped and the walk reports success having re-owned nothing, after which code-server starts demoted against root-owned files. Restricting the skip to the expected errnos closes it.

2. **`editor_unix.go:208`** `mkdirOpenAt` fchmods an existing directory to `0700` unconditionally, while `writeFileAt` preserves the mode an admin set on a file. Opposite policies in one file. The directory side is defensible given what lands there, but it is worth a word in the comment.

3. **`editor_unix_test.go:747`** A `settings.json` from an older install keeps `0644` through the carry-over, so "readable only by its user" rests on the `0700` directory rather than on the file mode. Not reachable, but the test pre-creates only stale directories, so the file half of the upgrade path is uncovered.

4. **`editor_unix.go:257`** `err == unix.ENOENT` is still there while `mkdirOpenAt` moved to `errors.Is`. Consistency only, carried over from the last round.

## Highlights

- **Finding `139db67` yourself** rather than taking the reviewer's shape and stopping.
- **`seen == 1` in `TestEachNameStopsMidDirectory`**, with `readdirBatch + 1` entries laid down so the batch boundary is actually crossed. It pins that the cancel lands at the next entry and not the next batch.
- **Picking a FIFO for `TestChownUserDataDirSkipsHardLinkedNonRegularFiles`**, which puts the "link() does not care about the type" argument directly into a test.
- **Skipping the over-deep subtree instead of failing**, which turns a silent breakage that depended on readdir order into a warning.

---

**Verdict**: APPROVED, with the two test gaps to land before merge

**Checklist**:
- [x] Follows Go conventions (gofmt, vet and golangci-lint clean; one `errors.Is` left)
- [ ] Tests: green, but two do not constrain the behavior they name (both demonstrated above)
- [x] No security vulnerabilities (the hard-link race, the FIFO wait and the `0755` exposure are all closed)
- [x] Performance: readdir batched at 512, walk cancellable, descent bounded
- [x] No documentation change needed

pkg/runner/editor_unix_test.go
jisung-02: Fixed in b57666a8, and the demonstration is what makes the point: with `if !isDir || depth <= 0` turned into `if true` the whole package stayed green, so the assertion could not fail for the reason its message gave.

The witness is where you put it, `root/deep/settings.json`, read into `beforeNested` before the walk and asserted changed after. That gives the negative assertion below it something to stand against: `beforeDeepest` unchanged now means the walk stopped at the limit rather than never having entered. The doc comment says which of the two the witness is there to separate.

Seen RED: with the recursion removed the case fails on `the walk must re-own what it descends into, which is why it descends`; green on 61686880.

pkg/runner/editor_unix_test.go
jisung-02: Fixed in 759fe616, with the literals from your suggestion. You are right that the expectation was reading the same two constants the implementation reads, so it moved with the code instead of holding it.

Seen RED: with `userDataDirMode` back at `0755` and `userDataFileMode` at `0644` all four assertions fail (`expected: 0x1c0, actual: 0x1ed`); green on 61686880.

Your third suggestion belongs beside this one, so it landed as its own commit: 61686880 adds `TestSetupUserDataDirKeepsAnOlderFileMode` for the file half of the upgrade path. It pins both halves in one case—the `0644` an older install left on `config.yaml` survives the rewrite, and the directory around it is `0700`—because the carried-over mode on its own reads as the exposure 5ab9de4 closed rather than as the admin-intent carry-over it is.

pkg/runner/editor_unix.go
```diff
@@ -0,0 +1,302 @@
...
+	return eachName(ctx, dir, func(name string) error {
+		path := filepath.Join(dir.Name(), name)
+		fd, err := unix.Openat(dirfd, name, entryFlags, 0)
+		if err != nil {
+			// ELOOP is a symlink, ENXIO a socket, ENOENT one a live code-server
+			// just removed, and none is worth leaving the settings.json elsewhere
+			// in the tree root-owned. Anything else is about the process rather
+			// than the entry—EMFILE and ENFILE would skip every one of them and
+			// still report success—so it fails the walk.
+			if errors.Is(err, unix.ELOOP) || errors.Is(err, unix.ENXIO) || errors.Is(err, unix.ENOENT) {
```
geunwoonoh, line 151: Suggestion: this list is Linux-shaped, and darwin answers differently for a socket.

Opening a unix socket with `entryFlags` returns `EOPNOTSUPP` on darwin, not `ENXIO`, so it falls through to the error return and fails the whole walk:

```
open(socket, entryFlags) = operation not supported on socket (ENXIO=false EOPNOTSUPP=true)
chownUserDataDir over a tree containing a socket = open .../s: operation not supported on socket
```

No production impact: `setupUserDataDir` gates the chown on `runtime.GOOS != "darwin"`, so the walk never runs there. The macOS test job does call `chownUserDataDir` directly though, and it is green only because nothing plants a socket in a walked tree. Since `TestChownUserDataDirSkipsHardLinkedNonRegularFiles` already reasons from sockets, that is a natural case to add later, and it would fail on macOS alone.

```suggestion
			if errors.Is(err, unix.ELOOP) || errors.Is(err, unix.ENXIO) ||
				errors.Is(err, unix.EOPNOTSUPP) || errors.Is(err, unix.ENOENT) {
```

geunwoonoh (approved):
Approving. Both warnings are closed, and closed in the way that matters: the mutations I used to demonstrate the gaps now make the tests red.

| Item | Commit | Reverting it |
| --- | --- | --- |
| Warning 1, nothing pinned the descent | `b57666a` | recursion removed, **FAIL** `the walk must re-own what it descends into` |
| Warning 2, the mode test read its own constants | `759fe61` | constants widened to `0755`/`0644`, **FAIL** `expected: 0x1c0` |
| Suggestion 1, every open failure skipped the entry | `73f3346` | skip is now the three errnos, everything else fails the walk with the path |
| Suggestion 2, opposite mode policies | `6278637` | comment |
| Suggestion 3, the file half of the upgrade path | `6168688` | carry-over dropped, **FAIL** `a mode an admin set on the file survives the rewrite` |
| Suggestion 4, `err == unix.ENOENT` | `e242956` | applied |

Verified on `6168688`: `gofmt`, `go vet` for linux, darwin and windows, and `go test ./pkg/runner/` all clean, CI 26/26.

The `nested` witness in `b57666a` is placed well. The positive and the negative assertion now hold each other up inside one test, so the `deepest` check can no longer pass on its own.

Suggestion 1 came out better than what I asked for. The skip is exactly the three errnos the comment always claimed, and saying in the PR body that no test reaches EMFILE is the right call rather than implying otherwise: getting there means lowering `RLIMIT_NOFILE` for a binary every other case in the package shares under `-p 1`.

## Suggestions (1)

**`pkg/runner/editor_unix.go:151`** The skip list is Linux-shaped. On darwin, opening a unix socket with `entryFlags` gives `EOPNOTSUPP`, not `ENXIO`, so a socket fails the whole walk there:

```
open(socket, entryFlags) = operation not supported on socket (ENXIO=false EOPNOTSUPP=true)
chownUserDataDir over a tree containing a socket = open .../s: operation not supported on socket
```

Production is unaffected, since `setupUserDataDir` gates the chown on `runtime.GOOS != "darwin"` and the walk never runs there. But the macOS job does call `chownUserDataDir` directly in several tests, and it is green only because none of them plants a socket. `TestChownUserDataDirSkipsHardLinkedNonRegularFiles` already argues from sockets, so the case is a natural one for someone to add, and it would go red on macOS alone. One more errno makes the comment true on both platforms the tests run on.

## Highlights

- **Where the witness went in `b57666a`.** Neither assertion can carry the test by itself now.
- **Not claiming coverage that is not there.** The EMFILE path is stated as untested in the PR body instead of being left to be found later. That has held for five rounds.
- **`6278637` staying a comment.** A review point that did not call for a code change did not get one.
- **The comment in `759fe61`.** It says why the literals are literals, which is what stops the next person folding them back into the constants.

---

**Verdict**: APPROVED

The condition on my last approve is met. What is left is a darwin-only portability nit that does not reach the production path.

**Checklist**:
- [x] Follows Go conventions (gofmt, vet and golangci-lint clean; `errors.Is` now consistent)
- [x] Tests pass, and each constrains the behavior it names (confirmed by reverting)
- [x] Performance: readdir batched, walk cancellable, descent bounded
- [x] No documentation change needed

pkg/runner/editor_unix.go
jisung-02: Not applying this one. `chownUserDataDir` is gated off on darwin at `pkg/runner/editor.go:647`, so `EOPNOTSUPP` is reachable only from a test that calls the walk directly. There is no production path behind it on that platform.

The errno is per-entry and would fit the list, so this is a judgment about when to add it rather than whether it is right. Suggestion 1 narrowed the skip to the errnos the walk actually meets, and I would rather keep it to those than carry one for a case nothing exercises. If someone adds the socket case to a macOS test, the errno goes in alongside it, where the failure it prevents is visible instead of asserted in a comment.

PR #458, opened 2026-09-03
test: take the wall clock out of time-dependent tests

jisung-02:
## Background

Tests across the agent waited on real time in three ways, and each one can turn a healthy build red on a loaded machine.

Some slept a fixed span and then asserted, so the assertion only held if the work finished inside that guess. Some drove a timer the test could not hurry, such as the migration watchdog. The rest were timeout guards like `case <-time.After(2 * time.Second)`. A guard does not fire on a healthy run, which is why they looked safe, but whether it fires is still decided by the wall clock, so "it has not flaked yet" is not the same as clock-independent.

## Changes

Sleep-then-assert waits now wait on the signal the work itself produces: a channel the job closes, or a WaitGroup it drains.

Everything that depends on a timer runs inside a `testing/synctest` bubble, where `time.Now`, timers, and sleeps read a fake clock that advances only once every goroutine in the bubble is blocked. The sleeps stay in the source because they still express the scenario, but they no longer cost anything, and a timeout guard now fires on a real block instead of on a slow runner.

An exact clock let four assertions get stronger. `postChunk`'s drop is pinned to `maxWait` rather than "at least `maxWait`", `Ping`'s timestamp is compared to the current time rather than to a one-second window, `TestContextTimeout` and `TestContextDeadline` compare the elapsed span to the timeout itself, and the suppressed session events prove no emit fired by settling every goroutine through `synctest.Wait`—a 100ms sleep only ever proved that none fired inside that window.

`TestDetect_ContextDeadlineRespectedBetweenProbes` carried a comment claiming its error assertion would catch a removed in-loop ctx check. It would not: the post-loop `ctx.Err` check in `Detect` reports `DeadlineExceeded` either way. It now asserts the third provider went unprobed, which does catch it.

`pkg/executor/resource_test.go` is deleted. Five of its seven tests asserted wall-clock or memory thresholds that either duplicated an existing benchmark or could not fail for the reason their names suggest: the lookup test allowed 100us for an operation that benchmarks at 13.68ns, the startup test allowed a second for work that takes microseconds while discarding every `Register` error, and the idle-memory test read the whole process heap.

The two that checked real behavior survive with exact assertions instead of one-sided ones. `TestWorkerPool_SpreadsCommandsAcrossWorkers` stays in `pkg/executor` and times the queue against the bubble's clock. The saturation check moved on to `internal/pool` as `TestPool_SaturatesWithoutExceedingWorkers`, because it uses no executor type at all, and it replaces `TestPoolConcurrency` and `TestPool_MaxGoroutineEnforcement` there: three tests asserted one property across two packages, and the two older ones bounded it with `LessOrEqual`, which passes for a pool that has serialized down to a single worker.

Two invariants that were promises in a comment are now build errors. `drainBudgetOutlastsDelay` in the system handler tests fails to compile if the pool drain budget stops outlasting `delayedActionDelay`, and `noticeSpellsOutDelay` in `pkg/executor/handlers/system/system.go` fails if that delay stops matching the "1 second" the restart, quit, reboot and shutdown messages spell out to the operator. Both read as `constant -N overflows uint`.

Failure paths inside a bubble had a shape of their own worth fixing. A `require` that fires runs `runtime.Goexit` on the bubble's main goroutine, so anything still parked—a held worker, a watchdog on its timer, a blocked send—turns one assertion into `deadlock: main bubble goroutine has exited but blocked goroutines remain`, and the panic takes the rest of the package with it. Every teardown that releases such a goroutine is deferred now, in `internal/pool`, `pkg/executor`, `pkg/migrate` and `pkg/runner`, and the discarded `pool.Submit` errors that could leave a wait unsatisfiable are checked.

Assertions across the touched files moved to testify, and the comments left behind are the ones the code cannot state: why a wait needs no guard inside a bubble, why a defer sits where it does, that `net.Pipe` is unbuffered, how sudo's `use_pty` reshapes the process tree.

The rule is written into the testing-conventions paragraph, which is mirrored in `CLAUDE.md`, `README.md` and `.github/copilot-instructions.md`.

## Not bubbled, on purpose

A bubble does not count network, file, or process waits as blocked, so the clock stalls or the test deadlocks. The httptest-backed emit tests in `pkg/runner/auth_session_event_test.go` stay on the real clock for that reason.

`TestFinalizeRequest_RacesHandleResponse` exists to let two goroutines race, and a deterministic scheduler would hand the same one the win every time.

Two bubbles that hold for a non-obvious reason now say so in a comment: `TestWaitForDaemonReadyClosedSession` is safe only because its unix dial fails synchronously against a socket that was never created, and `TestInfoHandler_Ping` compares a second-precision round trip only because the bubble's clock starts on a whole second.

## Testing

`go test ./... -p 1` passes, `go vet ./...` and `gofmt -l .` are clean, and the packages touched here also pass under `-race -count=2`.

Three packages got faster because their tests no longer spend real seconds. Measured on the pushed head against a detached worktree at `origin/main`, running the same three packages in both:

| Package | origin/main | this branch |
| --- | --- | --- |
| `pkg/migrate` | 13.922s | 0.406s |
| `pkg/agent` | 1.316s | 0.226s |
| `pkg/executor/handlers/system` | 7.473s | 0.284s |

The system handler tests were the third. `handleRestart` and its siblings submit a pool job that sleeps a second before acting, and `Pool.Shutdown` cannot interrupt a sleeping job, so a one-second drain budget always lost the race and the job outlived the test. Every one of those tests discarded the `Shutdown` error, which is why the timeout went unnoticed on `main`. Removing this branch's 100ms sleep from `Restart_Default` only took away a head start that had been hiding the same defect in one of six places. All six now drain inside a bubble against a budget that outlasts the scheduled second, and the result is asserted instead of discarded.

The bubble also changes what a regression looks like. Making `executeJob`'s `recover` block instead of returning fails `TestPool_NoLeakAfterPanic` here as `panic: deadlock: all goroutines in bubble are blocked` in 0.429s; the same mutation against `origin/main` waits out the shutdown budget and reports a bare 5.20s failure with nothing naming the stuck worker.

`BenchmarkRegistry_GetManyHandlers` was built for this branch and reports 15.45 ns/op at 0 allocs, against 13.68 ns/op for the single-entry `BenchmarkRegistry_Get`. Note that neither runs in CI, which only calls `go test -v ./... -p 1`; they are there to be run by hand.


pkg/executor/handlers/system/system_test.go
```diff
@@ -238,8 +238,6 @@ func TestSystemHandler_Restart_Default(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, 0, exitCode)
 	assert.Contains(t, output, "restart")
```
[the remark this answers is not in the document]
jisung-02: Correct, and it was worse than a risk: with the sleep gone this test failed 10 out of 10 runs with `shutdown timeout after 1s`.

It is not specific to this test, though. `Restart_Alpamon`, `Quit`, `Reboot`, `Shutdown` and `Upgrade_SelfUpdate` all schedule the same one-second delayed action behind a one-second drain budget, and all of them took the timeout path on `main` too. The `_ =` on `Shutdown` is what kept it quiet. The 100ms sleep here was only ever a head start that hid the same defect in this one test, so restoring it would have papered over five others.

Fixed in 4b694059 for all six. They now run inside a `synctest` bubble, the drain gets a budget that outlasts the scheduled second, and the result is asserted rather than discarded, so a job that genuinely fails to drain fails the test instead of leaking. Verified RED by shortening the budget below the scheduled delay: each of the six then reports the timeout. The package drops from 7.47s to 0.30s, and `-race -count=2` is clean.

internal/pool/pool_test.go
```diff
@@ -53,27 +55,25 @@ func TestPoolQueueFull(t *testing.T) {
...
 	})
-	if err != nil {
-		t.Fatalf("failed to submit blocking task: %v", err)
-	}
+	require.NoError(t, err, "failed to submit blocking task")
 
-	// Give worker time to pick up the first job
-	time.Sleep(10 * time.Millisecond)
+	// The worker has taken the job off the queue, so the queue is empty again.
+	<-started
```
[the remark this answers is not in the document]
jisung-02: You are right, and the bound is the part I got wrong last round.

I answered earlier that `TestPoolQueueFull` stays on the real clock on purpose because it has no timer to fake. That reasoning was about the clock, and your point is about the wait: `<-started` is unbounded either way, so a worker that never takes the job hangs the test until the package timeout and reports it as one.

Fixed in bfbeddc5, for this test and for `TestPool_NoLeakQueueFull` in `pool_leak_test.go`, which has the same shape. Both go into a bubble, which reports that regression as a deadlock naming the goroutines instead. The bubble is there for the bound rather than for a fake clock, and the comment above each `<-started` now says so, so the next reader does not undo it on the "no timer here" reasoning I used. A `time.After` guard would have worked too, but it puts back the wall-clock dependency this branch exists to remove.

The direct `t.Error` and `t.Errorf` calls in both functions moved to testify while they were open.

internal/pool/pool_test.go
```diff
@@ -5,198 +5,198 @@ import (
...
-			time.Sleep(10 * time.Millisecond)
-			return nil
-		}); err != nil {
-			t.Errorf("failed to submit job: %v", err)
+				// Hold the worker long enough for the jobs to actually overlap.
+				time.Sleep(10 * time.Millisecond)
+				return nil
+			}); err != nil {
+				wg.Done()
+				t.Errorf("failed to submit job: %v", err)
```
geunwoonoh: Warning: this file now disagrees with itself on the testify convention. `TestPoolBasic`, `TestPoolQueueFull` and `TestPoolPanicRecovery` moved to `assert.NoError` / `require.NoErrorf` in this PR, but this submit path, the deferred shutdown at line 98, and `TestPoolShutdown`'s submit path at line 187 stayed on `t.Errorf`.

The same PR tightens the rule in `CLAUDE.md`, `README.md` and `.github/copilot-instructions.md` ("convert one while you are already editing it"), so leaving these behind in a body that was rewritten anyway is the inconsistency the rule is meant to prevent.

geunwoonoh (commented):
## Review summary

**Scope**: test-only refactor. The one production change is hoisting five scattered `1*time.Second` literals in `system.go` into a `delayedActionDelay` constant, so runtime behavior is unchanged.

The PR removes three ways tests depended on the real clock: a fixed sleep followed by an assertion, a timer the test could not hurry, and `case <-time.After(...)` guards. The instrument is `testing/synctest`: inside a bubble the fake clock advances only once every goroutine is blocked, so a timeout guard fires on a real block rather than on a slow runner, and a `wg.Wait` the code can no longer satisfy surfaces as a deadlock instead of a package timeout.

Direction and execution are both good. The exact clock made several assertions stronger rather than merely faster, and recording the bubbles that were skipped (with reasons, in the PR body and in code comments) is the right artifact for a change like this.

---

### Critical issues (0)

None.

### Warnings (3)

**1. Tests this PR rewrote still call `t.Error` / `t.Errorf` / `t.Fatal` directly.**

The same PR tightens the testify convention in `CLAUDE.md`, `README.md` and `.github/copilot-instructions.md` ("convert one while you are already editing it"), but tests whose bodies were replaced wholesale keep the old style:

- `internal/pool/pool_test.go:98,128,187` - the file disagrees with itself. `TestPoolBasic`, `TestPoolQueueFull` and `TestPoolPanicRecovery` moved to `assert.NoError` / `require.NoErrorf`, while `TestPoolConcurrency`'s deferred shutdown and submit path plus `TestPoolShutdown`'s submit path stayed on `t.Errorf`.
- `pkg/runner/auth_manager_cleanup_test.go:84,90,180,186,261,282,330,333,359,362,365,390` - all six bubbled tests. `if resp := expectSingleResponse(t, client); resp != want { t.Errorf(...) }` is one `assert.Equal(t, want, expectSingleResponse(t, client))`, and the failure prints as a diff.
- `pkg/agent/context_test.go:58,64,131,151,171,189,200,237,248,307`
- `pkg/runner/auth_session_event_test.go` - the three bubbled tests

Not a functional problem, but a PR that strengthens the rule ends up not following it in the files it touched most.

**2. `pkg/executor/handlers/system/system_test.go:715` keeps a 5-second wall-clock guard.**

```go
case <-time.After(5 * time.Second):
```

`TestSystemHandler_Uninstall` carries exactly the kind of timeout guard this PR set out to remove, it sits in a file the PR restructures, and it is not in the "Not bubbled, on purpose" list. It also looks bubble-able: the test passes `nil` for `apiSession`, so `unregisterFromConsole` returns before any HTTP call, and the rest of `executeUninstall` goes through `MockCommandExecutor`. The only real system call left is the stat inside `hasSystemd()`, which does not block a bubble. Either bubble it or add it to the exception list with the reason.

**3. `pkg/executor/worker_pool_test.go:29-78` has no deferred cleanup.**

`TestWorkerPool_SpreadsCommandsAcrossWorkers` shuts the pool and the context manager down at the end of the loop body (lines 77-78). If either `require.NoError` above (lines 39, 42) fails, `FailNow` runs `runtime.Goexit` on the bubble's main goroutine while the workers are still alive, and the run reports `deadlock: main bubble goroutine has exited but blocked goroutines remain` instead of the assertion message. That is the same failure shape this branch deliberately used to see the leak test RED, so here it hides the real cause. The sibling test at line 94 does defer its shutdown, so the new file is inconsistent with itself.

### Suggestions (6)

1. **`system_test.go:33-48`** - `requireDrainOutlastsDelay(t)` is called six times (254, 281, 308, 333, 358, 817), always immediately before `defer drainThen(...)`. Folding the precondition into `drainThen` leaves one call site and makes it impossible to forget.
2. **`system_test.go:255-258` and five siblings** - defer ordering is load-bearing but unstated. `defer drainThen(...)()` is registered first, so LIFO runs `ctxManager.Shutdown()` first and cancels `poolCtx` while the scheduled job is still sleeping. The assertions hold only because `time.Sleep` does not watch ctx and `MockCommandExecutor` ignores it. Making `scheduleDelayedAction` cancellable later (reasonable, since today a cancelled context cannot stop a pending reboot) would break all six for a reason unrelated to what they assert.
3. **`pkg/executor/worker_pool_test.go:88`** - `TestWorkerPool_SaturatesWithoutExceedingWorkers` uses no executor type; it tests `internal/pool` alone, and `TestPool_MaxGoroutineEnforcement` plus `TestPoolConcurrency` already assert the same property there. Three tests of one property across two packages. The new one is stronger (`Equal` rather than `LessOrEqual`), so moving it into `internal/pool` and dropping the weaker two is the better shape.
4. **`dispatcher_benchmark_test.go:20,44,66,85,103`** - `require.NoError(b, err, "register handler")` works, since testify takes `require.TestingT`, and matches the convention this PR is tightening.
5. **`internal/retry/retry_test.go:80`** - on a fake clock the cancel lands while `Retry` is parked on its first backoff timer, so `calls` is exactly 1. `assert.GreaterOrEqual` keeps the one-sided bound the PR body itself argues against elsewhere.
6. **`pkg/scheduler/queue_test.go:50,67`** - the guard duration equals `postChunk`'s own `maxWait` (both `time.Second`). If `drainOne` ever stopped dropping the queue below high water, `postChunk` would give up at t=1s and `close(done)` at the same instant the guard fires, and the tie resolves arbitrarily.

### Verified

The three exact-clock assertions were checked against the implementations, and all three hold:

| Assertion | Why it holds |
|---|---|
| `TestRetry_MaxElapsedTime` == 50ms | `Retry` has no jitter; `InitialInterval == MaxInterval == 10ms` pins `NextBackOff` at 10ms, and the post-operation `time.Since(start) >= MaxElapsedTime` check lands at t=50ms |
| `TestPostChunk_DropsAfterMaxWait` == 30ms | 1ms poll through t=29, drop at t=30 on `>= maxWait` |
| `TestDetect_...BetweenProbes` == 50ms, `third.probeCalls == 0` | 30ms probes against a 50ms deadline: the in-loop `ctx.Err` ends the walk during the second probe, leaving the third unprobed |

`StartWatchdog` has no jitter either, so dropping "plus jitter" from the comment and waiting 4 seconds is accurate. `net.Pipe` is channel-backed and `handleConnection` sets no deadline, so the two `pkg/logger` bubbles are safe. Sweeping the edited files, the only timer-bearing tests left on the real clock are `TestSystemHandler_Uninstall` (warning 2) and the httptest-backed `TestHandleSessionEvent_AcksAndEmits`, which the PR explains.

Deleting `resource_test.go` is right. The five threshold tests are each covered by an existing benchmark (`BenchmarkHandler_Execute`, `BenchmarkContextManager_*`, `BenchmarkRegistry_*`), and the two that checked real behavior moved with stronger assertions.

### Highlights

- `TestDetect_ContextDeadlineRespectedBetweenProbes` is the most valuable change here. Its comment promised a guarantee the assertion did not provide, since `Detect`'s post-loop `ctx.Err` reports `DeadlineExceeded` either way. Asserting `third.probeCalls == 0` makes the test earn its name.
- `requireDrainOutlastsDelay` turns an invariant that lived in a comment into an executable precondition: if `poolDrainWait` ever drops below `delayedActionDelay`, the test says why instead of failing as an unexplained shutdown timeout.
- `TestInfoHandler_Ping` asserts the whole-second premise rather than trusting it.
- The five rounds of Copilot follow-up were all substantive, especially catching that five system handler tests asserted only the response string and so never checked whether the job reached the pool.

### Performance

- **Production**: no impact, constant extraction only.
- **Tests**: `pkg/migrate` 13.97s to 0.49s, `pkg/executor/handlers/system` 7.47s to 0.30s, `pkg/agent` 1.30s to 0.39s. The bigger win is that roughly thirty tests no longer have their outcome decided by runner load.
- **New benchmark**: `BenchmarkRegistry_GetManyHandlers` reports 15.06 ns/op at 100 handlers against 13.32 ns/op for the single-entry case, so lookup stays effectively constant time. CI runs only `go test -v ./... -p 1`, so this is a manual check rather than a regression gate.

### Security

Not applicable. No input surface, auth path, or credential handling is touched.

---

**Verdict**: COMMENTED. Nothing blocks the merge. The three warnings can be follow-ups, though warning 3 (missing defers in `worker_pool_test.go`) is worth folding in now, since its effect is to obscure the cause when a test does fail.

**Checklist**
- [x] Follows stack conventions - except the testify conversion consistency in warning 1
- [x] Tests included, all 20 CI jobs green
- [x] Documentation updated - the three mirrored testing-conventions copies are in sync

pkg/runner/auth_manager_cleanup_test.go
```diff
@@ -52,41 +53,43 @@ func expectSingleResponse(t *testing.T, client net.Conn) SudoApprovalResponse {
...
+			Username:  "alice",
+			Groupname: "wheel",
+			PID:       4242,
+			PPID:      4200,
+			Command:   "sudo systemctl restart nginx",
+			SessionID: "sess-1",
+			Reason:    "Communication error",
+		}
+		if resp := expectSingleResponse(t, client); resp != want {
+			t.Errorf("response: got %+v, want %+v", resp, want)
```
geunwoonoh: Warning: all six bubbled tests in this file kept `t.Errorf` / `t.Error` / `t.Fatal` even though their bodies were rewritten (lines 84, 90, 180, 186, 261, 282, 330, 333, 359, 362, 365, 390).

This particular shape collapses to one line and gains a struct diff on failure:

```go
assert.Equal(t, want, expectSingleResponse(t, client))
```

Same convention point as `internal/pool/pool_test.go` and `pkg/agent/context_test.go`.

pkg/agent/context_test.go
```diff
@@ -30,35 +33,37 @@ func TestContextManagerCreation(t *testing.T) {
...
-			// Expected
-		case <-time.After(100 * time.Millisecond):
-			t.Errorf("context %d not cancelled after shutdown", i+1)
+		// All contexts should be cancelled
+		for i, ctx := range []context.Context{ctx1, ctx2, ctx3} {
+			select {
+			case <-ctx.Done():
+				// Expected
+			case <-time.After(100 * time.Millisecond):
+				t.Errorf("context %d not cancelled after shutdown", i+1)
```
geunwoonoh: Warning: the tests bubbled here still call `t.Errorf` / `t.Error` directly (lines 58, 64, 131, 151, 171, 189, 200, 237, 248, 307), while `TestContextTimeout` and `TestContextDeadline` in the same PR did move to `assert.ErrorIs`.

The bare `t.Fatal` in a `select` timeout branch is fine per the convention; the ones with a value to assert on are the ones worth converting while the body is already being edited.

pkg/executor/worker_pool_test.go
```diff
@@ -0,0 +1,137 @@
...
+				}
+			}
+
+			wg.Wait()
+
+			assert.Equal(t, int32(taskCount), completed.Load(), "workers=%d: not every task ran", workers)
+			assert.Equal(t, time.Duration(taskCount/workers)*jobTime, time.Since(start),
+				"workers=%d: the pool did not spread the tasks across its workers", workers)
+
+			require.NoError(t, workerPool.Shutdown(5*time.Second), "workers=%d: shutdown failed", workers)
```
geunwoonoh: Warning: neither shutdown is deferred, so a failing `require` above turns into a confusing panic instead of an assertion message.

`require.NoError` at lines 39 and 42 calls `FailNow`, which runs `runtime.Goexit` on the bubble's main goroutine while the pool workers are still alive. synctest then reports `deadlock: main bubble goroutine has exited but blocked goroutines remain` and the real reason is lost. That is the same failure shape this branch used deliberately to see the leak test RED, so here it works against you. An early failure in the second iteration also leaks that iteration's pool.

Wrapping the loop body in `t.Run(fmt.Sprintf("workers=%d", workers), ...)` (or a closure) and deferring both shutdowns fixes it. `TestWorkerPool_SaturatesWithoutExceedingWorkers` at line 94 already does this, so the new file is inconsistent with itself.

pkg/executor/worker_pool_test.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	})
+}
+
+// TestWorkerPool_SaturatesWithoutExceedingWorkers verifies the pool runs exactly as many jobs at
+// once as it has workers—never more, and never fewer once enough work is queued
+// to occupy them all.
+func TestWorkerPool_SaturatesWithoutExceedingWorkers(t *testing.T) {
+	synctest.Test(t, func(t *testing.T) {
+		const (
```
geunwoonoh: Suggestion: this test uses no executor type at all - it exercises `internal/pool` only, and the property is already asserted twice over there by `TestPool_MaxGoroutineEnforcement` and `TestPoolConcurrency`. That leaves three tests of one property spread across two packages.

This version is the strongest of the three (`Equal` rather than `LessOrEqual`, which is exactly the point the PR body makes about one-sided checks), so moving it into `internal/pool` and dropping the two weaker ones would consolidate rather than duplicate.

pkg/executor/handlers/system/system_test.go
```diff
@@ -20,6 +21,32 @@ import (
...
+// restart, quit, reboot, shutdown and the post-self-update restart all hand the
+// pool a job that sleeps delayedActionDelay before acting, and Shutdown cannot
+// interrupt a sleeping job. The drain wait therefore has to outlast that delay,
+// which requireDrainOutlastsDelay makes executable rather than a promise in a
+// comment. Inside a synctest bubble neither span costs real time.
+const poolDrainWait = 5 * time.Second
+
+// requireDrainOutlastsDelay fails with the reason rather than as a bare shutdown
+// timeout if delayedActionDelay ever grows past the budget these tests drain on.
+func requireDrainOutlastsDelay(t *testing.T) {
```
geunwoonoh: Suggestion: `requireDrainOutlastsDelay(t)` is called six times (lines 254, 281, 308, 333, 358, 817), and every one sits immediately before `defer drainThen(...)`. Folding the precondition into `drainThen` leaves a single call site and makes it impossible to add a seventh test that forgets it.

The invariant itself is a good idea - turning a comment-level promise into something that fails with its reason is the best small change in this file.

pkg/executor/handlers/system/system_test.go
```diff
@@ -219,109 +246,132 @@ func TestSystemHandler_Restart_Collector(t *testing.T) {
...
-
-	handler := NewSystemHandler(mockExec, mockWS, ctxManager, workerPool, newMockVersionResolver(), nil)
-	ctx := context.Background()
+	synctest.Test(t, func(t *testing.T) {
+		mockExec := common.NewMockCommandExecutor(t)
+		mockWS := &MockWSClient{}
+		ctxManager := agent.NewContextManager()
+		workerPool := pool.NewPool(2, 10)
+		requireDrainOutlastsDelay(t)
+		defer drainThen(t, workerPool, func() {
```
geunwoonoh, line 254: Suggestion: the defer ordering here is load-bearing but nothing says so. `defer drainThen(...)()` is registered first and `defer ctxManager.Shutdown()` second, so LIFO runs `ctxManager.Shutdown()` first and cancels `poolCtx` while the scheduled job is still sleeping out `delayedActionDelay`.

The assertion survives only because `time.Sleep` does not watch ctx and `MockCommandExecutor` ignores the ctx it is handed. If `scheduleDelayedAction` is ever made cancellable - reasonable, since today a cancelled context cannot stop a pending reboot - all six of these tests break for a reason unrelated to what they assert.

Registering `ctxManager.Shutdown` first, or naming the dependency in the `drainThen` comment, would pin it.

pkg/executor/dispatcher_benchmark_test.go
```diff
@@ -16,14 +17,42 @@ func BenchmarkRegistry_Get(b *testing.B) {
 		name:     "test",
 		commands: []string{"cmd1", "cmd2", "cmd3"},
 	}
-	_ = registry.Register(handler)
+	if err := registry.Register(handler); err != nil {
```
geunwoonoh: Suggestion: testify accepts `*testing.B` through `require.TestingT`, so this and the three sibling checks (lines 44, 66, 85, 103) can read as `require.NoError(b, err, "register handler")`, matching the convention this PR tightens.

Failing the benchmark on a dropped `Register` error is the right call either way.

internal/retry/retry_test.go
```diff
@@ -52,46 +56,49 @@ func TestRetry_PermanentError(t *testing.T) {
...
+		})
 
-	if !errors.Is(err, context.Canceled) {
-		t.Fatalf("expected context.Canceled, got %v", err)
-	}
-	if calls < 1 {
-		t.Fatal("expected at least 1 call")
-	}
+		require.ErrorIs(t, err, context.Canceled)
+		assert.GreaterOrEqual(t, calls, 1, "expected at least 1 call")
```
geunwoonoh: Suggestion: the fake clock makes this exact. `synctest.Wait` returns once `Retry` is parked on its first backoff timer, so `cancel()` always lands after exactly one call.

The PR body argues this same point for `TestRetry_MaxElapsedTime` ("a `>=` here would also pass for a retry loop that overran it"); the one-sided bound here has the same weakness.

```suggestion
		assert.Equal(t, 1, calls, "the cancel lands on the first backoff, so exactly one call")
```

pkg/scheduler/queue_test.go
```diff
@@ -38,43 +41,46 @@ func TestPostChunk_EnqueuesBelowHighWater(t *testing.T) {
...
+		case <-done:
+			t.Fatal("postChunk returned at high-water; expected backpressure")
+		default:
+		}
+
+		drainOne(t) // drop below high-water
+
+		select {
+		case <-done:
+		case <-time.After(time.Second):
```
geunwoonoh: Suggestion: this guard is the same duration as the `maxWait` passed to `postChunk` at line 50 - both `time.Second`. If `drainOne` ever stopped dropping the queue below high water, `postChunk` would give up at exactly t=1s and `close(done)` at the same instant this timer fires, and the select would resolve the tie arbitrarily: sometimes a pass, sometimes a failure whose message points at the wrong thing.

Making the guard outlast `maxWait` keeps them from coinciding.

```suggestion
		case <-time.After(5 * time.Second):
```

internal/pool/pool_test.go
jisung-02: Right, and the file was mid-conversion in a body I had rewritten anyway. `TestPoolShutdown`'s submit path moved to `require.NoErrorf` in c27ea639 — require rather than assert, because the count asserted after the loop means nothing once a submit was rejected.

The other two you named went away rather than converting: the deferred shutdown at line 98 and the submit path at line 128 both belonged to `TestPoolConcurrency`, which 02008491 removed as part of the consolidation you asked for below.

pkg/runner/auth_manager_cleanup_test.go
jisung-02: Applied in 897f80ee, including the collapse you wrote out — `assert.Equal(t, want, expectSingleResponse(t, client))` at four sites.

I went past the six bubbled tests and converted the whole file. Leaving the non-bubbled half on `t.Errorf` would have left the file disagreeing with itself for a second time, which is the thing your comment is about.

The 2s guard around the completion receive went too: on a fake clock it fires when the clock advances, so it was measuring nothing. A bare receive lets the bubble name the goroutine that never signalled.

pkg/agent/context_test.go
jisung-02: Applied in 62fee264, and I took the same "convert the whole file" line as in `auth_manager_cleanup_test.go`. The bare `t.Fatal` in a select timeout branch stayed, as you say.

Two wall-clock guards in this file went with the conversion, for the same fake-clock reason. The exact clock also let `TestContextTimeout` and `TestContextDeadline` compare the elapsed span to the timeout itself instead of checking that the context merely expired — seen RED at 60ms: `expected: 50ms actual: 60ms`.

pkg/executor/worker_pool_test.go
jisung-02: Applied in 02008491. The loop body is a `t.Run(fmt.Sprintf("workers=%d", workers), ...)` subtest now, with both `ctxManager.Shutdown` and the pool shutdown deferred inside its bubble.

You are right that the branch was using that failure shape deliberately elsewhere and had it working against itself here. Seen RED by looking up a command the registry does not hold: the failure now reads `no handler for command: no_such_cmd` with its own message and no deadlock panic behind it.

The teardown fix and the move below land in one commit because they are the same function body — there was no way to stage them apart.

pkg/executor/worker_pool_test.go
jisung-02: Agreed, and done in 02008491. `TestPool_SaturatesWithoutExceedingWorkers` now lives in `internal/pool`, and `TestPoolConcurrency` and `TestPool_MaxGoroutineEnforcement` are gone.

Seen RED by building the pool one worker short: `expected: 10 actual: 9`. That is the check the two `LessOrEqual` versions could not make.

pkg/executor/handlers/system/system_test.go
jisung-02: Taken, but not the way you wrote it — the precondition is gone rather than folded into `drainThen`, in dde86234:

```go
const _ = uint(poolDrainWait - delayedActionDelay - 1)
```

I tried the fold first and it reproduced the failure shape you objected to two comments up. Inside `drainThen` the check fires with the pool already running, so `require` exits the bubble's main goroutine with the workers alive and synctest buries the message under a deadlock panic. The invariant is a relation between two constants, so it belongs in the build: one declaration, no call site to forget, nothing to run.

Seen RED by shrinking `poolDrainWait` to `delayedActionDelay`: `constant -1 overflows uint`.

It is spelled with the blank identifier as of 1effbb29 — golangci-lint's `unused` reported the named version as a dead declaration.

pkg/executor/handlers/system/system_test.go
jisung-02: Applied in a39818c6. `ctxManager.Shutdown` is registered first now, so LIFO drains the pool before anything cancels it, and the `drainThen` doc names the dependency instead of leaving it to be rediscovered.

Your reading of why it held was exactly right: `time.Sleep` does not watch a context and `MockCommandExecutor` ignores the one it is handed. Making `scheduleDelayedAction` cancellable would have broken all six for a reason unrelated to what they assert.

pkg/executor/dispatcher_benchmark_test.go
jisung-02: Applied in 67fc2334 — all five now read `require.NoError(b, err, "register handler")` through `require.TestingT`.

internal/retry/retry_test.go
jisung-02: Applied as suggested in dc332292.

Seen RED by letting 3.5s of fake time pass before the cancel instead of waiting for the park: `expected: 1 actual: 4`. The `>=` version passed that mutation, which is your point exactly.

pkg/scheduler/queue_test.go
jisung-02: Applied as suggested in 133906a2. On the bubble's clock the longer guard costs nothing, and it keeps the two from firing at the same instant.

geunwoonoh (approved):

PR #467, opened 2026-09-11
fix(executor): create missing parent directories under privilege demotion

### LINKED ISSUE #463: fix(file): demoted writeFileAs never creates the parent directory, so transfers into a missing directory fail
## Description

`writeFileAs` in `pkg/executor/handlers/file/file_io_unix.go` creates the parent directory only on the non-demoted branch. When `sysProcAttr` is set it runs `sh -c "tee <path> > /dev/null"` as the requested user and never creates the directory, so a transfer into a directory that does not exist yet fails before the extract worker (which does `os.MkdirAll(destDir)` as that user) ever runs.

Every transfer that comes through the server carries a `username`, so the demoted branch is the normal path, not the edge case. The non-demoted branch and the Windows implementation both call `os.MkdirAll(filepath.Dir(path), 0755)` first.

The asymmetry dates from dcf7fdc6 (2026-04-11): 590c1151 had made the write unconditional with `MkdirAll` in front of it, and the follow-up that restored `tee` for demotion moved `MkdirAll` under `if sysProcAttr == nil`. Before that, the demoted command was `tee ... && unzip -n ... -d ...; rm ...`, which did not create the directory either, so the demoted path has never done it. No test exercises the `tee` branch; `TestWriteFileAs_DirectPath_CreatesParentDir` covers only the direct branch.

## Steps to Reproduce

1. On a Linux host with alpamon running as root, pick a target directory that does not exist, for example `/tmp/test-upload-dir/`.
2. From a machine with the CLI, upload a folder as a demoted user: `alpacon cp -r ./test-dir <server>:/tmp/test-upload-dir/ -u <user>`. Any user other than the agent's own identity takes the demoted branch.
3. Observe the failure. A single-file upload into the same missing directory fails the same way.

The alpacon-cp-action CI reproduces it on every run of `test-recursive-upload`, for example https://github.com/alpacax/alpacon-cp-action/actions/runs/34090551602.

## Expected Behavior

The demoted branch creates the missing parent directories as the requested user, the same way the direct branch does, and the transfer succeeds. The zip is then extracted by the worker as before.

## Actual Behavior

`tee` cannot open the target file because its directory is missing. The raw `tee` error is returned to the console because the exit status is 1, not `EACCES`, so the `"You do not have permission to write to the directory, or directory does not exist"` message at `file.go:289` is never reached.

## Error Message

```
Failed to upload to '<server>:/tmp/test-upload-dir/':
  exit status 1: tee: /tmp/test-upload-dir/test-dir.zip: No such file or directory
```

## Notes for the fix

- Create the directories with the demoted credentials, not as the agent, so ownership and permissions match what the user could have created themselves. `mkdir -p` in the same `sh -c` as `tee`, or a small worker like the extract one, both work.
- `SanitizePath` cleans the string only. It does not resolve symlinks, and `tee` follows them, so a symlink component in the target path is followed today. The fix should not widen that: create the directory as the demoted user and do not add any agent-level `MkdirAll` on this path. Consider a follow-up that resolves the target the way `UnzipReader` resolves `destDir`, but that is a separate change.
- Add a test for the `tee` branch with a missing parent directory alongside `TestWriteFileAs_DirectPath_CreatesParentDir`.

jisung-02:
## Background

writeFileAs in pkg/executor/handlers/file/file_io_unix.go ran the demoted (privilege-lowered) write through `sh -c "tee <path> > /dev/null"` and never created the destination's parent directory. The non-demoted branch and the Windows implementation both call os.MkdirAll first. Every transfer coming through the server carries a username, so the demoted branch is the normal path, not the edge case—so any transfer into a directory that didn't exist yet failed outright. Issue #463 has the full repro and root-cause analysis.

## Changes

- The demoted branch's shell command is now `mkdir -p <parentDir> && tee <path> > /dev/null`, so the parent directory is created as the requesting user rather than as the agent (root), which keeps ownership correct.
- Added firstMissingAncestor, which walks up from the parent directory to find the highest ancestor mkdir -p would actually create. If tee subsequently fails, writeFileAs removes only that ancestor, leaving directories that already existed before the call untouched. Unconditionally removing the parent directory was rejected because it could delete a directory the caller never created.
- firstMissingAncestor now walks ancestors with os.Lstat instead of os.Stat, so an existing symlink in the parent chain (dangling or not) is treated as present rather than as something this call created. firstMissingAncestor's second return value was later dropped: every return path already satisfied `ok == (missing != "")`, so it carried no information the string didn't already carry.
- The failure-path cleanup no longer removes path with os.Remove when path is itself a directory (an empty one, for example, that already existed before the call), and no longer runs os.RemoveAll(createdRoot) unless the whole created tree still contains only plain directories.
- writeFileAs never removes a target this call didn't create. mkdir -p exits 0 on an existing parent regardless of permissions, so a tee that fails to open a pre-existing file used to leave it untouched on disk but still get deleted by an unconditional `os.Remove(path)` on the failure path—as the agent, i.e. root, so it succeeded against files the requester had no permission to write. writeFileAs now checks with os.Lstat before running the command and only includes a removal for a target that doesn't exist yet.
- That removal, when it does run, now happens inside the same demoted shell command instead of later in Go: the script becomes `mkdir -p P && { tee F > /dev/null || { rm -f F; exit 1; }; }` for a target this call is about to create. Doing it there, rather than as a later `os.Remove` running as root, closes a symlink-redirection window: `os.Remove` resolves every intermediate path component at unlink time, so a requester who controls a directory in the chain could otherwise swap it for a symlink between the check and the removal and have root unlink through it.
- The one removal path that still runs on the Go side—the source read failing, rather than tee itself failing—turned out to share both problems the shell-side fix above closed, and both are fixed now:
  - `erc.err != nil` does not prove tee ever opened path. A tee that can't open a pre-existing file still drains stdin to EOF and exits nonzero, so a source read failure can arrive with the target completely untouched. This removal used to run whenever erc.err was non-nil, with no check at all, so it deleted a pre-existing file the requester had no permission to write—the same class of bug fixed above for the tee-failure path. It's now gated on the same pre-existence check (createdTarget) that already decides which script to run, so it only ever removes a target this call created.
  - That removal also ran as root. It's now routed through a new removeAsRequester helper, which execs `rm -f path` with the caller's own SysProcAttr—the same permission boundary the script's own rm -f already runs under—using context.WithoutCancel since context cancellation is one of the ways this cleanup path is reached and the removal still has to run after ctx is done.
- writeFileAs now rejects a non-absolute path outright instead of silently anchoring it. It previously relied on `filepath.Clean("/" + path)` to always produce an absolute path, so `""` became `"/"` and `"relative/x"` became `"/relative/x"` instead of surfacing an error. No current caller hits this—fileDownload sanitizes first—but the doc comment claimed path always arrived sanitized, which wasn't actually enforced.
- dirTreeIsAllDirs's WalkDir callback is now typed `fs.DirEntry`, the type the `fs.WalkDirFunc` contract actually names (`os.DirEntry` is a type alias for it, so this was never a compile error, only a misleading signature).
- writeFileAs now starts with `path = filepath.Clean("/" + path)`, the exact shape CodeQL's go/path-injection query treats as a sanitizer. For the absolute path SanitizePath always produces this is a no-op. CodeQL was flagging every filesystem call in the function because fileDownload writes the sanitized value back into args.Path and CodeQL's field-based taint model does not treat that store as clearing the original wire taint. The `// lgtm[go/path-injection]` markers in the function never suppressed anything (alerts fired on lines carrying them) and are removed.
- The failure-handling collapse from `runErr != nil || erc.err != nil` to `runErr != nil` rested on cmd.Wait always returning the stdin-copy goroutine's error on a clean process exit and the process's own exit error otherwise. That invariant depends on two os/exec internals, not one—Cmd.Wait's own ordering, and `skipStdinCopyError` discarding EPIPE from the same copy—so writeFileAs now falls back to reporting `erc.err` explicitly after the `runErr != nil` block instead of silently returning nil if the invariant is ever violated.
- The `os.RemoveAll(createdRoot)` call now carries a comment stating why it still runs as root instead of through removeAsRequester: `dirTreeIsAllDirs` already limits what it can remove to a tree of plain directories, so the worst case is deleting empty directories, not a file. That reasoning previously lived only in the review thread.

## Safety

The directories mkdir -p creates get their mode from **alpamon's own inherited umask**, not the requesting user's. `sh -c` runs a non-login, non-interactive shell that never reads the requesting user's profile, so the mode is whatever alpamon itself inherited (typically 0755, matching the direct-write branch's `os.MkdirAll(dir, 0755)`, but not user-controlled and not deterministic across deployments—a systemd `UMask=` setting changes it). An earlier version of this section attributed the umask to the requesting user; that was wrong.

One race window remains open; a second one, thought closed earlier, wasn't and is now:

- `os.RemoveAll(createdRoot)`, the directory-tree cleanup on failure, still runs in Go as the agent (root) and still resolves intermediate path components at unlink time, the same symlink-redirection shape closed below for file removal. `dirTreeIsAllDirs` narrows the window (a write landing in the instant between its check and the RemoveAll call is still possible) but doesn't close it. Closing it fully needs a symlink-safe walk in the style of `pkg/utils/archive.go`'s `mkdirAllInside`/`createFile`, which is a bigger change than this cleanup path warrants right now—now noted inline at the call site so the reasoning doesn't depend on finding this section.
- The Go-side `os.Remove(path)` reached on a source-read failure was previously assumed safe on the premise that tee had already opened and truncated the target by the time that code ran. That premise was wrong—a tee that can't open a pre-existing file still drains stdin and exits nonzero, so this path could delete a file the requester has no permission to write, the exact bug already fixed for the tee-failure path. It's now gated on createdTarget and routed through removeAsRequester, so it only ever removes a target this call created, and only under the requester's own permissions.

## Testing

Coverage in pkg/executor/handlers/file/file_io_unix_test.go for the demoted tee path:
- creating a missing parent directory, and reusing it on a later write
- rejecting an unwritable parent directory, and a non-absolute path
- mkdir -p succeeding but tee failing on a path that is itself a pre-existing directory, and that directory surviving the failure (asserted via the path tee names, not its locale-dependent message text)
- a pre-existing, unwritable file surviving a tee failure that never opened it
- a pre-existing, unwritable file surviving a source read failure that arrives before tee ever gets a chance to fail on its own
- removing only the directory this call created, on failure
- removing a target this call freshly created when the source read fails mid-copy
- leaving a pre-existing parent directory untouched, on that same failure
- leaving a pre-existing target that tee already opened and truncated in its truncated state, rather than removing it, when the source read then fails
- firstMissingAncestor directly, over an existing directory, a missing chain, a dangling symlink parent, a non-directory parent, and a permission-denied stat
- dirTreeIsAllDirs directly, over a tree of plain directories and one containing a file
- the demoted user owning newly created parents (requires root, skipped otherwise)

Every new test and every extended assertion was seen failing once before being trusted: disabling cleanup entirely, forcing cleanup to always treat the immediate parent as newly created, reverting the os.Lstat directory guard on os.Remove, making dirTreeIsAllDirs always return true, removing the createdTarget/absolute-path guards, reverting firstMissingAncestor's os.Lstat to os.Stat, no-opping removeAsRequester, dropping the createdTarget gate on the source-read-failure removal, and—for the truncated-target test added this round—dropping that same gate again and watching the file disappear with "no such file or directory" instead of coming back empty.

The demoted-owner test's chmod on the temp directory's parent is restored via t.Cleanup; this environment isn't root, so that specific test still only SKIPs here rather than exercising red/green.

```
$ go build ./pkg/executor/...
$ go vet ./pkg/executor/handlers/file/...
$ go test ./pkg/executor/handlers/file/... -p 1
ok  	github.com/alpacax/alpamon/v2/pkg/executor/handlers/file	0.556s
```

Run against the pushed head (`9b1f70e8`). GitHub's own checks (Build and Test across all platforms, golangci-lint, govulncheck, CodeQL) were still queued/pending on this head at the time of writing—see the PR's Checks tab for their result.

CodeQL, run locally with the CodeQL CLI 2.27.0 and codeql/go-queries 1.6.10 against Security/CWE-022/TaintedPath.ql: a database built from the tree before the sanitizer wrapper reports 7 go/path-injection findings in pkg/executor/handlers/file/file_io_unix.go (the 4 the PR check raised on lines 64, 136, 137, 140, plus 3 in the non-demoted branch that predate this branch); a database built from the current head reports none in the file. The PR's CodeQL check on the pushed head is what confirms this on GitHub's side.

Closes #463

### DIFF
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -121,14 +121,17 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	createdRoot := firstMissingAncestor(parentDir)
 	// Only a target this call creates may be removed on failure: mkdir -p exits 0 on an
 	// existing parent, so a tee that fails to open a pre-existing file leaves it untouched,
-	// and removing it would delete something the caller never created.
-	_, targetStatErr := os.Lstat(path)
-	createdTarget := os.IsNotExist(targetStatErr)
+	// and removing it would delete something the caller never created. When we do remove it,
+	// the removal runs inside this same demoted shell, under the requester's real permissions,
+	// not later in Go as the agent (root)--which would otherwise let a symlink swapped into a
+	// path component redirect a root-privileged unlink.
+	script := fmt.Sprintf("mkdir -p %s && tee %s > /dev/null", utils.Quote(parentDir), utils.Quote(path))
+	if _, err := os.Lstat(path); os.IsNotExist(err) {
+		script = fmt.Sprintf("mkdir -p %s && { tee %s > /dev/null || { rm -f %s; exit 1; }; }",
+			utils.Quote(parentDir), utils.Quote(path), utils.Quote(path))
+	}
 	// Create parents as the requesting user to preserve filesystem permissions.
-	cmd := exec.CommandContext(ctx, "sh", "-c", fmt.Sprintf(
-		"mkdir -p %s && tee %s > /dev/null",
-		utils.Quote(parentDir), utils.Quote(path),
-	))
+	cmd := exec.CommandContext(ctx, "sh", "-c", script)
 	cmd.SysProcAttr = sysProcAttr
 	// Wrap src to preserve its read error even if a subsequent broken-pipe write
 	// overwrites it before cmd.Wait collects the goroutine result.
@@ -142,7 +145,10 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
 	// goroutine's error on a clean exit, and the process's own exit error otherwise.
 	if runErr != nil {
-		if createdTarget {
+		if erc.err != nil {
+			// tee already opened (and truncated) path before the source read failed, so the
+			// original content is already gone; a target tee never reached is instead
+			// protected by omitting the rm -f clause above.
 			if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
 				_ = os.Remove(path)
 			}
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -77,6 +77,18 @@ func firstMissingAncestor(dir string) string {
 	}
 }
 
+// removeAsRequester unlinks path under the requesting user's own permissions instead of the
+// agent's (root), so a symlink swapped into a path component after the caller's Lstat check
+// cannot redirect a root-privileged unlink. Run errors are swallowed: rm -f already reports
+// success on a missing target, and there is no recovery available from a demoted rm failing.
+// context.WithoutCancel is deliberate--context cancellation is one of the ways this cleanup
+// path is reached, and the removal must still run even after ctx is done.
+func removeAsRequester(ctx context.Context, path string, sysProcAttr *syscall.SysProcAttr) {
+	cmd := exec.CommandContext(context.WithoutCancel(ctx), "rm", "-f", path)
+	cmd.SysProcAttr = sysProcAttr
+	_ = cmd.Run()
+}
+
 // dirTreeIsAllDirs reports whether root and everything under it are plain directories,
 // so removing it cannot discard a file a concurrent writer placed there.
 func dirTreeIsAllDirs(root string) bool {
@@ -150,11 +162,11 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	if runErr != nil {
 		// erc.err alone does not prove tee opened path: a tee that cannot open a pre-existing
 		// file still drains stdin to EOF and exits nonzero, so a source read failure can arrive
-		// with the target untouched. Only remove it when this call created it--the same target
-		// the demoted rm -f in the script above would have covered had tee itself failed.
+		// with the target untouched. Only remove it when this call created it, and--same as the
+		// rm -f in the script above--under the requester's own permissions, not root's.
 		if erc.err != nil && createdTarget {
 			if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
-				_ = os.Remove(path)
+				removeAsRequester(ctx, path, sysProcAttr)
 			}
 		}
 		if createdRoot != "" && dirTreeIsAllDirs(createdRoot) {
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -125,8 +125,10 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	// the removal runs inside this same demoted shell, under the requester's real permissions,
 	// not later in Go as the agent (root)--which would otherwise let a symlink swapped into a
 	// path component redirect a root-privileged unlink.
+	_, targetStatErr := os.Lstat(path)
+	createdTarget := os.IsNotExist(targetStatErr)
 	script := fmt.Sprintf("mkdir -p %s && tee %s > /dev/null", utils.Quote(parentDir), utils.Quote(path))
-	if _, err := os.Lstat(path); os.IsNotExist(err) {
+	if createdTarget {
 		script = fmt.Sprintf("mkdir -p %s && { tee %s > /dev/null || { rm -f %s; exit 1; }; }",
 			utils.Quote(parentDir), utils.Quote(path), utils.Quote(path))
 	}
@@ -146,10 +148,11 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	// goroutine's error on a clean exit, and the process's own exit error otherwise. The
 	// fallback after this block guards that invariant instead of relying on it silently.
 	if runErr != nil {
-		if erc.err != nil {
-			// tee already opened (and truncated) path before the source read failed, so the
-			// original content is already gone; a target tee never reached is instead
-			// protected by omitting the rm -f clause above.
+		// erc.err alone does not prove tee opened path: a tee that cannot open a pre-existing
+		// file still drains stdin to EOF and exits nonzero, so a source read failure can arrive
+		// with the target untouched. Only remove it when this call created it--the same target
+		// the demoted rm -f in the script above would have covered had tee itself failed.
+		if erc.err != nil && createdTarget {
 			if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
 				_ = os.Remove(path)
 			}
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -80,7 +81,7 @@ func firstMissingAncestor(dir string) string {
 // so removing it cannot discard a file a concurrent writer placed there.
 func dirTreeIsAllDirs(root string) bool {
 	safe := true
-	_ = filepath.WalkDir(root, func(_ string, d os.DirEntry, err error) error {
+	_ = filepath.WalkDir(root, func(_ string, d fs.DirEntry, err error) error {
 		if err != nil || !d.IsDir() {
 			safe = false
 			return filepath.SkipAll
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -88,25 +88,24 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	cmd.Stderr = errW
 	runErr := cmd.Run()
 
-	if runErr != nil || erc.err != nil {
+	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
+	// goroutine's error on a clean exit, and the process's own exit error otherwise.
+	if runErr != nil {
 		// lgtm[go/path-injection]: path sanitized via SanitizePath, which
 		// rejects null bytes, UNC/device prefixes, and literal ".." after
 		// cleaning. Wire input is admin-authenticated.
 		_ = os.Remove(path) // lgtm[go/path-injection]
-		if runErr != nil {
-			var details []string
-			if msg := strings.TrimSpace(errW.buf.String()); msg != "" {
-				details = append(details, msg)
-			}
-			if erc.err != nil && erc.err != runErr {
-				details = append(details, erc.err.Error())
-			}
-			if len(details) > 0 {
-				return fmt.Errorf("%w: %s", runErr, strings.Join(details, "; "))
-			}
-			return runErr
+		var details []string
+		if msg := strings.TrimSpace(errW.buf.String()); msg != "" {
+			details = append(details, msg)
 		}
-		return fmt.Errorf("failed to read source: %w", erc.err)
+		if erc.err != nil && erc.err != runErr {
+			details = append(details, erc.err.Error())
+		}
+		if len(details) > 0 {
+			return fmt.Errorf("%w: %s", runErr, strings.Join(details, "; "))
+		}
+		return runErr
 	}
 	return nil
 }
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -119,7 +119,9 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 		// lgtm[go/path-injection]: path sanitized via SanitizePath, which
 		// rejects null bytes, UNC/device prefixes, and literal ".." after
 		// cleaning. Wire input is admin-authenticated.
-		_ = os.Remove(path) // lgtm[go/path-injection]
+		if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
+			_ = os.Remove(path) // lgtm[go/path-injection]
+		}
 		if tracksCreation {
 			_ = os.RemoveAll(createdRoot) // lgtm[go/path-injection]
 		}
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -56,20 +56,21 @@ func readFileAs(ctx context.Context, path string, sysProcAttr *syscall.SysProcAt
 	return rc, st.Size(), nil
 }
 
-// firstMissingAncestor returns the highest ancestor of dir that mkdir -p would create.
-// ok is false when dir exists or its state is unknown, so callers never remove a directory they didn't create.
-func firstMissingAncestor(dir string) (missing string, ok bool) {
+// firstMissingAncestor returns the highest ancestor of dir that mkdir -p would create,
+// or "" when dir exists or its state is unknown, so callers never remove a directory they didn't create.
+func firstMissingAncestor(dir string) string {
+	missing := ""
 	cur := dir
 	for {
 		if _, err := os.Lstat(cur); err == nil {
-			return missing, missing != ""
+			return missing
 		} else if !os.IsNotExist(err) {
-			return "", false
+			return ""
 		}
 		missing = cur
 		parent := filepath.Dir(cur)
 		if parent == cur {
-			return missing, true
+			return missing
 		}
 		cur = parent
 	}
@@ -113,7 +114,7 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 		return err
 	}
 	parentDir := filepath.Dir(path)
-	createdRoot, tracksCreation := firstMissingAncestor(parentDir)
+	createdRoot := firstMissingAncestor(parentDir)
 	// Create parents as the requesting user to preserve filesystem permissions.
 	cmd := exec.CommandContext(ctx, "sh", "-c", fmt.Sprintf(
 		"mkdir -p %s && tee %s > /dev/null",
@@ -135,7 +136,7 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 		if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
 			_ = os.Remove(path)
 		}
-		if tracksCreation && dirTreeIsAllDirs(createdRoot) {
+		if createdRoot != "" && dirTreeIsAllDirs(createdRoot) {
 			_ = os.RemoveAll(createdRoot)
 		}
 		var details []string
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -92,9 +92,12 @@ func dirTreeIsAllDirs(root string) bool {
 }
 
 // writeFileAs streams src to a file, demoting via tee when sysProcAttr is set. Caller owns src.
-// path is sanitized by callers via utils.SanitizePath, which rejects null bytes,
-// UNC/device prefixes, and literal ".." after cleaning.
+// path must already be absolute: callers sanitize it via utils.SanitizePath, which rejects
+// null bytes, UNC/device prefixes, and literal ".." after cleaning.
 func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *syscall.SysProcAttr) error {
+	if !filepath.IsAbs(path) {
+		return fmt.Errorf("invalid argument: path must be absolute: %s", path)
+	}
 	// No-op for the absolute path SanitizePath produces; it is the sanitizer shape CodeQL recognizes.
 	path = filepath.Clean("/" + path)
 	if sysProcAttr == nil {
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -61,7 +61,7 @@ func readFileAs(ctx context.Context, path string, sysProcAttr *syscall.SysProcAt
 func firstMissingAncestor(dir string) (missing string, ok bool) {
 	cur := dir
 	for {
-		if _, err := os.Stat(cur); err == nil {
+		if _, err := os.Lstat(cur); err == nil {
 			return missing, missing != ""
 		} else if !os.IsNotExist(err) {
 			return "", false
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -119,6 +119,11 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	}
 	parentDir := filepath.Dir(path)
 	createdRoot := firstMissingAncestor(parentDir)
+	// Only a target this call creates may be removed on failure: mkdir -p exits 0 on an
+	// existing parent, so a tee that fails to open a pre-existing file leaves it untouched,
+	// and removing it would delete something the caller never created.
+	_, targetStatErr := os.Lstat(path)
+	createdTarget := os.IsNotExist(targetStatErr)
 	// Create parents as the requesting user to preserve filesystem permissions.
 	cmd := exec.CommandContext(ctx, "sh", "-c", fmt.Sprintf(
 		"mkdir -p %s && tee %s > /dev/null",
@@ -137,8 +142,10 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
 	// goroutine's error on a clean exit, and the process's own exit error otherwise.
 	if runErr != nil {
-		if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
-			_ = os.Remove(path)
+		if createdTarget {
+			if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
+				_ = os.Remove(path)
+			}
 		}
 		if createdRoot != "" && dirTreeIsAllDirs(createdRoot) {
 			_ = os.RemoveAll(createdRoot)
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -131,12 +131,8 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	}
 	parentDir := filepath.Dir(path)
 	createdRoot := firstMissingAncestor(parentDir)
-	// Only a target this call creates may be removed on failure: mkdir -p exits 0 on an
-	// existing parent, so a tee that fails to open a pre-existing file leaves it untouched,
-	// and removing it would delete something the caller never created. When we do remove it,
-	// the removal runs inside this same demoted shell, under the requester's real permissions,
-	// not later in Go as the agent (root)--which would otherwise let a symlink swapped into a
-	// path component redirect a root-privileged unlink.
+	// Only a target this call created may be removed on failure, and only under the
+	// requester's own permissions, never as root--see removeAsRequester and the script's own rm -f.
 	_, targetStatErr := os.Lstat(path)
 	createdTarget := os.IsNotExist(targetStatErr)
 	script := fmt.Sprintf("mkdir -p %s && tee %s > /dev/null", utils.Quote(parentDir), utils.Quote(path))
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -56,6 +56,25 @@ func readFileAs(ctx context.Context, path string, sysProcAttr *syscall.SysProcAt
 	return rc, st.Size(), nil
 }
 
+// firstMissingAncestor returns the highest ancestor of dir that mkdir -p would create.
+// ok is false when dir exists or its state is unknown, so callers never remove a directory they didn't create.
+func firstMissingAncestor(dir string) (missing string, ok bool) {
+	cur := dir
+	for {
+		if _, err := os.Stat(cur); err == nil {
+			return missing, missing != ""
+		} else if !os.IsNotExist(err) {
+			return "", false
+		}
+		missing = cur
+		parent := filepath.Dir(cur)
+		if parent == cur {
+			return missing, true
+		}
+		cur = parent
+	}
+}
+
 // writeFileAs streams src to a file, demoting via tee when sysProcAttr is set. Caller owns src.
 // path is sanitized by callers via utils.SanitizePath, which rejects null bytes,
 // UNC/device prefixes, and literal ".." after cleaning.
@@ -77,7 +96,13 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 		}
 		return err
 	}
-	cmd := exec.CommandContext(ctx, "sh", "-c", fmt.Sprintf("tee %s > /dev/null", utils.Quote(path)))
+	parentDir := filepath.Dir(path)
+	createdRoot, tracksCreation := firstMissingAncestor(parentDir)
+	// Create parents as the requesting user to preserve filesystem permissions.
+	cmd := exec.CommandContext(ctx, "sh", "-c", fmt.Sprintf(
+		"mkdir -p %s && tee %s > /dev/null",
+		utils.Quote(parentDir), utils.Quote(path),
+	))
 	cmd.SysProcAttr = sysProcAttr
 	// Wrap src to preserve its read error even if a subsequent broken-pipe write
 	// overwrites it before cmd.Wait collects the goroutine result.
@@ -95,6 +120,9 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 		// rejects null bytes, UNC/device prefixes, and literal ".." after
 		// cleaning. Wire input is admin-authenticated.
 		_ = os.Remove(path) // lgtm[go/path-injection]
+		if tracksCreation {
+			_ = os.RemoveAll(createdRoot) // lgtm[go/path-injection]
+		}
 		var details []string
 		if msg := strings.TrimSpace(errW.buf.String()); msg != "" {
 			details = append(details, msg)
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -143,7 +143,8 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	runErr := cmd.Run()
 
 	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
-	// goroutine's error on a clean exit, and the process's own exit error otherwise.
+	// goroutine's error on a clean exit, and the process's own exit error otherwise. The
+	// fallback after this block guards that invariant instead of relying on it silently.
 	if runErr != nil {
 		if erc.err != nil {
 			// tee already opened (and truncated) path before the source read failed, so the
@@ -167,6 +168,9 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 			return fmt.Errorf("%w: %s", runErr, strings.Join(details, "; "))
 		}
 		return runErr
+	}
+	if erc.err != nil {
+		return fmt.Errorf("failed to read source: %w", erc.err)
 	}
 	return nil
 }
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -75,6 +75,20 @@ func firstMissingAncestor(dir string) (missing string, ok bool) {
 	}
 }
 
+// dirTreeIsAllDirs reports whether root and everything under it are plain directories,
+// so removing it cannot discard a file a concurrent writer placed there.
+func dirTreeIsAllDirs(root string) bool {
+	safe := true
+	_ = filepath.WalkDir(root, func(_ string, d os.DirEntry, err error) error {
+		if err != nil || !d.IsDir() {
+			safe = false
+			return filepath.SkipAll
+		}
+		return nil
+	})
+	return safe
+}
+
 // writeFileAs streams src to a file, demoting via tee when sysProcAttr is set. Caller owns src.
 // path is sanitized by callers via utils.SanitizePath, which rejects null bytes,
 // UNC/device prefixes, and literal ".." after cleaning.
@@ -122,7 +136,7 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 		if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
 			_ = os.Remove(path) // lgtm[go/path-injection]
 		}
-		if tracksCreation {
+		if tracksCreation && dirTreeIsAllDirs(createdRoot) {
 			_ = os.RemoveAll(createdRoot) // lgtm[go/path-injection]
 		}
 		var details []string
--- a/pkg/executor/handlers/file/file_io_unix.go
+++ b/pkg/executor/handlers/file/file_io_unix.go
@@ -93,11 +93,13 @@ func dirTreeIsAllDirs(root string) bool {
 // path is sanitized by callers via utils.SanitizePath, which rejects null bytes,
 // UNC/device prefixes, and literal ".." after cleaning.
 func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *syscall.SysProcAttr) error {
+	// No-op for the absolute path SanitizePath produces; it is the sanitizer shape CodeQL recognizes.
+	path = filepath.Clean("/" + path)
 	if sysProcAttr == nil {
 		if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
 			return err
 		}
-		f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644) // lgtm[go/path-injection]
+		f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
 		if err != nil {
 			return err
 		}
@@ -106,7 +108,7 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 			err = cerr
 		}
 		if err != nil {
-			_ = os.Remove(path) // lgtm[go/path-injection] drop partial write so retry isn't blocked by AllowOverwrite=false
+			_ = os.Remove(path) // drop partial write so retry isn't blocked by AllowOverwrite=false
 		}
 		return err
 	}
@@ -130,14 +132,11 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
 	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
 	// goroutine's error on a clean exit, and the process's own exit error otherwise.
 	if runErr != nil {
-		// lgtm[go/path-injection]: path sanitized via SanitizePath, which
-		// rejects null bytes, UNC/device prefixes, and literal ".." after
-		// cleaning. Wire input is admin-authenticated.
 		if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
-			_ = os.Remove(path) // lgtm[go/path-injection]
+			_ = os.Remove(path)
 		}
 		if tracksCreation && dirTreeIsAllDirs(createdRoot) {
-			_ = os.RemoveAll(createdRoot) // lgtm[go/path-injection]
+			_ = os.RemoveAll(createdRoot)
 		}
 		var details []string
 		if msg := strings.TrimSpace(errW.buf.String()); msg != "" {
--- a/pkg/executor/handlers/file/file_io_unix_test.go
+++ b/pkg/executor/handlers/file/file_io_unix_test.go
@@ -66,7 +66,8 @@ func TestWriteFileAs_TeePath_SurfacesTeeFailureAfterMkdirSucceeds(t *testing.T) 
 
 	err := writeFileAs(t.Context(), path, strings.NewReader("payload"), &syscall.SysProcAttr{})
 	require.Error(t, err)
-	assert.ErrorContains(t, err, "directory")
+	// tee's "Is a directory" message is locale-dependent; the path it names is not.
+	assert.ErrorContains(t, err, path)
 	assert.DirExists(t, path)
 }
 
--- a/pkg/executor/handlers/file/file_io_unix_test.go
+++ b/pkg/executor/handlers/file/file_io_unix_test.go
@@ -36,6 +36,12 @@ func TestWriteFileAs_TeePath_CreatesParentDir(t *testing.T) {
 	}
 }
 
+func TestWriteFileAs_RejectsNonAbsolutePath(t *testing.T) {
+	err := writeFileAs(t.Context(), "relative/x", strings.NewReader("payload"), nil)
+	require.Error(t, err)
+	assert.ErrorContains(t, err, "absolute")
+}
+
 func TestWriteFileAs_TeePath_RejectsUnwritableParent(t *testing.T) {
 	if os.Geteuid() == 0 {
 		t.Skip("root can write through directory mode restrictions")
--- a/pkg/executor/handlers/file/file_io_unix_test.go
+++ b/pkg/executor/handlers/file/file_io_unix_test.go
@@ -17,7 +17,7 @@ import (
 
 func TestWriteFileAs_TeePath_CreatesParentDir(t *testing.T) {
 	for _, dir := range []string{"nested/deep", "space and 'quote'/$literal;name"} {
-		t.Run(dir, func(t *testing.T) {
+		t.Run(strings.ReplaceAll(dir, "/", "_"), func(t *testing.T) {
 			path := filepath.Join(t.TempDir(), dir, "out 'file'.bin")
 			// Select the subprocess branch without changing credentials or requiring root.
 			err := writeFileAs(t.Context(), path, strings.NewReader("payload"), &syscall.SysProcAttr{})
--- a/pkg/executor/handlers/file/file_io_unix_test.go
+++ b/pkg/executor/handlers/file/file_io_unix_test.go
@@ -61,6 +61,7 @@ func TestWriteFileAs_TeePath_SurfacesTeeFailureAfterMkdirSucceeds(t *testing.T) 
 	err := writeFileAs(t.Context(), path, strings.NewReader("payload"), &syscall.SysProcAttr{})
 	require.Error(t, err)
 	assert.ErrorContains(t, err, "directory")
+	assert.DirExists(t, path)
 }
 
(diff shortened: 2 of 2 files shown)


pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -88,25 +113,27 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
...
 	runErr := cmd.Run()
 
-	if runErr != nil || erc.err != nil {
+	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
+	// goroutine's error on a clean exit, and the process's own exit error otherwise.
+	if runErr != nil {
 		// lgtm[go/path-injection]: path sanitized via SanitizePath, which
 		// rejects null bytes, UNC/device prefixes, and literal ".." after
 		// cleaning. Wire input is admin-authenticated.
 		_ = os.Remove(path) // lgtm[go/path-injection]
```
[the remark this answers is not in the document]
jisung-02: Fixed in 5a67e0e3. You were right that os.Remove also deletes empty directories—the failure path now checks os.Lstat(path) first and skips the removal when path is a directory, so a pre-existing empty directory survives a tee failure. Extended TestWriteFileAs_TeePath_SurfacesTeeFailureAfterMkdirSucceeds, which already exercised this exact case, with assert.DirExists(t, path).

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -88,25 +113,27 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
...
-			}
-			if erc.err != nil && erc.err != runErr {
-				details = append(details, erc.err.Error())
-			}
-			if len(details) > 0 {
-				return fmt.Errorf("%w: %s", runErr, strings.Join(details, "; "))
-			}
-			return runErr
+		if tracksCreation {
+			_ = os.RemoveAll(createdRoot) // lgtm[go/path-injection]
```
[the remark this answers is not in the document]
jisung-02: Fixed in 731ee37b. Added dirTreeIsAllDirs, which walks createdRoot and reports false if anything other than a plain directory shows up in it; RemoveAll now only runs when the whole tree still looks like an empty mkdir -p chain. This narrows the race rather than closing it—a write landing in the instant between the check and the RemoveAll call is still possible, closing that fully needs locking the directory chain for the call, which felt like a bigger change than this cleanup path warrants. Said so in the PR body under Safety.

pkg/executor/handlers/file/file_io_unix_test.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func TestWriteFileAs_DemotedPath_CreatesParentsAsUser(t *testing.T) {
+	if os.Geteuid() != 0 {
+		t.Skip("credential demotion requires root")
+	}
+	// Only the demoted user should own the newly created parents.
+	const uid, gid = 65534, 65534
+	dir := t.TempDir()
+	require.NoError(t, os.Chmod(filepath.Dir(dir), 0755))
```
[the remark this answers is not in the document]
jisung-02: Fixed in 32bd5812. Recorded the parent's mode with os.Stat before chmod and restored it via t.Cleanup, matching how TestWriteFileAs_TeePath_RejectsUnwritableParent already restores what it changes. Could not verify red/green here since this test requires root and this environment is not root, so it still only SKIPs locally.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -56,6 +56,25 @@ func readFileAs(ctx context.Context, path string, sysProcAttr *syscall.SysProcAt
 	return rc, st.Size(), nil
 }
 
+// firstMissingAncestor returns the highest ancestor of dir that mkdir -p would create.
+// ok is false when dir exists or its state is unknown, so callers never remove a directory they didn't create.
+func firstMissingAncestor(dir string) (missing string, ok bool) {
+	cur := dir
+	for {
+		if _, err := os.Stat(cur); err == nil {
```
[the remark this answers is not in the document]
jisung-02: Fixed in 5d01777b. Switched firstMissingAncestor from os.Stat to os.Lstat, so an existing symlink in the ancestor chain (dangling or not) reports as present instead of ENOENT and is never classified as something this call created.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -88,25 +129,26 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
...
-			}
-			if len(details) > 0 {
-				return fmt.Errorf("%w: %s", runErr, strings.Join(details, "; "))
-			}
-			return runErr
+	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
+	// goroutine's error on a clean exit, and the process's own exit error otherwise.
+	if runErr != nil {
+		if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
+			_ = os.Remove(path)
```
geunwoonoh: Critical: this removes a pre-existing target that the demoted write never opened, and it runs as root.

`mkdir -p` exits 0 on an existing parent regardless of permissions, so when `tee` then fails with `EACCES` the target file is left completely untouched. `os.Lstat` reports it as a regular file, the `!fi.IsDir()` guard passes, and `os.Remove` deletes it as the agent, i.e. root—so it succeeds on files the requester has no permission to write.

Verified locally, with `dir` at 0555 and the target at 0444:

```
runErr=exit status 1 stderr="tee: …/precious.conf: Permission denied"
after failed tee: content="ORIGINAL CONTENT\n"
cleanup: os.Remove(…/precious.conf) -> <nil>
target still present? false
```

In production that is a transfer to `/etc/motd` with `AllowOverwrite=true` demoted to `alice`: `mkdir -p /etc` exits 0, `tee` fails with `Permission denied`, and the cleanup deletes a file `alice` cannot write. A requester can delete any file the agent can reach by aiming a transfer they know will fail at it.

This predates the branch—the old code called `os.Remove(path)` unconditionally—so the PR strictly improves the situation. But it is this PR's own stated principle ("removing the parent directory was rejected because it could delete a directory the caller never created") applied to directories and not to files, and the fix reuses the `createdRoot` pattern already in the diff. Note the direct branch at line 110 does not have this hole: it only reaches `os.Remove` after its own `OpenFile` succeeded, so the file is provably one this call created.

Two edits, next to the existing ones:

```go
	parentDir := filepath.Dir(path)
	createdRoot, tracksCreation := firstMissingAncestor(parentDir)
	// Only a target this call created may be removed on failure: mkdir -p exits 0
	// on an existing parent, so a tee that fails to open a pre-existing file
	// leaves it untouched and it must survive the rollback.
	_, targetStatErr := os.Lstat(path)
	createdTarget := os.IsNotExist(targetStatErr)

	// ... unchanged ...

	if runErr != nil {
		if createdTarget {
			if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
				_ = os.Remove(path)
			}
		}
```

A regression guard for it would be the file mirror of `TestWriteFileAs_TeePath_SurfacesTeeFailureAfterMkdirSucceeds`: non-writable file at the target, expect the error, assert the original content survives.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -88,25 +129,26 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
...
+		}
+		if tracksCreation && dirTreeIsAllDirs(createdRoot) {
+			_ = os.RemoveAll(createdRoot)
```
geunwoonoh, line 169: Warning: this removal, and the `os.Remove` three lines up, resolve intermediate path components at unlink time, so a requester who controls a parent directory can redirect them.

`os.Remove` does not follow the *final* component, but it does follow every directory above it, and the `os.Lstat` guard on line 135 is a separate syscall from the removal. A requester who owns a directory in the chain (their own home, the usual case) can make `tee` fail deterministically, then swap that directory for a symlink to `/etc` and have root unlink `/etc/<basename>`. The window is small and winning it needs a retry loop, but every failed transfer is a free attempt and both ends are under their control.

This repo already refuses exactly this class elsewhere: `mkdirAllInside` and `createFile` in `pkg/utils/archive.go:553-604` walk one component at a time, reject a symlink component outright, and open with `O_NOFOLLOW` where the host has it.

The cleanest fix is to stop removing anything with privilege: move the `tee` rollback into the demoted shell, so the kernel enforces the requester's own permissions on the removal too.

```sh
mkdir -p P && { tee F > /dev/null || { rm -f F; exit 1; }; }
```

That does not cover the source-read-error case (there `tee` sees EOF and exits 0, so the `||` branch never runs), which still needs the Go-side removal—but that case is one where `tee` provably created the file, which is the safe subset.

Pre-existing, race-dependent, and a follow-up issue is a reasonable home for it. Recording it here because this PR is what rewrote these lines.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -88,25 +129,26 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
...
-			}
-			if erc.err != nil && erc.err != runErr {
-				details = append(details, erc.err.Error())
-			}
-			if len(details) > 0 {
-				return fmt.Errorf("%w: %s", runErr, strings.Join(details, "; "))
-			}
-			return runErr
+	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
+	// goroutine's error on a clean exit, and the process's own exit error otherwise.
```
geunwoonoh: Warning: the collapse is correct, but the failure mode when the invariant does not hold is a silent success.

`Cmd.Wait` reports the copy goroutine's error only when the process itself exited cleanly (`os/exec/exec.go:940-945`), which is what this comment says. What it omits is the other half: the stdin copier also discards `EPIPE` via `skipStdinCopyError`. So the invariant rests on two `os/exec` internals, and if it is ever violated `writeFileAs` returns `nil` for a truncated file. The old code returned `failed to read source: …` on that path.

No reachable trigger was found—a child that exits 0 without draining stdin does not exist for `mkdir -p … && tee …`—so this is robustness, not a live bug, and the dead branch was worth deleting. A one-line fallback keeps the simplification while removing the silent-nil outcome:

```go
	if erc.err != nil {
		return fmt.Errorf("failed to read source: %w", erc.err)
	}
	return nil
```

Worth extending the comment to name `skipStdinCopyError` too, since that is the part a future reader is least likely to reconstruct.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -56,15 +56,50 @@ func readFileAs(ctx context.Context, path string, sysProcAttr *syscall.SysProcAt
 	return rc, st.Size(), nil
 }
 
+// firstMissingAncestor returns the highest ancestor of dir that mkdir -p would create.
+// ok is false when dir exists or its state is unknown, so callers never remove a directory they didn't create.
+func firstMissingAncestor(dir string) (missing string, ok bool) {
```
geunwoonoh: Suggestion: the second return value carries no information. Every return path satisfies `ok == (missing != "")`—`("", false)` on a stat error, `(missing, missing != "")` when `cur` exists, and `(missing, true)` at the root where `missing` is never empty. So `tracksCreation` at the call site is always `createdRoot != ""`.

Dropping it removes a variable, a doc line, and the chance of the two drifting apart:

```go
// firstMissingAncestor returns the highest ancestor of dir that mkdir -p would create,
// or "" when dir exists or its state is unknown, so callers never remove a directory they didn't create.
func firstMissingAncestor(dir string) string {
```

with the call site becoming `if createdRoot != "" && dirTreeIsAllDirs(createdRoot)`.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -56,15 +56,50 @@ func readFileAs(ctx context.Context, path string, sysProcAttr *syscall.SysProcAt
...
+		}
+		cur = parent
+	}
+}
+
+// dirTreeIsAllDirs reports whether root and everything under it are plain directories,
+// so removing it cannot discard a file a concurrent writer placed there.
+func dirTreeIsAllDirs(root string) bool {
+	safe := true
+	_ = filepath.WalkDir(root, func(_ string, d os.DirEntry, err error) error {
```
geunwoonoh: Minor: the `fs.WalkDirFunc` contract is `func(string, fs.DirEntry, error) error`. `os.DirEntry` is an alias so this compiles, but `fs.DirEntry` is what the signature is implementing, and it reads that way to someone checking the callback against the interface.

The short-circuit on `err != nil` before `d.IsDir()` is right, by the way—`d` is nil on a walk error, and this ordering is what keeps that from panicking. Worth keeping exactly as is.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -56,15 +56,50 @@ func readFileAs(ctx context.Context, path string, sysProcAttr *syscall.SysProcAt
...
+			return filepath.SkipAll
+		}
+		return nil
+	})
+	return safe
+}
+
 // writeFileAs streams src to a file, demoting via tee when sysProcAttr is set. Caller owns src.
 // path is sanitized by callers via utils.SanitizePath, which rejects null bytes,
 // UNC/device prefixes, and literal ".." after cleaning.
```
geunwoonoh: Suggestion: after this change the doc comment above states a precondition the body now silently repairs instead of relying on. `filepath.Clean("/" + path)` is a no-op for the absolute path `SanitizePath` produces (confirmed), but for a relative one it root-anchors it: `relative/x` becomes `/relative/x`, and `""` becomes `/`.

No current caller does that—`fileDownload` sanitizes first, and the tests pass `t.TempDir()` paths—so this is about the next caller, not this one. Rejecting a non-absolute path would make the comment true and cost nothing:

```go
	if !filepath.IsAbs(path) {
		return fmt.Errorf("invalid argument: path must be absolute: %s", path)
	}
```

Also worth noting the asymmetry this introduces with `file_io_windows.go:26-41`, which is otherwise byte-identical to the `sysProcAttr == nil` branch below and does not get the same normalization.

pkg/executor/handlers/file/file_io_unix_test.go
```diff
@@ -0,0 +1,127 @@
...
+func TestWriteFileAs_TeePath_SurfacesTeeFailureAfterMkdirSucceeds(t *testing.T) {
+	dir := t.TempDir()
+	// mkdir -p is a no-op here since the parent already exists, so this exercises
+	// "mkdir succeeds, tee fails" rather than the mkdir-failure branch above.
+	path := filepath.Join(dir, "target")
+	require.NoError(t, os.Mkdir(path, 0755))
+
+	err := writeFileAs(t.Context(), path, strings.NewReader("payload"), &syscall.SysProcAttr{})
+	require.Error(t, err)
+	assert.ErrorContains(t, err, "directory")
```
geunwoonoh: Suggestion: this asserts on `tee`'s localized `strerror` text. Under a non-C locale the message is translated (`Ist ein Verzeichnis`) and the test fails for a reason unrelated to the code. The path in `tee`'s message is locale-independent and discriminates just as well between "tee failed on this target" and any other failure:

```suggestion
	assert.ErrorContains(t, err, path)
```

pkg/executor/handlers/file/file_io_unix_test.go
```diff
@@ -0,0 +1,127 @@
...
+	assert.DirExists(t, path)
+}
+
+func TestDirTreeIsAllDirs(t *testing.T) {
```
geunwoonoh, line 153: Suggestion: `firstMissingAncestor` has no direct test, and the symlink guarantee its doc comment makes—the reason for `5d01777`—has no regression guard.

Through `writeFileAs` the two protections overlap. For a dangling symlink parent, reverting `os.Lstat` to `os.Stat` changes `createdRoot` from `""` to the symlink path, but `dirTreeIsAllDirs` then rejects it on its own (`WalkDir` lstats the root, sees a symlink, `!d.IsDir()`), so nothing is removed either way and every current test stays green. The guard that would catch the regression lives in a different function than the one that was fixed.

A direct table test pins it: existing directory, missing chain, dangling symlink parent, non-directory parent, and a stat error that is not `ENOENT`.

This is also the one change the PR body's mutation list does not actually cover—the entry there ("reverting the os.Lstat directory guard on os.Remove") is the cleanup `Lstat` at line 135, a different call site.

pkg/executor/handlers/file/file_io_unix_test.go
```diff
@@ -0,0 +1,127 @@
...
+	t.Cleanup(func() { require.NoError(t, os.Chmod(dir, 0700)) })
+	path := filepath.Join(dir, "missing", "out.bin")
+
+	err := writeFileAs(t.Context(), path, strings.NewReader("payload"), &syscall.SysProcAttr{})
+	require.Error(t, err)
+	_, err = os.Stat(filepath.Dir(path))
+	assert.ErrorIs(t, err, os.ErrNotExist)
+}
+
+func TestWriteFileAs_TeePath_SurfacesTeeFailureAfterMkdirSucceeds(t *testing.T) {
```
geunwoonoh, line 60: Suggestion: this test got `assert.DirExists` when the Copilot finding was fixed, which is right—but the file case has no equivalent, and it is the one that currently misbehaves.

The mirror of this test is the regression guard for the critical issue on line 136: put a non-writable *file* at the target instead of a directory, expect the error, and assert the original content survives. It fails today, because `os.Remove` deletes a file `tee` never managed to open.

```go
func TestWriteFileAs_TeePath_KeepsUnwritableTargetOnTeeFailure(t *testing.T) {
	if os.Geteuid() == 0 {
		t.Skip("root can write through file mode restrictions")
	}
	path := filepath.Join(t.TempDir(), "precious.conf")
	require.NoError(t, os.WriteFile(path, []byte("ORIGINAL"), 0444))

	err := writeFileAs(t.Context(), path, strings.NewReader("payload"), &syscall.SysProcAttr{})
	require.Error(t, err)
	got, readErr := os.ReadFile(path)
	require.NoError(t, readErr)
	assert.Equal(t, "ORIGINAL", string(got))
}
```

geunwoonoh (changes requested):
## Summary

The demoted write path never created the destination's parent directory, so every transfer into a not-yet-existing directory failed. The fix folds `mkdir -p` into the same demoted `sh -c` as `tee`, which is the right call: parents end up owned by the requesting user, and it costs no extra process spawn beyond `mkdir` itself. The rollback work (`firstMissingAncestor`, `dirTreeIsAllDirs`) is careful, and the four Copilot findings were each fixed at the cause rather than papered over.

Verification done for this review: all 27 checks pass on head `8de7e62`, and `code-scanning/alerts?state=open` returns nothing for this file, so the PR's CodeQL claim holds. `filepath.Clean("/" + path)` is confirmed a no-op for the absolute paths `SanitizePath` produces. The `os/exec` `Wait` ordering in GOROOT confirms the collapsed `erc.err` branch was dead. No case was found where `mkdir -p` fails where the old `tee`-only command would have succeeded—`mkdir -p` exits 0 on an existing directory even without write access to it, checked directly.

One reachable data-loss path remains in the rewritten cleanup, described below.

**Performance**: net cost is a handful of `os.Lstat` calls per demoted write plus one `mkdir` exec; `dirTreeIsAllDirs` walks only on the failure path and stops early via `filepath.SkipAll`. Nothing here needs optimizing. Skipping the `mkdir -p` exec when `firstMissingAncestor` already proved the parent exists would save one fork per transfer, but it reintroduces a TOCTOU window for a sub-millisecond gain—not worth it.

## Critical issues (1)

**1. [`file_io_unix.go:135-137`] The failure cleanup removes a pre-existing target that the demoted write never opened, as root.**

`mkdir -p` exits 0 on an existing parent regardless of permissions, so when `tee` then fails with `EACCES` the target file is left completely untouched—and `os.Lstat` reports it as a regular file, so `os.Remove` deletes it. That removal runs as the agent, i.e. root, so it succeeds on files the requester has no permission to write.

Failure scenario, verified locally: a transfer to `/etc/motd` with `AllowOverwrite=true` demoted to `alice`. `mkdir -p /etc` exits 0. `tee /etc/motd` fails with `Permission denied`, file intact. Cleanup deletes `/etc/motd`. A requester can delete any file the agent can reach by aiming a transfer they know will fail at it.

```
runErr=exit status 1 stderr="tee: …/precious.conf: Permission denied"
after failed tee: content="ORIGINAL CONTENT\n"
cleanup: os.Remove(…/precious.conf) -> <nil>
target still present? false
```

This predates the branch—the old code called `os.Remove(path)` unconditionally—so the PR strictly improves the situation rather than regressing it. It is still the PR's own stated principle ("removing the parent directory was rejected because it could delete a directory the caller never created") applied to directories and not to files, and the fix is the same three lines the `createdRoot` logic already uses: record pre-existence before the run, and only remove what this call created. The direct branch at line 110 does not have this hole, because it only reaches `os.Remove` after its own `OpenFile` succeeded.

## Warnings (3)

**1. [`file_io_unix.go:136,139`] The root-privileged removal resolves intermediate path components at unlink time, so it can be redirected by a parent the requester controls.**

`os.Remove` does not follow the *final* component, but it does follow every directory above it; the `os.Lstat` guard on line 135 is a separate syscall from the `os.Remove` on 136. A requester who owns a directory in the chain (their own home, the usual case) can make `tee` fail deterministically, then swap that directory for a symlink to `/etc` and have root unlink `/etc/<basename>`. The window is small and winning it needs a retry loop, but each failed transfer is a free attempt. The same shape applies to `os.RemoveAll(createdRoot)`.

This repo already has a precedent for refusing exactly this class: `mkdirAllInside` and `createFile` in `pkg/utils/archive.go:553-604` walk one component at a time and refuse to write through a symlink component, with `O_NOFOLLOW` where the host has it. The cleanest fix is to stop doing privileged removal at all: put the `tee` rollback inside the demoted shell (`mkdir -p P && { tee F > /dev/null || { rm -f F; exit 1; }; }`), so the kernel enforces the requester's own permissions on the removal too. Pre-existing, and a follow-up issue is a reasonable place for it—but worth recording while this path is being rewritten.

**2. [`file_io_unix.go:132-134`] The collapse to `runErr != nil` is correct, but its consequence when the invariant does not hold is a silent success rather than a wrong error.**

`Cmd.Wait` reports the copy goroutine's error only when the process itself exited cleanly (`exec.go:940-945`), and the stdin copier discards `EPIPE` via `skipStdinCopyError`. So the comment's invariant depends on two `os/exec` internals, not just the one it names, and if it is ever violated `writeFileAs` returns `nil` for a truncated file—the old code returned `failed to read source: …` there. No reachable trigger was found (a child that exits 0 without draining stdin does not exist for this command), so this is robustness, not a live bug. A one-line `if erc.err != nil { return fmt.Errorf("failed to read source: %w", erc.err) }` after the `runErr` block keeps the simplification and removes the silent-nil outcome.

**3. [PR body, "Safety" section] The umask attribution is wrong and will mislead the next reader.**

`sh -c` is a non-login, non-interactive shell: it never reads the requesting user's profile, so the mode of the new directories comes from **alpamon's** inherited umask, not the user's. The outcome is usually `0755` and therefore matches the direct branch's `os.MkdirAll(dir, 0755)`—but it is not user-controlled, and it is not deterministic across deployments (a systemd `UMask=0027` changes it). If parity with the direct branch is the actual intent, `mkdir -m 755 -p` pins the final component (intermediates still follow umask). At minimum, the sentence should be corrected.

## Suggestions (5)

**1. [`file_io_unix.go:61,116`] `firstMissingAncestor`'s second return value carries no information.** Every return path satisfies `ok == (missing != "")`, so `tracksCreation` is always `createdRoot != ""`. Dropping it removes a variable and a doc line.

**2. [`file_io_unix_test.go:67`] `firstMissingAncestor` has no direct test, and the symlink guarantee in its doc comment—the reason for `5d01777`—has no regression guard.** Through `writeFileAs` the two protections overlap: `dirTreeIsAllDirs` rejects a symlink root on its own, so reverting `os.Lstat` to `os.Stat` still leaves the symlink intact and every current test stays green. Only a direct table test over (existing dir, missing chain, dangling symlink parent, non-directory parent, stat error) pins the behavior the comment promises. This is also the one change the PR body's mutation list does not cover—the entry there ("reverting the os.Lstat directory guard on os.Remove") is the cleanup `Lstat` at line 135, a different call.

**3. [`file_io_unix_test.go:54`] The directory case got `assert.DirExists` when the Copilot finding was fixed; the file case has no equivalent.** A test that puts a non-writable file at the target, expects the error, and asserts the original content survives is the regression guard for the critical issue above—and it fails today.

**4. [`file_io_unix_test.go:63`] `assert.ErrorContains(t, err, "directory")` asserts on `tee`'s localized `strerror` text.** Under a non-C locale the message is translated (`Ist ein Verzeichnis`) and the test fails for a reason unrelated to the code. Asserting the path instead keeps the same discrimination and is locale-independent.

**5. Minor, bundled.** `dirTreeIsAllDirs`'s callback takes `os.DirEntry` where the `fs.WalkDirFunc` contract is `fs.DirEntry` (an alias, so it compiles—`fs.DirEntry` reads as the intended type); the subtest name `"nested/deep"` at test line 19 becomes two `-run` path segments, which makes the case harder to target individually; and the doc comment at lines 92-94 still states `path` arrives sanitized while the body now silently repairs it—`filepath.Clean("/" + "relative/x")` yields `/relative/x`, so a relative path becomes root-anchored instead of erroring. Rejecting a non-absolute path would make the comment true.

Outside the diff, on dead and duplicated logic: `file.go:289`'s `os.IsPermission(err) || errors.Is(err, syscall.EACCES)` can never match on the demoted path (the error is always an `ExitError`), so that friendly message is unreachable for what the PR body correctly calls the normal path. And the `sysProcAttr == nil` branch here is byte-identical to `file_io_windows.go:26-41`; the two have now started to diverge (the `Clean` line exists only on Unix), which is the usual precursor to one being fixed and the other not.

## Highlights

- Folding `mkdir -p` into the existing `sh -c` instead of spawning a second demoted process is the right shape, and doing the creation demoted rather than as root is what keeps ownership correct. The rejected alternative (root `MkdirAll` then `chown`) would have been racier and wrong.
- `"space and 'quote'/$literal;name"` as a subtest name genuinely exercises `utils.Quote`; `pattern` (`[^\w@%+=:,./-]`) quotes all of it and escapes the embedded `'` correctly. That is the test most PRs in this area forget.
- Each Copilot finding was fixed at the cause with its own commit and an honest reply, including the one where the residual race was documented rather than claimed closed.
- `TestWriteFileAs_DemotedPath_CreatesParentsAsUser` chmods the *per-test* `MkdirTemp` root, not `/tmp`—Copilot's premise was wrong about which directory that is (Go's `t.TempDir` returns `<MkdirTemp 0700>/001`, so the parent is the private per-test root, verified). The chmod is both necessary and correctly restored.
- `assert.ErrorIs(t, err, os.ErrNotExist)` off `os.Stat` at test line 50 rather than `assert.NoFileExists` follows this repo's testing convention precisely.

---

The change itself is a net improvement and does not regress anything found in this review. The blocker is narrow: the cleanup this PR rewrote still deletes a pre-existing file that the demoted write never opened, as root. It predates the branch, so treating it as a separate issue is a defensible call—but it is three lines using the `createdRoot` pattern already in the diff, in a branch dedicated to making exactly this cleanup path safe.

**Checklist**:
- [x] Tests included and passing (27/27 checks green on `8de7e62`)
- [ ] No security vulnerabilities—root-privileged removal of a target the call never created (critical 1) and symlink-redirectable unlink (warning 1)
- [x] Documentation updated if needed—the PR body's umask rationale needs the correction in warning 3

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Fixed in e1ebfc13. You were right: mkdir -p succeeding on an existing parent meant tee's open failure left the target completely untouched, while the unconditional os.Remove(path) still deleted it as root. Added an os.Lstat check before running the command, so the target is only removed on failure when it didn't exist beforehand. Added TestWriteFileAs_TeePath_KeepsUnwritableTargetOnTeeFailure, seen RED first against the old unconditional removal ("no such file or directory" reading back a file that had existed moments earlier).

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Partially fixed in 00c152d9. For the file removal: moved the rm -f into the same demoted shell command instead of a later Go-side os.Remove, so the unlink runs under the requester's real uid and is bound by their own directory permissions rather than root's, closing the redirection window for that path. The os.RemoveAll(createdRoot) directory-tree cleanup you also flagged still runs in Go as root with the same exposure; closing that fully needs a symlink-safe walk like archive.go's mkdirAllInside/createFile, which is a bigger change than this cleanup path warrants right now. Documented as an open, pre-existing risk in the PR body's Safety section rather than filing a separate issue - let me know if you'd rather track it there instead.

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Fixed in 30499117. Added a fallback after the runErr != nil block: if erc.err is ever non-nil there, it now surfaces as "failed to read source" instead of silently returning nil. Agreed this is defensive rather than a live bug - no trigger exists today, since the invariant holds for this specific command - so no test was added; there's no way to trigger it without faking os/exec's own internals.

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Fixed in 41ec0375, exactly as suggested: dropped the second return value and switched the call site to createdRoot != "".

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Fixed in 487aa9d4, switched to fs.DirEntry. Also folded in the "nested/deep" subtest-name fix you noted in the review body's bundled note - renamed to nested_deep so it can't split into two -run path segments.

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Fixed in 546a84f4. writeFileAs now rejects a non-absolute path outright instead of relying on filepath.Clean("/" + path) to silently anchor it. Added TestWriteFileAs_RejectsNonAbsolutePath, seen RED first: without the guard, the write went ahead against "/relative" and failed on a read-only-filesystem error instead of catching the argument mistake.

pkg/executor/handlers/file/file_io_unix_test.go
jisung-02: Fixed in a8da4e5b, asserting on the path tee names instead of the word "directory". Tried to reproduce the original failure with LC_ALL=de_DE.UTF-8 first: it didn't reproduce on this darwin environment, since macOS's tee is a BSD build with no localized message catalog. The path-based assertion is still the more robust choice for the glibc/coreutils case you described.

pkg/executor/handlers/file/file_io_unix_test.go
jisung-02: Fixed in d6cdeced. Added a table test covering an existing directory, a missing chain, a dangling symlink parent, a non-directory parent, and a permission-denied stat. Seen RED first: reverting os.Lstat to os.Stat failed the dangling-symlink case - exactly the regression you pointed at.

pkg/executor/handlers/file/file_io_unix_test.go
jisung-02: Fixed in e1ebfc13, alongside the critical fix - added the file-mirror test you sketched (TestWriteFileAs_TeePath_KeepsUnwritableTargetOnTeeFailure), same shape as your suggested snippet.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -88,24 +142,34 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
...
-				details = append(details, erc.err.Error())
+	// erc.err is only ever non-nil alongside a non-nil runErr: cmd.Wait returns the stdin-copy
+	// goroutine's error on a clean exit, and the process's own exit error otherwise. The
+	// fallback after this block guards that invariant instead of relying on it silently.
+	if runErr != nil {
+		if erc.err != nil {
+			// tee already opened (and truncated) path before the source read failed, so the
+			// original content is already gone; a target tee never reached is instead
+			// protected by omitting the rm -f clause above.
+			if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
```
geunwoonoh, line 164: Critical: on the read-error path this still deletes a pre-existing target that `tee` never opened, and it runs as root.

`erc.err != nil` means the source failed, not that `tee` opened `path`. When `tee` cannot open its file (a pre-existing target at 0444, say) it keeps draining stdin into `/dev/null` and exits 1 at EOF, so a source failure arrives with the target untouched, and this branch removes it. The comment above ("tee already opened (and truncated) path") is where it breaks. That premise came from my comment on line 139, which said this case "is one where `tee` provably created the file". That was wrong.

Reproduced against this head's `writeFileAs`:

```
writeFileAs err = exit status 1: tee: …/precious.conf: Permission denied; connection reset
open …/precious.conf: no such file or directory
```

The pre-existence check on line 129 already has the answer. Keeping it in a variable and gating this removal on it closes the gap:

```go
	_, targetStatErr := os.Lstat(path)
	createdTarget := os.IsNotExist(targetStatErr)
	script := fmt.Sprintf("mkdir -p %s && tee %s > /dev/null", utils.Quote(parentDir), utils.Quote(path))
	if createdTarget {
		script = fmt.Sprintf("mkdir -p %s && { tee %s > /dev/null || { rm -f %s; exit 1; }; }",
			utils.Quote(parentDir), utils.Quote(path), utils.Quote(path))
	}
	// ...
	if runErr != nil {
		if erc.err != nil && createdTarget {
```

A pre-existing target that `tee` did open and truncate is then left truncated rather than removed. That is acceptable: reaching it required `AllowOverwrite=true`, so a retry is not blocked, and nothing `tee` didn't touch gets deleted.

Optional hardening on the same line: this is the one remaining root-privileged unlink, so the symlink-swap concern the comment on lines 124-127 describes still applies here. Running `rm -f` through `exec.CommandContext(context.WithoutCancel(ctx), "rm", "-f", path)` with the same `SysProcAttr` removes it under the requester's permissions instead. `WithoutCancel` matters because context cancellation is one of the ways into this branch.

Regression test. `errors` and `testing/iotest` are already imported:

```go
func TestWriteFileAs_TeePath_KeepsUnwritableTargetOnSourceReadFailure(t *testing.T) {
	if os.Geteuid() == 0 {
		t.Skip("root can write through file mode restrictions")
	}
	path := filepath.Join(t.TempDir(), "precious.conf")
	require.NoError(t, os.WriteFile(path, []byte("ORIGINAL"), 0444))

	err := writeFileAs(t.Context(), path, iotest.ErrReader(errors.New("connection reset")), &syscall.SysProcAttr{})
	require.Error(t, err)
	got, readErr := os.ReadFile(path)
	require.NoError(t, readErr)
	assert.Equal(t, "ORIGINAL", string(got))
}
```

It fails on this head and passes with the gate above.

geunwoonoh (changes requested):
## Summary

Re-review of head `a8da4e5` against the previous review on `8de7e62`. All 27 checks are green. Nearly every finding landed well, and the move of the `tee`-failure rollback into the demoted shell is a better fix than the one the previous review proposed. One path of the original critical issue survives, narrowed but with the same outcome, so this stays at changes requested for one small fix.

## Critical issues (1)

**1. [`file_io_unix.go:149-155`] A source read failure still makes root delete a pre-existing target that `tee` never opened.**

The `tee`-only failure is now handled correctly. But when a pre-existing file the requester cannot write meets a source stream failure in the same transfer, the Go-side `os.Remove` on line 154 deletes it as root. Reproduced against the real head `writeFileAs` with a 0444 target and a source that errors:

```
writeFileAs err = exit status 1: tee: …/precious.conf: Permission denied; connection reset
open …/precious.conf: no such file or directory
```

The same setup with a clean source leaves the file intact, which is what the new `TestWriteFileAs_TeePath_KeepsUnwritableTargetOnTeeFailure` checks. The premise in the comment on lines 150-152 is where it breaks: `erc.err != nil` means the source failed, not that `tee` opened the path. `tee` that cannot open its file keeps draining stdin into `/dev/null` and exits 1 at EOF, so the read error arrives with the target untouched.

That premise came from the previous review: the comment on line 139 said the read-error case "is one where `tee` provably created the file". It isn't, and that sentence was wrong.

Triggers are a dropped connection, `download too large` from `limitedReadCloser` mid-stream, or context cancellation. Narrower than before, same result: root deletes a file the requester has no permission to write. Fix and regression test are in the inline comment.

## Resolved from the previous review

- Silent `nil` on a violated invariant: fallback added on lines 172-174.
- PR body umask attribution: corrected.
- `firstMissingAncestor`'s redundant `ok` return: dropped.
- Direct `firstMissingAncestor` coverage: table test now covers the dangling-symlink parent and the non-`ENOENT` stat error, so the `5d01777` fix has its own guard.
- File-case regression test for the `tee`-only failure: added.
- Locale-dependent assertion: now asserts on the path.
- `fs.DirEntry`, the `/` in the subtest name, and rejecting a non-absolute path: all done, with a test for the last one.
- Root-privileged unlink redirected through a swapped symlink: fixed for the `tee`-failure path by running `rm -f` in the demoted shell. Line 154 still unlinks as root on the read-error path; the inline comment covers that too.

---

One narrow fix, about three lines plus a test. With it in, this is ready to approve.

**Checklist**:
- [x] Tests included and passing (27/27 checks green on `a8da4e5`)
- [ ] No security vulnerabilities—root-privileged removal of an untouched pre-existing target on the read-error path

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Fixed in 45598c09 and 37327108.

You were right on both counts. erc.err != nil never proved tee opened path—a tee that can't open a pre-existing file still drains stdin to EOF and exits nonzero, so this removal could reach a file the requester has no permission to write. 45598c09 reuses the same os.Lstat pre-existence check that already picks the script variant (now kept as createdTarget) and gates the removal on it, so it only ever fires for a target this call created.

37327108 also takes the optional hardening: the removal now goes through a new removeAsRequester helper that execs `rm -f path` with the caller's SysProcAttr instead of os.Remove, so it runs under the requester's own permissions, not root's—closing the symlink-redirection window on this last path too. context.WithoutCancel is there because context cancellation is one of the ways this cleanup path is reached, and the removal still has to happen after ctx is done.

Regression test is TestWriteFileAs_TeePath_KeepsUnwritableTargetOnSourceReadFailure, seen failing against the pre-45598c09 code with exactly the repro you posted (`open …/precious.conf: no such file or directory`). TestWriteFileAs_TeePath_RemovesFreshTargetOnSourceReadFailure pins the other side—a target this call did create still gets removed—and was seen failing with removeAsRequester turned into a no-op.

pkg/executor/handlers/file/file_io_unix.go
```diff
@@ -88,24 +152,35 @@ func writeFileAs(ctx context.Context, path string, src io.Reader, sysProcAttr *s
...
+		// with the target untouched. Only remove it when this call created it, and--same as the
+		// rm -f in the script above--under the requester's own permissions, not root's.
+		if erc.err != nil && createdTarget {
+			if fi, statErr := os.Lstat(path); statErr == nil && !fi.IsDir() {
+				removeAsRequester(ctx, path, sysProcAttr)
 			}
-			return runErr
 		}
+		if createdRoot != "" && dirTreeIsAllDirs(createdRoot) {
+			_ = os.RemoveAll(createdRoot)
```
geunwoonoh, line 169: Suggestion: this is now the last root-side removal in the function, and the PR body documents it well—the deferral rationale and the `mkdirAllInside`/`createFile` precedent are both there. Not a blocker: `dirTreeIsAllDirs` only lets an all-directories tree through, so the worst case here is removing empty directories, not a file, and the file unlink right above it is demoted now.

Worth opening a follow-up issue so the asymmetry with `removeAsRequester` does not sit in the code unexplained once the PR body scrolls out of view. A one-line pointer here would also do it:

```go
	// Still root-side: bounded to an all-directories tree by dirTreeIsAllDirs. See #<issue>.
	if createdRoot != "" && dirTreeIsAllDirs(createdRoot) {
```

pkg/executor/handlers/file/file_io_unix_test.go
```diff
@@ -0,0 +1,213 @@
...
+	require.NoError(t, os.WriteFile(path, []byte("ORIGINAL"), 0444))
+
+	err := writeFileAs(t.Context(), path, iotest.ErrReader(errors.New("connection reset")), &syscall.SysProcAttr{})
+	require.Error(t, err)
+	got, readErr := os.ReadFile(path)
+	require.NoError(t, readErr)
+	assert.Equal(t, "ORIGINAL", string(got))
+}
+
+func TestWriteFileAs_TeePath_RemovesFreshTargetOnSourceReadFailure(t *testing.T) {
```
geunwoonoh, line 102: Suggestion: with the `createdTarget` gate in place, a pre-existing target that `tee` did open and truncate is now left truncated on a source read failure rather than removed.

That is the right trade—reaching it requires `AllowOverwrite=true`, so a retry is not blocked, and it is what keeps a target `tee` never touched from being deleted—but no test states it. A future change could route that case back through removal and nothing would fail.

```go
func TestWriteFileAs_TeePath_LeavesTruncatedPreexistingTargetOnSourceReadFailure(t *testing.T) {
	path := filepath.Join(t.TempDir(), "out.bin")
	require.NoError(t, os.WriteFile(path, []byte("ORIGINAL"), 0644))

	err := writeFileAs(t.Context(), path, iotest.ErrReader(errors.New("boom")), &syscall.SysProcAttr{})
	require.Error(t, err)
	// tee opened and truncated it, so the old content is gone either way; what matters
	// is that a target this call did not create is not unlinked.
	got, readErr := os.ReadFile(path)
	require.NoError(t, readErr)
	assert.Empty(t, string(got))
}
```

geunwoonoh (dismissed):
## Summary

Re-review of head `4aaf20c` against the previous review on `a8da4e5`. All 27 checks green, and the one remaining critical path is closed. Approving.

The fix went past what the previous review asked for. `createdTarget` gates the removal, and `removeAsRequester` takes root out of it entirely: the unlink now runs as a demoted `rm -f` with `context.WithoutCancel`, which matters because cancellation is one of the ways into that branch. The absolute-path check earlier in the function is what makes the bare `rm -f <path>` safe from option injection. The comment that carried the wrong premise was rewritten to state what is actually true, and the PR body records why the old premise was wrong rather than quietly dropping it.

Verified independently against this head rather than from the diff:

```
--- PASS: TestProbe_UnwritableTargetSurvivesSourceReadFailure   (FAILs on a8da4e5)
--- PASS: TestProbe_PartialFreshTargetStillRemoved
```

The first is the case that failed last round: a pre-existing 0444 target plus a stream that drops mid-transfer. The second checks the opposite direction, that the gate did not narrow the cleanup into uselessness: a partial write this call created is still removed. Both probes use a reader that serves a chunk before failing, so `tee` has really written something by the time the read error lands, which the new tests' `iotest.ErrReader` does not exercise.

## Resolved from the previous review

- The read-error path no longer deletes an untouched pre-existing target, and no longer unlinks as root. Both new tests (`KeepsUnwritableTargetOnSourceReadFailure`, `RemovesFreshTargetOnSourceReadFailure`) pin the two directions.
- The false invariant in the old comment is gone, and the correction is in the PR body.

## Suggestions (2)

Neither blocks this PR.

**1. [`file_io_unix.go:169`] `os.RemoveAll(createdRoot)` is now the last root-side removal in this function.** The PR body documents it honestly, including the deferral rationale and the `mkdirAllInside`/`createFile` precedent that would close it. Impact is bounded by `dirTreeIsAllDirs`, which only lets an all-directories tree through, so the worst case is removing empty directories rather than a file. Worth a follow-up issue so the asymmetry with `removeAsRequester` does not sit here unexplained once the PR body scrolls out of view.

**2. [`file_io_unix_test.go`] The truncated-pre-existing-target outcome is intentional but untested.** With the gate in place, a pre-existing target that `tee` did open and truncate is left truncated on a source read failure instead of being removed. That is the right trade (reaching it requires `AllowOverwrite=true`, so a retry is not blocked, and nothing `tee` never touched gets deleted), but no test states it, so a future change could flip it back to removal without anything failing.

Still open outside the diff, unchanged from the previous review and fine as follow-ups: `file.go:289`'s `os.IsPermission(err) || errors.Is(err, syscall.EACCES)` cannot match on the demoted path, and the `sysProcAttr == nil` branch remains duplicated in `file_io_windows.go:26-41`.

## Highlights

- Taking root out of the file removal instead of merely gating it is the stronger fix, and `context.WithoutCancel` shows the cancellation path was actually thought through rather than copied.
- `RemovesFreshTargetOnSourceReadFailure` is the test most reviewers would not have asked for: it guards the fix against being over-applied, which is the usual way a safety gate regresses behavior.
- The PR body names a premise it previously got wrong and says so plainly. That is what made this round quick to verify.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (27/27 checks green on `4aaf20c`)
- [x] No security vulnerabilities—the root-privileged file unlink is gone; the documented residual on line 169 is bounded to empty directories

pkg/executor/handlers/file/file_io_unix.go
jisung-02: Added the pointer inline in 029c0fe7: `dirTreeIsAllDirs` bounds this to a tree of plain directories, so the worst case is deleting empty directories, not a file—that reasoning was only in this thread before. Left `os.RemoveAll(createdRoot)` as root-side rather than routing it through `removeAsRequester`; a symlink-safe walk is the actual fix and is a bigger change than this cleanup path warrants right now, as you said.

pkg/executor/handlers/file/file_io_unix_test.go
jisung-02: Added in 9b1f70e8: `TestWriteFileAs_TeePath_LeavesTruncatedPreexistingTargetOnSourceReadFailure`. It writes a pre-existing file, fails the source mid-read, and asserts the file comes back empty rather than removed. Seen RED first by dropping the `createdTarget` gate on that removal call—the file disappeared and `os.ReadFile` returned "no such file or directory" instead; restored, it passes. You were right that this direction had no guard: the gate already made it the safe behavior, but nothing pinned it.

geunwoonoh (approved):
## Summary

Re-review of head `9b1f70e` against the approval on `4aaf20c`. Both notes from that approval are addressed, the `main` merge is clean, and this stays approved.

- `029c0fe` states in the code why `os.RemoveAll(createdRoot)` remains root-side, so the asymmetry with `removeAsRequester` is explained where a reader meets it rather than only in the PR body.
- `9b1f70e` pins the truncated-target outcome: a pre-existing target `tee` did open and truncate is left in place on a source read failure. That is the passive half of the `createdTarget` gate, the half that usually has no test and so regresses quietly.
- `f0b0a65` merges `main`. `git diff 4aaf20c f0b0a65 -- pkg/executor/handlers/file/` is empty and there are no conflict markers, so it brought nothing into this PR's files.

The probes from the previous round were not re-run. Across `4aaf20c..9b1f70e` the only change to `file_io_unix.go` is one comment line, so the executable code is identical to what was already verified.

## Checks

Two checks are red for an infrastructure reason, not for the code. `Build and Test (amazonlinux-2023-arm64)` could not pull `amazonlinux:2023` from Docker Hub:

```
Error response from daemon: Head "https://registry-1.docker.io/v2/library/amazonlinux/manifests/2023":
… context deadline exceeded
##[error]Docker pull failed with exit code 1
```

All three pull attempts timed out, and the aggregate `build-and-test` gate failed behind it. The other 25 checks pass, including the remaining arm64 jobs (`rocky-9-arm64`, `ubuntu-24.04-arm64`), `golangci-lint`, and `govulncheck`. Re-running the failed jobs should clear it.

## Still open, unchanged and out of scope

Follow-up material from the earlier rounds, none of it blocking: the symlink-safe walk that would close the root-side `RemoveAll`, `file.go:289`'s `os.IsPermission` branch that cannot match on the demoted path, and the duplicated direct-write branch in `file_io_windows.go:26-41`.

## Highlights

- Testing the negative ("this file is not unlinked") is the assertion most review rounds never ask for, and it is what keeps a safety gate from being quietly widened later.
- Merging `main` rather than force-pushing keeps the review history on this PR readable across four rounds.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (25/27; the two failures are a Docker Hub pull timeout)

PR #469, opened 2026-09-11
fix(utils): use the default file mode when a zip entry carries none

### LINKED ISSUE #440: fix(utils): extractFile treats a zero entry mode as 0000, unlike applyDirModes
## Summary

`extractFile` passes a zip entry's mode straight to `createFile`, so an entry written by a Unix creator with a zero external attribute extracts as a `0000` file. `applyDirModes` guards exactly this case for directory entries, so the two paths now disagree about what a zero mode means.

## Where

`pkg/utils/archive.go`, `extractFile`:

```go
outFile, err := createFile(fpath, f.Mode().Perm())
```

The directory path, a few lines up in `applyDirModes`, reads:

```go
// Perm only, for the reason extractFile gives. Zero is an archive that
// carried no mode at all, not a directory nobody may enter.
mode := dir.file.Mode().Perm()
if mode == 0 {
    continue
}
```

## Reproduce

Hand-build an archive whose file entry declares `CreatorVersion` of Unix and an external attribute of 0, then extract it:

```
P3 unzip=<nil>  resulting file mode=----------
```

The extraction reports no error and leaves a file nobody can read.

## Why it matters

The same reasoning the directory guard states applies to files: a zero mode is an archive that carried no mode at all, not a file nobody may open. A user who uploads such an archive gets their own data back unreadable, with no error to explain it.

## Not a regression

`extractFile` on `main` already passed `f.Mode()` through unchanged. This PR only made the inconsistency visible by adding the guard on the directory half. Reported as an observation on #435 and deliberately left out of that PR's scope.

## Suggested fix

Give `extractFile` the same guard the directory path has: fall back to a sane default (the mode `createFile` would otherwise be given for an entry with no mode) when `f.Mode().Perm()` is zero and the creator is Unix, and cover it with a test alongside `TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault`.

Reported by @geunwoonoh in https://github.com/alpacax/alpamon/pull/435#discussion_r3878114375

jisung-02:
## Background
Directory entries already fell back to a default mode when none was carried, but file entries had no equivalent handling. A Unix or macOS zip entry with a zero mode in its external attributes has no mode set at all, and `extractFile` passed that 0 straight through as the file's permissions, producing a file nobody could read or write.

## Changes
In `extractFile`, fall back to 0666 whenever the decoded mode is 0—the same default `archive/zip` gives files with no Unix mode. This value still goes through the process's umask.

The initial fix gated the fallback on the creator being `zipCreatorUnix` or `zipCreatorMacOSX`, which left the same bug in place for any creator `archive/zip` does not decode a mode from (OpenVMS and others): those entries carry a zero mode too and were never covered by the gate. FAT, NTFS and VFAT creators never need a gate, since their branch of `FileHeader.Mode()` never returns a zero permission for a file—checking `mode == 0` alone already excludes them. The gate is gone, and the `zipCreatorUnix`/`zipCreatorMacOSX` doc comment is corrected: only FAT, NTFS and VFAT creators derive a mode from FAT attributes; every other creator gets none, which is what made the gate look safe.

## Testing
`TestUnzip_FileEntryWithoutModeKeepsTheDefault` now covers the Unix, MacOSX and OpenVMS creators. Against the pre-review code the OpenVMS case fails with `expected: 0x1a4, actual: 0x0`. The reference file is now built with `os.WriteFile` plus `os.Stat`, matching the `os.Mkdir` plus `os.Stat` reference `TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault` already uses.

Confirmed `go test -race -v -shuffle=on ./pkg/utils/... -run TestUnzip` passes.

Closes #440

### DIFF
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -30,8 +30,8 @@ const (
 
 	// zipCreatorUnix and zipCreatorMacOSX are the creator halves of a zip
 	// header's CreatorVersion under which archive/zip reads a Unix mode out of
-	// ExternalAttrs; SetMode writes the first. Every other creator gets a mode
-	// derived from FAT attributes instead.
+	// ExternalAttrs; SetMode writes the first. FAT, NTFS and VFAT creators get a
+	// mode derived from FAT attributes instead; every other creator gets none.
 	zipCreatorUnix   = 3
 	zipCreatorMacOSX = 19
 
@@ -756,9 +756,10 @@ func extractFile(f *zip.File, fpath string) error {
 	// or sticky off an entry would land on a root-owned file. newZipEntry drops
 	// the same bits on the way in.
 	mode := f.Mode().Perm()
-	if creator := f.CreatorVersion >> 8; mode == 0 && (creator == zipCreatorUnix || creator == zipCreatorMacOSX) {
-		// As with directories, zero means no mode was carried. Use the same
-		// default archive/zip gives files without Unix modes; umask still applies.
+	if mode == 0 {
+		// Zero is an entry that carried no mode: a Unix header left empty, or
+		// a creator archive/zip does not decode. Use the 0666 it gives a FAT
+		// entry; umask still applies.
 		mode = 0666
 	}
 	outFile, err := createFile(fpath, mode)
--- a/pkg/utils/archive.go
+++ b/pkg/utils/archive.go
@@ -755,7 +755,13 @@ func extractFile(f *zip.File, fpath string) error {
 	// Perm only: extraction runs as root on a normal install, so setuid, setgid
 	// or sticky off an entry would land on a root-owned file. newZipEntry drops
 	// the same bits on the way in.
-	outFile, err := createFile(fpath, f.Mode().Perm())
+	mode := f.Mode().Perm()
+	if creator := f.CreatorVersion >> 8; mode == 0 && (creator == zipCreatorUnix || creator == zipCreatorMacOSX) {
+		// As with directories, zero means no mode was carried. Use the same
+		// default archive/zip gives files without Unix modes; umask still applies.
+		mode = 0666
+	}
+	outFile, err := createFile(fpath, mode)
 	if err != nil {
 		return err
 	}
--- a/pkg/utils/archive_test.go
+++ b/pkg/utils/archive_test.go
@@ -1171,11 +1171,13 @@ func TestUnzip_FileEntryWithoutModeKeepsTheDefault(t *testing.T) {
 	}{
 		{"Unix", zipCreatorUnix},
 		{"MacOSX", zipCreatorMacOSX},
+		{"OpenVMS", 2},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			var buf bytes.Buffer
 			w := zip.NewWriter(&buf)
-			// No SetMode: preserve zero external attributes from a Unix creator.
+			// No SetMode: preserve the zero external attributes archive/zip
+			// leaves for a creator it does not decode a mode from.
 			zw, err := w.CreateHeader(&zip.FileHeader{
 				Name:           "file.txt",
 				CreatorVersion: tc.creator << 8,
--- a/pkg/utils/archive_test.go
+++ b/pkg/utils/archive_test.go
@@ -1176,8 +1176,8 @@ func TestUnzip_FileEntryWithoutModeKeepsTheDefault(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			var buf bytes.Buffer
 			w := zip.NewWriter(&buf)
-			// No SetMode: preserve the zero external attributes archive/zip
-			// leaves for a creator it does not decode a mode from.
+			// No SetMode: zero external attributes read back as mode 0 whether
+			// archive/zip reads a Unix mode for the creator or decodes nothing.
 			zw, err := w.CreateHeader(&zip.FileHeader{
 				Name:           "file.txt",
 				CreatorVersion: tc.creator << 8,
--- a/pkg/utils/archive_test.go
+++ b/pkg/utils/archive_test.go
@@ -1194,12 +1194,11 @@ func TestUnzip_FileEntryWithoutModeKeepsTheDefault(t *testing.T) {
 			out := filepath.Join(dir, "out")
 			require.NoError(t, UnzipReader(r, out))
 
-			// os.Create uses 0666, subject to the process's umask, just like
-			// extraction of a file from an archive with no Unix mode.
-			ref, err := os.Create(filepath.Join(dir, "ref"))
-			require.NoError(t, err)
-			want, err := ref.Stat()
-			require.NoError(t, ref.Close())
+			// A file written at 0666 goes through the same umask as extraction,
+			// so it is the reference, not a literal 0644.
+			ref := filepath.Join(dir, "ref")
+			require.NoError(t, os.WriteFile(ref, nil, 0666))
+			want, err := os.Stat(ref)
 			require.NoError(t, err)
 			path := filepath.Join(out, "file.txt")
 			fi, err := os.Stat(path)
--- a/pkg/utils/archive_test.go
+++ b/pkg/utils/archive_test.go
@@ -1160,6 +1160,56 @@ func TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault(t *testing.T) {
 	assert.Equal(t, "hi", string(content))
 }
 
+func TestUnzip_FileEntryWithoutModeKeepsTheDefault(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Unix-specific behavior")
+	}
+
+	for _, tc := range []struct {
+		name    string
+		creator uint16
+	}{
+		{"Unix", zipCreatorUnix},
+		{"MacOSX", zipCreatorMacOSX},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			w := zip.NewWriter(&buf)
+			// No SetMode: preserve zero external attributes from a Unix creator.
+			zw, err := w.CreateHeader(&zip.FileHeader{
+				Name:           "file.txt",
+				CreatorVersion: tc.creator << 8,
+			})
+			require.NoError(t, err)
+			_, err = zw.Write([]byte("hi"))
+			require.NoError(t, err)
+			require.NoError(t, w.Close())
+			r, err := zip.NewReader(bytes.NewReader(buf.Bytes()), int64(buf.Len()))
+			require.NoError(t, err)
+			require.Zero(t, r.File[0].Mode().Perm())
+
+			dir := t.TempDir()
+			out := filepath.Join(dir, "out")
+			require.NoError(t, UnzipReader(r, out))
+
+			// os.Create uses 0666, subject to the process's umask, just like
+			// extraction of a file from an archive with no Unix mode.
+			ref, err := os.Create(filepath.Join(dir, "ref"))
+			require.NoError(t, err)
+			want, err := ref.Stat()
+			require.NoError(t, ref.Close())
+			require.NoError(t, err)
+			path := filepath.Join(out, "file.txt")
+			fi, err := os.Stat(path)
+			require.NoError(t, err)
+			assert.Equal(t, want.Mode().Perm(), fi.Mode().Perm())
+			content, err := os.ReadFile(path)
+			require.NoError(t, err)
+			assert.Equal(t, "hi", string(content))
+		})
+	}
+}
+
 func TestUnzipReader_ExtractsAnInMemoryArchive(t *testing.T) {
 	var buf bytes.Buffer
 	w := zip.NewWriter(&buf)


geunwoonoh (commented):
## PR Review: #469 - fix(utils): use the default file mode when a zip entry carries none

**Author**: @jisung-02
**Branch**: 440-extractfile-zero-mode-fallback → main
**Stack**: golang
**Changes**: +57 -1 in 2 files

---

### Summary
Closes #440. `extractFile` handed `f.Mode().Perm()` straight to `createFile`, so a Unix- or macOS-created entry with empty external attributes extracted as a `0000` file. This PR falls back to `0666` (umask still applies), mirroring the zero-mode guard in `applyDirModes`, and adds `TestUnzip_FileEntryWithoutModeKeepsTheDefault` for both creators. The fix is correct for the reported case. The creator gate leaves the same bug in place for other creator bytes (W1).

### Critical issues (0)
None.

### Warnings (1)

1. **[pkg/utils/archive.go:759]** The creator gate still extracts `0000` files for every creator `archive/zip` does not decode. It also leaves those files disagreeing with directories.

   `FileHeader.Mode()` sets permission bits for only two groups:
   - Unix and macOS (3, 19), read from the high 16 bits of `ExternalAttrs`.
   - FAT, NTFS and VFAT (0, 11, 14), through `msdosModeToFileMode`.

   Any other creator byte falls through the switch with no permission bits. `Perm()` is then 0: the same bug, reached through a different header field. Probe at ae9eb4b through `go test -overlay` (one file entry, zero external attributes, umask 022):

   | Creator | `Mode().Perm()` | base | this PR | gate removed |
   |---|---|---|---|---|
   | 0 FAT, 11 NTFS, 14 VFAT | 0666 | `-rw-r--r--` | `-rw-r--r--` | `-rw-r--r--` |
   | 3 Unix, 19 macOS | 0000 | `----------` | `-rw-r--r--` | `-rw-r--r--` |
   | 1, 2, 6, 7, 10, 16, 18 | 0000 | `----------` | `----------` | `-rw-r--r--` |

   Creator 10 is Windows NTFS in APPNOTE (Go follows Info-ZIP's 11), so a writer that follows the spec lands in the last row. `applyDirModes` gives a directory from any of those creators the default mode. For those creators, a directory and the file inside it still disagree, which is the inconsistency #440 was filed to remove.

   The gate does no work. `msdosModeToFileMode` returns 0666 or 0444 for a file, never 0, so `mode == 0` alone already leaves the FAT family out. With the gate removed, the full `pkg/utils` suite passes under `-race -shuffle=on`. Removing it also drops the second copy of the Unix/macOS check. The suggested change is in the inline comment.

   Two related changes:
   - Add a case such as `{"OpenVMS", 2}` to the test table, and reword the "from a Unix creator" comment at `archive_test.go:1178`. On the current code that case fails with `expected: 0x1a4, actual: 0x0`; with the suggested change it passes.
   - Correct the constants comment at `archive.go:33-34`. "Every other creator gets a mode derived from FAT attributes instead" holds only for FAT, NTFS and VFAT; every other creator gets no permission bits. This wrong premise is what made the gate look safe.

### Suggestions (1)

1. **[pkg/utils/archive_test.go:1195-1201]** Building the reference file takes five lines, and `Stat`'s error is checked only after `Close`. `os.WriteFile` plus `os.Stat` needs four lines and matches the `os.Mkdir` plus `os.Stat` reference in the directory test. The suggested change is in the inline comment.

### Conventions, dead code, and performance

- **Go conventions:** `gofmt -l` and `go vet` are clean, and golangci-lint passes in CI. The `if creator := …; cond` form, the `0666` octal style, and table-driven subtests (`require` for setup, `assert` for checks) all match the file and CLAUDE.md.
- **Duplicated logic:** the PR adds a second copy of the Unix/macOS creator check (`archive.go:496` in `applyDirModes`, `archive.go:759`). W1 removes it. If the gate stays, a helper such as `carriesUnixMode(h *zip.FileHeader) bool` keeps the two copies from drifting apart.
- **Unused code:** nothing is added unused or left orphaned. `zipCreatorUnix` and `zipCreatorMacOSX` are still used by `applyDirModes` and the tests.
- **Performance:** no change. The PR adds one integer comparison per file entry, still reads `f.Mode()` once, and adds no allocation or syscall. Decompression, `io.Copy`, and the per-entry `Lstat` and `OpenFile` dominate this path; nothing in the diff is worth optimizing.
- **Test duplication:** the new test does not duplicate `TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault`; it covers a different entry kind against a different reference. Building the archive inline is required, because `writeEntryZip` always calls `SetMode`.

### Considered, not flagged

- A file set to mode `000` on purpose now extracts as `0666 &^ umask`. `SetMode(0)` writes `ExternalAttrs>>16 = 0x8000`; base extracts `----------`, this PR `-rw-r--r--`. `Perm()` can't tell that from an empty header, though `ExternalAttrs>>16 == 0` could. Not flagged:
  - `applyDirModes` and its test fixture already treat zero permission bits as "no mode".
  - #440 asked for the same meaning on files.
  - Splitting the two cases would mean changing the file and directory paths together.

### Highlights

- `require.Zero(t, r.File[0].Mode().Perm())` pins the test's precondition, so a change in `archive/zip` can't make it pass for the wrong reason.
- The expected mode comes from a reference file created under the same umask, not a literal `0644`, the same technique the directory test uses.
- It is a real regression test: against the base `archive.go` it fails with `expected: 0x1a4, actual: 0x0`.
- Passing an in-memory reader to `UnzipReader` means the test needs no temporary archive file.

---

**Verdict**: COMMENTED

**Checklist**:
- [x] Tests included and passing (CI green; `windows-amd64` still pending, and the test skips on Windows)
- [ ] Documentation updated if needed: constants comment at `archive.go:33-34` (W1)

pkg/utils/archive.go
```diff
@@ -755,7 +755,13 @@ func extractFile(f *zip.File, fpath string) error {
 	// Perm only: extraction runs as root on a normal install, so setuid, setgid
 	// or sticky off an entry would land on a root-owned file. newZipEntry drops
 	// the same bits on the way in.
-	outFile, err := createFile(fpath, f.Mode().Perm())
+	mode := f.Mode().Perm()
+	if creator := f.CreatorVersion >> 8; mode == 0 && (creator == zipCreatorUnix || creator == zipCreatorMacOSX) {
+		// As with directories, zero means no mode was carried. Use the same
+		// default archive/zip gives files without Unix modes; umask still applies.
+		mode = 0666
+	}
```
geunwoonoh, line 764: Warning: The creator gate still extracts `0000` files for every creator `archive/zip` does not decode (1, 2, 6, 7, 10, 16, 18, ...). For those creators a file also disagrees with its directory, which `applyDirModes` gives the default mode. The gate protects nothing: `msdosModeToFileMode` never returns 0 for a file, so `mode == 0` alone already leaves the FAT family out. With the gate removed, the full `pkg/utils` suite passes under `-race -shuffle=on`. The probe table is in the review summary.

```suggestion
	mode := f.Mode().Perm()
	if mode == 0 {
		// Zero is an entry that carried no mode: a Unix header left empty, or
		// a creator archive/zip does not decode. Use the 0666 it gives a FAT
		// entry; umask still applies.
		mode = 0666
	}
```

pkg/utils/archive_test.go
```diff
@@ -1160,6 +1160,56 @@ func TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault(t *testing.T) {
...
+			out := filepath.Join(dir, "out")
+			require.NoError(t, UnzipReader(r, out))
+
+			// os.Create uses 0666, subject to the process's umask, just like
+			// extraction of a file from an archive with no Unix mode.
+			ref, err := os.Create(filepath.Join(dir, "ref"))
+			require.NoError(t, err)
+			want, err := ref.Stat()
+			require.NoError(t, ref.Close())
+			require.NoError(t, err)
```
geunwoonoh: Suggestion: `os.WriteFile` plus `os.Stat` builds the reference in four lines instead of five, avoids checking `Stat`'s error after `Close`, and matches the `os.Mkdir` plus `os.Stat` reference in the directory test.

```suggestion
			// A file written at 0666 goes through the same umask as extraction,
			// so it is the reference, not a literal 0644.
			ref := filepath.Join(dir, "ref")
			require.NoError(t, os.WriteFile(ref, nil, 0666))
			want, err := os.Stat(ref)
			require.NoError(t, err)
```

pkg/utils/archive.go
jisung-02: Fixed in cc9b1d5. You're right that the gate did no work: `msdosModeToFileMode` never returns 0 for a file, so `mode == 0` alone already excludes the FAT family, and the gate only left OpenVMS and the other undecoded creators stuck at 0000. Dropped the creator check as suggested, corrected the doc comment on `zipCreatorUnix`/`zipCreatorMacOSX` that made the gate look safe, and added an OpenVMS case to the test—against the pre-fix code it fails with `expected: 0x1a4, actual: 0x0`.

pkg/utils/archive_test.go
jisung-02: Fixed in ecf8a7b. Switched to `os.WriteFile` plus `os.Stat`, matching the `os.Mkdir`/`os.Stat` reference the directory test already uses.

pkg/utils/archive_test.go
```diff
@@ -1160,6 +1160,57 @@ func TestUnzip_DirectoryEntryWithoutModeKeepsTheDefault(t *testing.T) {
...
+	}{
+		{"Unix", zipCreatorUnix},
+		{"MacOSX", zipCreatorMacOSX},
+		{"OpenVMS", 2},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			var buf bytes.Buffer
+			w := zip.NewWriter(&buf)
+			// No SetMode: preserve the zero external attributes archive/zip
+			// leaves for a creator it does not decode a mode from.
```
geunwoonoh: Suggestion (non-blocking): This now describes only the OpenVMS row. For Unix and macOS, `archive/zip` does read a Unix mode, and the mode it reads is zero.

```suggestion
			// No SetMode: zero external attributes read back as mode 0 whether
			// archive/zip reads a Unix mode for the creator or decodes nothing.
```

geunwoonoh (dismissed):
## PR Review: #469 - fix(utils): use the default file mode when a zip entry carries none

**Author**: @jisung-02
**Branch**: 440-extractfile-zero-mode-fallback → main
**Stack**: golang
**Changes**: +61 -3 in 2 files

---

### Summary
Re-reviewed at ecf8a7b. Both items from the previous review are resolved:

- **W1 (cc9b1d5):** the creator gate is gone, so `mode == 0` alone selects the fallback, and files now agree with directories for every creator. The constants comment matches `archive/zip`. The new OpenVMS case fails against ae9eb4b's `extractFile` (`expected: 0x1a4, actual: 0x0`) and passes here.
- **Suggestion (ecf8a7b):** the reference file is built with `os.WriteFile` and `os.Stat`, matching the directory test.

At ecf8a7b the full `pkg/utils` suite passes under `-race -shuffle=on`, and `gofmt` and `go vet` are clean.

### Critical issues (0)
None.

### Warnings (0)
None.

### Suggestions (2, non-blocking)

1. **[pkg/utils/archive_test.go:1179-1180]** The reworded comment now describes only the OpenVMS row. For Unix and macOS, `archive/zip` does read a Unix mode; the mode it reads is zero. Inline suggestion attached.
2. **[pkg/utils/archive.go:493-495]** Outside the diff: the `applyDirModes` comment still says `archive/zip` "derives 0666 from any other creator's attributes," which now contradicts the corrected constants comment. Skipping every non-Unix creator is still correct, so only the wording is off. Fine to fix here or in a follow-up.

### Highlights
- The fallback is one condition that covers every entry `archive/zip` can't supply a mode for.
- Each follow-up commit does one thing, and the PR description was updated to match the new design.
- The OpenVMS case was checked against the pre-fix code, so it guards the regression rather than only adding coverage.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (CI: all 27 checks pass)

pkg/utils/archive_test.go
jisung-02: Fixed in c2fd767. Reworded to state the shared fact—zero external attributes read back as mode 0 whether the creator gets a Unix mode decoded (Unix, macOS) or none at all (OpenVMS)—rather than implying no creator here gets decoded.

geunwoonoh (approved):
## PR Review: #469 - fix(utils): use the default file mode when a zip entry carries none

**Author**: @jisung-02
**Branch**: 440-extractfile-zero-mode-fallback → main
**Stack**: golang
**Changes**: +61 -3 in 2 files

---

### Summary
Re-reviewed at c2fd767. The comment-wording suggestion from the previous review was taken as written, and `pkg/utils/archive.go` is byte-identical to ecf8a7b, which I already approved. The comment now states the fact both rows share: zero external attributes read back as mode 0 whether `archive/zip` decodes a Unix mode for the creator or decodes nothing.

At c2fd767 the full `pkg/utils` suite passes under `-race -shuffle=on`, and `gofmt` and `go vet` are clean.

### Critical issues (0)
None.

### Warnings (0)
None.

### Suggestions (0)
None. The `applyDirModes` comment at `archive.go:493-495` stays a follow-up, as agreed; a one-line issue would keep it from getting lost.

### Highlights
- Three rounds of review, each answered with a single-purpose commit whose message says what changed and why.
- The fallback is one condition covering every entry `archive/zip` cannot supply a mode for, with the regression pinned by a test that fails against the pre-fix code.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (CI: 24 passed, ubuntu-20.04 and ubuntu-24.04 still running at review time; the only change since ecf8a7b, which passed every job, is a test comment)

PR #471, opened 2026-09-11
fix(scheduler): clear the backing-array slot on priority queue pop

### LINKED ISSUE #470: fix: priorityQueue.Pop retains processed request payloads in the backing array
priorityHeap.Pop reduces the slice length without clearing the removed PriorityEntry:
```go
  item := old[n-1]
  *h = old[:n-1]
  return item
```
  The backing array still retains references to the processed request’s data, headers, URL, and other
  fields. As a result, the garbage collector cannot reclaim payloads after requests are dequeued.

  In a local reproduction with 8,192 queued requests containing 4 KiB payloads, approximately 33 MiB
  remained retained after the queue was drained. Clearing the removed slot reduced retained heap usage
  to approximately 1 MiB.

  Suggested fix:
```go
  item := old[n-1]
  old[n-1] = PriorityEntry{}
  *h = old[:n-1]
  return item
```
  Add a regression test to verify that the removed backing-array slot no longer contains the processed
  entry.

jisung-02:
## Background

`priorityHeap.Pop` only reslices the backing array after removing the min-priority entry (`pkg/scheduler/queue.go:48-56`). The removed `PriorityEntry`, including its `data` payload and `headers` pointer, stays reachable through the backing array until a later `Push` overwrites that slot. Under sustained load this lets processed request bodies linger in memory far longer than needed.

## Changes

- `priorityHeap.Pop` now zeroes the vacated backing-array slot (`old[n-1] = PriorityEntry{}`) right after taking the item, so the GC can reclaim the payload immediately instead of waiting for the slot to be reused.

## Testing

- `go test ./pkg/scheduler/... -v -p 1` — all 5 tests pass.
- `TestPriorityQueue_GetReleasesOnlyRemovedEntry` offers three entries, pops the min-priority one, and asserts the vacated backing-array slot is zeroed while the two still-queued entries keep their payload. It then drains the rest and asserts `/b` then `/c` come out in priority order, so a damaged heap invariant (for example in `Less`'s `due` tiebreak) fails here too, not just a slot zeroed at the wrong index. Verified RED twice: once by reverting the fix (the zero-value assertion fails), and once by flipping the priority comparison in `Less` (the drain order assertion fails, expecting `/c` and getting `/a`).
- `drainOne`, `TestPostChunk_EnqueuesBelowHighWater`, and `TestPostChunk_DropsOnContextCancel` now use `require.NoError`/`assert.Equal` instead of bare `t.Fatalf`/`t.Errorf`, per this repo's testing convention of converting a neighbouring assertion while already editing the file. The bare `t.Fatal` in `TestPostChunk_BlocksUntilSpaceFrees`'s `select` timeout branch is left as is, since there is no error value to assert on there.

Closes #470

### DIFF
--- a/pkg/scheduler/queue.go
+++ b/pkg/scheduler/queue.go
@@ -49,6 +49,8 @@ func (h *priorityHeap) Pop() any {
 	old := *h
 	n := len(old)
 	item := old[n-1]
+	// Clear the backing-array slot so processed request data can be reclaimed.
+	old[n-1] = PriorityEntry{}
 	*h = old[:n-1]
 	return item
 }
--- a/pkg/scheduler/queue_test.go
+++ b/pkg/scheduler/queue_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func queueSize() int {
@@ -27,6 +28,31 @@ func drainOne(t *testing.T) {
 func fill(n int) {
 	for range n {
 		Rqueue.Post("/other", nil, 10, time.Time{})
+	}
+}
+
+func TestPriorityQueue_GetReleasesOnlyRemovedEntry(t *testing.T) {
+	queue := newPriorityQueue(3)
+	headers := Headers{"X-Test": "value"}
+	entries := []PriorityEntry{
+		{priority: 3, url: "/c", data: make([]byte, 4096), headers: &headers},
+		{priority: 1, url: "/a", data: make([]byte, 4096), headers: &headers},
+		{priority: 2, url: "/b", data: make([]byte, 4096), headers: &headers},
+	}
+	for _, e := range entries {
+		require.NoError(t, queue.Offer(e))
+	}
+
+	vacated := len(queue.h) - 1
+	got, err := queue.Get()
+	require.NoError(t, err)
+	assert.Equal(t, "/a", got.url)
+
+	assert.Len(t, queue.h, 2)
+	assert.Equal(t, PriorityEntry{}, queue.h[:cap(queue.h)][vacated])
+	for _, remaining := range queue.h {
+		assert.NotEqual(t, PriorityEntry{}, remaining)
+		assert.Len(t, remaining.data, 4096)
 	}
 }
 


pkg/scheduler/queue_test.go
```diff
@@ -30,6 +31,31 @@ func fill(n int) {
...
+
+	vacated := len(queue.h) - 1
+	got, err := queue.Get()
+	require.NoError(t, err)
+	assert.Equal(t, "/a", got.url)
+
+	assert.Len(t, queue.h, 2)
+	assert.Equal(t, PriorityEntry{}, queue.h[:cap(queue.h)][vacated])
+	for _, remaining := range queue.h {
+		assert.NotEqual(t, PriorityEntry{}, remaining)
```
geunwoonoh: Suggestion: this assertion is subsumed by the `assert.Len` on the next line. A zero-valued `PriorityEntry` has `data == nil`, so `assert.Len(t, remaining.data, 4096)` already fails on a zeroed entry, and strictly more besides. This line can never fail while the `Len` passes.

```suggestion
```

pkg/scheduler/queue_test.go
```diff
@@ -30,6 +31,31 @@ func fill(n int) {
...
+		assert.Len(t, remaining.data, 4096)
+	}
+}
```
geunwoonoh, line 63: Suggestion: consider draining the rest and asserting the order.

`Pop` is where the heap invariant is maintained, and no test in this package asserts the ordering contract—`Less`'s `due` tiebreak is entirely uncovered. The current test pops once and checks `/a`, which catches a slot zeroed at the wrong index but not a damaged invariant. Two extra lines buy the stronger guard.

```suggestion

	// Pop maintains the heap invariant; assert the rest still drains in priority order.
	for _, want := range []string{"/b", "/c"} {
		next, err := queue.Get()
		require.NoError(t, err)
		assert.Equal(t, want, next.url)
	}
}
```

pkg/scheduler/queue.go
```diff
@@ -49,6 +49,8 @@ func (h *priorityHeap) Pop() any {
 	old := *h
 	n := len(old)
 	item := old[n-1]
+	// Clear the backing-array slot so processed request data can be reclaimed.
+	old[n-1] = PriorityEntry{}
```
geunwoonoh, line 53: Informational, no action needed.

The ordering here is what makes this correct: the copy is taken before the zeroing, matching the stdlib `container/heap` idiom (`old[n-1] = nil // avoid memory leak`), with the zero value standing in for `nil` on a value element. Reversing the two lines would return a zero-valued entry and break every dequeue.

One residual worth recording: the backing array's capacity is never released, so the array shell survives a full drain—120 bytes x `MaxQueueSize` (36000) is roughly 4.12 MB worst case. That is exactly the ~1 MiB residual measured in #470 at 8,192 entries, so it is already known and accepted. The payloads were the dominant term and this PR removes them; shrinking the array would trade reallocation churn for a much smaller win.

geunwoonoh (dismissed):
## Summary

`priorityHeap.Pop` took the min-priority entry and resliced, leaving the removed `PriorityEntry` reachable through the backing array until a later `Push` reused the slot. The fix assigns `PriorityEntry{}` to the vacated slot after copying the item out, releasing the `data` payload and `headers` pointer at pop time. This is the exact fix and regression test requested in #470.

## Verification performed

- Reverted the fix and re-ran the new test: it fails on the zero-value assertion, and the failure output still shows the 4096-byte payload and the `headers` pointer retained in the slot. RED confirmed.
- Searched the repo for the same pattern. This is the only `heap.Interface` implementation; the other slice-shrink sites (`pkg/utils/archive_worker.go`, `pkg/utils/wirepath.go`, `pkg/utils/archive.go`) all slice strings, which carry no pointer retention. The scope of the fix is correct.
- `gofmt` and `go vet` clean. All 5 tests in `pkg/scheduler` pass.
- Measured `unsafe.Sizeof(PriorityEntry{})` = 120 bytes, which makes the ~1 MiB residual reported in #470 exactly the backing-array shell at 8,192 entries.

## Critical issues (0)

None.

## Warnings (0)

None.

## Suggestions (4)

1. **[pkg/scheduler/queue_test.go:54]** `assert.NotEqual(t, PriorityEntry{}, remaining)` is subsumed by the line below it. A zero-valued `PriorityEntry` has `data == nil`, so `assert.Len(t, remaining.data, 4096)` already fails on a zeroed entry, and strictly more besides. The `NotEqual` can never fail while the `Len` passes.

2. **[pkg/scheduler/queue_test.go:34-57]** Consider draining the rest and asserting `/b` then `/c`. The fix edits `Pop`, which is where the heap invariant is maintained, and no test in this package asserts the ordering contract—`Less`'s `due` tiebreak is entirely uncovered. The current test pops once and checks `/a`, which catches a slot zeroed at the wrong index but not a damaged invariant. Two extra lines buy the stronger guard.

3. **[pkg/scheduler/queue_test.go:24,65,122]** The repo's CLAUDE.md testing section asks for converting a neighbouring `t.Errorf`/`t.Fatalf` "while you are already editing it rather than as a separate sweep." This PR edits the file without converting any. Three convert cleanly: `drainOne`'s `t.Fatalf("drain: %v", err)` to `require.NoError`, and the two size comparisons at lines 65 and 122 to `assert.Equal`. Note that lines 84 and 94 should stay as-is—the same convention keeps a bare `t.Fatal` in a `select` timeout branch, where there is no error value to assert on.

4. **[pkg/scheduler/queue.go:53]** Informational, no action needed. The backing array's capacity is never released, so the array shell survives a full drain: 120 bytes x `MaxQueueSize` (36000) is roughly 4.12 MB worst case. This is exactly the ~1 MiB residual measured in #470 at 8,192 entries, so it is already known and accepted. The payloads were the dominant term and this PR removes them; shrinking the array would trade reallocation churn for a much smaller win.

## Highlights

- Matches the documented `container/heap` idiom precisely. The stdlib example writes `old[n-1] = nil // avoid memory leak` for pointer elements; value elements correctly take the zero value instead.
- The copy is taken before the zeroing. Reversing those two lines would return a zero-valued entry and break every dequeue—the ordering is right.
- The comment carries the *why* (GC reclamation), which the bare assignment does not convey. It earns its line rather than restating the code.
- `require` for the preconditions (`Offer` and `Get` must succeed or the rest is meaningless) and `assert` for the actual checks matches the repo's testify convention exactly.
- The test builds a local `newPriorityQueue(3)` instead of mutating the package-global `Rqueue` the way the neighbouring `TestPostChunk_*` tests do, so it cannot leak state into them.
- The index expression `queue.h[:cap(queue.h)][vacated]` is safe in every case, not just today's: `cap` is at least the pre-pop length and reslicing never reduces it, so `vacated` is always in bounds. Confirmed the growth path empirically (cap 0 to 1 to 2 to 4).
- RED verification is documented in the PR body and reproduces independently.

## Security

No concerns. No credential handling, injection surface, or logging changes. Zeroing processed payloads marginally reduces how long request bodies sit readable in process memory—a small positive.

## Performance

One 120-byte zero-store per pop, negligible against the HTTP round trip that follows each dequeue. The retention win is large: #470 measured 33 MiB down to ~1 MiB across 8,192 entries with 4 KiB payloads.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Code follows stack conventions (`gofmt` and `go vet` clean; `golangci-lint` not available locally, left to CI)
- [x] Tests included and passing (5/5 in `pkg/scheduler`, RED verified by reverting the fix)
- [x] Documentation updated if needed (no user-facing surface changed)

pkg/scheduler/queue_test.go
jisung-02: Removed. You're right that this line can never fail while `assert.Len(t, remaining.data, 4096)` below it stands—a zeroed entry always has `data == nil`, so the `Len` check already covers it and more. Fixed in 5e61cd8e.

pkg/scheduler/queue_test.go
jisung-02: Applied as suggested. The test now drains the remaining two entries and asserts `/b` then `/c` come out in order. Verified this actually guards the invariant by flipping the priority comparison in `Less`: the test failed expecting `/c` and got `/a`, confirming it wasn't just checking the slot index. Fixed in 53df0e85.

pkg/scheduler/queue.go
jisung-02: Agreed, no action needed here—the copy-before-zero ordering is correct and the array-shell residual is an accepted tradeoff. Thanks for spelling out the math on the ~4.12 MB worst case; leaving this as is.

geunwoonoh (approved):
## Summary

Re-review of `b74361a4`. All three actionable suggestions from the previous review are applied correctly, and `queue.go` is unchanged from the version already reviewed. Nothing new to flag.

## Suggestion follow-up

| # | Suggestion | Commit | Result |
|---|---|---|---|
| 1 | Drop the subsumed assertion | `5e61cd8e` | `assert.NotEqual` removed; the `assert.Len` stands alone |
| 2 | Assert the full drain order | `53df0e85` | `/b` then `/c` asserted after the retention checks |
| 3 | Convert bare `t.Fatalf`/`t.Errorf` | `b74361a4` | Three converted; the two documented exceptions preserved |
| 4 | Backing-array residual | — | Informational, acknowledged, no change needed |

Suggestion 3 went further than asked. The CLAUDE.md convention invites converting one `t.Errorf`/`t.Fatalf` while already editing the file; all three convertible sites were handled, and the two bare `t.Fatal` calls in `select` timeout branches (lines 88 and 98) were correctly left alone, which is the exception the same convention carves out.

## Verification performed

Two mutation tests confirm the new assertions carry weight rather than decorate:

- Reverting the fix entirely still fails at line 51, the retention assertion. RED holds.
- Zeroing the wrong slot (`old[0]` instead of `old[n-1]`) fails at lines 51, 53 **and 58**. Line 58 is one of the drain-order assertions added by suggestion 2, so that suggestion is now guarding the heap invariant that `Pop` maintains.

Also checked:

- All 5 tests in `pkg/scheduler` pass; `gofmt` clean; `go vet ./pkg/scheduler/...` exits 0.
- All 27 CI checks green, covering 17 distribution builds.

## Critical issues (0)

None.

## Warnings (0)

None.

## Suggestions (0)

None. The two drain blocks repeat three near-identical lines each and a short loop would collapse them, but that is style-equivalent to the surrounding code and not worth a round trip.

## Highlights

- The ordering assertions were placed after the retention assertions, so the zeroed-slot and intact-payload checks still run against a partially drained queue rather than an empty one. The sequencing is deliberate and correct.
- Converting `drainOne` to `require.NoError` preserves the abort semantics the previous `t.Fatalf` had, including inside the `synctest` bubble that calls it.
- `assert.Equal(t, 1, queueSize(), ...)` compares `int` to `int`, avoiding the `int`/`int64` mismatch that silently fails testify comparisons.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Code follows stack conventions (`gofmt` clean, `go vet` clean on the changed package)
- [x] Tests included and passing (5/5 locally, 27/27 CI checks green, RED and mutation-verified)
- [x] Documentation updated if needed (no user-facing surface changed)

Issue #452, opened 2026-09-01
WebsocketClient has no owner for Conn: unsynchronized field, two readers, and a ShutDown that panics twice


jisung-02:
## Background

`WebsocketClient` has no owner for its connection. `wc.Conn` is written by `Connect()` and read from four other methods, `Conn` itself may be read by two goroutines at once, and `ShutDownChan` is closed by a method six call sites invoke. Every one of these is a data race or a panic waiting for the right interleaving, and they share one cause: nothing says which goroutine owns the client's state.

PR #444 surfaced all three while fixing an unrelated fd leak. It captured `conn := wc.Conn` once at the top of `Close()` (`pkg/runner/client.go:232`) so that one method acts on a single connection, but that narrows the window rather than closing it. The three items below are pre-existing and were left out of that PR deliberately: an fd-leak fix is the wrong place to change the concurrency model.

`ControlClient` already solves the same problem with a `sync.Mutex` (`pkg/runner/control_client.go:29`), so the shape of the answer is in the repo.

## 1. `wc.Conn` is read without synchronization

`Connect()` assigns the field at `pkg/runner/client.go:209`. These read it, unsynchronized:

| Site | Method |
| --- | --- |
| `client.go:136` | `RunForever` (`SetReadDeadline`) |
| `client.go:178` | `ReadMessage` |
| `client.go:232` | `Close` |
| `client.go:357` | `WriteJSON` |

The writer runs on the `RunForever` goroutine by way of `CloseAndReconnect`. The readers do not all run there. `Close()` is called from outside `pkg/runner` on other goroutines:

- `cmd/alpamon/command/root.go:231` (`gracefulShutdown`, main goroutine) against `root.go:187`, which starts `go wsClient.RunForever(ctx)`
- `pkg/plugin/plugin.go:207`
- `pkg/pluginclient/client.go:197`

A pointer read is atomic on the platforms Alpamon ships, so the observed value is old or new and never garbage. That is why this has not produced a crash. It is still a race by the memory model, `-race` will report it as soon as a test drives both paths together, and the behavior it allows is real: `gracefulShutdown` can close connection A while `RunForever` has already moved to B.

## 2. Two goroutines can read the socket at once

`drainCloseReply` calls `Conn.NextReader` (`pkg/runner/client.go:255`) while `RunForever` may be blocked in `Conn.ReadMessage` (`client.go:178`). gorilla/websocket allows one concurrent reader and one concurrent writer, no more. Its own docs say so.

[...]

PR #473, opened 2026-09-11
feat(wsclient): add a leaf WebSocket transport package

### LINKED ISSUE #466: Add pkg/wsclient leaf package and tag release
This is the `mon/01` task of the Kubernetes support implementation plan—alpamon's P0 deliverable, and the leaf package `alpamon-kube` pins to start its own P0 work.

| Item | Value |
|---|---|
| Task ID | `mon/01` |
| Phase | P0 |
| Size | M |
| Depends on | None (leads the whole plan) |
| Blocks | `alpamon-kube` TASK-01, `alpamon` TASK-02 |
| Basis | Plan 06 P0 alpamon row and release-order row 0, 01 §1-§1.1 |
| Task file | `archive/plan/kubernetes plan/tasks/alpamon/TASK-01_p0_wsclient-leaf-extraction-and-tag-release.md` |

## Goal

Once merged and tagged, alpamon has a pure WS-transport leaf, `pkg/wsclient`, as an exported public API, and the separate `alpacax/alpamon-kube` repo can pin that version and start its own P0 (scaffold, registration, connect). Unlike dhcp-plugin's wholesale dependency on `pkg/runner`, `pkg/wsclient` has no alpamon-internal dependency besides `pkg/tunnel` and `pkg/version`.

### In scope

- Add `pkg/wsclient`: WS dial, `Authorization: id="..." key="..."` header, exponential backoff with jitter, read timeout, reconnect.
- Take config as a struct the caller injects, not the global `config.GlobalSettings`.
- Add a CI guard job that runs `go list -deps ./pkg/wsclient` and checks the alpamon-internal dependency set is limited to `pkg/tunnel` and `pkg/version`.
- Verify with the same guard that `pkg/wsclient` and its dependency graph carry no pty or systemd symbols.
- Cut a semver tag carrying the new API, the version `alpamon-kube` pins.

### Out of scope

- Moving dhcp-plugin onto `pkg/wsclient`, which is `alpamon` TASK-02 (`mon/02`).
- Refactoring `pkg/runner/client.go NewWebsocketClient` or its callers (`pkg/plugin`, `pkg/pluginclient`). The existing path stays as is; this PR only adds the new leaf package.
- `alpamon-kube`-side work (registration flow, RBAC, Helm chart), which belongs to `alpamon-kube` TASK-01/02/03.

## Files changed

| Type | Path | Notes |
|---|---|---|
| New | `pkg/wsclient/client.go` | WS dial, header, read timeout, reconnect loop; no `internal` types in exported constructors or signatures |
| New | `pkg/wsclient/backoff.go` | Exponential backoff with jitter, reimplemented (not imported) from `internal/retry/retry.go ExponentialBackoff` |
| New | `pkg/wsclient/config.go` | Caller-injected config struct (URL, id, key, timeout values) |
| Modified | `.github/workflows/build-and-test.yml` | New job checking `go list -deps ./pkg/wsclient` output and pty/systemd symbol absence |
| New | `pkg/wsclient/client_test.go`, `backoff_test.go` | testify-based unit tests |

## Definition of done

- [ ] `pkg/wsclient` implements WS dial, the `Authorization: id/key` header, exponential backoff with jitter, read timeout, and reconnect.
- [ ] `pkg/wsclient`'s public signatures take a config struct and expose no `internal` types.
- [ ] `go list -deps ./pkg/wsclient` shows no alpamon-internal dependency beyond `pkg/tunnel` and `pkg/version`.
- [ ] That check is automated as a CI guard job that fails on regression.
- [ ] The same guard confirms no pty or systemd symbols in the dependency graph.
- [ ] Unit tests (backoff, jitter, reconnect) pass.
- [ ] Existing `build-and-test.yml` jobs pass with no regression.
- [ ] A semver tag is published.

## Verification

- Unit tests cover the backoff+jitter sequence (cap reached, jitter range), reconnect after dial failure, and read-timeout triggering, using testify (`require`/`assert`).
- Use `synctest` with a virtual clock to verify reconnect timing (backoff intervals).
- New regression guard: a CI job runs `go list -deps ./pkg/wsclient`, confirms the result is limited to `pkg/tunnel`, `pkg/version`, the standard library, and `gorilla/websocket`, and greps for the absence of `pkg/pty`, `pkg/utils`, `internal/*`, and systemd-related symbols.
- No regression in existing `build-and-test.yml` jobs (`verify-tidy`, build, test); `go mod tidy` produces a zero diff.

## Related issues

- alpacax/alpamon#452: `WebsocketClient` has no owner for its `Conn`—an unsynchronized field, two readers, and a `ShutDown` that can panic twice, all in `pkg/runner/client.go`, the code this extraction is modeled on. Decide before merge whether `pkg/wsclient` adopts an ownership model (the `ControlClient` mutex shape that issue points to) or reproduces the current one. Exporting the racy shape as a public leaf API makes it harder to change once `alpamon-kube` pins the tag.

## PR meta

- Branch: `feature/kube-wsclient-leaf`
- PR title: `Add pkg/wsclient leaf package and tag release`

Task file: `archive/plan/kubernetes plan/tasks/alpamon/TASK-01_p0_wsclient-leaf-extraction-and-tag-release.md`

geunwoonoh:
## Summary

Adds `pkg/wsclient`, alpamon's WebSocket transport as a package a module outside this repository can import, and a CI job that keeps it that way. This is the first PR of the Kubernetes plan: `alpacax/alpamon-kube` cannot start until alpamon ships this and tags it.

Today the only WS client is in `pkg/runner/client.go`. `NewWebsocketClient` takes an `internal/pool.Pool`, so no other module can build an argument for it, and importing `pkg/runner` pulls in `internal/{pool,protocol,retry}`, `pkg/{agent,config,executor/handlers/common,scheduler,signing,utils}`, creack/pty and gopsutil. `alpamon-dhcp-plugin` is the standing evidence: pinned to pre-`/v2` `v1.3.0`, compiling the whole host stack to open a socket.

Measured with a throwaway module importing each:

| Imports | Module-backed packages compiled |
|---|---|
| `pkg/wsclient` | 3: gorilla/websocket, `pkg/version`, itself |
| `pkg/runner` | 63, including creack/pty, gopsutil, smux, zerolog, validator |

Nothing existing changes. `pkg/runner`, `pkg/plugin` and `pkg/pluginclient` are untouched, and no alpamon code calls the new package yet.

## Issue #452

#466 makes this a merge gate, because exporting a racy shape freezes it for every module that pins the tag. The new package does not reproduce any of the three defects: `conn` is unexported and guarded by a mutex, `Run` is the only reader, writes hold a write mutex, and `Shutdown` is a `sync.Once`. No exported method touches the read side — `Shutdown` and `Reconnect` push the read deadline into the past under the same mutex the read loop arms it with, so a request cannot be lost to a re-arm.

**#452 stays open.** This PR sidesteps those defects in new code; `pkg/runner` still has them.

## Deliberate differences from pkg/runner/client.go

None can regress a caller, because there is no caller yet.

- Nothing calls `os.Exit`, and there is no retry exhaustion. `pkg/runner`'s `Connect` gives up after `maxRetryTimeout` (3 days) and exits the process. `Run` keeps redialing until `Shutdown` (returns nil), its context ending (returns `ctx.Err()`), or a handler error (returns that error). A caller that wants a give-up policy counts attempts in `OnRetry` and calls `Shutdown`, or bounds the context.
- The backoff paces attempts, not just failed dials. A connection that ends before it proved itself waits like a failed dial does; a peer that accepts the handshake and closes at once would otherwise be redialed thousands of times a second, per agent. Proving itself means staying up for `MinBackoff`, measured before the close handshake. Carrying a frame does not count, because a peer that sends one byte and hangs up would otherwise reset the backoff every round. A drop after that, or a close the caller asked for with `Reconnect`, still redials immediately. A failed write is paced like any other failure.
- `HandshakeTimeout` defaults to 30s, and a dial stalled after TCP is aborted when `Shutdown` or the context ends. gorilla/websocket stops watching the context once the socket is up, so otherwise a shutdown waits out the handshake, or forever when a caller's dialer sets no timeout.
- Proxies from the environment are honored; every dialer in the repo ignores them today. gorilla/websocket v1.5.3 panics on a CONNECT status line with no reason phrase (`proxy.go` indexes the reason phrase unchecked), so a dial that panics is returned as an error rather than taking the agent down.
- Writes carry a deadline, and a failed write replaces the connection, since gorilla fails every later write on it. `WriteMessage` accepts data frames only: a close frame handed to it would end the connection behind the read loop's back.
- A server answering with an extension the client never offered is refused. gorilla enables decompression for it regardless of `EnableCompression`, and a read limit counts compressed bytes, so together they turn a small frame into an unbounded allocation.

Logging is by hooks (`OnConnect`, `OnDisconnect`, `OnRetry`) rather than a logger: `pkg/logger` would pull `pkg/config`, `pkg/scheduler`, `pkg/utils`, gopsutil and smux into the graph, and a library should not write into somebody else's global. alpamon-kube drives its `alpamonkube_backhaul_connected` gauge off the two edges.

The backoff is `internal/retry`'s jitter from #468, reimplemented rather than imported, because Go's internal rule puts `internal/retry` out of reach of the modules this package exists for.

## The guard

`leaf-guard` asserts the module-backed dependency graph of `pkg/wsclient` stays within `pkg/tunnel`, `pkg/version` and gorilla/websocket, and is in the `build-and-test` aggregator's `needs` list, which is the required status check. That set is a ceiling, not an inventory: what ships imports `pkg/version` and gorilla and nothing else, and `pkg/tunnel` is allowed for a caller that wants `tunnel.WebSocketConn` around what `Dial` returns without this package reaching for it. `release.yml` calls this workflow, so it gates tags too.

It is written not to pass on air (the package must appear in its own graph first), not to fail open (plain string comparison, because `if unexpected=$(grep -vE ...)` reads grep's exit 2 for a bad pattern the same as exit 1 for no matches), and not to check only one platform (every target `.goreleaser.yaml` ships, both architectures, plus a cgo-on pass).

Run locally against each violation it claims to catch. All fail:

| Case | Result |
|---|---|
| clean tree | pass |
| `import _ "pkg/config"` | fail, names `pkg/config` and its tail |
| `import _ "pkg/utils"` | fail |
| `import _ "internal/retry"` | fail |
| `import _ "zerolog"` | fail |
| `import _ "creack/pty"` | fail, with the pty and systemd message |
| `//go:build windows` file importing `pkg/config` | fail, on `windows/amd64/0` |
| `//go:build arm64` file importing `pkg/config` | fail, on `linux/arm64/0` |
| package renamed away | fail |
| pattern matching no packages | fail, as a vacuous pass |
| allowlist emptied or left as a stale regex | fail |
| `import _ "pkg/tunnel"` (allowed) | pass |

## Testing

84 tests. Backoff and config are table tests; the reconnect schedule is timed under `testing/synctest`'s fake clock, which works because the dialer fails in-process and the only thing `Run` blocks on is its backoff timer; everything socket-backed stays on the real clock, as `CLAUDE.md` requires.

- `go test ./... -p 1`: 48 packages, no failures.
- `go test -race ./pkg/wsclient -count=15`, four times over, plus `-cpu 1,2,8`: clean.
- Cross-compiles for linux/amd64, linux/arm64, darwin/amd64, darwin/arm64, windows/amd64.
- `go mod tidy` is a zero diff; no new module enters `go.mod`.
- 26 deliberate mutations of the code were each caught by a test: dropping the write mutex, an interrupt that frees nothing, a non-idempotent `Shutdown`, jitter fed back into the doubling base, a lost backoff reset, an overflowing double, a requested close reported as a failure, `armRead` ignoring requests, a missing write deadline, control frames let through, the unconditional `Close` moved back inside the success branch (the PR #444 fd leak), the header's comma, the pacing rule, the handshake abort, the panic recovery, the compression refusal, and the rest.

## Not in this PR

- The `v2.7.0` tag. Nothing needs editing to cut it: the version comes from ldflags and there is no CHANGELOG. **This issue stays open until that tag exists**, since `alpamon-kube` TASK-01 pins it.
- Moving dhcp-plugin onto this API, which happens in its own repo.

## Note for alpamon-kube TASK-01

Its env table lists `WS_MIN_BACKOFF`, `WS_MAX_BACKOFF` and `WS_MAX_BACKOFF_CEILING`, but this package exposes two knobs. Map `WS_MIN_BACKOFF` to `MinBackoff` and `WS_MAX_BACKOFF_CEILING` to `MaxBackoff`, and drop `WS_MAX_BACKOFF`: a 60s cap cannot spread a 500-cluster herd, so the ceiling is the number that matters. It must also set `UserAgent`, or it reports itself as alpamon.

Refs #466, #452

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -157,6 +157,35 @@ jobs:
             printf '%s\n' "$deps" | sed 's/^/  /'
           done
 
+  # pkg/wsclient holds one connection open across reconnects with a read
+  # loop, callers writing from their own goroutines, and a watcher that
+  # interrupts a parked read. That ownership model is the whole point of the
+  # package, and getting it wrong is what issue #452 is about in pkg/runner.
+  # The suite below runs without the race detector on every platform, so a
+  # race reintroduced here would pass it.
+  #
+  # Cheap enough to be worth its own job: no cgo of its own, no database and
+  # no generated code, so it needs neither a distro container nor the ent
+  # codegen step. `cache: false` for the same reason as leaf-guard, so this
+  # job cannot claim the shared setup-go cache key for a near-empty cache.
+  race-wsclient:
+    name: Race detector (pkg/wsclient)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+          cache: false
+
+      # Twice over: a race that needs an unlucky interleaving does not always
+      # show up on the first scheduling.
+      - name: Run tests under the race detector
+        run: go test -race -count=2 ./pkg/wsclient
+
   # Renamed from `build-and-test` so the flat aggregator at the bottom of this
   # file can take that id: a job id and the aggregate check name cannot be the
   # same string, and the aggregate is the one that has to be `build-and-test`.
@@ -414,6 +443,7 @@ jobs:
     needs:
       - verify-tidy
       - leaf-guard
+      - race-wsclient
       - build-and-test-linux
       - build-and-test-windows
       - build-and-test-macos
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -29,6 +29,134 @@ jobs:
           go mod tidy
           git diff --exit-code go.mod go.sum
 
+  # pkg/wsclient is imported by alpacax/alpamon-kube, a separate module that
+  # pins an alpamon tag, and its whole value is that it is a leaf: WebSocket
+  # transport and nothing else. The day it imports pkg/config, pkg/utils,
+  # internal/* or anything that drags in creack/pty or the systemd probes,
+  # every importer has to compile the host agent just to open a socket, which
+  # is how alpamon-dhcp-plugin ended up pinned to v1.3.0 for over a year. No
+  # linter here enforces an import boundary (there is no .golangci.yml, so no
+  # depguard), so this job is the only thing holding the line.
+  #
+  # It deliberately skips ./.github/actions/setup-go-generate. The ent package
+  # is gitignored and loading pkg/wsclient must not need it; if this job ever
+  # fails for want of generated code, the package has grown a dependency it
+  # should not have, and red is the right answer.
+  #
+  # To run it yourself, save the step's script to a file and run it with
+  # `PKG=... ALLOWED='...' bash that-file`. It needs nothing but Go. Do not
+  # paste it into an interactive shell: `set -e` and `exit 1` would take the
+  # shell down with it.
+  leaf-guard:
+    name: Verify pkg/wsclient stays a leaf
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+          # This job downloads almost nothing, and setup-go's cache key is the
+          # same one the ubuntu jobs using setup-go-generate compute. Saving
+          # from here would claim that key for a near-empty cache and leave the
+          # jobs that need it with nothing to restore.
+          cache: false
+
+      - name: Check the dependency graph
+        env:
+          PKG: github.com/alpacax/alpamon/v2/pkg/wsclient
+          # The only packages pkg/wsclient may reach outside the standard
+          # library. pkg/tunnel and pkg/version are leaves themselves.
+          ALLOWED: >-
+            github.com/alpacax/alpamon/v2/pkg/wsclient
+            github.com/alpacax/alpamon/v2/pkg/tunnel
+            github.com/alpacax/alpamon/v2/pkg/version
+            github.com/gorilla/websocket
+        run: |
+          set -euo pipefail
+
+          # Plain string comparison rather than a regex. `if unexpected=$(grep
+          # -vE "$ALLOWED" ...)` would read grep's exit 2, which is what a
+          # mistyped pattern gives, the same as exit 1, and let the guard pass
+          # on a package that had already drifted.
+          allowed() {
+            local candidate="$1" entry
+            for entry in $ALLOWED; do
+              if [ "$entry" = "$candidate" ]; then
+                return 0
+              fi
+            done
+            return 1
+          }
+
+          # go list resolves build constraints for one target at a time, so a
+          # file tagged //go:build windows could import anything and a Linux
+          # runner would never see it. Cover every target .goreleaser.yaml
+          # ships, which build with CGO off, plus the cgo-on pass that matches
+          # how this repo's own Linux jobs build.
+          for target in linux/amd64/1 linux/amd64/0 linux/arm64/0 darwin/amd64/0 darwin/arm64/0 windows/amd64/0; do
+            IFS=/ read -r goos goarch cgo <<<"$target"
+
+            # {{if .Module}} prints only packages that come from a module,
+            # which leaves out the standard library, including the
+            # GOROOT-vendored vendor/golang.org/x/... packages crypto/tls and
+            # net/http pull in. Those print as empty lines, which sed drops.
+            # -deps skips test-only imports on purpose: tests may use testify,
+            # and no importer compiles them, so do not add -test.
+            deps="$(GOOS="$goos" GOARCH="$goarch" CGO_ENABLED="$cgo" \
+              go list -deps -f '{{if .Module}}{{.ImportPath}}{{end}}' "$PKG" | sed '/^$/d')"
+
+            # Unquoted on purpose: one dependency per word.
+            # shellcheck disable=SC2086
+            set -- $deps
+
+            # A package that no longer resolves would leave nothing to check,
+            # and every test below would pass on an empty list.
+            self=no
+            for dep in "$@"; do
+              if [ "$dep" = "$PKG" ]; then
+                self=yes
+              fi
+            done
+            if [ "$self" = no ]; then
+              echo "::error::$PKG is missing from its own dependency list on $target, so this guard would have passed without checking anything"
+              printf '%s\n' "$deps"
+              exit 1
+            fi
+
+            coupling=
+            unexpected=
+            for dep in "$@"; do
+              case "$dep" in
+                *pty*|*systemd*) coupling="$coupling  $dep"$'\n' ;;
+              esac
+              if ! allowed "$dep"; then
+                unexpected="$unexpected  $dep"$'\n'
+              fi
+            done
+
+            # Checked before the allowlist, which would refuse these too:
+            # naming the two couplings that made a separate alpamon-kube
+            # necessary says more than a generic message about whichever
+            # transitive import happened to bring them in.
+            if [ -n "$coupling" ]; then
+              echo "::error file=pkg/wsclient/doc.go::on $target, pkg/wsclient's dependency graph reaches pty or systemd code:"
+              printf '%s' "$coupling"
+              exit 1
+            fi
+            if [ -n "$unexpected" ]; then
+              echo "::error file=pkg/wsclient/doc.go::pkg/wsclient must stay a leaf, but on $target its dependency graph reaches:"
+              printf '%s' "$unexpected"
+              echo "Allowed: $ALLOWED, and the standard library."
+              exit 1
+            fi
+
+            echo "pkg/wsclient on $target:"
+            printf '%s\n' "$deps" | sed 's/^/  /'
+          done
+
   # Renamed from `build-and-test` so the flat aggregator at the bottom of this
   # file can take that id: a job id and the aggregate check name cannot be the
   # same string, and the aggregate is the one that has to be `build-and-test`.
@@ -285,6 +413,7 @@ jobs:
   build-and-test:
     needs:
       - verify-tidy
+      - leaf-guard
       - build-and-test-linux
       - build-and-test-windows
       - build-and-test-macos
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -34,7 +34,7 @@ jobs:
   # transport and nothing else. The day it imports pkg/config, pkg/utils,
   # internal/* or anything that drags in creack/pty or the systemd probes,
   # every importer has to compile the host agent just to open a socket, which
-  # is how alpamon-dhcp-plugin ended up pinned to v1.3.0 for over a year. No
+  # is what keeps alpamon-dhcp-plugin on a pre-/v2 alpamon today. No
   # linter here enforces an import boundary (there is no .golangci.yml, so no
   # depguard), so this job is the only thing holding the line.
   #
--- a/pkg/wsclient/backoff.go
+++ b/pkg/wsclient/backoff.go
@@ -1,0 +1,59 @@
+package wsclient
+
+import (
+	"math/rand/v2"
+	"time"
+)
+
+// backoff is the reconnect schedule: exponential with jitter. It mirrors
+// internal/retry.ExponentialBackoff rather than importing it, because Go's
+// internal rule keeps internal/ out of reach of the modules this package
+// exists for.
+//
+// The base doubles from initial up to max, and only the value handed back is
+// jittered. Keeping the base pure means an unlucky draw never drags the
+// sequence down, so it reaches max on the same schedule every time.
+type backoff struct {
+	initial time.Duration
+	max     time.Duration
+	rand    func() float64 // returns [0, 1); nil means math/rand/v2
+
+	current time.Duration
+}
+
+// next returns the next wait: the doubling base multiplied by a factor in
+// [0.5, 1.5), clamped to [initial, max]. Without the random factor every
+// agent reconnecting after the same event, such as a backhaul restart,
+// would retry in lockstep.
+func (b *backoff) next() time.Duration {
+	switch {
+	case b.current == 0:
+		b.current = b.initial
+	case b.current > b.max/2: // doubling would pass max, or overflow
+		b.current = b.max
+	default:
+		b.current *= 2
+	}
+
+	randFn := b.rand
+	if randFn == nil {
+		randFn = rand.Float64
+	}
+
+	// Clamp in float64 before converting: a factor near 1.5 on a large max
+	// would overflow time.Duration, and !(x < max) also catches a NaN from
+	// a caller-supplied source.
+	jittered := float64(b.current) * (0.5 + randFn())
+	if !(jittered < float64(b.max)) {
+		return b.max
+	}
+	if d := time.Duration(jittered); d > b.initial {
+		return d
+	}
+	return b.initial
+}
+
+// reset restarts the schedule at initial, after a connection succeeds.
+func (b *backoff) reset() {
+	b.current = 0
+}
--- a/pkg/wsclient/backoff.go
+++ b/pkg/wsclient/backoff.go
@@ -22,9 +22,15 @@ type backoff struct {
 }
 
 // next returns the next wait: the doubling base multiplied by a factor in
-// [0.5, 1.5), clamped to [initial, max]. Without the random factor every
-// agent reconnecting after the same event, such as a backhaul restart,
-// would retry in lockstep.
+// [1.0, 1.5), clamped to [initial, max]. Without the random factor every
+// agent reconnecting after the same event, such as a backhaul restart, would
+// retry in lockstep.
+//
+// The factor starts at 1.0 rather than 0.5, which is where internal/retry
+// starts it, because the clamp to initial would otherwise swallow the whole
+// lower half: at the first attempt the base is initial, so every draw below
+// 1.0 returns exactly initial and half of a fleet retries on the same tick.
+// Spreading upward keeps initial a real floor and the distribution intact.
 func (b *backoff) next() time.Duration {
 	switch {
 	case b.current == 0:
@@ -43,7 +49,7 @@ func (b *backoff) next() time.Duration {
 	// Clamp in float64 before converting: a factor near 1.5 on a large max
 	// would overflow time.Duration, and !(x < max) also catches a NaN from
 	// a caller-supplied source.
-	jittered := float64(b.current) * (0.5 + randFn())
+	jittered := float64(b.current) * (1.0 + 0.5*randFn())
 	if !(jittered < float64(b.max)) {
 		return b.max
 	}
--- a/pkg/wsclient/client.go
+++ b/pkg/wsclient/client.go
@@ -319,11 +319,12 @@ func (c *Client) serve(ctx context.Context, conn *websocket.Conn, h Handler) (di
 		messageType, payload, err := conn.ReadMessage()
 		if err != nil {
 			cause, ending := c.closeRequest(ctx)
-			// Freeing a read produces a timeout and nothing else, so any
-			// other error is the connection's own ending, even if a request
-			// happened to be pending: report that rather than the request.
-			var timeout net.Error
-			if ending && errors.As(err, &timeout) && timeout.Timeout() {
+			// Freeing a read produces a timeout, or net.ErrClosed when the
+			// socket refused the deadline and freeRead closed it instead. Any
+			// other error is the connection's own ending, a peer's close
+			// frame among them, even if a request happened to be pending:
+			// report that rather than the request.
+			if ending && wokenOnPurpose(err) {
 				err = cause
 			}
 			c.release(conn, err)
@@ -376,14 +377,25 @@ func (c *Client) rearm(ctx context.Context, conn *websocket.Conn) {
 	_ = conn.SetReadDeadline(time.Now().Add(c.s.readTimeout))
 }
 
-// freeRead unblocks a read parked on conn. Pushing the deadline into the
-// past is the ordinary way, but a caller-supplied net.Conn may refuse a
-// deadline, and then closing the socket is the only thing left that ends the
-// read. Both are safe to call while another goroutine is reading.
+// freeRead unblocks a read parked on conn from a goroutine other than the
+// reader. It works on the underlying net.Conn: gorilla/websocket counts its
+// own SetReadDeadline among the read methods that only one goroutine may
+// call, while net.Conn promises that any of its methods may be called
+// concurrently. Pushing the deadline into the past is the ordinary way, but a
+// caller-supplied net.Conn may refuse a deadline, and then closing the socket
+// is the only thing left that ends the read; gorilla's Close is safe to call
+// concurrently too.
 func freeRead(conn *websocket.Conn) {
-	if err := conn.SetReadDeadline(aLongTimeAgo); err != nil {
+	if err := conn.UnderlyingConn().SetReadDeadline(aLongTimeAgo); err != nil {
 		_ = conn.Close()
 	}
+}
+
+// wokenOnPurpose reports whether a failed read is one this package caused
+// to free it, as opposed to one the connection failed on by itself.
+func wokenOnPurpose(err error) bool {
+	var timeout net.Error
+	return (errors.As(err, &timeout) && timeout.Timeout()) || errors.Is(err, net.ErrClosed)
 }
 
 // closeRequest reports whether the live connection is ending, and the cause
@@ -468,8 +480,10 @@ func closeConn(conn *websocket.Conn) {
 	err := conn.WriteControl(websocket.CloseMessage,
 		websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""),
 		time.Now().Add(closeTimeout))
-	if err == nil {
-		_ = conn.SetReadDeadline(time.Now().Add(drainTimeout))
+	// Only drain when the drain can be bounded: a conn that refuses the
+	// deadline would leave NextReader waiting on a silent peer forever, and
+	// Close, which is what actually matters, would never be reached.
+	if err == nil && conn.SetReadDeadline(time.Now().Add(drainTimeout)) == nil {
 		for {
 			if _, _, err := conn.NextReader(); err != nil {
 				break
--- a/pkg/wsclient/client.go
+++ b/pkg/wsclient/client.go
@@ -1,0 +1,427 @@
+package wsclient
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/gorilla/websocket"
+)
+
+// closeTimeout bounds each step of closing a connection: sending the close
+// frame, and waiting for the peer's reply to it.
+const closeTimeout = 5 * time.Second
+
+// aLongTimeAgo is a read deadline already in the past. Setting it frees a
+// parked read at once, whatever the clock says.
+var aLongTimeAgo = time.Unix(1, 0)
+
+var (
+	// ErrNotConnected is returned by a write made while no connection is up,
+	// for example between a drop and the reconnect that follows it.
+	ErrNotConnected = errors.New("wsclient: not connected")
+
+	// ErrAlreadyRunning is returned by Run on a Client whose Run has already
+	// been called. A Client is single-use.
+	ErrAlreadyRunning = errors.New("wsclient: Run has already been called")
+)
+
+// Handler receives each inbound frame. It runs on the Run goroutine, so the
+// next frame is not read until it returns: hand long work to another
+// goroutine. ctx is done once Run starts stopping. A non-nil error stops Run,
+// which returns it; call Shutdown for an ordinary stop.
+type Handler func(ctx context.Context, messageType int, payload []byte) error
+
+// Client keeps one WebSocket connection up: it dials, reconnects with
+// jittered backoff whenever the connection drops, and hands every inbound
+// frame to a Handler.
+//
+// The Client owns its connection. Run is the only goroutine that reads it,
+// writes are serialized, and Reconnect and Shutdown only signal the read
+// loop, so gorilla/websocket's one-reader, one-writer rule holds without the
+// caller's help. All methods are safe for concurrent use.
+type Client struct {
+	s settings
+
+	// mu guards conn and the pending-reconnect pair. Nothing slower than
+	// SetReadDeadline happens while it is held.
+	mu   sync.Mutex
+	conn *websocket.Conn
+	// reconnectPending asks Run to replace conn; reconnectCause is what
+	// OnDisconnect then reports: nil for Reconnect, the write error for a
+	// connection a failed write abandoned. Both reset whenever conn changes.
+	reconnectPending bool
+	reconnectCause   error
+
+	// writeMu serializes data frames. Lock order is writeMu, then mu.
+	writeMu sync.Mutex
+
+	started      atomic.Bool
+	shutdownOnce sync.Once
+	shutdown     chan struct{}
+	done         chan struct{}
+}
+
+// New validates cfg and returns a Client ready to Run. It does not dial.
+func New(cfg Config) (*Client, error) {
+	s, err := cfg.resolve()
+	if err != nil {
+		return nil, err
+	}
+	return &Client{
+		s:        s,
+		shutdown: make(chan struct{}),
+		done:     make(chan struct{}),
+	}, nil
+}
+
+// Run dials, then feeds every inbound frame to h until Shutdown is called,
+// ctx is done, or h returns an error, reconnecting whenever the connection
+// drops. It keeps trying for as long as it runs; OnRetry reports every wait,
+// so a caller that wants to give up can call Shutdown.
+//
+// The backoff paces attempts, not just failed dials. A connection that ends
+// before it proved itself is followed by the same jittered wait a failed dial
+// gets. A connection proves itself by carrying an inbound frame or by staying
+// up for MinBackoff, and one that does restarts the schedule, as does a
+// reconnect the caller asked for. Without that, a peer that accepts the
+// handshake and closes at once would be redialed in a tight loop.
+//
+// Run returns nil after Shutdown, ctx.Err() once ctx is done, and h's error
+// when h stopped it. The connection is closed before Run returns. A Client
+// runs once: a second call returns ErrAlreadyRunning.
+func (c *Client) Run(ctx context.Context, h Handler) error {
+	if !c.started.CompareAndSwap(false, true) {
+		return ErrAlreadyRunning
+	}
+	defer close(c.done)
+	if h == nil {
+		return errors.New("wsclient: Run needs a non-nil Handler")
+	}
+
+	// runCtx also ends on Shutdown, so one context aborts a dial in flight,
+	// a backoff wait and the handler alike. A parked read ignores contexts,
+	// which is why the watcher also interrupts it.
+	runCtx, cancel := context.WithCancel(ctx)
+	var wg sync.WaitGroup
+	wg.Go(func() {
+		select {
+		case <-c.shutdown:
+			cancel()
+		case <-runCtx.Done():
+		}
+		c.interrupt()
+	})
+	defer func() {
+		cancel()
+		wg.Wait()
+	}()
+
+	b := &backoff{initial: c.s.minBackoff, max: c.s.maxBackoff, rand: c.s.rand}
+	var (
+		attempt int   // attempts since the last connection that worked
+		wait    bool  // whether the next attempt waits out a backoff first
+		cause   error // what made it wait
+	)
+	for {
+		// Checked before the wait as well as before the dial, so a Shutdown
+		// neither dials again nor reports a retry that will never happen.
+		if c.stopping(runCtx) {
+			return ctx.Err()
+		}
+		if wait {
+			attempt++
+			delay := b.next()
+			c.s.onRetry(attempt, delay, cause)
+			if !c.sleep(runCtx, delay) {
+				return ctx.Err()
+			}
+		}
+
+		conn, _, err := c.s.dial(runCtx)
+		if err != nil {
+			wait, cause = true, err
+			continue
+		}
+		if !c.install(runCtx, conn) {
+			closeConn(conn)
+			return ctx.Err()
+		}
+		c.s.onConnect()
+
+		ended, handlerErr := c.serve(runCtx, conn, h)
+		if handlerErr != nil {
+			return handlerErr
+		}
+		if ended.requested || ended.proven {
+			b.reset()
+			attempt, wait, cause = 0, false, nil
+			continue
+		}
+		wait, cause = true, ended.cause
+	}
+}
+
+// sleep waits out d, and reports false if Run should stop instead.
+func (c *Client) sleep(ctx context.Context, d time.Duration) bool {
+	timer := time.NewTimer(d)
+	defer timer.Stop()
+	select {
+	case <-timer.C:
+		return true
+	case <-ctx.Done():
+		return false
+	case <-c.shutdown:
+		return false
+	}
+}
+
+// WriteMessage sends one data frame on the live connection. messageType must
+// be websocket.TextMessage or websocket.BinaryMessage: control frames belong
+// to the Client. It returns ErrNotConnected while no connection is up, and
+// never queues a frame for a later connection. A failed write leaves the
+// connection unable to send, so it also makes Run replace the connection.
+func (c *Client) WriteMessage(messageType int, data []byte) error {
+	if messageType != websocket.TextMessage && messageType != websocket.BinaryMessage {
+		// gorilla/websocket would send a close frame given here, and every
+		// later write would then fail behind the read loop's back.
+		return fmt.Errorf("wsclient: WriteMessage sends data frames only, got message type %d", messageType)
+	}
+
+	c.writeMu.Lock()
+	defer c.writeMu.Unlock()
+
+	c.mu.Lock()
+	conn := c.conn
+	c.mu.Unlock()
+	if conn == nil {
+		return ErrNotConnected
+	}
+	// gorilla/websocket keeps the write deadline in a plain field that the
+	// next write reads, so setting it is only safe under writeMu.
+	if err := conn.SetWriteDeadline(time.Now().Add(c.s.writeTimeout)); err != nil {
+		return err
+	}
+	if err := conn.WriteMessage(messageType, data); err != nil {
+		// gorilla/websocket fails every later write once one has failed, so
+		// this connection can no longer send. Replace it now rather than
+		// wait for the read side to notice.
+		c.abandon(conn, err)
+		return err
+	}
+	return nil
+}
+
+// WriteJSON sends v, encoded as JSON, as one text frame.
+func (c *Client) WriteJSON(v any) error {
+	data, err := json.Marshal(v)
+	if err != nil {
+		return err
+	}
+	return c.WriteMessage(websocket.TextMessage, data)
+}
+
+// Reconnect drops the live connection so that Run dials a fresh one, as when
+// the server asks the agent to reconnect. While no connection is up it does
+// nothing, since Run is already dialing. It does not block.
+func (c *Client) Reconnect() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.conn == nil || c.reconnectPending {
+		return
+	}
+	c.reconnectPending = true
+	c.reconnectCause = nil
+	_ = c.conn.SetReadDeadline(aLongTimeAgo)
+}
+
+// Shutdown makes Run close the connection and return. It does not wait; Done
+// does. Calling it more than once, before Run, or after Run returned is safe.
+func (c *Client) Shutdown() {
+	c.shutdownOnce.Do(func() { close(c.shutdown) })
+}
+
+// Done is closed once Run has returned. It never closes if Run is never called.
+func (c *Client) Done() <-chan struct{} {
+	return c.done
+}
+
+// Connected reports whether a connection is up. It is a snapshot: the
+// connection may drop the moment after it returns true.
+func (c *Client) Connected() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.conn != nil
+}
+
+// install makes conn the live connection, unless Run started stopping while
+// conn was being dialed.
+func (c *Client) install(ctx context.Context, conn *websocket.Conn) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.stopping(ctx) {
+		return false
+	}
+	c.conn = conn
+	c.reconnectPending, c.reconnectCause = false, nil
+	return true
+}
+
+// disconnect says how a connection ended, which is what decides whether Run
+// redials at once or waits out a backoff first.
+type disconnect struct {
+	requested bool  // Shutdown, Reconnect or the Run context asked for it
+	proven    bool  // it carried a frame, or stayed up for MinBackoff
+	cause     error // what ended it; nil when the client closed a healthy connection
+}
+
+// serve reads conn into h until the connection ends, and releases it. It
+// returns h's error when h stopped it, and nil otherwise.
+func (c *Client) serve(ctx context.Context, conn *websocket.Conn, h Handler) (disconnect, error) {
+	opened := time.Now()
+	frames := 0
+	// A connection that carried traffic, or lasted as long as the shortest
+	// backoff, has shown the endpoint is worth redialing at once.
+	proven := func() bool { return frames > 0 || time.Since(opened) >= c.s.minBackoff }
+
+	for {
+		if cause, requested := c.armRead(ctx, conn); requested {
+			c.release(conn, cause)
+			return disconnect{requested: true, proven: proven(), cause: cause}, nil
+		}
+		messageType, payload, err := conn.ReadMessage()
+		if err != nil {
+			cause, requested := c.closeRequest(ctx)
+			// Freeing a read produces a timeout and nothing else, so any
+			// other error is the connection's own ending, even if a request
+			// happened to be pending: report that rather than the request.
+			var timeout net.Error
+			if requested && errors.As(err, &timeout) && timeout.Timeout() {
+				err = cause
+			} else {
+				requested = false
+			}
+			c.release(conn, err)
+			return disconnect{requested: requested, proven: proven(), cause: err}, nil
+		}
+		frames++
+		if err := h(ctx, messageType, payload); err != nil {
+			// A handler that honors its context reports the cancellation that
+			// is already stopping Run. That is not a failure of its own, and
+			// Run must still return nil for a Shutdown.
+			if ctxErr := ctx.Err(); ctxErr != nil && errors.Is(err, ctxErr) {
+				c.release(conn, nil)
+				return disconnect{requested: true, proven: proven()}, nil
+			}
+			c.release(conn, err)
+			return disconnect{proven: proven(), cause: err}, err
+		}
+	}
+}
+
+// armRead re-arms conn's read deadline, unless a close has been requested,
+// in which case it reports that and the cause instead. Checking and arming
+// under one lock is what keeps a request from landing between the two and
+// being overwritten by a fresh ReadTimeout.
+func (c *Client) armRead(ctx context.Context, conn *websocket.Conn) (cause error, requested bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if cause, requested := c.closeRequestLocked(ctx); requested {
+		return cause, true
+	}
+	_ = conn.SetReadDeadline(time.Now().Add(c.s.readTimeout))
+	return nil, false
+}
+
+// closeRequest reports whether the live connection is ending on request
+// rather than on a failed read, and the cause OnDisconnect should report.
+func (c *Client) closeRequest(ctx context.Context) (cause error, requested bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.closeRequestLocked(ctx)
+}
+
+func (c *Client) closeRequestLocked(ctx context.Context) (cause error, requested bool) {
+	if c.reconnectPending {
+		return c.reconnectCause, true
+	}
+	return nil, c.stopping(ctx)
+}
+
+// stopping reports whether Run should wind down. It reads the shutdown
+// channel itself instead of waiting for the watcher to cancel ctx, so no
+// dial can start in the gap between the two.
+func (c *Client) stopping(ctx context.Context) bool {
+	select {
+	case <-c.shutdown:
+		return true
+	default:
+		return ctx.Err() != nil
+	}
+}
+
+// abandon asks Run to replace conn after a write on it failed, unless conn has
+// already been replaced or a reconnect is already on its way.
+func (c *Client) abandon(conn *websocket.Conn, cause error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.conn != conn || c.reconnectPending {
+		return
+	}
+	c.reconnectPending = true
+	c.reconnectCause = cause
+	_ = conn.SetReadDeadline(aLongTimeAgo)
+}
+
+// interrupt frees a read parked on the live connection. It holds mu, so it
+// serializes with armRead: either armRead sees the request behind this call,
+// or this call lands after armRead and cuts its deadline short.
+func (c *Client) interrupt() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.conn != nil {
+		_ = c.conn.SetReadDeadline(aLongTimeAgo)
+	}
+}
+
+// release takes conn out of service. It clears the field first, so writers
+// get ErrNotConnected rather than a closing connection, then closes conn and
+// reports the disconnect. cause is nil for a requested close.
+func (c *Client) release(conn *websocket.Conn, cause error) {
+	c.mu.Lock()
+	if c.conn == conn {
+		c.conn = nil
+		c.reconnectPending, c.reconnectCause = false, nil
+	}
+	c.mu.Unlock()
+	closeConn(conn)
+	c.s.onDisconnect(cause)
+}
+
+// closeConn sends a close frame, waits for the peer's reply only when that
+// frame went out, and closes the socket regardless, so a broken connection
+// cannot leak its fd. Only the goroutine that was reading conn calls it, so
+// the drain is never a second concurrent reader.
+//
+// After a failed read, including one freed early by Reconnect or Shutdown,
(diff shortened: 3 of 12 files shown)


pkg/wsclient/client.go
```diff
@@ -0,0 +1,427 @@
...
+// in which case it reports that and the cause instead. Checking and arming
+// under one lock is what keeps a request from landing between the two and
+// being overwritten by a fresh ReadTimeout.
+func (c *Client) armRead(ctx context.Context, conn *websocket.Conn) (cause error, requested bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if cause, requested := c.closeRequestLocked(ctx); requested {
+		return cause, true
+	}
+	_ = conn.SetReadDeadline(time.Now().Add(c.s.readTimeout))
```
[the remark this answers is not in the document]
geunwoonoh: Valid, and worse than "the timeout is not armed": every mechanism that frees a parked read sets that same deadline, so `Reconnect`, `Shutdown` and context cancellation would all be unable to end it. `Run` and `Done` block for good. Reachable because `Config.Dialer` invites a caller-supplied `net.Conn`, which need not implement deadlines.

Fixed in 027fd80, both halves as suggested:

- `armRead` now returns the error as the disconnect cause, so the connection is given up instead of entering a read nothing can end, and the redial is paced like any other failure.
- The interrupt paths (`interrupt`, `Reconnect`, `abandon`) go through a new `freeRead`, which falls back to `conn.Close()` when the deadline is refused. That covers a refusal starting mid-connection, where the read is already parked on a deadline that was accepted once; without it the loop would sit there until `ReadTimeout`.

Two tests, each verified to fail against the unfixed code: `TestClient_GivesUpAConnectionWhoseDeadlineWontArm` (refused from the first arm) and `TestClient_ShutdownClosesAConnThatWontTakeADeadline` (accepted once, refused after, with a 30s `ReadTimeout` so only the `Close` fallback can end it).

pkg/wsclient/config.go
```diff
@@ -0,0 +1,297 @@
...
+		userAgent = "alpamon/" + version.Version
+	}
+	header.Set("User-Agent", userAgent)
+
+	var dialer *websocket.Dialer
+	if c.Dialer == nil {
+		dialer = DefaultDialer()
+	} else {
+		copied := *c.Dialer // later edits to the caller's dialer must not reach a running client
+		dialer = &copied
```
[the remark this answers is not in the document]
geunwoonoh: Right on both fields. The TLS half landed in 726386c before this thread was re-read (`TLSClientConfig` is cloned now, since gorilla re-reads it on every dial, so a caller flipping `InsecureSkipVerify` on a shared `tls.Config` would have disabled verification on the next reconnect).

`Subprotocols` was still aliased, and is cloned in 027fd80. `TestResolve_DoesNotAliasCallerState` now mutates `dialer.Subprotocols[0]` after `New` and asserts the resolved copy is unchanged; it fails without the clone.

The other reference fields are deliberately left shared: `Jar` and `WriteBufferPool` are stateful by design and meant to be shared, and `Proxy`/`NetDial*` are function values. The comment no longer claims more than the copy delivers.

pkg/wsclient/client.go
```diff
@@ -0,0 +1,480 @@
...
+}
+
+// freeRead unblocks a read parked on conn. Pushing the deadline into the
+// past is the ordinary way, but a caller-supplied net.Conn may refuse a
+// deadline, and then closing the socket is the only thing left that ends the
+// read. Both are safe to call while another goroutine is reading.
+func freeRead(conn *websocket.Conn) {
+	if err := conn.SetReadDeadline(aLongTimeAgo); err != nil {
+		_ = conn.Close()
```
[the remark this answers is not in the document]
geunwoonoh: Valid, fixed in d19bde9. You are right about the contract: gorilla lists `SetReadDeadline` among the read methods only one goroutine may call. It was safe in practice only because v1.5.3 forwards straight to the `net.Conn` (`return c.conn.SetReadDeadline(t)`), which is why `-race` never fired, but that is an implementation detail and not something a package pinned for the life of v2 should lean on.

`freeRead` now calls `conn.UnderlyingConn().SetReadDeadline(aLongTimeAgo)`. That rests on the `net.Conn` guarantee instead ("Multiple goroutines may invoke methods on a Conn simultaneously"), which is the same idiom `net/http` uses to abort a blocked read. The `Close` fallback stays on the gorilla conn, since its `Close` is documented as safe to call concurrently. The read loop arms its own deadline on the gorilla conn from the reader goroutine, which is inside the contract, so that path is unchanged.

There is no behavioral test for this one: with v1.5.3 both calls reach the same socket, so no test can tell them apart. The race suite still passes, 5 times over.

pkg/wsclient/client.go
```diff
@@ -0,0 +1,480 @@
...
+		messageType, payload, err := conn.ReadMessage()
+		if err != nil {
+			cause, ending := c.closeRequest(ctx)
+			// Freeing a read produces a timeout and nothing else, so any
+			// other error is the connection's own ending, even if a request
+			// happened to be pending: report that rather than the request.
+			var timeout net.Error
+			if ending && errors.As(err, &timeout) && timeout.Timeout() {
+				err = cause
+			}
```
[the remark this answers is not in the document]
geunwoonoh: Valid, and it was introduced by the previous commit's `Close` fallback. Fixed in d19bde9.

The substitution is now `ending && wokenOnPurpose(err)`, and `wokenOnPurpose` accepts a timeout or `net.ErrClosed`. Only this package closes the socket while the loop is reading, so `net.ErrClosed` mid-read can only be our own wakeup. A peer's `*websocket.CloseError` is neither, so it still outranks a pending request, as you asked.

The consequence was worse than the wrong `OnDisconnect` value: with a non-nil cause, `Run` treated a `Reconnect` as a failure and paced it by `MinBackoff` instead of redialing at once. Tests, each verified to fail against the unfixed code:
- `TestClient_ReconnectOnAConnThatWontTakeADeadlineIsStillARequest` sets a one-hour `MinBackoff`, so any paced wait would hang it.
- `TestClient_ShutdownClosesAConnThatWontTakeADeadline` now also asserts `OnDisconnect(nil)`.

I also confirmed that widening the rule to accept everything fails `TestClient_PrefersThePeersReasonOverAPendingRequest`, so the peer-error precedence is pinned.

pkg/wsclient/client.go
```diff
@@ -0,0 +1,480 @@
...
+		websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""),
+		time.Now().Add(closeTimeout))
+	if err == nil {
+		_ = conn.SetReadDeadline(time.Now().Add(drainTimeout))
+		for {
+			if _, _, err := conn.NextReader(); err != nil {
+				break
+			}
+		}
+	}
```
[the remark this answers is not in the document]
geunwoonoh: Valid, fixed in d19bde9: the drain only runs when its deadline can be armed.

    if err == nil && conn.SetReadDeadline(time.Now().Add(drainTimeout)) == nil {

This is the same gap as the `armRead` fix one layer down. A refused deadline plus a peer that never answers the close frame left `NextReader` blocking forever, and `Close`, the part that matters, comes after it. `TestClient_DrainGivesUpOnAConnThatWontTakeADeadline` pairs a deadline-refusing conn with a server that accepts and then never answers the close frame. It fails without the fix and passes with it, because the client gets as far as redialing.

pkg/wsclient/config.go
```diff
@@ -0,0 +1,326 @@
...
+	ReadLimit int64
+
+	// ReadTimeout is re-armed before every read. Zero means DefaultReadTimeout.
+	ReadTimeout time.Duration
+
+	// WriteTimeout bounds every write. Zero means DefaultWriteTimeout.
+	WriteTimeout time.Duration
+
+	// MinBackoff and MaxBackoff bound the wait between reconnect attempts.
+	// Each wait is the doubling base times a random factor in [0.5, 1.5),
```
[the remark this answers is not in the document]
geunwoonoh: Right, stale doc. When the factor moved to `[1.0, 1.5)` I updated `backoff.go` and missed `Config`. Fixed in d19bde9, and the doc now also says why the range starts at 1.0: `MinBackoff` is a true floor. Below 1.0 the clamp swallowed the whole lower half, so half of a fleet's first retry landed on exactly `MinBackoff` (49.8% of 10,000 draws), in the one scenario jitter exists for. I grepped for other stale mentions and found none.

pkg/wsclient/client.go
```diff
@@ -0,0 +1,530 @@
...
+			return handlerErr
+		}
+		// A close with no failure behind it is one the caller asked for, and
+		// it redials at once. Everything else is a failure, and only a
+		// connection that lasted earns the same treatment.
+		if ended.cause == nil || ended.proven {
+			b.reset()
+			attempt, wait, cause = 0, false, nil
+			continue
+		}
```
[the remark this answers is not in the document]
geunwoonoh: This one is working as documented, and the storm it describes is bounded. Not changing it, but here is the reasoning.

The fast path needs the connection to have lasted `MinBackoff`. `proven` is `time.Since(opened) >= c.s.minBackoff`, taken before release (client.go:309-315), so a redial is immediate only when the connection it replaces stayed up that long. The dial rate is therefore capped at one per `MinBackoff` per agent, five seconds by default, whatever ended it. A broken write path cannot beat that cap: if the caller writes early, as a heartbeat-on-connect does, the connection dies short of `MinBackoff`, is unproven, and gets the full jittered backoff, which is what `TestClient_PacesARedialAfterAFailedWrite` pins.

That cap is the same one every other failure mode has, including the peer that drops at `MinBackoff + 1ms` on purpose. Singling out write failures would not lower it.

On the documented rule: `Run` says a wait follows anything that ends a connection **short of MinBackoff**, "whether the dial was refused, the peer hung up, or a write failed", and that only a connection which lasted at least that long, or a `Reconnect`, redials at once (client.go:94-101). A failed write is paced exactly like any other failure. It is proof-of-uptime, not the kind of failure, that decides.

The trade in the other direction is real: reserving the fast path for peer drops would add five seconds of downtime to every transient write error on an endpoint that has just shown it works. For an agent whose value is being reachable, faster recovery is worth more than a cap that does not move.

Two things this review round did change, both raised in the same review body: the pacing tests asserted only the delay `OnRetry` reported and never that the redial waited for it, so replacing `sleep` with a `continue` left them green. They now measure from the retry report to the next upgrade, and that mutation fails them. And `reconnectCause`'s comment overstated itself; it now says the cause is reported only when the read ended on our own wakeup. Both in c75b950.

pkg/wsclient/dial.go
```diff
@@ -0,0 +1,242 @@
...
+	// gorilla/websocket v1.5.3 reads the reason phrase out of a proxy's
+	// CONNECT status line without checking there is one (proxy.go), so a
+	// proxy answering "HTTP/1.1 407" panics whichever goroutine dialed. For
+	// an agent that goroutine is the one keeping the process alive, and a
+	// proxy is reachable by default through the environment, so turn it into
+	// the error it should have been.
+	defer func() {
+		if r := recover(); r != nil {
+			conn, resp, err = nil, nil, fmt.Errorf("wsclient: dial panicked, most likely on a malformed proxy response: %v", r)
+		}
```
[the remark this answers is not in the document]
geunwoonoh: No leak: gorilla closes the socket itself, on the line before the one that panics.

`httpProxyDialer.Dial` (proxy.go:71-75 in v1.5.3):

```go
if resp.StatusCode != 200 {
    conn.Close()                              // :72
    f := strings.SplitN(resp.Status, " ", 2)  // :73  ["HTTP/1.1", "407"] -> len 1 for "HTTP/1.1 407"
    return nil, errors.New(f[1])              // :74  panics here
}
```

`conn` there is what `forwardDial` returned, which is the socket `abortState.track` recorded, so it is already closed by the time the panic reaches `dial`'s `recover`. `finish` then calls `SetDeadline` on a closed socket, gets `net.ErrClosed`, and discards it, which is why it only has to clear deadlines.

The same holds on every other exit: a dial hook that fails tracks nothing; a failed handshake is closed by gorilla's own `defer func() { if netConn != nil { netConn.Close() } }()`; a CONNECT rejection that *does* carry a reason phrase takes the same `conn.Close()` at :72 and returns an ordinary error; and a handshake that succeeds hands the socket to the caller.

You are right that nothing here proved it, though, and this is the third round the question has come back. 8763bf3 dials `TestDial_SurvivesAProxyWithoutAReasonPhrase` through `closeTrackingDialer` and asserts `allClosed()`, which also proves a socket was opened at all rather than passing on an empty list. Neutering `closeTrackingConn.Close` fails it, so the assertion is load-bearing: if a future gorilla moves that `conn.Close()` after the `SplitN`, this test goes red instead of the fd count going up.

pkg/wsclient/config_test.go
```diff
@@ -0,0 +1,303 @@
...
+		})
+	}
+}
+
+// TestResolve_AcceptsASendableHost keeps the Host check to what net/http
+// really refuses: an internationalized name it can punycode, a port, and a
+// value with a line break in it, which net/http neutralizes rather than
+// refuses, all dial fine.
+func TestResolve_AcceptsASendableHost(t *testing.T) {
+	for _, host := range []string{"backhaul.example.com", "bücher.example", "bücher.example:8443", "good.example\r\nX: 1"} {
```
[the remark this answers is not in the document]
geunwoonoh: The stated reason is wrong, but the conclusion is right, so this is fixed in 068e864.

`net/http` does not reject a Host containing CR/LF from `Request.Write`. It zeroes the field and returns nil (request.go:620-639):

```go
if !httpguts.ValidHostHeader(host) {
    // ... Instead, zero out the Host header entirely if it isn't valid.
    if !usingProxy {
        host = ""
    } else {
        return errors.New("http: invalid Host header")
    }
}
```

So the assertion did not make the test fail. It passed locally and on all 17 Build and Test jobs; you can see `TestResolve_AcceptsASendableHost/good.example__X:_1` passing in every one of them.

What it did do is worse than a failing test. Serializing that value gives:

```
Host="good.example\r\nX: 1"  Write err: <nil>
wire: "GET / HTTP/1.1\r\nHost: \r\nUser-Agent: ...\r\n\r\n"
```

The whole Host is gone. There is no header injection, which is what I checked when this came up in an earlier round, and I stopped there. But a handshake that goes out with an empty Host is a 400 from all but any server, so `Run` would have retried that config error for as long as it ran, which is precisely what the probe was added to prevent. `bad host`, `host/path` and `host\x00x` all behave the same way, and none of them were caught either.

The fix keeps net/http as the oracle rather than reimplementing `ValidHostHeader`, which the leaf dependency set would not allow anyway. `Write` and `WriteProxy` differ in exactly this case -- one zeroes the field, the other returns `http: invalid Host header` -- so `resolveDial` now probes with `WriteProxy`. Nothing is proxied; it is used only because it is the half that answers.

Checked in both directions before changing it: `backhaul.example.com`, `bücher.example`, `bücher.example:8443`, `xn--bcher-kva.example`, `[::1]:8443` and `[fe80::1%25en0]:443` all still resolve, and the four unsendable shapes now fail in `New`. `TestResolve_AcceptsASendableHost` gained the ASCII-punycode and IPv6-zone cases, the line-break value moved to `TestResolve_Rejects` alongside the space and slash ones, and putting `Write` back fails those three.

This narrows what `New` accepts. That is a contract change, but only over configurations whose every dial already failed, and the package has no in-repo caller and no tag yet, so now is the time.

pkg/wsclient/dial.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+
+	base := dialer.NetDialContext
+	switch {
+	case base != nil:
+	case dialer.NetDial != nil:
+		netDial := dialer.NetDial
+		base = func(_ context.Context, network, addr string) (net.Conn, error) { return netDial(network, addr) }
+	default:
```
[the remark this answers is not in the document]
geunwoonoh: The property is real, the attribution is not, and the documentation was wrong about it. Fixed as documentation in ad32719; the dial path is unchanged, deliberately.

`NetDial` is context-blind in gorilla too. Without this wrapper, gorilla calls it the same way:

```go
} else if d.NetDial != nil {
    netDial = d.NetDial
}
```

and its `HandshakeTimeout` reaches the socket only as a deadline set *after* the hook returns, so a `NetDial` that blocks blocks the dial there as well. The adapter does not make the hook context-blind; it was. What the adapter adds is that the socket such a hook returns is now tracked and abortable, including when it arrives after the abort has already fired -- `state.track` force-aborts it -- which is the second half of what you suggested, already in place.

So the residual is: a caller callback that never returns holds `Run`. That is not specific to `NetDial`. `Handler`, `OnConnect`, `OnDisconnect`, `OnRetry` and a `NetDialContext` that ignores its context all run on `Run`'s goroutine and all do the same thing.

On the two remedies:

- Rejecting `NetDial` in `New` would refuse `(&net.Dialer{Timeout: 5 * time.Second}).Dial`, which is correct code, and would not help against a `NetDialContext` that ignores its context. It buys the guarantee back only for callers who were not the problem.
- Running the hook on its own goroutine and abandoning it on cancellation lets `Run` return, but the hook keeps running, so the goroutine leaks and the socket it may eventually open is one nobody is holding. For a process that is terminating that is close to free; for the reconnect loop, which does this on every dial, it is a leak with a trigger the package cannot see. Trading a caller bug the caller can fix for a library leak it cannot is the wrong way round.

What was actually wrong is that neither doc said any of this. `Dialer` claimed a caller's dialer is "used as given except for three things" while quietly wrapping the dial hooks and rewriting `NetDial` into a `NetDialContext` -- a fourth thing, and the one your concern lives in. `Run` promised to return on `Shutdown` or a done context without saying whose code that depends on. Both now say it, `Dialer` names `NetDialContext` as the one to prefer and suggests giving a `NetDial` a timeout of its own, and `Run` names the dial hook as the one to watch, because it can be waiting on a socket that does not exist yet and so cannot be freed.

The suppressed `config.go:179` comment in the same review was a different matter and is fixed as code in the same commit; details in the PR comment.

pkg/wsclient/config.go
```diff
@@ -0,0 +1,393 @@
...
+	// and treat what it will not write as the config error it is.
+	for name, values := range header {
+		if name == "Host" {
+			continue // sent as the request's Host, and probed as one below
+		}
+		for _, v := range values {
+			if strings.ContainsAny(v, "\r\n") {
+				return dialSettings{}, fmt.Errorf("wsclient: header %s has a value with a line break, which net/http rewrites rather than sends", name)
+			}
+		}
```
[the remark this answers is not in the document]
geunwoonoh: `net/http` does not reject them. Swept over every byte value, `Request.Write` returns nil for all 256 and rewrites exactly two:

```
byte  10 (0xa): rewritten -> "X-Custom: a b"
byte  13 (0xd): rewritten -> "X-Custom: a b"
bytes that error or get rewritten: 2 of 256
```

NUL goes on the wire verbatim, as does every other control byte. So there is no case where a value passes `New` and then fails serialization, and no permanent config error hiding behind one.

That is also why the check is split the way it is rather than by carelessness. `net/http` has two ways of not sending what it was given, and they do not fall on the same axis:

- a **name** it will not write is **dropped**, and dropping is decided by the name alone -- which is why the probe can carry a literal value and still answer the only question being asked of it.
- a **value** is never dropped and never refused; two bytes are **rewritten**, and those two are checked directly.

A byte outside that set reaches the server exactly as the caller wrote it. Refusing it would be this package deciding the caller may not send a header that `net/http` is willing to send and the server may well accept, which is a different policy from the one the check states: refuse what will not go out as given.

You are right that nothing proved the boundary, and this is the second round it has come up, so `TestResolve_RefusesTheHeaderValueBytesNetHTTPRewrites` (ecf5d76) now runs that sweep as a test -- all 256 values through `net/http` and through `resolve`, asserting the two agree on every one. If a future Go widens the set, or starts returning an error, it fails there instead of in a handshake nobody can account for.

The three suppressed comments in the same review were a different matter: two were real and are fixed in ecf5d76, with details in the PR comment.

pkg/wsclient/client.go
```diff
@@ -0,0 +1,544 @@
...
+		time.Now().Add(closeTimeout))
+	// Only drain when the drain can be bounded: a conn that refuses the
+	// deadline would leave NextReader waiting on a silent peer forever, and
+	// Close, which is what actually matters, would never be reached.
+	if err == nil && conn.SetReadDeadline(time.Now().Add(drainTimeout)) == nil {
+		for {
+			if _, _, err := conn.NextReader(); err != nil {
+				break
+			}
+		}
```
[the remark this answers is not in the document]
geunwoonoh: The loop is bounded, by the one property the description leaves out: `SetReadDeadline` takes an **absolute** instant, not a per-read budget. It is not re-armed inside the loop, so once `time.Now().Add(drainTimeout)` passes, the next read off the socket fails whether or not the peer is mid-sentence. A peer that keeps handing over complete frames does keep `NextReader` from blocking, but it cannot hold the loop past that instant: refilling gorilla's 4 KiB reader is itself a socket read, and it is the read that the deadline stops.

Measured rather than argued. b4d069c adds a peer that answers nothing and writes frames in a tight loop for as long as the socket lives, and `TestClient_TheCloseDrainEndsOnAPeerThatKeepsTalking` asserts both halves: the drain ran for its full second, and it ended there. It comes out at **1.00s**, every run. Widening the deadline to a minute fails it.

Getting that test to exercise the drain at all took one more thing, which is worth recording because the first attempt passed in 0.00s and proved nothing. The drain only runs for a connection still healthy when the close was decided. `Shutdown` normally frees a parked read by pushing the deadline into the past, and gorilla stores that failure and returns it to every later read -- so the drain ends on the stored error, not on anything under test. The test now parks `Run` inside the handler first, so `armRead` sees the request with no read in flight and the connection reaches `closeConn` healthy.

So there is no unbounded path here, and no independent timer to add: an absolute deadline already is one. What was missing was anything saying so, which is now a test rather than a claim in a comment.

pkg/wsclient/config.go
```diff
@@ -0,0 +1,393 @@
...
+	// an edit to an entry after New would change later handshakes.
+	dialer.Subprotocols = slices.Clone(dialer.Subprotocols)
+	if dialer.TLSClientConfig != nil {
+		// Shared, this is the one field whose later mutation would change
+		// how an already-running client verifies certificates: gorilla reads
+		// it again on every dial, so a caller flipping InsecureSkipVerify on
+		// a tls.Config it also uses elsewhere would silently disable
+		// verification on the next reconnect.
+		dialer.TLSClientConfig = dialer.TLSClientConfig.Clone()
+	}
```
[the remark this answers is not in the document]
geunwoonoh: Correct on the facts, and the field said more than `Clone` buys. Fixed as documentation in b4d069c; the clone is deliberately left one level deep.

`tls.Config.Clone` settles the fields the struct holds itself and copies the rest as the pointers they are. `RootCAs` and `ClientCAs` are the ones you name, and `Certificates` and the verify callbacks are in the same position. So `Dialer` saying "later edits to them cannot change how a running client verifies certificates" was wrong: it holds for `InsecureSkipVerify`, `MinVersion`, `ServerName` and the rest of the value fields, and not for anything reached through a pointer.

Not cloning the pools, though, and the reason is the case your remedy would break. A caller who rotates trust by adding to a pool it also handed us is doing something deliberate and reasonable, and snapshotting it would stop that working with no diagnostic at all -- reconnects would simply start failing to verify some time after a rotation that looked fine. Between "a mutation you meant reaches the client" and "a mutation you meant is silently ignored", the first is the better failure, and it is the one a caller can reason about.

A deep clone could not be complete anyway. `GetClientCertificate`, `GetCertificate` and `VerifyPeerCertificate` are function values; whatever they close over stays the caller's no matter what this package copies. So the honest line is not "deeper" but "one level, and named": value fields are settled at `New`, pointer-reachable state is shared, and editing it is editing a running client rather than a copy of its config. `Dialer` and the clone site both say that now.

pkg/wsclient/backoff.go
```diff
@@ -0,0 +1,62 @@
...
+// [1.0, 1.5), clamped to [initial, max]. Without the random factor every
+// agent reconnecting after the same event, such as a backhaul restart, would
+// retry in lockstep.
+//
+// The factor starts at 1.0 rather than 0.5, which is where internal/retry
+// starts it, because the clamp to initial would otherwise swallow the whole
+// lower half: at the first attempt the base is initial, so every draw below
+// 1.0 returns exactly initial and half of a fleet retries on the same tick.
+// Spreading upward keeps initial a real floor and the distribution intact.
+func (b *backoff) next() time.Duration {
```
jisung-02, line 45: Critical: the jitter disappears entirely once the base reaches `MaxBackoff`, which is exactly the sustained-outage case it exists for.

The factor is drawn from [1.0, 1.5), so `jittered >= b.current` always holds. Once `b.current == b.max`, `!(jittered < float64(b.max))` is always true and `next()` returns exactly `b.max` on every later attempt. With `MinBackoff == MaxBackoff`, a pairing `resolve()` accepts, it returns a flat value from the very first attempt.

Transcribing `next()` and running it, with the defaults:

```
attempt 1: 5.149693441s
attempt 2: 11.587691326s
attempt 3: 22.503312631s
attempt 4: 40.886407875s
attempt 5: 1m0s
attempt 6: 1m0s
attempt 7: 1m0s
attempt 8: 1m0s
```

A backhaul restart is precisely when every agent's schedule has already reached the ceiling, so the whole fleet then redials on the same tick — the lockstep this file's own comment says the jitter prevents. `internal/retry` does not have this property: its factor starts at 0.5, so draws below the ceiling survive the clamp.

Neither test covers it. `TestRun_BackoffScheduleUnderAFakeClock` pins `Rand` to 0, and at factor 1.0 a correct implementation and this one agree; `TestRun_JitterDesynchronizesClients` samples only the first wait; `backoff_test.go` never sets `MinBackoff == MaxBackoff`.

One way out, if the floor behavior is to stay as documented, is to cap the base low enough that the jitter still fits underneath the ceiling:

```suggestion
	case b.current > b.max*2/3: // the jitter must still fit under max
		b.current = b.max * 2 / 3
```

That keeps `MinBackoff` a true floor and spreads the ceiling across [2/3·max, max). Whichever shape you pick, the test that pins it draws several samples at the ceiling and asserts they differ.

pkg/wsclient/conn.go
```diff
@@ -0,0 +1,292 @@
...
+// socket is stored as fatal and is not. A locally declared interface, not
+// net.Error, because that interface's Temporary method is deprecated.
+func temporary(err error) bool {
+	var t interface{ Temporary() bool }
+	return errors.As(err, &t) && t.Temporary()
+}
+
+// wokenOnPurpose reports whether a failed read is one this package caused
+// to free it, as opposed to one the connection failed on by itself.
+func wokenOnPurpose(err error) bool {
```
jisung-02, line 202: `wokenOnPurpose` separates "a read this package freed" from "a connection that ended on its own" by the error's shape alone, and an organic `ReadTimeout` has the same shape as a freed one: a `net.Error` whose `Timeout()` is true.

Interleaving: `ReadMessage` returns a genuine timeout at conn.go:108. Before the goroutine reaches `closeRequest` at conn.go:110, a concurrent `WriteMessage` fails and calls `abandon(conn, errW)`, which stores `reconnectCause = errW`. `closeRequest` then reports `ending=true, cause=errW`, `wokenOnPurpose(err)` is true because the real timeout is also a timeout, and conn.go:116 substitutes it. `OnDisconnect` is told the write killed a connection the read timeout killed.

Pacing is unaffected, since both causes are non-nil — the cost is that the reported cause can name the wrong failure. Not a data race, so `-race` will not find it; it needs a generation token or a distinct sentinel from the paths that free a read, rather than sniffing the error's shape. The design note at conn.go:111-115 currently does not cover this case.

pkg/wsclient/client.go
```diff
@@ -0,0 +1,278 @@
...
+// message to a Handler.
+//
+// The Client owns its connection. Run is the only goroutine that reads it,
+// writes are serialized, and Reconnect and Shutdown only signal the read
+// loop, so gorilla/websocket's one-reader, one-writer rule holds without the
+// caller's help. All methods are safe for concurrent use.
+type Client struct {
+	s settings
+
+	// mu guards conn and the pending-reconnect pair. Nothing slower than
```
jisung-02: This promises that "nothing slower than SetReadDeadline happens while it is held," and one path does not keep that promise.

`Reconnect` (client.go:250-259), `abandon` (conn.go:231-240) and `interrupt` (conn.go:245-251) all call `freeRead` with `mu` held, and `freeRead` falls back to `conn.Close()` whenever the socket refuses a deadline — a case the package deliberately supports (conn.go:173-176). A caller-supplied `net.Conn` with a slow `Close` holds `mu` for that long, and `WriteMessage`, `armRead`, `rearm` and `Connected` all queue behind it.

Either move the fallback out from under the lock, or narrow the comment to what the code actually guarantees.

pkg/wsclient/config.go
```diff
@@ -0,0 +1,404 @@
...
+	// hook has to return that socket first: NetDialContext is handed a
+	// context that ends with the client's, but NetDial takes none and is
+	// adapted to the wrapper, so nothing can interrupt it before it returns.
+	// Prefer NetDialContext, and give a NetDial a timeout of its own.
+	Dialer *websocket.Dialer
+
+	// ReadLimit is the largest inbound message accepted, in bytes: its
+	// frames' payloads summed as they arrive, before any decompression. Zero
+	// means DefaultReadLimit; a negative value removes the limit.
+	ReadLimit int64
```
jisung-02, line 102: The doc above is accurate that the limit counts payload bytes "before any decompression," but it does not say what that means for a caller who sets `EnableCompression` on their own dialer.

gorilla/websocket v1.5.3 counts only the compressed frame bytes against the limit (`conn.go:924`) and then calls `io.ReadAll` on the decompression reader (`conn.go:1091`), so a small frame can inflate without bound. The default path is safe — `DefaultDialer()` never enables compression and `dial.go:76` rejects an extension the client did not offer — which is exactly why a caller who opts in deserves one sentence of warning here or on the `Dialer` field.

pkg/wsclient/client.go
```diff
@@ -0,0 +1,278 @@
...
+// the next message is not read until it returns: hand long work to another
+// goroutine. ctx is done once Run starts stopping. A non-nil error stops Run,
+// which returns it; call Shutdown for an ordinary stop.
+type Handler func(ctx context.Context, messageType int, payload []byte) error
+
+// Client keeps one WebSocket connection up: it dials, reconnects with
+// jittered backoff whenever the connection drops, and hands every inbound
+// message to a Handler.
+//
+// The Client owns its connection. Run is the only goroutine that reads it,
```
jisung-02: This restates doc.go:18-21's Ownership section almost verbatim, and conn.go's file header already names that section as the canonical statement ("The Ownership section of the package doc is the summary"). Keeping one copy and leaving the caller-facing consequence here would say the same thing once.

Two more of the same kind elsewhere: the inline note at config.go:231-240 re-derives the `Dialer` field doc at config.go:73-83, and "a Client retries a config error forever" appears five times in config.go (145, 223, 252, 262, 290) where one statement in `resolveDial`'s doc would cover every check.

pkg/wsclient/dial.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	s.aborted = true
+	for _, c := range s.opened {
+		_ = forceAbort(c)
+	}
+}
+
+// setDeadline installs t on c, unless the dial has been aborted, in which
+// case the abort's deadline is the one that stands.
+func (s *abortState) setDeadline(c net.Conn, t time.Time) error {
```
jisung-02, line 202: The doc above is the three-line body translated into prose: `if s.aborted { return forceAbort(c) }; return c.SetDeadline(t)`. A reader of the function learns nothing new from it.

Same pattern in a few other places, so it is worth one pass rather than one fix: `track` (dial.go:174) narrates its two statements, `sleep` (client.go:173) expands its signature while the real explanation sits inline below it, `reset` (backoff.go:59) describes `b.current = 0`, and `closeConn` (conn.go:267-269) lists its own three calls in order before reaching the sentence that matters ("so a broken connection cannot leak its fd").

Three comments also explain that a `select` picks among ready cases at random (client.go:179-181, client_test.go:987-988, client_test.go:998-999) — that is a Go language rule the reader already has. The non-obvious half, why the code re-checks instead of trusting the timer, is the part worth keeping.

pkg/wsclient/dial.go
```diff
@@ -0,0 +1,249 @@
...
+// it. It suits a connection with its own lifecycle, such as a tunnel data
+// plane wrapped in tunnel.WebSocketConn for smux.
+//
+// Dial uses URL, ID, Key, Origin, UserAgent, Header, Dialer and ReadLimit
+// from cfg, and ignores the timeout, backoff and hook fields. When the server
+// rejects the handshake, the error wraps websocket.ErrBadHandshake and the
+// response is returned alongside it, with the Authorization header stripped
+// from the request hanging off it so that logging the response cannot spill
+// the credential.
+func Dial(ctx context.Context, cfg Config) (*websocket.Conn, *http.Response, error) {
```
jisung-02, line 26: `Dial` is exported with no caller anywhere in the repository, and doc.go:33-35 says the `pkg/tunnel` consumer it exists for "is not reached from here."

The whole package has no in-repo caller yet, so that is not proof on its own. But the branch name says this goes out as a tag, and a tag freezes the public surface for every module that pins it — worth deciding deliberately rather than by default. `Client.Run` already uses the unexported `dialSettings.dial`, so nothing here depends on it being public.

pkg/wsclient/config.go
```diff
@@ -0,0 +1,404 @@
...
+	// MinBackoff and MaxBackoff bound the wait between reconnect attempts.
+	// Each wait is the doubling base times a random factor in [1.0, 1.5),
+	// clamped to this range, so MinBackoff is a true floor. Zero means
+	// DefaultMinBackoff and DefaultMaxBackoff.
+	MinBackoff time.Duration
+	MaxBackoff time.Duration
+
+	// Rand returns the value in [0, 1) the jitter factor is drawn from. Nil
+	// means math/rand/v2. Tests pin it for a deterministic schedule.
+	Rand func() float64
```
jisung-02, line 131: Every assignment to `Rand` in this PR is a test pinning determinism (client_test.go:970, 1563, 1608, 1681, 1708 and reconnect_test.go:49), and those tests are `package wsclient`, so they can already reach `c.s.rand` without an exported field.

The counterargument is real: a module outside this repository cannot reach it, and deterministic backoff in its own tests is a fair thing to want — this package exists for exactly those modules. Either way the field's doc currently reads as if it were for this package's tests, which is the one audience that does not need it.

pkg/wsclient/config.go
```diff
@@ -0,0 +1,404 @@
...
+)
+
+// Defaults applied to zero-valued Config fields.
+const (
+	// DefaultHandshakeTimeout bounds the opening handshake in DefaultDialer.
+	DefaultHandshakeTimeout = 30 * time.Second
+
+	// DefaultReadTimeout is how long a read may wait for a frame before the
+	// connection is treated as dead. The Alpacon backhaul pings well inside it.
+	DefaultReadTimeout = 35 * time.Minute
```
jisung-02, line 24: These numbers now exist twice. `pkg/runner/client.go:28-30` carries the same 35-minute read timeout and the same 5s/60s reconnect bounds, and the `id="%s", key="%s"` header format is written out in both files (config.go:200 and runner/client.go:67). If the backhaul's ping interval moves, one side can be updated and the other forgotten.

Is removing that duplication a planned follow-up? `pkg/runner` is free to import `pkg/wsclient`, so the constants could live here and be used there, and the same question applies to the close sequence (`conn.go:277` against runner's `Close`/`drainCloseReply`) and to the two retry schedules, which have already diverged: `internal/retry/retry.go:68` draws its factor from [0.5, 1.5) and `backoff.go` from [1.0, 1.5). Whether that convergence is next, later, or deliberately out of scope, one sentence in `doc.go` would tell the next reader how to treat the two copies.

.github/workflows/build-and-test.yml
```diff
@@ -29,6 +29,163 @@ jobs:
...
+          # from here would claim that key for a near-empty cache and leave the
+          # jobs that need it with nothing to restore.
+          cache: false
+
+      - name: Check the dependency graph
+        env:
+          PKG: github.com/alpacax/alpamon/v2/pkg/wsclient
+          # The only packages pkg/wsclient may reach outside the standard
+          # library. pkg/tunnel and pkg/version are leaves themselves.
+          ALLOWED: >-
```
jisung-02, line 74: The allowed-import list is maintained by hand here and again in prose at doc.go:27-30. They agree today — both name the standard library, gorilla/websocket, `pkg/tunnel` and `pkg/version` — but nothing fails when only one of them is updated.

Having this job's failure message point at doc.go, or doc.go point at this job as the authority, keeps the two from drifting apart in silence.

Running the guard's logic locally on three of the six targets, the graph today is `pkg/version`, `gorilla/websocket` and the package itself — `pkg/tunnel` is allowed but not reached, which doc.go already says ("That list is the ceiling, not an inventory").

jisung-02 (changes requested):
## Review summary

A new leaf package, `pkg/wsclient`, holds one WebSocket connection to the Alpacon backhaul open across reconnects: `Client` owns the connection and never exposes it, `Run` is the only reader, writes are serialized, and `Shutdown`/`Reconnect` only signal the read loop. `Dial` is a second entry point that hands one connection to a caller outright. Everything arrives through `Config`, so nothing reads `config.GlobalSettings`, and connects, drops and retries are reported through hooks instead of a logger. Two CI jobs back the boundary: `leaf-guard` walks `go list -deps` over six build targets and fails if the graph reaches anything but the standard library, gorilla/websocket, `pkg/tunnel` and `pkg/version`; `race-wsclient` runs the package under `-race`.

Reviewed at `3f6beda`, stack golang, +4231 -0 across 13 files.

### Critical (1)

The reconnect jitter disappears once the backoff reaches its ceiling, which is the sustained-outage case the jitter exists for. Details inline at `pkg/wsclient/backoff.go`.

### Warnings (4)

- `conn.go:194` — an organic read timeout and a freed read are indistinguishable by error shape, so a concurrent write failure can take credit for a disconnect it did not cause.
- `client.go:42` — the comment on `mu` promises more than the code delivers on the deadline-refusing path.
- `config.go:96` — `ReadLimit` stops bounding memory the moment a caller enables compression, and nothing says so.
- Comments: the same fact is stated in several places, and a group of doc comments narrate the code they sit on. Details inline.

### Suggestions (5)

- `dial.go:26` — `Dial` is exported with no caller; a tag freezes that surface.
- `config.go:115` — `Config.Rand` is set only by this package's own tests.
- `config.go:24` — timeouts, backoff bounds and the Authorization format are now written out in both this package and `pkg/runner/client.go`.
- `.github/workflows/build-and-test.yml:72` — the allowed-import list is maintained by hand in two places.
- Commit `f12f529b`'s body says a connection proves itself by "carrying a frame or lasting MinBackoff"; the code that shipped proves it by duration alone (`conn.go:69`). Worth fixing before the sentence reaches the PR description or the release notes.

### Question

Is there a follow-up planned to remove the duplication this creates against `pkg/runner/client.go`? The read timeout, the backoff bounds, the Authorization header format, and the close-frame-then-drain-then-close sequence now exist in both places, and the retry schedules have diverged: `internal/retry` draws its jitter factor from [0.5, 1.5) and `backoff.go` from [1.0, 1.5). `pkg/runner` is free to import `pkg/wsclient`, so converging is possible in that direction. Knowing whether that is the plan, a later step, or deliberately out of scope would tell the next reader how to treat the two copies — and it is the kind of thing worth one sentence in `doc.go` either way.

### Highlights

- The dial path is hardened against what a hostile backhaul or proxy can do, each with the failure it prevents named: the `recover` at `dial.go:39-49` turns gorilla v1.5.3's panic on a proxy CONNECT response with no reason phrase into an error, `Authorization` is stripped from `resp.Request` before the response is handed back, and an extension the client never offered fails the connection rather than enabling decompression the read limit cannot bound.
- Configuration errors are caught at `New` rather than becoming a dial that fails forever: credentials carrying a quote or a line break, headers the handshake sets itself, and header names and `Host` values that `net/http` would silently drop or rewrite, probed by asking `net/http` to write them.
- The ownership model is enforced where it can be checked. Cross-goroutine wakeups go through `UnderlyingConn()` because `net.Conn` allows concurrent use where gorilla's own read methods do not, and `install`, `armRead` and `rearm` re-check the stop condition under the same lock that arms the deadline, so a request cannot be lost to a re-arm.
- `leaf-guard` is written to fail loudly rather than pass vacuously: it sweeps six OS/arch/cgo combinations because build constraints hide imports from a single target, and it fails when the package is missing from its own dependency list, which is the state in which every other check would pass while testing nothing.

### Checklist

- [ ] Documentation updated if needed

pkg/wsclient/backoff.go
geunwoonoh: Fixed in a75e446, and it measured worse than the transcription suggested. Fifty clients taken past the ceiling produced **one** distinct wait:

```
distinct value: 1개 -> map[1m0s:50]
```

Not a spread that degrades at the ceiling, but a single instant, in the state a backhaul restart finds a fleet in.

Took the shape you suggested. The base now stops at the largest value whose whole `[1.0, 1.5)` window still fits under max, which is two thirds of it:

```go
ceiling := b.max / 3 * 2 // divide first: b.max*2 can overflow
if ceiling < b.initial {
	ceiling = b.initial
}
```

The `initial` guard is for a config whose floor sits above two thirds of its ceiling, where the cap would otherwise move the base backwards. Same fifty clients now draw fifty distinct waits across `[40s, 60s)`.

`MinBackoff == MaxBackoff` still returns that one value, and I left it: the wait has to be in `[30s, 30s]`, so there is nowhere to spread. `next`'s doc now says so rather than leaving it to be discovered.

`TestBackoff_JittersAtTheCeiling` draws fifty samples past the ceiling, asserts each lands in `[2/3 max, max)`, and fails if fewer than 45 are distinct. Putting the base back at `max` fails it on the collapse assertion.

You were right that nothing covered this. Three existing tests did pin the old ceiling, though — `BaseDoublingReachesCeiling`, `ExtremeDrawsClamp` and `ResetRestartsTheSequence`. Rather than rewrite their expectations I moved each one's `MaxBackoff` so that the new two-thirds point lands on the round number they were already asserting (500ms→600ms, and 8s→12s in the fake-clock schedule), which left every expected schedule in them unchanged. `ExtremeDrawsClamp`'s ceiling case is the one real edit: it asserted `~1.5x clamps down to max`, and now asserts the top of the window reaches just under max without touching it, which is the property that replaced it.

pkg/wsclient/conn.go
geunwoonoh: Confirmed, and the interleaving is exactly as you describe. Changed the comment rather than the code, and I want to be explicit that this is a decision and not an oversight.

Telling the two apart needs the free to leave something the read can compare against — a generation counter armed under `mu` in `armRead` and checked after `ReadMessage` returns. That is a counter on every read, on the hot path, to improve the accuracy of a value that only reaches `OnDisconnect`. Both causes are non-nil, so the pacing decision is identical either way; what is wrong is the name in a log line, in a window bounded by the distance between conn.go:108 and conn.go:110.

A distinct sentinel does not work here either: `freeRead` pushes the deadline into the past, and the socket answers with the same `os.ErrDeadlineExceeded` it gives an organic timeout. Only closing produces a distinguishable error, and closing is what the package deliberately avoids when a deadline will do.

The note now carries the case instead of implying it cannot happen:

> An organic ReadTimeout wears the same shape as a freed read, so a write that fails in the window between the read returning and this line takes the credit for a disconnect the timeout caused. Only the reported cause is wrong: both are failures, so the pacing is the same either way. Telling them apart would need the free to carry a token the read could compare, and that is not worth a generation counter on every read.

If you read the trade differently, the counter is a contained change and I will make it.

pkg/wsclient/client.go
geunwoonoh: Fixed the comment. You are right that the code cannot be fixed here instead: `mu` is what makes `armRead` see a request rather than re-arm past it, so moving the `freeRead` fallback out from under it would trade a latency bound for a lost wakeup.

The comment now names the path rather than claiming it does not exist:

> mu guards conn and the pending-reconnect pair. What runs under it is a SetReadDeadline on the socket, except on the path where a caller-supplied conn refuses one and freeRead closes it instead: a slow Close on such a conn holds every other method here behind it. Moving that fallback out from under mu is not an option, since it is the lock that keeps a request from being lost to a re-arm.

pkg/wsclient/config.go
geunwoonoh: Fixed. Checked the two lines you cite and they say what you say: v1.5.3 counts the compressed frame against the limit at conn.go:924 and then hands the decompression reader to `io.ReadAll` at conn.go:1091, so the limit stops bounding the allocation the moment a caller opts in.

The field now carries that rather than only the accurate-but-incomplete "before any decompression":

> It bounds memory only while compression is off, which is the default: gorilla/websocket counts the compressed bytes against the limit and then reads the decompressed stream whole, so a caller who sets EnableCompression on their own Dialer is bounding the frame, not the allocation it expands into. Nothing here can bound the second.

Put it on `ReadLimit` rather than on `Dialer` because that is the field whose promise changes.

pkg/wsclient/client.go
geunwoonoh: Fixed, all three.

`Client`'s doc kept the half a caller acts on and handed the mechanism back to the section that owns it:

> All methods are safe for concurrent use, and the connection is never exposed, so a caller cannot break gorilla/websocket's one-reader, one-writer rule by accident. The Ownership section of the package doc says how that is arranged; conn.go holds the arrangement.

The inline note at config.go that re-derived the `Dialer` field doc is gone, and "a Client retries a config error forever" now appears once, in `resolveDial`, instead of at each check.

pkg/wsclient/dial.go
geunwoonoh: Fixed, and took the pass rather than the one fix.

`setDeadline` and `track` now say why they hold the lock rather than what the three lines under them do:

```go
// setDeadline installs t on c under the same lock the abort takes, so a
// deadline already on its way down cannot land after an abort and undo it.

// track records a socket the dial hook just opened, aborting it on the spot
// if the abort already went past.
```

`reset` says what calls it and why rather than restating `b.current = 0`. `closeConn` leads with the sentence that matters — it closes the socket whatever else fails, so a broken connection cannot leak its fd — and mentions the frame and the drain after. `sleep`'s doc no longer expands its signature.

The three `select` comments: you are right that "picks among ready cases at random" is a language rule the reader has. All three now state only the consequence, that the timer carries half those draws, and keep the part that is not a language rule — why the code asks again instead of trusting the answer it got.

pkg/wsclient/dial.go
geunwoonoh: Kept it, deliberately rather than by default, since you are right that the tag is what makes this the moment to decide.

`Dial` is the one entry point for a caller that wants to own the connection itself — the smux data plane wrapping it in `tunnel.WebSocketConn` is the case it exists for, and that caller cannot use `Client`, which never exposes its conn. Removing it would leave that consumer with no way in short of reimplementing the handshake, the header and the read limit, which is the duplication this package is meant to end.

doc.go:33-35 saying `pkg/tunnel` "is not reached from here" is about this package's own import graph, not about whether the caller exists: the wrapping happens on the caller's side, which is exactly why `Dial` returns a `*websocket.Conn` and stops there.

What you are right about is that it ships untested by a real consumer. Its own tests cover the handshake, the read limit, the rejected extension and the credential stripping, but nothing in this repository calls it, so the first real use will be someone else's module. If you would rather that surface not be frozen until a consumer exists, unexporting it now costs nothing and re-exporting later costs nothing either, and I will take that reading over mine if you hold it.

pkg/wsclient/config.go
geunwoonoh: Fixed the doc, kept the field. Your counterargument is the one I land on: a module outside this repository cannot reach `c.s.rand`, and reproducible reconnect timing in its own tests is a fair thing for it to want. This package exists for those modules.

The doc named the one audience that does not need it, which is what made it read as accidental:

> Rand returns the value in [0, 1) the jitter factor is drawn from. Nil means math/rand/v2. It is here for a caller that wants its own reconnect timing to be reproducible in its own tests, which a module outside this repository cannot arrange any other way.

pkg/wsclient/config.go
geunwoonoh: Answered in doc.go, and the answer is that convergence is the plan.

Mapping it before writing that down, the duplication is wider than the constants: there are three clients, not one.

| | lines | consumers |
|---|---|---|
| `pkg/runner/client.go` | 400 | `root.go`, `pkg/plugin`, `pkg/executor/handlers/system` |
| `pkg/runner/control_client.go` | 218 | `root.go` |
| `pkg/pluginclient/client.go` | 303 | external plugins |

The third is the SDK an out-of-tree plugin builds against, which makes it the one to move last.

Two things turned up while measuring. `runner.Connect` calls `os.Exit(1)` when its retry window expires, which `wsclient` deliberately does not do, so that is a behavior decision inside the migration rather than a mechanical step — the window is three days, so it is rarely reached, but it is reached by every agent that reaches it at the same time. And `ControlClient` is the natural first move: 218 lines, two call sites, and `Connect`, `CloseAndReconnect`, `WriteJSON`, `IsConnected` and `GetWSPath` have no caller outside their own file.

Not in this PR, though. Its stated scope is that nothing existing changes, and `v2.7.0` is the tag `alpamon-kube` pins; putting a rewrite of the agent's backhaul client under the same tag means a regression there poisons the thing this package was extracted to unblock.

doc.go now carries the short version, in a `# Relation to pkg/runner` section, including the drift you spotted between the two jitter ranges.

.github/workflows/build-and-test.yml
geunwoonoh: Fixed, in both directions, since neither copy is obviously the authority on sight.

The job's `ALLOWED` now says where the prose lives and which of the two decides:

```yaml
# The only packages pkg/wsclient may reach outside the standard
# library. pkg/tunnel and pkg/version are leaves themselves. The
# same list is written in prose in pkg/wsclient/doc.go, which says
# that this variable is the one that decides; update both together.
```

doc.go says the same from its side, and the failure message now tells whoever hits it what to do with both:

> pkg/wsclient must stay a leaf, but on $target its dependency graph reaches the following. If this is intended, widen ALLOWED in this job and the list in doc.go together:

Nothing enforces it still, which I think is right at this size — a check that the YAML and a doc comment agree is more machinery than two lines that point at each other.

pkg/wsclient/client_test.go
```diff
@@ -0,0 +1,1784 @@
...
+	recv(t, h.connects, "a connect")
+	require.Error(t, recv(t, h.disconnects, "the close the server sent"))
+	r := recv(t, h.retries, "the wait a connection that lasted no time at all should get")
+	assert.Equal(t, cfg.MinBackoff, r.delay)
+}
+
+// TestClient_ABackoffWaitPrefersAStopToItsTimer covers the tie the wait can
+// end in. A Shutdown or a done ctx that arrives as the backoff timer fires
+// leaves two of the select's cases ready at once, so the timer carried it
+// half the time and Run went on to dial. That dial is a connect to the backhaul and a call into a caller's
```
jisung-02: Nit: this line didn't get rewrapped along with the rest of the comment edit — it's 108 chars against ~78 for its neighbors.

```suggestion
	// half the time and Run went on to dial. That dial is a connect to the
	// backhaul and a call into a caller's own dial hook, both made after
```

pkg/wsclient/backoff.go
```diff
@@ -0,0 +1,77 @@
...
+// it. A base of max would put every draw at or above max, the clamp would
+// return max for all of them, and the schedule would go flat exactly where
+// the jitter matters most: a fleet that has been retrying long enough to
+// reach the ceiling is a fleet already in step. The base stops at two thirds
+// of max instead, which is the largest one whose whole window still fits
+// underneath, so the ceiling is a spread over [2/3 max, max) rather than a
+// single instant. A configuration with initial equal to max has no room for
+// any of this and gets that one value, which is what it asked for.
+func (b *backoff) next() time.Duration {
+	ceiling := b.max / 3 * 2 // divide first: b.max*2 can overflow
```
jisung-02, line 46: Small observation, not blocking: when `MinBackoff` is set above `2/3 * MaxBackoff`, `ceiling` falls back to `initial` and `current` stays pinned there, so the jittered result (`current * (1.0 + 0.5*r)`) exceeds `max` for roughly the top 20% of draws and gets clamped down to exactly `max` each time — a narrower version of the collapse this PR just fixed, only reachable when `MinBackoff` is already close to the ceiling. `backoff_test.go` doesn't cover that combination. Worth a follow-up test if it's easy, otherwise fine to leave as is given how narrow the window is.

jisung-02 (dismissed):
The fixes for the earlier round look right. Reran the transcribed `next()` logic by hand: the ceiling now spreads over [2/3·max, max) instead of collapsing to a single value, and `TestBackoff_JittersAtTheCeiling` pins it. `go vet` and `go test -race -count=2 ./pkg/wsclient` both pass. Two small things below, neither blocking.

pkg/wsclient/backoff.go
geunwoonoh: Fixed in 9a1373f, and the window is wider than the estimate. Measured over 50,000 runs per configuration, taking each one past the ceiling first:

```
initial=50s max=1m0s | multiply+clamp: 60.2% on max, 19903 distinct | window draw: 0.0%, 50000
initial=45s max=1m0s | multiply+clamp: 33.5% on max, 33236 distinct | window draw: 0.0%, 50000
initial=5s  max=1m0s | multiply+clamp:  0.0% on max, 50000 distinct | window draw: 0.0%, 50000
```

60% rather than 20% at the far end, and it degrades smoothly as `MinBackoff` approaches the ceiling rather than switching on at it.

Went further than a test because the cause turned out to be the formulation rather than that one configuration. The wait is now drawn uniformly from the base up to half again as much, or up to max where that would pass it, instead of multiplied and clamped back:

```go
top := b.max
if b.current <= b.max-b.current/2 { // else half again would pass max
	top = b.current + b.current/2
}
jittered := float64(b.current) + float64(top-b.current)*randFn()
```

The two agree wherever the whole window fits, which is every configuration the suite already had: all eleven backoff and reconnect tests pass with no change, including the fake-clock schedule and its exact 1s/2s/4s/8s waits. Where it does not fit, there is no overflow to clamp, so nothing lands on max by accident.

`TestBackoff_JittersWithAFloorNearTheCeiling` pins the combination at 50s/60s and asserts 200 draws stay inside `[MinBackoff, MaxBackoff)` with more than 190 distinct; putting the multiply back fails it on the spread.

Thanks for not letting this one go as "narrow" — the multiply was the thing worth replacing, and the narrow case is what showed it.

pkg/wsclient/client_test.go
geunwoonoh: Fixed in 9a1373f. Took the rewrap; my earlier comment edit shortened the line above it and left this one at 107 characters against 75 for its neighbours.

jisung-02 (approved):

PR #475, opened 2026-09-14
fix(signing): resolve dev from a workspace URL and send the env parameter the AI server reads

junho226:
Closes 474

## Summary

A dev agent held the prod signing key, so every signed command from a dev server failed verification. Monitor mode hid it behind a warning; the file lane (ADR 0053, #431) fails closed on it and refused with `key_unavailable`.

Two defects in `pkg/signing/keymanager.go`, either one enough on its own:

- `ResolveAuthEnv` returned `dev` only for the bare hostname `dev.alpacon.io`. An agent's configured URL is its workspace subdomain, `<workspace>.dev.alpacon.io`, so every dev agent resolved to prod.
- The public-key fetch sent `?auth_env=dev`. The AI server's endpoint reads `env` and ignores anything else, answering with the prod key.

## Changes

- `ResolveAuthEnv` returns `dev` for `dev.alpacon.io` and any subdomain of it (case-insensitive). A hostname that merely contains the string (`notdev.alpacon.io`, `dev.alpacon.io.example.com`) stays prod.
- `fetchKeyLocked` sends `?env=<authEnv>`.
- `TestResolveAuthEnv` gains the workspace-URL cases and the two look-alikes. `TestKeyManager_AuthEnvQueryParam` now pins `env`; it previously pinned the name the server ignores, which is why nothing caught this.

Prod agents are unaffected: their URLs do not end in `.dev.alpacon.io`, and a request with no parameter already returns the prod key. The AI server needs no change.

## Testing

- `go test ./pkg/signing/ ./pkg/runner/` pass; golangci-lint v2.9.0 (the CI pin) reports 0 issues on `pkg/signing`.
- Verified against the live AI server before the change: `?auth_env=dev` → `key-b9c40bc4e71366a6` (prod), `?env=dev` → `key-0ec988fe67979a78` (dev).
- After release, the ADR 0053 dev walkthrough (`alpacon exec --file` against `sudo-test`) is the end-to-end check; it currently stops at `key_unavailable`.


pkg/signing/keymanager.go
```diff
@@ -29,7 +34,8 @@ func ResolveAuthEnv(serverURL string) string {
 	if err != nil {
 		return ""
 	}
-	if strings.EqualFold(u.Hostname(), "dev.alpacon.io") {
+	host := strings.ToLower(u.Hostname())
+	if host == devDomain || strings.HasSuffix(host, "."+devDomain) {
```
geunwoonoh, line 39: Suggestion: a trailing-dot FQDN (`https://alpacax.dev.alpacon.io.`) keeps the dot in `u.Hostname()`, so it fails both the equality and suffix check here and falls through to prod. Config values come from `alpamon.conf`, a trusted local source, so this is unlikely in practice—not a blocker, just worth a `strings.TrimSuffix(host, ".")` if it ever comes up.

Suggestion: the `"."+devDomain` suffix string is rebuilt on every call. `ResolveAuthEnv` runs once per process (`cmd/alpamon/command/root.go:161`, `pkg/plugin/plugin.go:243`), so this has no measurable cost—flagging only because performance was explicitly in scope for this review.

geunwoonoh (dismissed):
### Summary
`ResolveAuthEnv` recognized only the bare host `dev.alpacon.io`, so a dev agent's actual configured URL—its workspace subdomain, `<workspace>.dev.alpacon.io`—resolved to prod. Separately, `fetchKeyLocked` sent `?auth_env=dev`, but the AI server's public-key endpoint reads `env` and ignores anything else, so it answered with the prod key regardless. Either defect alone was enough to make every dev agent sign with the prod key. This PR fixes both and pins the fix with new test cases.

### Critical issues (0)
None.

### Warnings (0)
None.

### Suggestions (2)

1. **pkg/signing/keymanager.go:38** A trailing-dot FQDN (`https://alpacax.dev.alpacon.io.`) keeps the dot in `u.Hostname()`, so it fails both the equality and suffix check and falls through to prod. Config values come from `alpamon.conf`, a trusted local source, so this is unlikely in practice—not a blocker, just worth a `strings.TrimSuffix(host, ".")` if it ever comes up.
2. **pkg/signing/keymanager.go:38** The suffix string is rebuilt on every call. `ResolveAuthEnv` runs once per process (`cmd/alpamon/command/root.go:161`, `pkg/plugin/plugin.go:243`), so this has no measurable cost—flagging only because performance was explicitly in scope for this review.

### Highlights
- Test coverage precisely targets the bug class: workspace subdomains, mixed-case hosts, and the two look-alike hosts (`notdev.alpacon.io`, `dev.alpacon.io.example.com`) that a naive `strings.Contains` fix would have mismatched.
- `TestKeyManager_AuthEnvQueryParam` previously pinned the query parameter name the server ignores (`auth_env`)—the PR calls this out directly as the reason the mismatch shipped unnoticed, and the updated assertion closes that gap.
- Comments explain the non-obvious parts: why a bare-domain check misses every real agent, and why the wrong query parameter name fails silently instead of erroring.
- Diff is scoped to exactly the two files that needed to change; no unrelated cleanup.

Verified: `go build ./pkg/signing/...`, `go test ./pkg/signing/... ./pkg/runner/...`, `go vet ./pkg/signing/...`, and `gofmt -l` all pass clean.

---
**Verdict**: APPROVED

**Checklist**:

pkg/signing/keymanager.go
junho226: Took the first one: `strings.TrimSuffix(host, ".")` before the match, with `https://alpacax.dev.alpacon.io.` added to `TestResolveAuthEnv`. Left the suffix concatenation as is, for the reason you gave: the function runs once per process, and a package-level constant would be the only alternative for a two-byte string.

pkg/signing/keymanager_test.go
```diff
@@ -20,8 +20,16 @@ func TestResolveAuthEnv(t *testing.T) {
 	}{
 		{"https://dev.alpacon.io", "dev"},
 		{"https://dev.alpacon.io/", "dev"},
+		// A workspace URL, which is what alpamon.conf actually carries.
+		{"https://alpacax.dev.alpacon.io", "dev"},
+		{"https://Alpacax.DEV.alpacon.io/", "dev"},
+		{"https://alpacax.dev.alpacon.io.", "dev"},
```
geunwoonoh, line 26: Suggestion: this new case covers the trailing dot on the subdomain form (`alpacax.dev.alpacon.io.`), which exercises the `HasSuffix` branch. The bare-domain form (`https://dev.alpacon.io.`) would exercise the `host == devDomain` branch instead and isn't covered here. Both branches share the same one-line trim above, so the risk is effectively zero—optional, not blocking.

geunwoonoh (dismissed):
### Summary
Follow-up commit `75e99fb` adds `strings.TrimSuffix(strings.ToLower(u.Hostname()), ".")` before the dev-domain comparison, so a trailing-dot FQDN (`https://alpacax.dev.alpacon.io.`) now resolves to `dev` instead of falling through to prod. A new test case pins this exact input. This is precisely the fix my prior Suggestion #1 called for—minimal, one line, no side effects on the other branches.

### Critical Issues (0)
None.

### Warnings (0)
None.

### Suggestions (1)

1. **pkg/signing/keymanager_test.go:26** The new case only covers the trailing dot on the subdomain form (`alpacax.dev.alpacon.io.`). The bare-domain form (`https://dev.alpacon.io.`) exercises the `host == devDomain` branch instead of the `HasSuffix` branch and isn't tested. Both branches share the same one-line trim, so the risk is effectively zero—optional, not blocking.

### Highlights
- The fix is scoped to exactly the reported gap: no unrelated cleanup, no scope creep.
- The added comment (`// A trailing-dot FQDN keeps its dot in Hostname(); it names the same host.`) states the non-obvious reason for the trim, not just what it does.
- The commit message explains the failure mode precisely and references why it matters (fails both equality and suffix checks, falls through to prod).
- Local verification against the new commit: `go build`, `go vet`, `gofmt -l` all clean; `TestResolveAuthEnv` and `TestKeyManager_AuthEnvQueryParam` pass, including the new trailing-dot case.

---
**Verdict**: APPROVED

**Checklist**:

pkg/signing/keymanager_test.go
junho226: Added `{"https://dev.alpacon.io.", "dev"}` so the equality branch is covered too.

geunwoonoh (approved):
### Summary
Commit `af12c35` adds `{"https://dev.alpacon.io.", "dev"}` to `TestResolveAuthEnv`, closing the one remaining gap from the prior review: the trailing-dot fix now has a test case for both branches it touches (`HasSuffix` via the subdomain form, `==` via the bare-domain form).

### Critical Issues (0)
None.

### Warnings (0)
None.

### Suggestions (0)
None outstanding.

### Highlights
- Test-only change, precisely scoped to the gap called out in review—no scope creep.
- Commit message states the rationale directly: "the bare domain goes through the equality branch, which the same trim feeds."
- All three rounds of this PR now form a clean, traceable sequence: fix → edge-case fix → test-coverage completion.
- Local verification: `go build`, `go vet`, `gofmt -l` clean; all 14 `TestResolveAuthEnv` subtests and `TestKeyManager_AuthEnvQueryParam` pass.

---
**Verdict**: APPROVED

**Checklist**:

PR #476, opened 2026-09-14
fix(utils): refuse a C1 control in zip names and escape control bytes on display

### LINKED ISSUE #441: fix(utils): hasControlBytes misses a raw 0x9b C1 CSI byte
## Summary

`hasControlBytes` is meant to reject "a byte a terminal would act on", but it iterates runes, so a raw `0x9b` (C1 CSI) is decoded as U+FFFD and passes. The byte itself still reaches the console inside a wrapped OS error.

## Where

`pkg/utils/archive.go`:

```go
// hasControlBytes reports whether s carries a byte a terminal would act on.
func hasControlBytes(s string) bool {
	return strings.ContainsFunc(s, func(r rune) bool { return r < 0x20 || r == 0x7f })
}
```

## Reproduce

```
hasControlBytes(ESC 0x1b)      = true
hasControlBytes(DEL 0x7f)      = true
hasControlBytes(raw byte 0x9b) = false
```

`strings.ContainsFunc` decodes UTF-8, and a lone `0x9b` is not valid UTF-8, so the callback sees U+FFFD instead of the byte.

## Why it matters

Small, and noted as such by the reviewer. Few UTF-8 terminals act on C1 controls today, so the practical exposure is low. It is filed because the function's stated job is to reject bytes a terminal would act on, and a raw C1 CSI is one of them.

## Suggested fix

Scan bytes rather than runes, so an invalid UTF-8 sequence cannot hide a control byte:

```go
func hasControlBytes(s string) bool {
	for i := 0; i < len(s); i++ {
		if b := s[i]; b < 0x20 || b == 0x7f || (b >= 0x80 && b <= 0x9f) {
			return true
		}
	}
	return false
}
```

Note that the `0x80`–`0x9f` range also matches the continuation bytes of legitimate multi-byte UTF-8, so the rule has to be decided deliberately: either reject only well-formed C1 sequences and lone `0x9b`, or accept that non-ASCII entry names are refused outright. Extend `TestUnzip_RejectionMessageEscapesControlBytes` with whichever rule is chosen.

Reported by @geunwoonoh in https://github.com/alpacax/alpamon/pull/435#discussion_r3878114380

jisung-02:
## Background

`hasControlBytes` scanned zip entry names for C0 controls and DEL only. U+009B is the CSI, which opens an escape sequence exactly as `ESC [` does, so a name carrying it extracted with the byte intact and reached the console whenever a later failure named the file.

Widening the check alone does not close the route, because the same byte range is ambiguous. A lone byte in `0x80`-`0x9f` is indistinguishable from a Shift-JIS or CP949 lead byte, so refusing it would make every legacy-encoded name Windows Explorer writes unextractable. The two halves of the range therefore need different answers: refuse what decodes to a control, and render what does not decode harmless where it is displayed.

### Escaping on display

`EscapeControlBytes` renders every C0 control, DEL and C1 control as `\xNN` and leaves everything else alone. A byte that decodes to nothing is judged by its raw value, so a lone `0x9b` is caught even though no rune reader sees one. Its output is plain ASCII, which makes it idempotent.

It is applied where text leaves for a display, not where it is produced:

- the extract worker's status line and the archive worker's error field, which carry wrapped OS errors that embed a raw path the surrounding `%q` never reaches
- the file handler's `Execute`, the single boundary every command result crosses. A `CreateZip` failure on the in-process path, a demotion error naming the requested user, the `rm` rejection that echoes back the path it refused, and the worker fallback that returns raw stderr all reach the console through it, and none of them quote what they carry. Escaping once there means a return added later cannot miss it.
- `statFileTransfer`, which reads the handler's result before it crosses that boundary

### Refusal at extraction

`hasControlBytes` now covers the C1 range, so an entry name or link target that decodes to a C1 control is refused with the same message the C0 case already used. A byte in that range that decodes to nothing passes on purpose and is handled by the escape above.

The predicate is `unicode.IsControl` rather than a hand-rolled range test. Every rune from `0x00` to `0xff` was checked against both and they never disagree, and both are false above `0xff`. This is the Cc category only: the Cf format characters `U+202E`, `U+200B` and `U+FEFF` are not refused, and the doc comment now says so instead of claiming to cover everything a terminal reacts to. Refusing Cf would cost every name carrying a BOM or a zero-width joiner, and visual filename spoofing with `U+202E` is a separate problem from escape sequence injection.

### Escape cost

`EscapeControlBytes` writes `\xNN` from a hex digit lookup rather than `fmt.Fprintf`. Passing `&b` to `Fprintf` hands the `strings.Builder` over as an `io.Writer`, so escape analysis moves it to the heap on every call, including the common one where the string holds no control byte and the escape branch never runs. `go build -gcflags=-m` over a standalone file holding both variants reports it for the `Fprintf` one only:

```
./esc.go:36:6: moved to heap: b
```

An 80 character path measures 312 ns/op and 1 alloc with the lookup against 355 ns/op and 2 allocs with `Fprintf`; a string that is a quarter control bytes measures 415 against 1404 ns/op. The function runs on every OS error and path crossing the file transfer and archive boundaries.

## Scope

Extraction behavior for a legacy-encoded name is unchanged: `TestUnzip_LegacyEncodedEntryNameIsExtracted` and `TestUnzip_LegacyEncodedLinkTargetIsRestored` pin the create-to-extract round trip.

## Testing

On `f66bdfad`. `gofmt -l pkg/utils pkg/executor/handlers/file`, `go vet ./pkg/utils/... ./pkg/executor/handlers/file/...` and `go build ./...` all print nothing.

```
$ go test ./pkg/utils/... ./pkg/executor/handlers/file/... -p 1 -count=1
ok  	github.com/alpacax/alpamon/v2/pkg/utils	0.526s
ok  	github.com/alpacax/alpamon/v2/pkg/executor/handlers/file	0.451s
```

The two control-byte rejection tests assert with `require.ErrorContains` before reading `err.Error()` on the next line, so a regression returning `nil` fails there instead of panicking.

## Deferred

Replacing four `range strings.Split(...)` loops with `strings.SplitSeq` is #477. None of them are in this branch's diff.

Closes #441


pkg/utils/archive.go
```diff
@@ -544,8 +545,12 @@ func recheckLinks(root string, links []deferredEntry) (rejected, left error) {
 }
 
 // hasControlBytes reports whether s carries a byte a terminal would act on.
+// UTF-8 validity is checked separately so a C1 byte isn't confused with a
+// continuation byte inside legitimate multi-byte UTF-8—both share 0x80-0x9f.
 func hasControlBytes(s string) bool {
-	return strings.ContainsFunc(s, func(r rune) bool { return r < 0x20 || r == 0x7f })
+	return !utf8.ValidString(s) || strings.ContainsFunc(s, func(r rune) bool {
+		return r < 0x20 || r == 0x7f || (r >= 0x80 && r <= 0x9f)
+	})
```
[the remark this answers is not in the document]
jisung-02: Fixed in 676df0db, taking the first of the two options you named: the check inspects malformed bytes only for the control ranges.

You were right on both halves of this. `utf8.ValidString` made the function reject every malformed sequence rather than just C0, DEL, and C1, and `addSymlinkToZip` writes `os.Readlink`'s target with no matching validation, so the create-to-extract round trip broke for targets carrying no control byte at all.

`hasControlBytes` now walks the string with `utf8.DecodeRuneInString`. Where a byte decodes to nothing, the raw byte itself is tested instead of the `U+FFFD` the decoder substitutes for it, so a lone `0x9b` is still caught while `0xa0` and a Latin-1 name now pass.

I did not take the second option. Applying the same encoding restriction in `addSymlinkToZip` would refuse links that archive fine today, which turns a terminal-safety check into a limit on which Unix paths can be archived at all.

One case stays rejected on purpose: a continuation byte left standing alone by a truncated sequence, for example the `0x95` of a cut-off `한`, cannot be told apart from a raw C1 control. The test names that case and the reason.

`TestHasControlBytes` gained `\xa0`, a Latin-1 `caf\xe9.txt`, `\xa0\x9b`, and the truncated `한`. Every changed expectation was seen red first: restoring the `utf8.ValidString` form failed the `\xa0` and Latin-1 cases, and dropping the raw-byte substitution failed the `0x9b`, `0x80`, `0x9f`, truncated `한`, and trailing-C1 cases. `go test ./pkg/utils/ -p 1` passes on the pushed head.

pkg/executor/handlers/file/file.go
```diff
@@ -462,7 +462,8 @@ func uploadSummary(count int, skipped utils.SkippedReport) string {
 	}
 	reasons := make([]string, 0, len(named)+1)
 	for _, entry := range named {
-		reasons = append(reasons, fmt.Sprintf("%s: %v", entry.Path, entry.Reason))
+		// Neither half is quoted, and the summary is the command result.
+		reasons = append(reasons, utils.EscapeControlBytes(fmt.Sprintf("%s: %v", entry.Path, entry.Reason)))
```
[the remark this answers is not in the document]
jisung-02: Fixed in 118b6770, and you found the shape of the problem rather than one instance of it. The escape was sitting at each place a string was built, so the summary had one and the twelve error returns beside it did not.

It now sits at the handler boundary instead. `Execute` is a thin wrapper that escapes the message the inner body returns, which is the single point every result crosses on its way to the console. That covers the `CreateZip` path you named, the demotion errors, `SanitizePath` failures that format the raw path into the message, the `rm` rejection that echoes back the path it refused, and anything a later return adds.

`statFileTransfer` reads the handler's result before it crosses that boundary, so it escapes separately. `uploadSummary` no longer escapes its own text: the boundary covers it, and one authority is easier to reason about than two.

On the logged error, I checked rather than adding a call. The stderr sink is `zerolog.ConsoleWriter` via `logger.PrettyWriter`, which escapes a C1 already; a run with a raw `\xc2\x9b` in the error shows `json sink holds raw C1: true` and `console sink holds raw C1: false`. The server sink is `logRecordWriter`, which ships only files listed in `logRecordFileHandlers` (command.go, commit.go, pty.go, shell.go, server.go), so this handler's logs are dropped there, and the `LogRecord` it builds carries `entry.Message` only, never the error value. Adding an escape would have suggested a hole that is not there.

pkg/utils/archive_worker.go
```diff
@@ -143,7 +143,7 @@ func RunArchiveWorker(in io.Reader, dst io.Writer, status io.Writer) int {
 			resp.Skipped = append(resp.Skipped, reported)
 		}
 		if err != nil {
-			resp.Error = err.Error()
+			resp.Error = EscapeControlBytes(err.Error())
```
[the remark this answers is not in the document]
jisung-02: Fixed in 118b6770, though not at this line.

Both routes you named end at the same place. `resp.Skipped` reaches the command result through `uploadSummary`, and the unparseable-status fallback reaches it as `fmt.Errorf("%w: %s", runErr, msg)` returned from `archiveWorkerResult` and handed back by `handleUpload` as `err.Error()`. Both are the handler's result string, so escaping at the `Execute` boundary covers them without sanitizing the skipped fields before encoding.

I chose the boundary over this line for the reason your other comment exposed: escaping at each producer is how the gap appeared in the first place. `TestFileHandler_Execute_EscapesTheResult` pins it, and it was seen red first, with the escape removed:

```
Error: "rm: refusing to remove path outside staging namespace: /tmp/evil\x9b[2J.sh" should not contain "\x9b"
```

pkg/utils/archive_test.go
```diff
@@ -451,34 +460,77 @@ func TestUnzip_RejectionMessageEscapesControlBytes(t *testing.T) {
...
-	require.NoError(t, os.MkdirAll(out, 0755))
+	for _, tt := range controlByteIntroducers {
+		t.Run(tt.name, func(t *testing.T) {
+			zipPath, out := newZipFixture(t, []zipEntry{{name: "d" + tt.byte + "[2J/x.txt", body: "x"}})
 
-	err := Unzip(zipPath, out)
-	assert.ErrorContains(t, err, "illegal file path in zip")
-	assert.NotContains(t, err.Error(), "\x1b")
+			err := Unzip(zipPath, out)
+			assert.ErrorContains(t, err, "illegal file path in zip")
```
geunwoonoh, line 480: Warning: `assert.ErrorContains` here is immediately followed by `err.Error()` on the next line (`assert.NotContains(t, err.Error(), tt.byte)`). If `Unzip` ever regresses to return `nil`, this panics with a nil-pointer dereference instead of failing with a clear message. This repo's own `CLAUDE.md` documents exactly this convention ("`require` when a failed check makes the rest of the test meaningless"), and Copilot flagged this same line in its latest review — it was not applied.

```suggestion
			require.ErrorContains(t, err, "illegal file path in zip")
```

pkg/utils/archive_test.go
```diff
@@ -451,34 +460,77 @@ func TestUnzip_RejectionMessageEscapesControlBytes(t *testing.T) {
...
 	// wraps the OS error, which embeds the raw target where %q cannot reach.
-	writeEntryZip(t, zipPath, []zipEntry{{name: "lnk", body: "\x1b[2Jx", isLink: true}})
+	for _, tt := range controlByteIntroducers {
+		t.Run(tt.name, func(t *testing.T) {
+			zipPath, out := newZipFixture(t, []zipEntry{{name: "lnk", body: tt.byte + "[2Jx", isLink: true}})
 
-	out := filepath.Join(dir, "out")
-	require.NoError(t, os.MkdirAll(out, 0755))
+			err := Unzip(zipPath, out)
+			assert.ErrorContains(t, err, "illegal link target in zip")
```
geunwoonoh, line 494: Warning: same issue as above — `assert.ErrorContains` followed by an unguarded `err.Error()` on the next line. A regression that makes `Unzip` return `nil` here panics instead of failing cleanly.

```suggestion
			require.ErrorContains(t, err, "illegal link target in zip")
```

pkg/utils/archive.go
```diff
@@ -543,9 +543,40 @@ func recheckLinks(root string, links []deferredEntry) (rejected, left error) {
 	return rejected, left
 }
 
-// hasControlBytes reports whether s carries a byte a terminal would act on.
+// hasControlBytes reports whether s decodes to a character a terminal acts
+// on. An undecodable byte is not one; see EscapeControlBytes.
 func hasControlBytes(s string) bool {
-	return strings.ContainsFunc(s, func(r rune) bool { return r < 0x20 || r == 0x7f })
+	return strings.ContainsFunc(s, isTerminalControl)
```
geunwoonoh, line 549: Suggestion: `isTerminalControl` (lines 552-556 below) reimplements `unicode.IsControl` byte for byte — verified directly that for every rune 0x00-0xFF, `r < 0x20 || r == 0x7f || (r >= 0x80 && r <= 0x9f)` matches `unicode.IsControl(r)` with zero mismatches, and both are `false` above `0xFF`. Swapping this call and the `isTerminalControl(r)` check inside `EscapeControlBytes` (line 568) to `unicode.IsControl` removes the duplicate, hand-rolled function (and its comment) for the cost of one `unicode` import.

```suggestion
	return strings.ContainsFunc(s, unicode.IsControl)
```

pkg/utils/archive.go
```diff
@@ -543,9 +543,40 @@ func recheckLinks(root string, links []deferredEntry) (rejected, left error) {
...
+		r, size := utf8.DecodeRuneInString(s[i:])
+		if r == utf8.RuneError && size == 1 {
+			r = rune(s[i])
+		}
+		if !isTerminalControl(r) {
+			b.WriteString(s[i : i+size])
+			i += size
+			continue
+		}
+		for _, c := range []byte(s[i : i+size]) {
```
geunwoonoh, line 573: Suggestion: this loop calls `fmt.Fprintf(&b, "\x%02x", c)` once per escaped byte. `fmt.Fprintf` re-parses the format string and goes through reflection on every call; since the output is always exactly `\xNN` from a single `byte`, a small hex-digit lookup table (or `hex.EncodeToString` on a 1-byte slice) would avoid that overhead. Low impact today given short error/path strings, but this runs on every OS-error-wrapping boundary in the file-transfer and archive/extract paths.

geunwoonoh (commented):
### Summary
Widens `hasControlBytes` to cover the C1 control range and adds `EscapeControlBytes` to neutralize control bytes wherever a raw OS error or path reaches the console (file handler `Execute`, `statFileTransfer`, and the archive/extract workers' status streams). Legacy-encoded (Shift-JIS/CP949/Latin-1) names still extract correctly because ambiguous raw bytes are escaped for display rather than refused at extraction time.

### Critical Issues (0)
None.

### Warnings (1)

1. **`pkg/utils/archive_test.go:480,494`** `assert.ErrorContains(t, err, ...)` is immediately followed by `err.Error()` on the next line (`assert.NotContains(t, err.Error(), tt.byte)`). If `Unzip` ever regresses to return `nil` here, this panics with a nil-pointer dereference instead of failing with a clear message. This repo's own `CLAUDE.md` states exactly this rule ("`require` when a failed check makes the rest of the test meaningless"), and Copilot flagged these same two lines in its latest review — it was not applied.

### Suggestions (2)

1. **`pkg/utils/archive.go:549`** `isTerminalControl` (lines 552-556) reimplements `unicode.IsControl` byte for byte: verified directly that for every rune 0x00-0xFF, `r < 0x20 || r == 0x7f || (r >= 0x80 && r <= 0x9f)` matches `unicode.IsControl(r)` with zero mismatches, and both are `false` above `0xFF`. Swapping both call sites (`hasControlBytes` here and the `isTerminalControl(r)` check inside `EscapeControlBytes` at line 568) to call `unicode.IsControl` directly removes the duplicate, hand-rolled function and its own comment, at the cost of one `unicode` import.
2. **`pkg/utils/archive.go:573-575`** (`EscapeControlBytes`) calls `fmt.Fprintf(&b, "\\x%02x", c)` once per escaped byte. `fmt.Fprintf` re-parses the format string and goes through reflection on every call; since the output is always exactly `\xNN` from a single `byte`, a small hex-digit lookup (or `hex.EncodeToString` on a 1-byte slice) would avoid that overhead. Low impact today given short error/path strings, but this runs on every OS-error-wrapping boundary in the file-transfer and archive/extract paths.

### Highlights

- The "refuse what decodes to a control, escape what doesn't" split is the right answer to the stated ambiguity (C1 range overlaps legacy lead bytes), and it's consistently applied: `CreateZip`/`Unzip` converge on the `Execute` boundary escape (or pre-escape before it) across every dispatch path — Windows in-process, Unix without demotion, and the Unix worker subprocess.
- `EscapeControlBytes`'s idempotency (verified by `TestEscapeControlBytes_LeavesTheRestAlone`) matters in practice, since some errors get escaped twice (once in the worker, again at `Execute`).
- The legacy-encoding regression the first review round caught is now pinned with `TestUnzip_LegacyEncodedEntryNameIsExtracted`/`LinkTargetIsRestored`.
- `go build`, `go vet`, and `go test -p1` pass on both touched packages; `gofmt -l` reports no diffs.

---

**Verdict**: COMMENTED

**Checklist**:
- [x] Code follows stack conventions (gofmt clean, exported/unexported naming correct)
- [x] Tests included and passing (`go test ./pkg/utils/... ./pkg/executor/handlers/file/... -p 1` -> ok)
- [x] No security vulnerabilities (this is itself a security fix)
- [ ] `require`/`assert` convention from this repo's own `CLAUDE.md` not fully applied (see Warning)

pkg/utils/archive_test.go
jisung-02: Applied as asked in a01660c2. You were right that the dependent line makes this more than a style preference: `assert` records the failure and keeps going, so `err.Error()` on the next line runs on a nil error.

Confirmed it panics rather than failing cleanly, with a stand-in for a regressed `Unzip`:

```
--- FAIL: TestDemoAssertPanics (0.00s)
        Error:      An error is expected but got nil.
panic: runtime error: invalid memory address or nil pointer dereference [recovered, repanicked]
```

The panic takes the whole package down, so the one focused assertion becomes an unreadable failure for every test after it.

pkg/utils/archive_test.go
jisung-02: Applied as asked in a01660c2, same as the entry-name case above. Both sites now stop at the missing error instead of dereferencing it on the next line.

pkg/utils/archive.go
jisung-02: Applied as asked in b772d1d7. Checked the equivalence independently before swapping and got the same result you did: no rune from `0x00` to `0xff` disagrees, and both are false above `0xff`.

Your comment also surfaced something the suggestion did not mention. Once the predicate is named `unicode.IsControl`, the doc comment above it is visibly wrong: "a character a terminal acts on" reads as covering the Cf format characters, and it does not.

```
U+202E: IsControl=false IsFormat=true
U+200B: IsControl=false IsFormat=true
U+FEFF: IsControl=false IsFormat=true
```

Narrowed the comment to the range actually tested rather than widening the check. Refusing Cf would cost every name carrying a BOM or a zero-width joiner, and `U+202E` filename spoofing is a different attack from escape sequence injection, so it is not something to settle here.

pkg/utils/archive.go
jisung-02: Applied as asked in 312f900c, and the measurement says the effect is larger than the reasoning in your comment predicted.

The format string parsing is real, but the dominant cost is not per escaped byte. Passing `&b` to `Fprintf` hands the `strings.Builder` over as an `io.Writer`, so escape analysis moves it to the heap on every call, whether or not the escape branch is ever entered. `go build -gcflags=-m` over a standalone file with both variants reports it for the `Fprintf` one only:

```
./esc.go:36:6: moved to heap: b
```

That is why the common case, a path with no control byte at all, still shows a difference (`-benchmem -count=6`, `for b.Loop()` with the result assigned to a sink so the calls are not optimized away):

| input | lookup | Fprintf |
|---|---|---|
| 80 char path, no control bytes | 312 ns/op, 64 B/op, 1 alloc | 355 ns/op, 96 B/op, 2 allocs |
| same, one ESC | 320 ns/op, 80 B/op, 1 alloc | 422 ns/op, 112 B/op, 2 allocs |
| a quarter control bytes | 415 ns/op, 240 B/op, 2 allocs | 1404 ns/op, 272 B/op, 3 allocs |

The absolute numbers are small, so this is a saved allocation per call rather than a hot path win. The heap escape is the part worth recording, so the loop carries a two-line comment saying why `Fprintf` is not used. The benchmark itself is not committed.

geunwoonoh (approved):
### Summary
All three items from the prior review are fixed, verified against the actual diff, not just claimed:

| Finding | Commit | Verified |
|---|---|---|
| Warning: `require.ErrorContains` before the dependent `err.Error()` | `a01660c` | `archive_test.go:480,494` now use `require.ErrorContains` |
| Suggestion: `unicode.IsControl` over the hand-rolled predicate | `b772d1d` | `isTerminalControl` removed; both call sites (`hasControlBytes`, `EscapeControlBytes`) call `unicode.IsControl` directly |
| Suggestion: hex-digit lookup over `fmt.Fprintf` | `312f900` | `fmt.Fprintf` replaced with a `hexDigits` lookup table and `WriteByte` |

Re-ran the equivalence check for the `unicode.IsControl` swap (every rune 0x00-0xFF, both false above 0xFF) against the current code, and `TestHasControlBytes` now exercises the exact boundary cases (`U+009F`/`U+00A0`) that would catch a regression there.

### Critical Issues (0)
None.

### Warnings (0)
None. The previously open item is fixed.

### Suggestions (0)
None outstanding. Both prior suggestions are applied; the fix for the `Fprintf` cost is more precise than my own diagnosis (the cost is dominated by the `strings.Builder`->`io.Writer` conversion forcing a heap escape, not primarily format-string parsing), and that reasoning is captured in a comment at the fix site.

### Highlights

- The doc comment on `hasControlBytes` was narrowed from "a character a terminal acts on" to "a C0, DEL, or C1 control rune" -- a real correction, not just a rename: `unicode.IsControl` (Cc category) does not cover Cf format characters (`U+202E`, `U+200B`, `U+FEFF`), so the old wording overclaimed. The PR explicitly scopes that out (visual spoofing via `U+202E` is a different problem from escape-sequence injection) rather than silently leaving a gap.
- The now-unnecessary comment on the link-target `hasControlBytes` call in `extractSymlink` was dropped in the same change -- a legitimate restatement removal, not scope creep.
- Independently confirmed: `gofmt -l pkg/utils pkg/executor/handlers/file` clean, `go vet` clean on both packages, `go build` clean, and `go test ./pkg/utils/... ./pkg/executor/handlers/file/... -p 1 -count=1` passes, including every control-byte-specific test.
- The `strings.Split`->`strings.SplitSeq` cleanup was correctly deferred to a separate issue (#477) rather than pulled into this branch.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Code follows stack conventions (gofmt clean, stdlib preferred over hand-rolled logic)
- [x] Prior review findings resolved and independently reverified

Issue #480, opened 2026-09-21
Metrics that fail to transmit are never retried after a restart


jisung-02:
## Overview

A metric that fails to transmit is never retried after the agent restarts. Failed sends are retried a few times from an in-memory queue, and when the process goes down, those pending sends go with it. The raw value stays in the local database, but that row exists for aggregation, not as a transmit backlog.

## Current behavior

A realtime check measures a value, writes it to SQLite, and publishes the result to `SuccessQueue`. For the CPU check, `saveCPUUsage` writes the row (`pkg/collector/check/realtime/cpu/cpu.go:69-73`) and `Execute` publishes it (`pkg/collector/check/realtime/cpu/cpu.go:32`).

`Collector` drains that queue and sends each metric to the server. On a transport failure it moves the metric to `FailureQueue` (`pkg/collector/collector.go:188-193`), and `failureQueueWorker` pulls one item every five seconds and retries it through `retryWithBackoff` (`pkg/collector/collector.go:210`, `:217`, `:230-248`).

That is where the chain ends.

`FailureQueue` lives only in memory. Once shutdown starts, `failureQueueWorker` returns on `ctx.Done()` (`pkg/collector/collector.go:207-208`) and `Stop()` closes the channel (`pkg/collector/collector.go:266`). Whatever was still queued is dropped without anyone reading it.

The database records no transmission state. `pkg/db/schema/cpu.go` defines two columns, `timestamp` and `usage`. Searching all of `pkg/db/schema/` for `sent`, `status`, `pending`, `published`, or `transmitted` returns nothing, so there is no way to tell an untransmitted row from a transmitted one.

Nothing replays past data at startup either. `NewCollector` and `initTasks` only construct and register the checks (`pkg/collector/collector.go:98-156`), and `AddTask` sets `nextRun` to `time.Now().Add(interval)` (`pkg/collector/scheduler/scheduler.go:71`). No last-run timestamp is persisted, so a restart never catches up on the interval it missed.

## Retention of raw rows

Two code paths delete raw rows, and both key on the timestamp alone:

```
pkg/collector/check/batch/hourly/cpu/cpu.go:124        tx.CPU.Delete().Where(cpu.TimestampLTE(now.Add(-1 * time.Hour)))
pkg/collector/check/batch/daily/cleanup/cleanup.go:93  same condition
```

[...]

PR #481, opened 2026-09-21
fix(collector): stop closing taskQueue while the dispatcher may still send

### LINKED ISSUE #453: Scheduler.Stop closes taskQueue while dispatcher may be sending, crashing the agent
## Background

`Scheduler.Stop()` closes `taskQueue` while `dispatcher()` may be blocked sending on it. Closing a channel that a live sender still holds is a panic the Go runtime does not let anyone recover from, so the whole agent goes down. The same shape can also hang the dispatcher instead, depending on which goroutine loses the race.

This is not a shutdown-only path. `Collector.Stop()` runs on every collector restart the server asks for, so a long-lived agent reaches it repeatedly.

## The code

`taskQueue` is unbuffered (`pkg/collector/scheduler/scheduler.go:53`):

```go
taskQueue: make(chan *ScheduledTask),
```

`dispatcher()` sends inside the `tasks.Range` callback with no guard (`scheduler.go:108` and `scheduler.go:112`):

```go
if now.After(task.nextRun) {
    task.nextRun = now.Add(task.interval)
    s.taskQueue <- task
}

if task.isRetryRequired(now) {
    s.taskQueue <- task
}
```

The enclosing `select` has `ctx.Done()` and `stopChan` arms (`scheduler.go:93-96`), but a goroutine parked in a send never gets back to them.

`worker()` receives with `range` and only then looks at the context (`scheduler.go:121-129`):

```go
for task := range s.taskQueue {
    select {
    case <-ctx.Done():
        return
    default:
        s.executeTask(ctx, task)
    }
}
```

`Stop()` closes both channels (`scheduler.go:83-86`):

```go
func (s *Scheduler) Stop() {
    close(s.stopChan)
    close(s.taskQueue)
}
```

`Collector.Stop()` cancels the context, then calls it (`pkg/collector/collector.go:235-241`):

```go
if c.cancel != nil {
    c.cancel()
}

c.scheduler.Stop()
c.wg.Wait()
```

`c.wg.Wait()` does not cover the scheduler. `Start` launches it with a bare `go c.scheduler.Start(c.ctx, c.buffer.Capacity)` (`collector.go:145`) and no matching `wg.Add(1)`, so `Wait()` returns without either the dispatcher or the workers having stopped.

## Two ways this goes wrong

Send on a closed channel. Workers are unbuffered receivers, so once every worker is busy the dispatcher parks in `s.taskQueue <- task`. If `close(s.taskQueue)` lands while it is parked, the send panics. Nothing in `dispatcher` recovers, and an unrecovered panic in any goroutine takes the process with it.

A dispatcher that never returns. After `c.cancel()`, each worker receives one more task, sees `ctx.Done()`, and returns. They leave one at a time, and after the last one there is no receiver left. The dispatcher blocks in the send forever and never reaches its `ctx.Done()` or `stopChan` arm.

## Reproducing it

`pkg/collector/scheduler/` has no test file at all, which is why neither path is covered. A test that fills every worker slot, cancels the context, and calls `Stop()` should reach one of the two outcomes; running it under `-race` with `-count` high enough makes the interleaving show up.

The path a user hits is `cmd/alpamon/command/root.go:214`, where a collector restart command from the server calls `metricCollector.Stop()` and then builds a new collector.

## What a fix looks like

Whoever sends on a channel should be the one to close it, and here the sender is the dispatcher while the closer is `Stop()`. Bringing those back together covers both failures:

- Wrap each send in a `select` with `ctx.Done()` and `s.stopChan`, so a parked dispatcher can still walk away.
- Have `worker()` select on `ctx.Done()` per iteration rather than checking after a receive, so workers do not silently disappear from under the dispatcher.
- Drop `close(s.taskQueue)` from `Stop()`, or move it to the dispatcher's own exit path once it is the only sender left.
- Track the scheduler goroutines in `Collector.wg` so `Stop()` actually waits for them.

jisung-02:
Closes #453.

## Problem

`Scheduler.Stop` closed `taskQueue` as well as `stopChan`. The dispatcher is the only sender on `taskQueue`, and it can be parked in a send when `Stop` runs, so it woke up on a closed channel and panicked. A send on a closed channel cannot be recovered, so the agent died during shutdown.

Closing from the receiving side is the root of it. The close carried no signal anyone needed either, since the workers can leave on `stopChan` and `ctx.Done()`.

## Change

`Stop` closes `stopChan` only, guarded by a `sync.Once`, and then waits on a `WaitGroup` that covers every worker and the dispatcher. Nothing closes `taskQueue`; it is reclaimed with the scheduler.

`Start` registers the goroutines with the `WaitGroup` on the caller's goroutine, so calling it with `go` would let `Stop` reach the wait first. A comment says so at the declaration. Both it and `Collector.Start` use `sync.WaitGroup.Go`, which arrived in Go 1.25 and makes an `Add` without its `Done` unspellable.

Every dispatcher send goes through `Scheduler.send`, which checks `stopChan` and `ctx` before blocking and again while blocked. The non-blocking pre-check is what gives the stop signal priority, because `select` picks at random among ready cases.

## Bounded shutdown join

`wg.Wait` with no bound would hand shutdown to the slowest check. A check stuck in a context-unaware syscall—`statfs` on a dead network mount during the disk usage check—would block `Collector.Stop`, and with it graceful shutdown and the collector-restart path. Before this branch the join did not exist, so shutdown was effectively bounded by the HTTP timeout.

`agent.ShutdownWaitBudget` bounds both joins at five seconds each. `Scheduler.Stop` reports whether it joined, and `Collector.Stop` logs a warning for each wait that ran out. Worst case shutdown is therefore ten seconds. The two waits cannot overlap: the collector's queue workers can only be known idle once the scheduler has stopped dispatching checks that publish to those queues.

The bound is why `Collector.Stop` no longer closes `SuccessQueue` and `FailureQueue`. A check that outlives the budget still reaches `publish()`, whose `select` has both a send on the queue and `ctx.Done` ready at once, so it would take the closed-channel send about half the time—the same unrecoverable panic this branch removes for `taskQueue`. The closes bought nothing: the channels are reclaimed with the collector, and the `ok` guards that received them are gone with them. `close(errorChan)` stays, because nothing anywhere sends to `errorChan` and `handleErrors` would otherwise leak on every restart.

## Other defects fixed in the same path

Both sit in code the shutdown fix had to rewrite, so they are not separable into their own commit.

`ScheduledTask` fields were read by the dispatcher and written by a worker with no lock. A check that ran longer than its interval was also queued again on the next tick and run concurrently with itself. A mutex guards the fields, and a `running` marker keeps one run of a task in flight at a time. A task whose check is still running is skipped on every tick until it finishes, and then runs once on the next tick rather than catching up on the ticks it missed.

`MaxRetries` capped nothing. `isRetryRequired` ignored the attempt count, so once the attempts were spent the due time stayed in the past and a failing check retried every tick until the expiry window closed. The field named the wrong fact: `isSuccess` became `retryPending`, which states whether a retry is scheduled, and exhausting the attempts clears it.

## Shutdown is quiet

On shutdown `Execute` returns `context.Canceled` for in-flight checks, and the scheduler logged it as `failed to execute check` and bumped retry state on a task about to be discarded. `executeTask` now returns early when `ctx.Err()` is set. It asks the context rather than matching the error value, so a check that returns `context.Canceled` for its own reason still logs and still retries.

## Collector

`Collector.Stop` is idempotent and starts the scheduler synchronously. Check publishing takes a context, so a check cannot block forever on a queue that shutdown has stopped draining.

## Tests

`pkg/collector/scheduler/scheduler_test.go`, `pkg/collector/check/base/base_test.go` and `pkg/agent/wait_test.go` are new. They cover a `Stop` while the dispatcher is parked in a send, a `Stop` while a worker is parked publishing to a full queue, a second `Stop`, `Stop` with zero workers, a `Stop` whose budget runs out on a check that never returns, the retry bound, a slow check never overlapping itself, a canceled check leaving retry state alone while a plain error bumps it, the metric queues staying open on both the joined and the timed-out path, and both branches of the bounded wait. The timer-dependent ones run in `testing/synctest` bubbles.

The `Race detector` job in `.github/workflows/build-and-test.yml` runs `go test -race -count=2` on `pkg/wsclient`, `pkg/collector/scheduler`, `pkg/collector`, `pkg/collector/check/base` and `pkg/agent`. It was named for `pkg/wsclient` alone and skipped the ent codegen step on purpose; the scheduler reaches `ent` through `base.CheckStrategy`, so the job now runs the shared `setup-go-generate` action. `go test` runs package targets concurrently, so the four added packages finish inside the existing `pkg/wsclient` run.

## Verification

Run on 93cd51b3.

```
gofmt -l .                                           # clean
go build ./...                                       # ok
go vet ./...                                         # ok
go test -race -count=2 ./pkg/wsclient ./pkg/collector/scheduler ./pkg/collector ./pkg/collector/check/base ./pkg/agent
ok  github.com/alpacax/alpamon/v2/pkg/wsclient               13.149s
ok  github.com/alpacax/alpamon/v2/pkg/collector/scheduler     1.273s
ok  github.com/alpacax/alpamon/v2/pkg/collector               1.270s
ok  github.com/alpacax/alpamon/v2/pkg/collector/check/base    1.206s
ok  github.com/alpacax/alpamon/v2/pkg/agent                   1.231s
go test ./... -p 1                                   # all packages ok
```

## Comments

The comments in `.github/workflows/build-and-test.yml` were trimmed to what the YAML cannot state on its own, which is a drive-by on a file this branch already touches. One sentence went with them by mistake and is back: which consumer the pre-/v2 leaf guard keeps working for.

## Out of scope

#480 covers metrics that fail to transmit and are never retried after a restart. It was found while reviewing this shutdown path and is a separate policy question.

#482 adds a bounded flush at the same point as the shutdown join. It will want to share `agent.ShutdownWaitBudget` rather than add a second budget beside it; that is left to #482.


eunyoung14 (commented):
Read the shutdown path end to end. The fix is right: no `close(taskQueue)` remains, every send goes through `Scheduler.send`, and `Collector.Stop`'s cancel → `scheduler.Stop` → `wg.Wait` → close order joins every producer of both queues before the close. Go 1.25 is already required by `go.mod`, so `WaitGroup.Go` and `synctest.Test` are fine on CI and goreleaser.

A few non-blocking notes, none of which should hold the merge:

1. `Scheduler.Stop` now waits on `wg.Wait()` without a bound. A check stuck in a context-unaware syscall (for example `statfs` on a dead network mount during the disk usage check) would block `Collector.Stop`, and with it graceful shutdown and the collector-restart command path; before this change the join did not exist, so shutdown was effectively bounded by the HTTP timeout. A bounded wait that logs and proceeds on expiry would restore that. #482 adds a bounded flush at the same point, so the two budgets are best designed together.
2. `TestStop_ReturnsAfterDispatcherWasParkedInASend` no longer reproduces the parked send on the fixed code: with one task and one worker the `running` marker makes the dispatcher skip the tick, so it idles in the ticker select. Two tasks with one worker would park the dispatcher on the second send. Renaming the test or adding the second task would make the name true.
3. `collector_test.go:135` has a `wg.Add(1)` with no `Done`; the test never waits on it, so it is dead rather than wrong, but it is the pattern the PR body says `WaitGroup.Go` makes unspellable.
4. The race job runs `./pkg/wsclient ./pkg/collector/scheduler`, while this PR adds concurrency tests to `pkg/collector` and `pkg/collector/check/base` too; adding those two packages to the job would run under `-race` what the body says was run locally.
5. On shutdown, `Execute` returns `context.Canceled` for in-flight checks and the scheduler logs it as `failed to execute check` and bumps retry state on a task about to be discarded; skipping the log and bookkeeping for `context.Canceled` would keep restarts quiet.
6. The `running` marker skips ticks while a check runs and then catches up once on the next tick; worth one sentence in the body since "one in flight at a time" does not say which.
7. The leaf-guard comment cleanup dropped the sentence about which consumer the pre-/v2 guard keeps working; that is the one consequence a future reader cannot recover from the code.

eunyoung14:
One refinement to note 1 above, from tracing the straggler path: if `Scheduler.Stop` ever gets a bounded wait, the three `close(...)` calls at the end of `Collector.Stop` have to go with it. A check that outlives the bound still ends in `publish()`, which sends on `SuccessQueue`; with the queue closed that send is the same unrecoverable panic this PR removes for `taskQueue`. The closes buy nothing (the channels are reclaimed with the collector and the `ok` guards exist only to receive that close), so dropping them is what makes a bounded wait safe. Not needed for this PR to merge; it is the shape #482's follow-up would want to build on.

eunyoung14 (dismissed):
Approved. The shutdown ordering is correct and every send-after-close path is closed by construction; the notes above are follow-ups, not blockers.

jisung-02:
Thanks for reading the shutdown path end to end. All seven notes are in, plus the refinement you added afterwards. The PR body is rewritten against the final diff.

| Note | Where it landed |
|---|---|
| 1. Unbounded `wg.Wait` in `Scheduler.Stop` | be8ff739 |
| Follow-up: the `close(...)` calls have to go with the bound | be8ff739 |
| 2. `TestStop_ReturnsAfterDispatcherWasParkedInASend` does not park | 9c158132 |
| 3. `wg.Add(1)` with no `Done` | 24b14181 |
| 4. Race job misses the new concurrency tests | 966e1d88 |
| 5. `context.Canceled` logged and retried on shutdown | 39ac761b |
| 6. Say which way the `running` marker catches up | PR body, no commit |
| 7. Dropped leaf-guard consumer sentence | 93cd51b3 |

Two of these landed differently from the wording of the note, both deliberately.

**Note 5.** Matching `context.Canceled` with `errors.Is` infers "we are shutting down" from the error value, so a check that returns `context.Canceled` for its own reason would have a real failure skip both the log and the retry bookkeeping. `executeTask` asks `ctx.Err()` instead. `stopChan` alone would not do: `Collector.Stop` cancels the context first and only then closes `stopChan`, so between the two there is a window where the context is done and `stopChan` is not. Writing the test exposed that the first version of it passed a live context and would have gone green either way; it now cancels the context it runs under.

**The follow-up.** `SuccessQueue` and `FailureQueue` are gone as you described, and the `ok` guards that only existed to receive those closes went with them. `close(errorChan)` stays. Nothing in `pkg/` ever sends to `errorChan`, so no straggler can race it, and without the close the `handleErrors` goroutine leaks on every collector restart.

Two things beyond the notes, both from reviewing the result:

- The budget lives in `pkg/agent` as `ShutdownWaitBudget`, five seconds per join, so worst case shutdown is ten seconds. #482's flush should take this constant rather than introduce a second budget beside it.
- The race job also gained `./pkg/agent`, since the bounded-wait helper that lives there spawns a goroutine.

Verification on 93cd51b3:

```
gofmt -l .                                           # clean
go build ./...                                       # ok
go vet ./...                                         # ok
go test -race -count=2 ./pkg/wsclient ./pkg/collector/scheduler ./pkg/collector ./pkg/collector/check/base ./pkg/agent
ok  github.com/alpacax/alpamon/v2/pkg/wsclient               13.149s
ok  github.com/alpacax/alpamon/v2/pkg/collector/scheduler     1.273s
ok  github.com/alpacax/alpamon/v2/pkg/collector               1.270s
ok  github.com/alpacax/alpamon/v2/pkg/collector/check/base    1.206s
ok  github.com/alpacax/alpamon/v2/pkg/agent                   1.231s
go test ./... -p 1                                   # all packages ok
```

eunyoung14 (approved):

Issue #438, opened 2026-08-26
rpm dependency: nftables|iptables alternation is silently dropped


eunyoung14:
## The defect

`.goreleaser.yaml`'s rpm override declares:

```yaml
- nftables | iptables
```

rpm has no `|` alternation. nfpm parses that string with the same regex it uses for `name >= version`, so it becomes a hard `Requires: nftables` with `| iptables` stranded in the (unused) version field. Verified by running nfpm's own parser (`rpmpack v0.7.1`, `toRelation` → `NewRelation`):

```
"nftables | iptables"    -> name="nftables" version="| iptables" sense=0
"(nftables or iptables)" -> name="(nftables or iptables)" version="" sense=0
```

Consequence: the alternative is dropped, and hosts that firewall with iptables are made to pull nftables in.

## Why it isn't fixed yet

rpm spells the alternation as a boolean dependency, `(nftables or iptables)`, which requires **rpm >= 4.13**. That covers RHEL/Rocky/Alma/Oracle 8+, Fedora, Amazon Linux 2023, SLES/Leap 15+ — but **not Amazon Linux 2 or CentOS 7 (rpm 4.11)**, and `README.md` lists Amazon Linux 2 as a supported platform.

The fix was prepared in #437 and then deliberately backed out of it (`ce1a32a`) so the upgrade-path fixes in that PR were not blocked on this decision.

## What needs deciding

1. Is Amazon Linux 2 still a supported target? If it is being dropped or moved to best-effort, `(nftables or iptables)` can land as-is.
2. If AL2 stays supported, the options are: leave the requirement as a hard `Requires: nftables` (today's behaviour, wrong for iptables hosts), drop the firewall dependency entirely and document the prerequisite, or emit a different rpm override per platform generation.

Note: neither AL2 nor CentOS 7 ships an nftables package, so today's requirement does not resolve on them either — this is a correctness/clarity problem rather than a live install failure on those two.

jisung-02:
Confirmed the rpm 4.13 boundary empirically while reviewing #483.

I built a test rpm with nfpm using the boolean form `(nftables or iptables)` and installed it in containers with only `iptables` present:

- openSUSE Leap 15.6: resolves, no nftables pulled in
- Rocky Linux 9: resolves, no nftables pulled in
- Amazon Linux 2 (rpm 4.11.3): fails with `Failed dependencies`, treating `(nftables or iptables)` as a literal package name

[...]

PR #483, opened 2026-09-21
ci(release): smoke-test the released rpm with zypper on openSUSE

### LINKED ISSUE #423: ci(release): smoke-test the released rpm with zypper on openSUSE
## Problem

`release.yml`'s `package-smoke-test` job only installs the built packages on `ubuntu:24.04` (deb) and `rockylinux:9` (rpm). Nothing verifies that the released rpm actually installs with zypper on openSUSE, even though README claims openSUSE Leap 15+ / SLES 15+ support and the registration guide's `install.suse.sh` installs exactly this rpm via zypper.

The nfpm rpm dependencies (`zip, sqlite, nftables | iptables`, `.goreleaser.yaml:136-138`) use EL package names. Today this happens to work: verified empirically with the v2.5.0 rpm — `zypper install --dry-run` succeeds on both `opensuse/leap:15.6` and `opensuse/leap:16.0`, with `sqlite` resolving to openSUSE's `sqlite3` package. But nothing in CI pins this. A future dependency addition using an EL-only name would ship a release that fails with "nothing provides ..." on every SUSE host, and we would only find out from user reports.

alpamon-pam already covers this: its `ci.yml` runs `zypper install ./dist/*.rpm` smoke jobs on `opensuse/leap:15.6` and `16.0`.

## Scope

1. Add an openSUSE entry to the `package-smoke-test` matrix in `release.yml` (e.g. `opensuse/leap:15.6`, optionally `16.0`), mirroring the alpamon-pam pattern: `zypper --non-interactive install ./dist/*.rpm`, then assert the binary and systemd unit land where expected.
2. While here, consider noting in README that SLES and SUSE arm64 are claimed but not CI-verified (source build in `build-and-test.yml` covers `opensuse-leap-15` amd64 only), or extend coverage if cheap.

## Priority

Low — no current breakage (verified above); this is regression protection for the rpm dependency names and postinstall behavior on SUSE.

jisung-02:
Closes #423

## Background

`package-smoke-test` installed the released packages on `ubuntu:24.04` with apt and on `rockylinux/rockylinux:9` with dnf. Nothing installed the rpm with zypper, even though the README lists openSUSE Leap 15+ and SLES 15+ as supported and `docs/opensuse.md` documents zypper as the install path there.

The nfpm rpm dependencies are EL names: `zip`, `sqlite`, and a firewall requirement. They resolve on openSUSE today—`sqlite` maps to the distro's `sqlite3` package—but nothing in CI pins that. A future dependency using an EL-only name would ship a release that fails with "nothing provides ..." on every SUSE host, and the first report would come from a user.

## Changes

Two openSUSE entries in the `package-smoke-test` matrix, `opensuse/leap:15.6` and `opensuse/leap:16.0`, installing with `zypper --non-interactive install --allow-unsigned-rpm`. The narrow flag is deliberate: `--no-gpg-checks` would also skip signature verification for the signed dependency packages zypper pulls from the distro repos, and only the locally built rpm is unsigned.

Both releases are covered because they differ in the layout the assertions read. Leap 15.6 is not usrmerged, so `/lib` is a real directory and its systemd 254 searches both `/usr/lib/systemd/system` and `/lib/systemd/system`. Leap 16.0 is usrmerged and its systemd 257 searches only `/usr/lib/systemd/system`. 15.6 also reached end of life on 2026-04-30, so 16.0 is the only Leap still supported.

The job's `container` moves to the mapping form with a pinned `PATH`, matching `build-and-test.yml`. The pin guards against a later step that prepends to `PATH`: the runner rebuilds `PATH` from the container's `Config.Env` only in that case, and neither the Leap nor the Rocky image sets one. No step in this job writes to `GITHUB_PATH` today, so nothing depends on the pin yet. The pinned value is the default `PATH` on `ubuntu:24.04`, so the existing entries are unaffected.

The smoke test goes past `alpamon --help` and asserts the binary and the systemd unit landed where systemd can read them. `set -x` names the failing assertion, which a bare `test` does not. The unit check accepts either `/usr/lib/systemd/system` or `/lib/systemd/system` rather than pinning nfpm's `dst`: systemd searches `/usr/lib` first on every target here, so a later `dst` change to that directory would otherwise fail this leg on a correct package, at release time, holding `packagecloud-deploy`. The assertions apply to all four matrix entries, not only the new ones.

`docs/opensuse.md` gains a short CI coverage section: what is actually verified on SUSE, and that SLES and arm64 are in the support table without being tested.

## Verification

`release.yml` only runs on a tag push, so this job does not run on the pull request. The matrix entries were reproduced locally against the released v2.7.1 packages on linux/amd64.

The rpm packages the unit at the literal `/lib` path:

```
$ rpm -qpl alpamon-2.7.1-1.linux.amd64.rpm | grep -E 'systemd|bin/alpamon'
/lib/systemd/system/alpamon-restart.service
/lib/systemd/system/alpamon-restart.timer
/lib/systemd/system/alpamon.service
/usr/bin/alpamon
```

That path is reachable at one directory or two depending on the image:

```
opensuse/leap:15.6         drwxr-xr-x 5 root root 4096 Apr 24 18:48 /lib
opensuse/leap:16.0         lrwxrwxrwx 1 root root 7 Jan 28  2025 /lib -> usr/lib
rockylinux/rockylinux:9    lrwxrwxrwx 1 root root 7 Nov  3  2024 /lib -> usr/lib
ubuntu:24.04               lrwxrwxrwx 1 root root 7 Apr 22  2024 /lib -> usr/lib
```

On Leap 15.6 the installed unit is visible only under `/lib/systemd/system`; on the other three both paths resolve to the same file.

zypper installs the rpm and the assertions pass, and deleting the unit still fails the step. Leap 16.0, install trimmed to its last lines:

```
(9/9) Installing: alpamon-2.7.1-1.x86_64 [...
Notice: systemd is not available. Skipping systemd service setup.
Notice: Environment variables not set. Skipping automatic setup.
Please run 'sudo alpamon register' to complete the registration.
done]
Running post-transaction scripts [...done]
=== smoke test ===
+ /usr/bin/alpamon --help
+ test -f /usr/lib/systemd/system/alpamon.service
exit=0
=== unit deleted ===
+ /usr/bin/alpamon --help
+ test -f /usr/lib/systemd/system/alpamon.service
+ test -f /lib/systemd/system/alpamon.service
exit=1
```

The deleted-unit run is what shows the assertion is not vacuous, and the trace shows `set -x` naming the check that failed. Leap 15.6 was reproduced the same way and passes through the second `test -f`, since the unit is reachable only under `/lib` there. Rocky 9 and Ubuntu 24.04 pass on the first. `sqlite` resolves to openSUSE's `sqlite3` on both Leap releases.

The `Config.Env` of each image backs the `PATH` comment:

```
opensuse/leap:15.6         null
opensuse/leap:16.0         null
rockylinux/rockylinux:9    ["container=oci"]
ubuntu:24.04               ["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"]
```

actionlint reports the same two pre-existing diagnostics before and after the change, both on the `actions/create-github-app-token@v3` step at line 69.

An earlier revision of this branch asserted only `/lib/systemd/system/alpamon.service` and explained that `test -f` follows the usrmerge symlink. That is true, but it checks nfpm's `dst` rather than a directory systemd reads, so the assertion was changed rather than kept.

## Not covered here

Local reproduction cannot cover the GitHub runner starting a Leap image as a job container and running `actions/download-artifact@v8` inside it. `build-and-test.yml` already runs a Leap image as a job container, which is the basis for expecting it to work. Cutting a `-dev` or `-rc` tag would exercise the job before a stable release does.

The rpm's firewall dependency is declared as `nftables | iptables`, which rpm does not support: nfpm writes it into the header as name `nftables` with the rest stranded in an unused version field, so the package hard-requires `nftables`. That predates this branch and is tracked in #438. This job does not catch it either—a fresh container has nothing installed, so a hard `Requires: nftables` resolves cleanly. Catching it needs a leg that pre-installs `iptables` and asserts `nftables` is not pulled in.


.github/workflows/release.yml
```diff
@@ -188,7 +188,16 @@ jobs:
             container: rockylinux/rockylinux:9
             arch: amd64
             install_cmd: dnf install -y
-    container: ${{ matrix.container }}
+          # Proves zypper still resolves the rpm's EL dependency names (zip, sqlite, nftables|iptables) on SUSE.
```
geunwoonoh: Warning: The comment names a dependency the rpm doesn't have. nfpm writes `nftables | iptables` into the rpm header as name `nftables`, flags `0`, version `| iptables`, and rpm ignores a version with no comparison flag, so the package requires plain `nftables`. With `iptables` already installed on Leap 15.6, `zypper install --dry-run` of alpamon-2.7.1 still adds `nftables`. The list also copies `.goreleaser.yaml`, so it goes stale the first time a dependency is added, which is the change this job is meant to catch.

```suggestion
          # Proves zypper resolves the rpm's EL dependency names on SUSE.
```

The packaging bug predates this PR and deserves its own issue: a rich dependency `(nftables or iptables)` built with nfpm resolves against `iptables` alone on Leap 15.6 and Rocky 9, but rpm 4.11.3 on Amazon Linux 2 (still in the README support table) rejects it as a literal package name.

.github/workflows/release.yml
```diff
@@ -188,7 +188,16 @@ jobs:
...
+          - pkg_type: rpm
+            container: opensuse/leap:15.6
+            arch: amd64
+            install_cmd: zypper --non-interactive install --allow-unsigned-rpm
+    # PATH is pinned because the runner rebuilds each exec PATH from the container config, and the Leap image alone defines none.
```
geunwoonoh: Warning: The stated cause doesn't hold for this job. The runner overrides an exec's `PATH` only when an earlier step has prepended to it (`actions/runner` `StepHost.cs`: `if (!string.IsNullOrEmpty(PrependPath))`), and no step here writes `GITHUB_PATH`. "The Leap image alone defines none" is also wrong: `rockylinux/rockylinux:9` sets only `container=oci`, and its leg passed in the v2.7.1 release run (35491260712) with no `PATH` in the container env.

Keep the pin, since `release.yml` runs only on tags and a future step that prepends to `PATH` would otherwise break both rpm legs at release time. Correct the reason, wrapped like the rest of the file (every other comment in `release.yml` fits in 80 columns):

```suggestion
    # Pinned so a later step that prepends to PATH keeps /usr/bin: the runner
    # then rebuilds PATH from the container's Config.Env, which neither the
    # Leap nor the Rocky image sets.
```

.github/workflows/release.yml
```diff
@@ -188,7 +188,16 @@ jobs:
             container: rockylinux/rockylinux:9
             arch: amd64
             install_cmd: dnf install -y
-    container: ${{ matrix.container }}
+          # Proves zypper still resolves the rpm's EL dependency names (zip, sqlite, nftables|iptables) on SUSE.
+          - pkg_type: rpm
+            container: opensuse/leap:15.6
+            arch: amd64
+            install_cmd: zypper --non-interactive install --allow-unsigned-rpm
```
geunwoonoh, line 195: Warning: Consider a Leap 16.0 leg next to this one. 15.6 reached end of life on 2026-04-30; 16.0 is the only supported Leap (until 2027-10-31), it falls under "Leap 15+" in the README, and issue #423 names it in scope. The same command and assertions pass on `opensuse/leap:16.0` against v2.7.1. It is not duplicate coverage: 16.0 is usrmerged (`/lib -> usr/lib`) and its systemd 257 searches only `/usr/lib/systemd/system`.

```suggestion
            install_cmd: zypper --non-interactive install --allow-unsigned-rpm
          - pkg_type: rpm
            container: opensuse/leap:16.0
            arch: amd64
            install_cmd: zypper --non-interactive install --allow-unsigned-rpm
```

docs/opensuse.md
```diff
@@ -2,6 +2,10 @@
 
 Support is best-effort. Alpamon detects openSUSE and SLES, runs `zypper` for its own package operations, and starts cleanly, but the console still composes `yum` for some server-driven operations (see Known limitations below).
 
+## CI coverage
+
+CI exercises SUSE on amd64 only. It builds Alpamon from source and runs the test suite on openSUSE Leap 15, and it installs the released rpm with `zypper` on Leap 15.6, checking that the binary and the systemd unit land where the package declares them. SLES and arm64 are in the support table but never tested, so a regression on either reaches users before it reaches us.
```
geunwoonoh: Warning: This lists SLES and arm64 as untested but leaves out Leap 16.0, which is in the support table and is the only supported Leap. Also, `opensuse/leap:15` in `build-and-test.yml` and `opensuse/leap:15.6` are the same image today (`sha256:88b078dba77b…`), so "Leap 15" and "Leap 15.6" read like two releases. With the 16.0 leg added:

```suggestion
CI exercises SUSE on amd64 only. It builds Alpamon from source and runs the test suite on openSUSE Leap 15.6, and it installs the released rpm with `zypper` on Leap 15.6 and 16.0, checking that the binary and the systemd unit land where the package declares them. SLES and arm64 are in the support table but never tested, so a regression on either reaches users before it reaches us.
```

If 16.0 stays out, add "Leap 16" to the untested sentence instead.

.github/workflows/release.yml
```diff
@@ -214,7 +223,11 @@ jobs:
         run: ${{ matrix.install_cmd }} ./${{ env.PKG_NAME }}
 
       - name: Smoke test
-        run: alpamon --help
+        run: |
```
geunwoonoh, line 232: Suggestion: A failing `test` prints nothing, so the log ends in `exit code 1` without saying which check failed. `set -x` names it (or an `::error` per check, as lines 114 and 138 do).

```suggestion
        run: |
          set -x
```

.github/workflows/release.yml
```diff
@@ -214,7 +223,11 @@ jobs:
...
+          alpamon --help
+          test -x /usr/bin/alpamon
```
geunwoonoh: Suggestion: These two lines check the same thing. Running the declared path directly proves the file exists, is executable, and runs; the `PATH` lookup adds nothing, since a fresh container has no other `alpamon` ahead of `/usr/bin`. The path is worth pinning because the unit's `ExecStart=/usr/bin/alpamon` depends on it.

```suggestion
          /usr/bin/alpamon --help
```

.github/workflows/release.yml
```diff
@@ -214,7 +223,11 @@ jobs:
...
+          # nfpm packages the unit to the literal /lib path: a real directory on Leap, a usrmerge symlink elsewhere.
+          test -f /lib/systemd/system/alpamon.service
```
geunwoonoh: Suggestion: This pins nfpm's `dst` rather than a directory systemd reads. On Leap 15.6, `/lib/systemd` does not exist until the package creates it, so moving `dst` to `/usr/lib/systemd/system` (which systemd searches first on every target here) would fail this leg on a correct package, at release time, holding `packagecloud-deploy`. Accepting either directory passes on Leap 15.6, Leap 16.0 and Rocky 9, and still fails when the unit is missing:

```suggestion
          # Leap 15's systemd searches both directories; usrmerged hosts have only one.
          test -f /usr/lib/systemd/system/alpamon.service || test -f /lib/systemd/system/alpamon.service
```

geunwoonoh (commented):
## PR review: #483 - ci(release): smoke-test the released rpm with zypper on openSUSE

**Stack**: common (GitHub Actions YAML, Markdown)
**Changes**: +19 -2 in 2 files

### Summary
The new leg works as the PR describes. `release.yml` only runs on tags, so I reproduced it locally against the v2.7.1 rpm on `opensuse/leap:15.6`, `opensuse/leap:16.0` and `rockylinux/rockylinux:9` (linux/amd64). On Leap 15.6, `zypper` exits 0 and every assertion passes. actionlint gives the same diagnostics before and after the change; only the line numbers move. The problems are all in the text: two comments state facts that aren't true, and the docs leave Leap 16.0 off the list of untested platforms.

### Critical Issues (0)

### Warnings (3)

1. **[.github/workflows/release.yml:191]** The comment names a dependency the rpm doesn't have. nfpm writes `nftables | iptables` into the rpm header as name `nftables`, flags `0`, version `| iptables`. rpm ignores a version that has no comparison flag, so the package really requires plain `nftables`. To confirm, I installed `iptables` on Leap 15.6 first, and `zypper install --dry-run` of alpamon-2.7.1 still added `nftables`. The comment also copies the dependency list from `.goreleaser.yaml`, so it goes stale the first time someone adds a dependency, which is exactly the change this job is meant to catch.

   The packaging bug was already there before this PR and deserves its own issue. I built a test rpm with nfpm using the rich dependency `(nftables or iptables)`. On Leap 15.6 and Rocky 9 it resolves with only `iptables` installed. But rpm 4.11.3 on Amazon Linux 2, which the README still lists as supported, treats it as a literal package name and fails with `Failed dependencies`.

2. **[.github/workflows/release.yml:196]** The reason the comment gives for pinning `PATH` doesn't hold for this job:
   - The runner only overrides an exec's `PATH` when an earlier step has added to it (`actions/runner` `StepHost.cs`: `if (!string.IsNullOrEmpty(PrependPath))`). No step in `package-smoke-test` writes to `GITHUB_PATH`.
   - "The Leap image alone defines none" is wrong. `rockylinux/rockylinux:9` doesn't set `PATH` either (its `Config.Env` is only `container=oci`), and its leg passed in the v2.7.1 release run (35491260712) without one.

   Keep the pin anyway. `release.yml` only runs on tags, so a future step that adds to `PATH` would break both rpm legs at release time. Fix the comment's reason, and wrap it like the rest of the file: every other comment in `release.yml` fits in 80 columns, while the three new ones are 112–129.

   `build-and-test.yml:197` has the same outdated wording since 560b196 moved Rocky to `rockylinux/rockylinux`. That's outside this diff, so it's a follow-up.

3. **[docs/opensuse.md:7]** The paragraph says SLES and arm64 are untested, but leaves out Leap 16.0.
   - Leap 16.0 is covered by the support table ("Leap 15+"), and it's the only Leap still supported: 15.6 reached end of life on 2026-04-30, and 16.0 is supported until 2027-10-31.
   - Issue #423 lists 16.0 in its scope, and alpamon-pam's CI already runs both releases.
   - The PR's command and assertions pass on `opensuse/leap:16.0` with v2.7.1.
   - It isn't duplicate coverage. 16.0 is usrmerged (`/lib -> usr/lib`) and its systemd 257 only searches `/usr/lib/systemd/system`. On 15.6, `/lib` is a real directory and systemd 254 searches both paths.

   `opensuse/leap:15` in `build-and-test.yml` and `opensuse/leap:15.6` are currently the same image (`sha256:88b078dba77b…`). As written, "Leap 15" and "Leap 15.6" read like two different releases. If you leave the 16.0 leg out, add "Leap 16" to the untested sentence instead.

### Suggestions (3)

1. **[.github/workflows/release.yml:227-228]** `alpamon --help` and `test -x /usr/bin/alpamon` check the same thing. Running the declared path directly proves the file exists, is executable, and runs. The `PATH` lookup adds nothing, because a fresh container has no other `alpamon` ahead of `/usr/bin`. Pinning the path does matter, because the unit's `ExecStart=/usr/bin/alpamon` depends on it.
2. **[.github/workflows/release.yml:229-230]** The unit check asserts nfpm's `dst` path, not a directory that systemd reads. On Leap 15.6, `/lib/systemd` doesn't exist until the package creates it. If someone later moves `dst` to `/usr/lib/systemd/system`, which systemd searches first on every target, this leg would fail even though the package is fine. Since the job only runs on tags, that false failure would show up at release time and block `packagecloud-deploy`.
3. **[.github/workflows/release.yml:226]** When `test` fails it prints nothing, so the log just ends with `exit code 1` and doesn't say which check failed. Add `set -x`, or an `::error` line per check like lines 114 and 138 do.

All three together. I checked this on Leap 15.6, Leap 16.0 and Rocky 9: it passes, and deleting the unit file makes it fail with exit 1.

```yaml
        run: |
          set -x
          /usr/bin/alpamon --help
          # Leap 15's systemd searches both directories; usrmerged hosts have only one.
          test -f /usr/lib/systemd/system/alpamon.service || test -f /lib/systemd/system/alpamon.service
```

### Additional checks
- **Go conventions**: don't apply, since the PR changes no Go files.
- **Duplicate or unused code**:
  - The `test -x` line repeats what `alpamon --help` already checks (Suggestion 1).
  - The comment's dependency list copies `.goreleaser.yaml` (Warning 1).
  - The `PATH` pin isn't needed today, but it's worth keeping as a guard (Warning 2).
  - The change leaves nothing unused.
- **Performance**: one more leg, and it runs in parallel. `packagecloud-deploy` waits for the slowest leg, which will be Leap 15.6.
  - Locally, refreshing its 7 repos downloaded 183 MB and took 48 s. Leap 16.0 has 2 repos: 34 MB in 12 s. The Rocky leg's `dnf install` took about 12 s in the v2.7.1 run.
  - That's under a minute added to each release. No change needed: disabling repos would test a narrower setup than real hosts have.
  - If a repo refresh fails transiently, zypper exits 106 and the leg fails. "Re-run failed jobs" fixes that, and the Leap job in `build-and-test` already has the same exposure.

### Highlights
- Using `--allow-unsigned-rpm` instead of `--no-gpg-checks` works as intended. The install log shows "Package header is not signed!" only for alpamon; the distro packages are still signature-checked.
- `packagecloud-deploy` needs `package-smoke-test`, so if SUSE can't resolve the dependencies, the `rpm_any` upload stops. It's a real gate, not just a red check.
- zypper exits 107 when `%post` fails, so this leg also catches a broken postinstall. On the Rocky leg, dnf treats that as a non-fatal warning.
- The verification in the PR body reproduced as written.

---

**Verdict**: COMMENTED

**Checklist**:
- [x] Code follows stack conventions (except comment width, Warning 2)
- [ ] Tests included and passing: the job only runs on tags, so I reproduced it locally
- [ ] Documentation updated if needed: Leap 16 is missing (Warning 3)

.github/workflows/release.yml
jisung-02: Applied in 3181b780. You are right that the comment named a dependency the rpm does not carry: rpm ignores a version with no comparison flag, so `| iptables` is stranded and the package requires plain `nftables`. The list also duplicated `.goreleaser.yaml`, which is exactly the thing this job is supposed to notice changing. The comment is now just the claim the job actually proves.

The packaging defect is already open as #438, which reached the same conclusion from nfpm's parser (`rpmpack v0.7.1`, name `nftables`, version `| iptables`, sense 0) and is blocked on whether Amazon Linux 2 stays supported, since `(nftables or iptables)` needs rpm 4.13. I added your Amazon Linux 2 result there, because #438 had only inferred it from the rpm version.

.github/workflows/release.yml
jisung-02: Applied in 3181b780, with your wording. Both corrections hold up: no step in this job writes to `GITHUB_PATH`, and the Leap image is not alone in defining no `PATH`. I checked the four images' `Config.Env` directly:

```
opensuse/leap:15.6         null
opensuse/leap:16.0         null
rockylinux/rockylinux:9    ["container=oci"]
ubuntu:24.04               ["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"]
```

The pin stays as the guard you described. All three new comments are now inside 80 columns, matching the rest of the file.

.github/workflows/release.yml
jisung-02: Added in d27661bc. The end-of-life argument settles it: 15.6 went end of life on 2026-04-30, so covering only that release means CI tests no supported Leap.

It is also not duplicate coverage, which I confirmed against the images rather than taking it on trust:

```
opensuse/leap:15.6         drwxr-xr-x 5 root root 4096 Apr 24 18:48 /lib
opensuse/leap:16.0         lrwxrwxrwx 1 root root 7 Jan 28  2025 /lib -> usr/lib
```

The rpm packages the unit at `/lib/systemd/system/alpamon.service`, so on 15.6 it is reachable at one path and on 16.0 at two. The two legs exercise different sides of the unit assertion.

docs/opensuse.md
jisung-02: Applied in d27661bc, with your sentence. The "Leap 15" versus "Leap 15.6" ambiguity was worth catching too, so the source-build line now says 15.6 as well, since `opensuse/leap:15` resolves to the same image.

.github/workflows/release.yml
jisung-02: Applied in 8b48d72e. The trace is what makes a failure readable: on Leap 16.0 with the unit removed, the step now prints which check ran before it gave up.

```
+ /usr/bin/alpamon --help
+ test -f /usr/lib/systemd/system/alpamon.service
+ test -f /lib/systemd/system/alpamon.service
exit=1
```

.github/workflows/release.yml
jisung-02: Applied in 8b48d72e. Running the declared path does cover both, and pinning it is the part worth keeping because the unit's `ExecStart` names it.

.github/workflows/release.yml
jisung-02: Applied in 8b48d72e. Your failure mode is real, and the image layouts show why: on Leap 15.6 `/lib` is a real directory and the installed unit is visible only under `/lib/systemd/system`, while the other three images are usrmerged and see it at both paths. Pinning `dst` would have failed the 15.6 leg the moment the package moved the unit to `/usr/lib`, at release time.

The `||` form passes on all four entries and still fails when the unit is gone, which I checked by deleting both paths after install:

```
=== smoke test ===
+ /usr/bin/alpamon --help
+ test -f /usr/lib/systemd/system/alpamon.service
exit=0
=== unit deleted ===
+ /usr/bin/alpamon --help
+ test -f /usr/lib/systemd/system/alpamon.service
+ test -f /lib/systemd/system/alpamon.service
exit=1
```

I wrapped the comment to two lines so it stays inside 80 columns.

jisung-02:
All seven comments are addressed. Each reply names its commit; the mapping:

| Comment | Change | Commit |
|---|---|---|
| Dependency comment names a dependency the rpm lacks | Comment reduced to what the job proves | 3181b780 |
| `PATH` pin reason does not hold | Reason corrected, wrapped to 80 columns | 3181b780 |
| Leap 16.0 leg missing | Second matrix entry added | d27661bc |
| Docs leave out Leap 16.0 | Sentence replaced, `Leap 15` also read as 15.6 | d27661bc |
| Failing `test` says nothing | `set -x` | 8b48d72e |
| `--help` and `test -x` overlap | Collapsed to `/usr/bin/alpamon --help` | 8b48d72e |
| Unit check pins nfpm's `dst` | Either systemd directory accepted | 8b48d72e |

Verification on the pushed head, linux/amd64 against the released v2.7.1 rpm. The rpm packages the unit at the literal `/lib` path:

```
$ rpm -qpl alpamon-2.7.1-1.linux.amd64.rpm | grep -E 'systemd|bin/alpamon'
/lib/systemd/system/alpamon-restart.service
/lib/systemd/system/alpamon-restart.timer
/lib/systemd/system/alpamon.service
/usr/bin/alpamon
```

Which resolves to one directory or two depending on the image:

```
opensuse/leap:15.6         drwxr-xr-x 5 root root 4096 Apr 24 18:48 /lib
opensuse/leap:16.0         lrwxrwxrwx 1 root root 7 Jan 28  2025 /lib -> usr/lib
rockylinux/rockylinux:9    lrwxrwxrwx 1 root root 7 Nov  3  2024 /lib -> usr/lib
ubuntu:24.04               lrwxrwxrwx 1 root root 7 Apr 22  2024 /lib -> usr/lib
```

Leap 16.0, install and both smoke-test runs:

```
(9/9) Installing: alpamon-2.7.1-1.x86_64 [...
Notice: systemd is not available. Skipping systemd service setup.
Notice: Environment variables not set. Skipping automatic setup.
Please run 'sudo alpamon register' to complete the registration.
done]
Running post-transaction scripts [...done]
=== smoke test ===
+ /usr/bin/alpamon --help
+ test -f /usr/lib/systemd/system/alpamon.service
exit=0
=== unit deleted ===
+ /usr/bin/alpamon --help
+ test -f /usr/lib/systemd/system/alpamon.service
+ test -f /lib/systemd/system/alpamon.service
exit=1
```

actionlint still reports only the two pre-existing diagnostics on the `actions/create-github-app-token@v3` step at line 69, unchanged by this branch.

On the points raised in the review body rather than inline:

- The rpm dependency defect is already open as #438, which reached the same conclusion from nfpm's parser and is blocked on whether Amazon Linux 2 stays supported. Your Amazon Linux 2 result is now a comment there, since #438 had inferred that boundary from the rpm version rather than measured it. I also noted there that this job cannot catch the dropped alternative, because a fresh container resolves a hard `Requires: nftables` cleanly.
- The stale `PATH` wording at `build-and-test.yml:197` is outside this diff and stays a follow-up, as you suggested. No issue filed for it yet.
- On release time: the second Leap leg runs in parallel with the first, and Leap 16.0 refreshes two repos rather than seven, so it should not become the leg `packagecloud-deploy` waits on.

geunwoonoh (approved):
## PR review: #483 - ci(release): smoke-test the released rpm with zypper on openSUSE

**Stack**: common
**Changes**: +26 -2 in 2 files

### Summary
All seven findings from the previous review are fixed in 3181b78, d27661b and 8b48d72, and each thread has a reply naming the commit. What I re-checked at 8b48d72:
- **All four matrix legs pass.** I used the install command and `Smoke test` block taken from `release.yml` at this head, against v2.7.1 (linux/amd64). `ubuntu:24.04`, `rockylinux/rockylinux:9`, `opensuse/leap:15.6` and `opensuse/leap:16.0` all install with exit 0 and pass the smoke test. Leap 15.6 passes on the second `test -f`, as the new comment says.
- **Leap 16.0 can run as a job container.** It hasn't been used as one anywhere in this repo's CI yet. Inside it, `tail` (the runner's container entrypoint) and `bash` are present, and the runner's bundled Node 24.21.0 runs and completes an HTTPS request, which is what `actions/download-artifact@v8` needs.
- **Lint and CI are clean.** actionlint reports only the diagnostics that were already there, and every comment in `release.yml` now fits in 80 columns. Build and Test is green on the PR, including `opensuse-leap-15`.
- **The nftables alternation is tracked separately** in #438. The PR body now says this job doesn't catch it, and what a leg that does would need.

### Critical Issues (0)

### Warnings (0)

### Suggestions (1)
1. **[commit 8b48d72]** The commit message has a spaced em dash ("on a correct package — at release time"). The repo's writing conventions rule that out. This repo squash-merges with the commit messages as the body, so the dash will land on main unless it's fixed in the squash message at merge time.

### Highlights
- Each fix commit explains its change with the evidence behind it, and the PR body was updated to match, including what the job still doesn't catch.
- Because the unit check accepts either directory, it tests what systemd reads rather than nfpm's `dst`. The deleted-unit run in the PR body shows it still fails when the unit is missing.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing (`release.yml` only runs on tags, so I reproduced all four legs locally)

PR #484, opened 2026-09-22
fix(runner): give WebsocketClient a single owner for its connection

### LINKED ISSUE #452: WebsocketClient has no owner for Conn: unsynchronized field, two readers, and a ShutDown that panics twice

jisung-02:
Closes #452.

## Problem

`WebsocketClient.Conn` had no owner. The read loop replaced the pointer from `Connect` while `Close` read it from another goroutine, and `CommandRequestHandler` closed the socket from inside the loop's own call stack. gorilla/websocket permits no concurrent reader and no concurrent writer, so these paths raced each other over one connection. `ShutDown` and `Restart` each closed a channel unconditionally, which panicked on a second call and, when both fired, left root.go's select to pick between quitting and restarting at random.

## Approach

The read loop is the connection's single owner. A mutex guards the `Conn` pointer, and `conn`/`swapConn` hand out the one connection a call should act on, so no caller works from a pointer that changed under it. I/O stays outside the lock.

A handler no longer closes or replaces the connection. It returns a `handlerOutcome`, and the read loop performs the reconnect, which is also what reapplies the read limit that a new connection starts without. `Close` sends a close frame without draining the peer's reply, because draining is a read and `Close` runs outside the loop; `closeAndDrain` keeps the drain for the one caller that owns the reads.

Peer-requested reconnects are paced. `connectForever` backs off only after a failed dial, so a console asking for a reconnect in a loop drove an unbounded stream of successful dials. Requests now wait out `peerReconnectMinInterval`, and the first one still reconnects at once.

`sync.Once` lets only the first of `ShutDown` and `Restart` close its channel.

## Compatibility

`Conn` stays exported, since plugin repos outside this one read it. It is marked `Deprecated`, with the note on its own paragraph so staticcheck recognizes it, and points those plugins at `SetReadDeadline`, `ReadMessage` and `WriteJSON` when they bump alpamon. staticcheck v0.8.1 on a throwaway package reading `wc.Conn` reports `wc.Conn is deprecated: use SetReadDeadline, ReadMessage and WriteJSON. (SA1019)`. Inside these packages, use the accessors.

Two exported methods are removed, which is a breaking change for the v2 module and belongs in the release notes:

- `runner.WebsocketClient.CommandRequestHandler`
- `pluginclient.Client.HandleMessage`

Once the read loop owns the reconnect, each wrapper could only discard the `handlerOutcome` it got back, so a reconnect frame arriving through it was accepted and then ignored. Keeping a method that silently does nothing is worse than a compile error. Nothing in this repo called either one, and the power, repository and cert-authority plugins each have their own unexported `commandRequestHandler`, so none of them breaks. Callers use `RunForever`, which acts on the outcome.

Only the `reconnect` query is paced. A peer that closes right after accepting takes the read-error path, where `connectForever` backs off only after a failed dial, and the external plugins run their own reconnect loops. That path predates this PR and is left as is.

## Tests

Every commit builds and passes on its own. The `race` job in `build-and-test.yml` now runs `pkg/runner` and `pkg/pluginclient` as well, twice over, so the concurrent-close test in `client_close_test.go` is checked by the race detector in CI and not only locally:

```
go test -race -count=2 ./pkg/runner ./pkg/pluginclient
ok  github.com/alpacax/alpamon/v2/pkg/runner        83.759s
ok  github.com/alpacax/alpamon/v2/pkg/pluginclient   3.463s
```

`TestRunForever_PacesRepeatedPeerReconnectRequests` in `pkg/runner` drives a real socket against a peer that asks for a reconnect on every connection, and measures the gap between the two close frames that reach the server. That gap leaves out the dial and the drain, so an unpaced loop lands under 1% of the interval and the assertion can keep a half-interval margin against CI jitter. Reverting the loop to the unpaced `CloseAndReconnect` fails it, so the pacing cannot be dropped silently. The pluginclient counterpart only shows that its loop takes the paced path: once the second connection lands, no third one arrives within a second. The peer server both tests use lives in `internal/testutil`.

The `pluginclient` handler tests call `handleMessage` directly and assert the `handlerOutcome` it returns, instead of only asserting that it does not panic.


pkg/pluginclient/client.go
```diff
@@ -113,17 +122,23 @@ func New(
 	}
 }
 
-// HandleMessage runs the shared dispatch for one inbound frame. It
+// HandleMessage is kept for external callers, but a reconnect request is
+// silently dropped here: only handleMessage's caller acts on it.
+func (c *Client) HandleMessage(ctx context.Context, message []byte) {
+	_ = c.handleMessage(ctx, message)
+}
```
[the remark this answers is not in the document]
jisung-02: Already addressed in e94021e2: the `HandleMessage` wrapper is removed rather than taught to close the socket. No caller in this repo or in the plugin repos invoked it (power-plugin and cert-authority have their own handlers; dhcp, dns and proxy hand everything to `pluginclient.RunForever`), and closing from outside the read loop is exactly the second owner this PR removes. The tests now call `handleMessage` and assert the outcome.

pkg/runner/client.go
```diff
@@ -268,15 +368,21 @@ func (wc *WebsocketClient) RestartCollector() {
 	}
 }
 
+// CommandRequestHandler is kept for external callers, but a reconnect request
+// is silently dropped here: only commandRequestHandler's caller acts on it.
 func (wc *WebsocketClient) CommandRequestHandler(message []byte) {
+	_ = wc.commandRequestHandler(message)
+}
```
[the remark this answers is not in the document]
jisung-02: Fixed in d3f1ecde by dropping the `CommandRequestHandler` wrapper, mirroring what e94021e2 did for `HandleMessage`. Closing from the wrapper would reintroduce a closer outside the read loop, which is what this PR removes on purpose. No caller in this repo or in the sibling repos used the exported method; the read loop calls `commandRequestHandler` and acts on the returned outcome.

pkg/runner/client.go
```diff
@@ -202,26 +268,60 @@ func (wc *WebsocketClient) Connect(ctx context.Context) error {
...
 	return wc.Connect(ctx)
 }
 
-// Cleanly close the websocket connection by sending a close message, waiting for
-// the peer's reply when that frame went out, and closing the connection either way.
-// Do not close quitChan, as the purpose here is to disconnect the WebSocket,
-// not to terminate RunForever.
+// CloseAndReconnectOnRequest paces a peer-requested reconnect so a
+// malfunctioning or compromised console cannot drive an unbounded dial loop.
+// Call it only from the read loop: it calls CloseAndReconnect, which owns the reads.
```
geunwoonoh: Warning: this comment promises more than the code delivers. Only the `reconnect` query is paced. A console that accepts the connection and closes it right away sends the loop through a `ReadMessage` error into `CloseAndReconnect`, and `connectForever` waits only after a failed dial (`auth_backoff.go:202-208`), so that path redials with no delay. In pluginclient, a frame over `MaxMessageSize` takes the same path. The power, repository and cert-authority plugins call `WsClient.Close()` from their own loop on `reconnect`, so they stay unpaced after they bump alpamon.

The unpaced path predates this PR, so it does not need closing here, but one of these should happen:

- (a) Narrow the comment to what the code does (suggestion below).
- (b) Move the pacing into `Connect` as a minimum interval since the last successful connect. That covers every path, the external plugins included, and removes the need for both `lastPeerReconnect` and the newly exported `CloseAndReconnectOnRequest`.

```suggestion
// CloseAndReconnectOnRequest paces a peer-requested reconnect, so a console
// repeating the reconnect query cannot drive an unbounded dial loop.
// Call it only from the read loop: it calls CloseAndReconnect, which owns the reads.
```

geunwoonoh (commented):
### Summary
Moving connection ownership to the read loop is the right direction, and the implementation is clean. Nothing here blocks the merge. Two comments promise more than the code does, the CI race job does not run the packages this PR changes, and the PR description no longer matches the last commit.

### Critical issues (0)
None.

### Warnings (4)
1. **pkg/runner/client.go:292-294**: the pacing comment covers more than the code does. Only the `reconnect` query is paced; a peer that closes right after accepting takes the read-error path with no delay. Details inline.
2. **pkg/runner/client_close_test.go:130**: the race test cannot fail in CI, because the `race` job in `build-and-test.yml` does not run `./pkg/runner` or `./pkg/pluginclient`. Details inline.
3. **PR description**: the Compatibility section says "`CommandRequestHandler` stays as a wrapper", but d3f1ecd removed it, and only the `HandleMessage` removal is described. Removing two exported methods from a v2 module is a breaking change and belongs in the release notes. Nothing actually breaks: the power, repository and cert-authority plugins each have their own unexported `commandRequestHandler`. The repository plugin was missing from the earlier reply, and it is unaffected too.
4. **pkg/runner/client.go:25-26**: the `handlerOutcome` comment says the handler never touches the socket, but it writes the pong. Details inline.

### Suggestions (11)
**Tests**
- `pkg/pluginclient/client_test.go:150`: duplicates the `reconnect` case of the test at line 94.
- `pkg/runner/client_reconnect_test.go:101`: passes with the pacing wait removed.
- `pkg/runner/client_reconnect_test.go:202`, `pkg/pluginclient/client_test.go:262`: the 50ms `dialJitter` margin can flake on Linux runners.
- `pkg/pluginclient/client_test.go:233`: waits the real 5s; 1s is enough.

**Go conventions**
- `pkg/runner/client.go:152`: `getConn` → `conn()`, matching `ControlClient`.
- `pkg/runner/client.go:27`: keep the enum type next to its const block.
- `pkg/runner/client.go:33`: mutex placement, and an optional `Deprecated:` on `Conn`.
- `pkg/runner/client.go:251`: `ReadMessage` and `WriteJSON` return `net.ErrClosed` without saying so.

**Duplicate and unused code**
- `pkg/runner/client_reconnect_test.go:118`: `reconnectPeerServer` is copied verbatim into pluginclient.
- `pkg/pluginclient/client.go:266`: unreachable nil check in `setReadLimit`.
- `pkg/runner/client.go:317-318`: the comment names its caller instead of its contract.
- `handlerOutcome` is defined in both packages. That is fine as is: it is an unexported two-value type, and sharing it would mean exporting it from runner.

### Performance impact

| Change | Effect |
|---|---|
| `Close()` no longer waits for the close reply | **Faster**: shutdown no longer waits up to 5s when the peer does not answer |
| `reconnect` handling | **Slightly faster**: previously close with drain → failed `ReadMessage` → a second close on an already-closed conn → dial. Now one close, then the dial |
| Pacing peer reconnect requests | **Better**: repeated requests no longer cause a dial storm. A second legitimate request within 5s now waits up to 5s, and the loop does not read meanwhile (intended) |
| `getConn` mutex | **Negligible**: one uncontended lock per message, tens of nanoseconds next to the JSON parse |
| Test time | **Slower**: the pluginclient test adds about 5s on every CI platform; 1s with the `require.Never` suggestion |

### Highlights
- The handler returns a `handlerOutcome` instead of closing the socket, so ownership is visible in the code. The same change fixes the read limit that pluginclient did not reapply after a reconnect.
- I/O stays outside the lock. `ControlClient.Close` holds its lock through a `WriteControl` of up to 5s; this PR avoids that.
- `swapConn` closes the replaced conn, so no fd leaks.
- One `sync.Once` ends the `ShutDown`/`Restart` race, and the PR names the actual consequence: root.go's select picking one at random.
- Tests assert the returned outcome instead of `NotPanics`. The test at `client_reconnect_test.go:85` sets `connectBackoff` so a regression fails on the assertion instead of panicking.

### Out of scope (pre-existing, not for this PR)
- The runner's main connection has no read limit: pluginclient is the only caller of `SetReadLimit`. A frame of any size is held in memory, and on a parse failure `commandRequestHandler` (`client.go:378`) logs the whole message, while pluginclient trims it to a 200-byte preview.
- `SendPingQuery` (`client.go:233`) has no caller in this repo. It is exported, so check external users before removing it.

---

**Verdict**: COMMENTED

**Checklist**:
- [x] Code follows stack conventions (minor nits in the suggestions)
- [ ] Tests included and passing: CI is green, but the race job skips `pkg/runner` and `pkg/pluginclient`
- [x] No security vulnerabilities (see warning 1 for the pacing scope)
- [ ] Documentation updated: the PR description's Compatibility section is stale

pkg/runner/client_close_test.go
```diff
@@ -99,10 +119,29 @@ func TestWebsocketClientClose_ClosesTheConnItStartedWith(t *testing.T) {
...
-	firstTracked.onWrite = func() { wc.Conn = second }
+	firstTracked.onWrite = func() { wc.swapConn(second) }
 
 	wc.Close()
 
 	require.True(t, firstTracked.closed.Load(), "Close() left the connection it started with open")
 	require.False(t, secondTracked.closed.Load(), "Close() closed the connection Connect() had just swapped in")
 }
+
+func TestWebsocketClientClose_IsRaceFreeAgainstTheReadLoop(t *testing.T) {
```
geunwoonoh, line 130: Warning: this test cannot fail in CI. Its comment says the real guard is the race detector, but the `race` job in `.github/workflows/build-and-test.yml` runs only `./pkg/wsclient`, the collector packages and `./pkg/agent`, not `./pkg/runner` or `./pkg/pluginclient`. The final `assert.Same` also holds under any interleaving, because `Close` never writes `Conn`. The `-race` run in the PR description was local.

Add both packages to the race job. At about 37s for `pkg/runner` locally, `-count=2` adds a little over a minute to CI.

pkg/runner/client.go
```diff
@@ -21,19 +22,20 @@ import (
...
-	maxConnectInterval    = 60 * time.Second
-	ConnectionReadTimeout = 35 * time.Minute
-
-	eventCommandAckURL    = "/api/events/commands/%s/ack/"
-	eventCommandFinURL    = "/api/events/commands/%s/fin/"
-	eventCommandRejectURL = "/api/events/commands/%s/reject/"
-	eventCommandChunkURL  = "/api/events/commands/%s/chunk/"
-)
+// handlerOutcome carries what a message handler wants done with the connection.
+// Only the read loop acts on it; the handler never touches the socket itself.
```
geunwoonoh: Warning: this comment is wider than the code. `commandRequestHandler` writes to the socket when it answers a ping with `SendPongResponse`. What the handler never does is close or replace the connection.

```suggestion
// handlerOutcome carries what a message handler wants done with the connection.
// Only the read loop acts on it; the handler never closes or replaces the connection.
```

pkg/pluginclient/client_test.go
```diff
@@ -140,3 +134,132 @@ func TestEnqueueConfigUpdate_LastWriteWins(t *testing.T) {
...
+	c := &Client{PluginName: "alpamon-test-plugin", WsClient: &runner.WebsocketClient{}}
+	msg, err := json.Marshal(map[string]string{"query": "reconnect"})
+	require.NoError(t, err)
+
+	outcome := c.handleMessage(context.Background(), msg)
+
+	assert.Equal(t, outcomeReconnect, outcome)
+}
+
+func TestHandleMessage_ReconnectWithoutWsClientDoesNotAskForReconnect(t *testing.T) {
```
geunwoonoh: Suggestion: this case already exists. `TestHandleMessage_PingQuitReconnectRestart_NoopWithoutWsClient` (line 94) sends `reconnect` with a nil `WsClient` and asserts `outcomeContinue`. Delete this test.

pkg/runner/client_reconnect_test.go
```diff
@@ -0,0 +1,205 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 50*time.Millisecond)
+	defer cancel()
+
+	err := wc.CloseAndReconnectOnRequest(ctx)
+
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+	assert.False(t, tracked.closed.Load(), "the second request dialled instead of waiting")
+}
+
+func TestCloseAndReconnectOnRequest_CancelledContextReturnsCanceledWithoutDialing(t *testing.T) {
```
geunwoonoh: Suggestion: this test passes with the pacing wait removed, because `CloseAndReconnect` returns `ctx.Err()` on its first line. It checks the same thing as `TestCloseAndReconnect_DoesNotReconnectAfterContextCancel` (line 52), and cancellation during the wait is already covered by `TestCloseAndReconnectOnRequest_SecondRequestRightAfterWaits` (line 85). Delete it.

pkg/runner/client_reconnect_test.go
```diff
@@ -0,0 +1,205 @@
...
+	cancel()
+	<-done
+
+	// The first peer-requested reconnect is never paced, so the gap to measure is
+	// the one after it.
+	conns := s.connections()
+	require.GreaterOrEqual(t, len(conns), 3)
+	// The pacing timer starts when the reconnect is requested, but the server only
+	// sees when each dial landed, so a slower dial before the gap shortens it.
+	const dialJitter = 50 * time.Millisecond
```
geunwoonoh: Suggestion: the 50ms `dialJitter` margin can flake on a loaded Linux runner (same at `pkg/pluginclient/client_test.go:262`). The measured gap is `interval + (second reconnect's duration) - (first reconnect's duration)`, so the test fails whenever the first close, drain and dial take 50ms longer than the second. Timing flakes like this tend to stay hidden on macOS and show up in Linux containers.

Record when the close frame arrives in the server's close handler and measure that gap instead. It leaves out the dial and the drain, so the only error left is the time to write one close frame.

pkg/pluginclient/client_test.go
```diff
@@ -140,3 +134,132 @@ func TestEnqueueConfigUpdate_LastWriteWins(t *testing.T) {
...
+
+func (s *reconnectPeerServer) connections() []time.Time {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	out := make([]time.Time, len(s.connectedAt))
+	copy(out, s.connectedAt)
+	return out
+}
+
+func TestRunForever_PacesRepeatedPeerReconnectRequests(t *testing.T) {
```
geunwoonoh: Suggestion: this test waits the real 5s. The pluginclient side only needs to show that the loop takes the paced path; the runner test already checks the interval. Once the second connection lands, assert with `require.Never` that no third one arrives within 1s. An unpaced loop redials within milliseconds, so 1s is a wide margin, and the test drops from 5s to 1s without depending on `dialJitter`. The comment at lines 181-182 describes this test, not the server type, so it can move here too.

pkg/runner/client.go
```diff
@@ -122,6 +149,41 @@ func (wc *WebsocketClient) SetOnAuthenticated(fn func()) {
 	wc.onAuthenticated.Store(&fn)
 }
 
+func (wc *WebsocketClient) getConn() *websocket.Conn {
```
geunwoonoh: Suggestion: `ControlClient` in the same package names this accessor `conn()` (`control_client.go:96`), and Effective Go leaves `Get` off getters. Rename it to `conn()` to match.

pkg/runner/client.go
```diff
@@ -21,19 +22,20 @@ import (
...
-	ConnectionReadTimeout = 35 * time.Minute
-
-	eventCommandAckURL    = "/api/events/commands/%s/ack/"
-	eventCommandFinURL    = "/api/events/commands/%s/fin/"
-	eventCommandRejectURL = "/api/events/commands/%s/reject/"
-	eventCommandChunkURL  = "/api/events/commands/%s/chunk/"
-)
+// handlerOutcome carries what a message handler wants done with the connection.
+// Only the read loop acts on it; the handler never touches the socket itself.
+type handlerOutcome int
```
geunwoonoh, line 27: Suggestion: the `WebsocketClient` struct now sits between `handlerOutcome` and its iota block (lines 82-85); `pkg/pluginclient/client.go` splits them the same way (lines 36 and 90-93). Go code usually puts the const block right after `type X int`. The type, const, var, func order is not a convention in this repo either: `pkg/wsclient/client.go` declares a var before its types, and `pkg/wsclient/config.go` mixes const, type and var. The reorder also grew the diff. At minimum, keep `handlerOutcome` and its constants together.

pkg/runner/client.go
```diff
@@ -21,19 +22,20 @@ import (
...
 
 type WebsocketClient struct {
-	Conn                 *websocket.Conn
+	Conn *websocket.Conn
+	// mu guards Conn: Connect writes it from the read loop; Close reads it on shutdown.
+	// Conn stays exported for outside readers; inside this package use the accessors.
+	mu sync.Mutex
```
geunwoonoh, line 36: Suggestion: a mutex conventionally sits right above the fields it guards; here it is below `Conn`. Separately, the external plugins read `Conn` only from their own read loop goroutine, so they are safe today, but nothing enforces that. Optionally, mark the field `// Deprecated: use SetReadDeadline, ReadMessage and WriteJSON.` so staticcheck points the plugins at the new accessors when they bump alpamon.

pkg/runner/client.go
```diff
@@ -182,12 +249,11 @@ func (wc *WebsocketClient) SendPongResponse() error {
 }
 
 func (wc *WebsocketClient) ReadMessage() (messageType int, message []byte, err error) {
```
geunwoonoh, line 255: Suggestion: `ReadMessage` and `WriteJSON` (line 450) now return `net.ErrClosed` when there is no connection, but neither has a doc comment saying so. `SetReadDeadline` (lines 177-178) documents the same contract.

pkg/runner/client_reconnect_test.go
```diff
@@ -0,0 +1,205 @@
...
+	err := wc.CloseAndReconnectOnRequest(ctx)
+
+	require.ErrorIs(t, err, context.Canceled)
+	assert.False(t, tracked.closed.Load(), "an already-cancelled context still dialled")
+}
+
+// reconnectPeerServer upgrades every connection, immediately sends a
+// "reconnect" frame, and records when each connection was accepted so the
+// test can measure the gap between successive peer-requested reconnects.
+type reconnectPeerServer struct {
```
geunwoonoh: Suggestion: `reconnectPeerServer`, `newReconnectPeerServer` and `connections()` are identical to the copy in `pkg/pluginclient/client_test.go` (about 50 lines). Move them to `internal/testutil`. `internal/runnertest` does not fit, because it imports `pkg/runner`, so the runner package's own tests would form an import cycle.

pkg/pluginclient/client.go
```diff
@@ -253,14 +263,14 @@ func (c *Client) configWorker(ctx context.Context) {
 }
 
 func (c *Client) setReadLimit() {
-	if c.WsClient == nil || c.WsClient.Conn == nil {
+	if c.WsClient == nil {
```
geunwoonoh: Suggestion: this nil check can never fire. The only caller, `RunForever`, already returns at line 279 when `WsClient` is nil, and `SetReadLimit` is now safe on a nil conn. Drop the wrapper and call `c.WsClient.SetReadLimit(MaxMessageSize)` at the four call sites.

pkg/runner/client.go
```diff
@@ -202,26 +268,60 @@ func (wc *WebsocketClient) Connect(ctx context.Context) error {
...
+
+// Close sends a close frame and closes the connection. It does not drain the
+// peer's reply: only the read loop may read, and Close runs outside it.
 func (wc *WebsocketClient) Close() {
-	conn := wc.Conn // Connect() may swap wc.Conn from the RunForever goroutine mid-close; every step below must act on the one connection we started with.
+	wc.closeConn(false)
+}
+
+// closeAndDrain also waits for the peer's close reply. Only CloseAndReconnect
+// calls it, because only the read loop owns the reads.
```
geunwoonoh: Suggestion: "Only CloseAndReconnect calls it" goes stale as soon as a second caller appears. State the function's own contract instead.

```suggestion
// closeAndDrain also waits for the peer's close reply, so it reads the socket:
// call it only from the goroutine that owns the reads.
```

pkg/runner/client.go
jisung-02: Took option (a) in e020e227: the comment now says the pacing covers only the reconnect query and must be called from the read loop. You are right that the read-error path and the external plugins' own loops redial without a delay; the PR body's Compatibility section now says so and leaves that path as it was.

pkg/runner/client_close_test.go
jisung-02: Fixed in 2a16cb5d: the race job now runs ./pkg/runner and ./pkg/pluginclient with -count=2 as well. You are right that the earlier -race run was local only, so the concurrent-close test could not fail anywhere CI would notice.

pkg/runner/client.go
jisung-02: Fixed in e020e227 with your wording. The handler does write the pong through SendPongResponse; what it never does is close or replace the connection.

pkg/pluginclient/client_test.go
jisung-02: Deleted in 6ea1ce69. It fed reconnect to a nil-WsClient Client and asserted outcomeContinue, which is exactly what the test at line 94 already does.

pkg/runner/client_reconnect_test.go
jisung-02: Deleted in 6ea1ce69. Confirmed before deleting: CloseAndReconnect returns ctx.Err() on its first line, so the pacing wait was never reached and the test held with that wait removed.

pkg/runner/client_reconnect_test.go
jisung-02: Fixed in 6ea1ce69. The shared peer server sets a close handler on each connection and records when the client's close frame arrives; the runner test now measures the gap between the first two close frames. lastPeerReconnect is stamped right before the close, so that gap is the interval plus the difference in writing one close frame, and the margin dropped from 50ms to 10ms. The test still fails with the loop reverted to the unpaced CloseAndReconnect (measured 328µs against the 190ms floor).

pkg/pluginclient/client_test.go
jisung-02: Fixed in 6ea1ce69. Once the second connection lands, assert.Never checks that no third one arrives within 1s at a 10ms tick; the test runs in about 1s and no longer reads dialJitter. It fails with the loop reverted to the unpaced call (a third connection lands within 20ms). The comment moved onto the test as you suggested.

pkg/runner/client.go
jisung-02: Renamed to conn() in e020e227, matching ControlClient.

pkg/runner/client.go
jisung-02: Fixed in e020e227 and b0020a86: the const block sits directly after type handlerOutcome int in both packages. Agreed that type, const, var, func was not a convention this repo follows; the PR body no longer claims it.

pkg/runner/client.go
jisung-02: Fixed in e020e227: mu now sits directly above Conn, and Conn carries a Deprecated note pointing at SetReadDeadline, ReadMessage and WriteJSON.

pkg/runner/client.go
jisung-02: Fixed in e020e227: ReadMessage and WriteJSON now document net.ErrClosed in the same form as SetReadDeadline.

pkg/runner/client_reconnect_test.go
jisung-02: Fixed in 6ea1ce69: the server lives in internal/testutil as ReconnectPeerServer, with Connections() and the new Closes() used by the timing fix. Not internal/runnertest, for the import cycle you named.

pkg/pluginclient/client.go
jisung-02: Fixed in b0020a86: the wrapper is gone and the four call sites call c.WsClient.SetReadLimit(MaxMessageSize).

pkg/runner/client.go
jisung-02: Fixed in e020e227 with your wording: the comment states the contract instead of naming the caller.

jisung-02:
@geunwoonoh Thanks for the review. Every item is answered inline; the map to commits:

| Item | Commit |
|---|---|
| Pacing comment narrowed to the `reconnect` query (option a) | e020e227 |
| `handlerOutcome` comment: handler never closes or replaces the connection | e020e227 |
| `getConn` renamed to `conn()`; const block beside its type; `mu` above `Conn`; `Deprecated` on `Conn`; `net.ErrClosed` documented on `ReadMessage` and `WriteJSON`; `closeAndDrain` states its contract | e020e227 |
| `setReadLimit` wrapper dropped; pluginclient const block beside its type | b0020a86 |
| Race job runs `./pkg/runner ./pkg/pluginclient` | 2a16cb5d |
| Shared `ReconnectPeerServer` in `internal/testutil`; pacing measured between close frames; pluginclient test asserts no third connection within 1s; two duplicate tests deleted | 6ea1ce69 |
| PR description: Compatibility section now lists both removed exported methods as a breaking change and names the unaffected plugins; the declaration-order claim is gone | PR body |

Both pacing tests were run with the loop reverted to the unpaced `CloseAndReconnect` before committing: runner failed with `"328.25µs" is not greater than or equal to "190ms"`, pluginclient failed on the `assert.Never` guard. On the pushed head:

```
go test -race -count=2 ./pkg/runner ./pkg/pluginclient
ok  	github.com/alpacax/alpamon/v2/pkg/runner	82.455s
ok  	github.com/alpacax/alpamon/v2/pkg/pluginclient	3.471s
```

The two out-of-scope items (no read limit on the runner's main connection, and `SendPingQuery` having no caller here) are not filed yet.

pkg/runner/client.go
```diff
@@ -21,19 +22,26 @@ import (
...
+	outcomeContinue handlerOutcome = iota
+	outcomeReconnect
 )
 
 type WebsocketClient struct {
-	Conn                 *websocket.Conn
+	// mu guards Conn: Connect writes it from the read loop; Close reads it on shutdown.
+	mu sync.Mutex
+	// Conn stays exported for outside readers; inside this package use the accessors.
+	// Deprecated: use SetReadDeadline, ReadMessage and WriteJSON.
```
geunwoonoh, line 39: Warning (non-blocking): staticcheck recognizes a deprecation only when `Deprecated: ` starts its own paragraph. It splits `doc.Text()` on blank lines and checks each part's prefix (`analysis/facts/deprecated/deprecated.go` in honnef.co/go/tools v0.8.1). Here the note shares a paragraph with the sentence above, so SA1019 never fires, and the PR description's "staticcheck points those plugins at `SetReadDeadline`, `ReadMessage` and `WriteJSON`" does not hold yet. I ran staticcheck v0.8.1 on two fields shaped like this one, and only the field with a blank `//` line before `Deprecated:` was reported.

The fix does not affect alpamon's own lint: SA1019 skips uses in the defining package and its tests, and nothing outside `pkg/runner` reads `WebsocketClient.Conn`.

```suggestion
	// Conn stays exported for outside readers; inside this package use the accessors.
	//
	// Deprecated: use SetReadDeadline, ReadMessage and WriteJSON.
```

geunwoonoh (dismissed):
### Summary
Re-reviewed at 6ea1ce6 against the 14 inline comments from the previous review. I checked each one against the diff from d3f1ecd, not only against the replies. 13 are resolved as described. The remaining one is the `Deprecated:` note on `Conn`, which is in place but written so staticcheck does not recognize it. CI is green, including the race job, which now runs `pkg/runner` and `pkg/pluginclient` (57s).

### Resolved

| Comment | Result |
|---|---|
| Pacing comment wider than the code | Option (a) taken; the PR description now says the read-error path is not paced |
| Race job skipped both packages | Added, and passing in CI |
| Stale Compatibility section | Both removed methods and the breaking change are listed |
| `handlerOutcome` comment | Fixed |
| Duplicate pluginclient test | Deleted |
| Test that passed without the pacing wait | Deleted |
| 50ms `dialJitter` margin | Now measured between close frames, with a 10ms margin |
| 5s real wait in pluginclient | `assert.Never` over 1s |
| `getConn` name | `conn()`, matching `ControlClient` |
| Enum split from its constants | Adjacent in both packages |
| `mu` placement and `Deprecated` | `mu` moved; the `Deprecated` note is not recognized (below) |
| `ReadMessage`/`WriteJSON` docs | Documented |
| Duplicated peer server | Moved to `internal/testutil` |
| Unreachable `setReadLimit` guard | Wrapper removed |
| `closeAndDrain` comment | Fixed |

I re-derived the new pacing measurement: the gap between close frames is `(L2 - L1) + (δ2 - δ1)`, where `L2 - L1 >= interval` because the timer never fires early, and δ is only the time from writing a close frame to the server's close handler running. Dial and drain are out of it.

### Warnings (1)
1. **pkg/runner/client.go:37-38**: the `Deprecated:` note shares a paragraph with the sentence above it, so staticcheck never reports SA1019 and the PR description's staticcheck claim does not hold yet. Non-blocking; details and a one-line fix inline. Please add it before merging.

### Suggestions (optional)
- `pkg/runner/client_reconnect_test.go:131`: 10ms is already far safer than before. With the pacing removed the gap drops to about 0.3ms against a 200ms interval, so a margin of half the interval would catch the same regression with no room left for a CI flake.
- `internal/testutil/fs.go`: the package comment still says the package provides filesystem assertions; it now also holds the WebSocket peer server. That file is outside the diff, so no inline comment.
- PR description, Approach section: "A handler no longer touches the socket" is the wording already fixed in the code comment. The handler writes the pong; what it never does is close or replace the connection.

### Highlights
- Every reply names its commit, and the test replies give the measured failure with the fix reverted (328µs against the 190ms floor; a third connection within 20ms). That made this re-review quick to verify.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing, including the race job
- [x] No security vulnerabilities (the pacing scope is documented in the PR description)
- [ ] Documentation updated: the `Deprecated:` note needs its own paragraph for the staticcheck claim to hold

pkg/runner/client.go
jisung-02: Fixed in e2894166: the `Deprecated:` note now starts its own paragraph. You were right that it was inert where it stood, so the PR description was claiming something staticcheck would never print.

Checked it both ways with staticcheck v0.8.1, against a throwaway package that reads `wc.Conn` from outside `pkg/runner`. Without the blank `//` line, staticcheck exits 0 and says nothing. With it:

```
probe.go:5:54: wc.Conn is deprecated: use SetReadDeadline, ReadMessage and WriteJSON.  (SA1019)
```

The PR description now carries that line instead of the bare claim.

jisung-02:
@geunwoonoh Thanks for the re-review. The warning and the three optional suggestions are all in.

| Item | Commit |
|---|---|
| `Deprecated:` note needs its own paragraph | e2894166 |
| Pacing margin: half the interval instead of 10ms | e0e73073 |
| `internal/testutil` package comment names only filesystem assertions | 1b487a42 |
| PR description: the handler writes the pong, so "no longer touches the socket" is wrong | PR body |

On the margin: the test interval is 200ms, so the floor moved from 190ms to 100ms. Reverting the read loop to the unpaced `CloseAndReconnect` still fails it, now with more room than before:

```
--- FAIL: TestRunForever_PacesRepeatedPeerReconnectRequests (0.01s)
    Error: "718.375µs" is not greater than or equal to "100ms"
```

On the deprecation, staticcheck v0.8.1 against a throwaway package reading `wc.Conn`: silent before the change, and after it

```
probe.go:5:54: wc.Conn is deprecated: use SetReadDeadline, ReadMessage and WriteJSON.  (SA1019)
```

The Approach section now says the handler never closes or replaces the connection, matching the code comment. The Tests section no longer claims the margin is only the time to write one close frame, since it is now half the interval.

On the pushed head:

```
go test -race -count=2 ./pkg/runner ./pkg/pluginclient
ok  	github.com/alpacax/alpamon/v2/pkg/runner	83.759s
ok  	github.com/alpacax/alpamon/v2/pkg/pluginclient	3.463s
```

The two out-of-scope items, no read limit on the runner main connection and `SendPingQuery` having no caller here, are still not filed.

jisung-02:
The two out-of-scope items are now filed together as #485.

One correction while writing it up: `SendPingQuery` is not unused. A code search across the org found six call sites outside this repo, two each in `alpamon-power-plugin`, `alpamon-repository-plugin` and `alpamon-cert-authority`, all discarding the returned error. So the question there is not removal but that an exported method on a published module has no test here, and a break would show up as a plugin that quietly stops pinging.

geunwoonoh (approved):
### Summary
Re-reviewed at 1b487a4. The three commits since my last approval address the one remaining warning and all three optional suggestions. Only comments and a test margin changed; runtime code is untouched. All 29 checks pass, including the race job.

### Resolved

| Comment | Result |
|---|---|
| `Deprecated:` note not recognized | Fixed in e289416; the reply shows staticcheck reporting SA1019 on an outside read of `wc.Conn` |
| Pacing test margin | Widened to half the interval in e0e7307. An unpaced gap is about 0.3ms, so the 100ms floor still catches the regression |
| `internal/testutil` package comment | Now names the peer server (1b487a4) |
| PR description wording | The Approach section says "closes or replaces the connection", and the staticcheck claim now quotes the actual output |

### Critical issues (0)
None.

### Warnings (0)
None.

### Suggestions (0)
None.

### Highlights
- Each fix was verified in both directions before it was reported: the deprecation note with and without the blank line, and the pacing tests with the fix reverted.

---

**Verdict**: APPROVED

**Checklist**:
- [x] Tests included and passing, including the race job
- [x] No security vulnerabilities (the pacing scope is documented in the PR description)
- [x] Documentation updated

Issue #486, opened 2026-09-22
net: interface reporting fixes (enp0s prefix, negative MTU)


eunyoung14:
## What happens

Two independent interface-reporting bugs, tracked together since both are collected in the same code path (`getNetworkInterfaces`, `pkg/runner/commit.go`).

### 1. `VirtualIfacePattern` matches `enp0s*` physical interfaces

`VirtualIfacePattern` (`pkg/utils/metrics.go:54`) contains the prefix `enp0s`:

```
^(lo|docker|veth|br-|virbr|vmnet|tap|tun|wg|zt|tailscale|enp0s|cni|utun|awdl|llw|bridge|anpi|ap|Loopback|isatap|Teredo|6to4)
```

`enp0s…` is an ordinary systemd predictable name for a **physical** NIC on PCI bus 0: `enp0s31f6` (a common onboard Intel NIC), `enp0s3` / `enp0s8` (VirtualBox defaults), `enp0s25`. The pattern is not anchored past the prefix, so all of them match and are treated as virtual.

The pattern gates three sites:

| site | effect |
|---|---|
| `pkg/runner/commit.go:594` (`getNetworkInterfaces`) | the interface is never registered in the inventory |
| `pkg/runner/commit.go:624` (`getNetworkAddresses`) | its addresses are never reported |
| `pkg/utils/metrics.go:177` (`FilterVirtualInterface`) | its traffic counters are never reported |

So on a host with such a NIC the agent reports no interface, no address and no traffic for it, and does so silently — the filter is a `continue`, not an error.

It is not deliberate. The last edit to this pattern only appended Windows adapter names; `enp0s` predates it.

**Why it is not just absence.** If **every** NIC on a host matches the pattern, `getNetworkInterfaces` returns an empty slice with no error, and the inventory delta (`compareListData`, `pkg/runner/commit.go:413-440`) then runs with an empty current set. An empty current set makes the delta a full removal: it sends a `DELETE` for every interface the server still holds for that host.

### 2. A negative MTU is reported as a measurement on Windows

`getNetworkInterfaces` (`pkg/runner/commit.go:581-608`) copies the MTU straight from the Go standard library:

```go
interfaces = append(interfaces, Interface{
    Name:      iface.Name,
    Flags:     getFlags(iface),
    MTU:       iface.MTU,
    Mac:       mac,
    Type:      0, // TODO
    LinkSpeed: 0, // TODO
})
```

[...]

PR #490, opened 2026-09-22
fix(runner): report an unknown MTU as unset instead of a negative number

eunyoung14:
## What

`net.Interface` holds `-1` for an adapter whose MTU the operating system does not give, which is what the standard library does on Windows. `getNetworkInterfaces` copied that value straight into the reported interface, so "unknown" went on the wire as though it were a measurement.

The whole interface list is reported in one payload, so a single adapter carrying `-1` puts every interface in that payload at risk wherever the MTU is read as a non-negative quantity.

## Change

- `Interface.MTU` becomes `*int` with `omitempty`, so an MTU that was not reported is an absent field rather than a number. An absent field was chosen over an explicit `null` because the field is read as a non-negative integer, not a nullable one.
- `reportedMTU` normalizes at the point of collection: `<= 0` becomes unset, everything else is passed through. The interface itself is still reported, since an unknown MTU is no reason to leave a NIC out of the inventory.
- The mapping from `net.Interface` moves into `buildInterfaces`, so a list can be handed to it directly. `getNetworkInterfaces` is now the `net.Interfaces()` call plus that function.
- `GetData` carries the same pointer, and the sync comparison runs through go-cmp, which compares what a pointer holds rather than which pointer it is. A test pins that down, since it is the part of this change that would quietly break if the comparison ever moved.

## Tests

- Table-driven over `buildInterfaces` with a fake interface list covering `1500`, `9000`, `-1` and `0`: the reported value for the first two, unset for the last two, and all four interfaces present in the output. The emitted JSON is checked too, so `mtu` is asserted absent for an unknown MTU rather than only nil in the struct.
- The existing `TestGetNetworkInterfaces` now asserts that an MTU, when reported, is positive.

Verified that the new test fails without the normalization.

## Note for the release

A host that used to report `-1` reports one interface change on the first run after the upgrade.

## Verification

`go build ./...`, `go vet ./...`, `GOOS=windows go build ./...`, `GOOS=windows go vet ./...`, `go test ./... -p 1` and `go test -race ./pkg/runner` all pass locally on macOS (arm64).

Part of #486

🤖 Generated with [Claude Code](https://claude.com/claude-code)


eunyoung14:
On the review note about the comparison test not exercising the production sync-comparison path:

`TestInterfaceComparisonFollowsTheMTUValue` asserts exactly the expression the sync comparison evaluates for each row, `cmp.Equal(current.GetComparableData(), remote.GetComparableData())`, which is what decides whether a row is re-sent. What it does not do is call `compareListData` itself, because that function's only output is a write onto the request queue, so driving it needs a session and a queue rather than a return value to assert on. That seam is worth having, but widening it is a larger change than this fix, and the assertion above is the part this PR could break.

The other path that reads the field, the sync fingerprint, goes through JSON, where a pointer and a value serialize alike, so it is unaffected by the type change.

